PR #33, opened 2015-07-22
Captcha

voodoorai2000:


xuanxu:
Merged master to have it updated.
@voodoorai2000 I can't see the captcha in my machine even with the captcha keys configured in the enviroment, is it working for you in local?

voodoorai2000:
Good catch @xuanxu!
Indeed it was only working after a refresh. 
It was missing a turbolinks option when following the new_debate_path.
Let's see how it's looking now.

PR #35, opened 2015-07-25
configures database_cleaner

voodoorai2000:


voodoorai2000:
Included in https://github.com/AyuntamientoMadrid/participacion/pull/40

PR #90, opened 2015-08-04
Adds badges for travis, code climate and gemnasium

kikito:


kikito:
Pending: Coveralls badge

PR #108, opened 2015-08-06
Adds CSS class to voted options

xuanxu:
Ref: #88 
After this @decabeza has to create the proper styles


app/views/debates/_votes.html.erb
```diff
@@ -1,7 +1,16 @@
+<% voted_classes = case @voted_values[debate.id]
+    when true
+      {in_favor: "voted", against: "no-voted"}
+    when false
+      {in_favor: "no-voted", against: "voted"}
+    else
+      {in_favor: "", against: ""}
+  end %>
```
kikito: This belongs in a helper.

xuanxu: so true, moved 37199125f

PR #182, opened 2015-08-17
Organizations 99

kikito:
Implements #99 

Adds a new "register as organization" in the registration footer:
participacion

The organization registration folder is different from the regular one - it has organization-related fields and it does not have first_name or last_name.
screen shot 2015-08-17 at 17 22 35

Administrators can Validate, Reject and Filter organizations in the administration menu:
screen shot 2015-08-17 at 17 20 06

The "My account" page always changes when your account is an organization account:
screen shot 2015-08-17 at 17 21 08


xuanxu:
:tada: :+1:

PR #209, opened 2015-08-20
Added ghostscript to Readme. Required for captcha

alejanderl:
One test was failing apparently cause ghostscript was missing.
All test passing after `brew install ghostscript`
 I thought would be nice to include in the README file


voodoorai2000:
¡Gracias @alejanderl!

PR #246, opened 2015-08-25
N plus 1

kikito:
Related with #228

- [x] Installs bullet and runs it on the feature specs
- [x] Removes default scopes and adds them when necessary
- [x] Adds cached fields for vote counts


kikito:
Cancelada por #255 y #253

PR #247, opened 2015-08-25
Add filters to debates index page

mlovic:
Allows user to select one of three filters to specify the order in which debates are listed on /debates page. 'Sort by newest' is the default. Issue #193

All tests pass except for "I18n does not have unused keys" as some of these are generated dynamically in app/views/debates/_filter_selector.erb


mlovic:
Borré el branch de mi repositorio estando el PR cerrado y ya no me deja volver a abrirlo... Abriré uno nuevo.

PR #250, opened 2015-08-26
Add filters to specify order of debates#index

mlovic:
Allows user to select one of three filters to specify the order in which debates are listed on /debates page. 'Sort by newest' is the default. Issue #193

All tests pass except for "I18n does not have unused keys" as some of these are generated dynamically in app/views/debates/_filter_selector.erb


mlovic:
Cualquier sugerencia es bienvenida.

No sé qué hacer con el test que no pasa. El problema es que no reconoce que haya usado los keys para los filtros al haberlos generado interpolando el nombre del filtro   `t("debates.index.filter_#{f}")`

Podría quitar el loop, "hardcode" los nombres de los filtros y evitar la interpolación.

xuanxu:
Thanks @mlovic, but the filters in the PR only reorders the debates of the page. This solution would only work if the index page shows all the debates, but that is not the case, we are showing just a few and paginating.

The point of the issue is to apply the filter to the db query, so you get the debates ordered from the db. In order to do that it's probably needed to add some scopes and apply them depending on the selected search filter.

mlovic:
Thanks @xuanxu, I understand what you mean. I had not thought about the pagination, sorry about that. I will comment again when I have it fixed!

xuanxu:
:+1: if you prefer you can close this PR and open another once your done

PR #349, opened 2015-09-03
Smaller, better, no reason for the image being so big?

dbillionlabs:
The biggest image of the site is this one, and I think there's no need for it to be so big. Also downscaling take CPU and time.


xuanxu:
:+1: Gracias!

PR #353, opened 2015-09-04
Hot score 347

kikito:
Adds:
* A new order in the select for debates#index and Home#index
* A hot_score field to Debate
* It is recalculated before_save
* It takes into account positive & negative votes, amount of votes, amount of comments and created_at
* It is automatically recalculated when voting or adding a comment to a debate
* There is also `rake debates:hot_score` for recalculating in all debates


db/schema.rb
```diff
@@ -131,6 +132,16 @@
 
   add_index "identities", ["user_id"], name: "index_identities_on_user_id", using: :btree
 
+  create_table "inappropiate_flags", force: :cascade do |t|
```
xuanxu: ojo @kikito que parece que tienes en tu bd esto viejo

PR #451, opened 2015-09-09
Adds utf-8 cleaner

kikito:
Do not merge, needs a test.

Prevents 'Invalid byte sequence in UTF-8' by cleaning params up before
they reach the app with a rack middleware.


kikito:
Cerrando esta pr. Precisamente las cadenas no-utf8 que nos están llegando ahora mismo son cadenas en el POST, que es precisamente [lo que utf8-cleaner deja sin tocar](https://github.com/singlebrook/utf8-cleaner/blob/master/spec/middleware_spec.rb#L42-L50)

PR #476, opened 2015-09-12
Adds whenever

kikito:
This PR only adds whenever with a dummy task which touches a file. I want to deploy and test it in all environments before adding “real” behaviour to it.

Related with #355 (won’t close it until we do hot_score calculations via cron)


kikito:
This has been tested in staging, by the way

PR #564, opened 2015-09-28
Splits ability.rb and specs into several smaller files

kikito:
I'm doing this in preparation for the "Manager" ability, which will
require even more refactors of the abilities (for example,
manager can not modify their own account)


voodoorai2000:
:clap:

PR #643, opened 2015-10-26
Add paragraphs to comments

### LINKED ISSUE #600: Respetar los párrafos en los comentarios
Cuando un usuario escribe un comentario utilizando párrafos, al mostrar el comentario se deberían respetar estos espacios, ya que mejoran la legibilidad del comentario.

Serabe:
Fix #600

### DIFF
--- a/app/helpers/text_with_links_helper.rb
+++ b/app/helpers/text_with_links_helper.rb
@@ -2,7 +2,7 @@ module TextWithLinksHelper
 
   def text_with_links(text)
     return unless text
-    sanitized = sanitize text, tags: [], attributes: []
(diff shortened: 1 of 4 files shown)


Serabe:
Failing tests are:

> 1) Verification::Management::Document#under_sixteen? returns false when the user is sixteen years old
>  2) Verification::Management::Document#valid_age? returns true when the user is sixteen years old
>  3) Stats Summary Users

First two seem to be a problem with dates and the last I'm not sure what the result should be.

I can reproduce 1 and 2 both in master and my branch locally, but I cannot reproduce 3 on either branch. Seems like db is not getting reset after tests though.

xuanxu:
:ok_hand:  Gracias!

PR #644, opened 2015-10-26
Refactor comment body

Serabe:
Refactor the body of the comment to use a couple of helpers instead of a long
list of `if`s in the html.

This will need an adjustment after merging #643

### DIFF
--- a/app/helpers/comments_helper.rb
+++ b/app/helpers/comments_helper.rb
@@ -17,4 +17,23 @@ module CommentsHelper
     comments.select{|c| c.parent_id == parent.id}
   end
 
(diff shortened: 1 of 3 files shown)


Serabe:
For failing tests, see #643.

For coverage decreasing, I see you are not covering helper methods for comments. Shall I add a spec for them?

Serabe:
Fix the issue after merging #643

voodoorai2000:
Thank you @Serabe! Yes please, those helper specs would be great!

Serabe:
Done. I should start doing these things earlier...

xuanxu:
ole!

PR #668, opened 2015-11-04
Update "More information" links to use PagesController

andion:
Paths were hardcoded instead of using PagesController actions and defined routes under routes.rb


xuanxu:
:+1:

PR #725, opened 2015-11-23
Featured debates

voodoorai2000:
Adds upcoming featured debate with Javier Barbero


voodoorai2000:
Making PR to Madrid fork.

PR #732, opened 2015-11-26
Fix double-submission prevention not working.

xJakub:
Since the submit buttons are asynchronous, the website should disable the submit button once clicked. However, this does not work due to an error in the code, so an user can accidentally submit a the same comment multiple times in a debate. This pull request fixes this error.

### DIFF
--- a/app/assets/javascripts/prevent_double_submission.js.coffee
+++ b/app/assets/javascripts/prevent_double_submission.js.coffee
@@ -21,7 +21,7 @@ App.PreventDoubleSubmission =
           button.data('text', null)
 
   initialize: ->
-    $('form').on('submit', event, ->
+    $('form').on('submit', (event) ->
       buttons = $(this).find(':button, :submit')
       App.PreventDoubleSubmission.disable_buttons(buttons)
     ).on('ajax:success', ->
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/prevent_double_submission.js.coffee
```diff
@@ -21,7 +21,7 @@ App.PreventDoubleSubmission =
           button.data('text', null)
 
   initialize: ->
-    $('form').on('submit', event, ->
```
xJakub: $('form').on('submit', event, function() {

app/assets/javascripts/prevent_double_submission.js.coffee
```diff
@@ -21,7 +21,7 @@ App.PreventDoubleSubmission =
...
+    $('form').on('submit', (event) ->
```
xJakub: $('form').on('submit', function(event) {

voodoorai2000:
Thank you very much @xJakub!

Could you write a spec to make sure this bug does not come back?

xJakub:
I've been working on it. However, Chrome and PhantomJS seem to execute the old code correctly, while Firefox treats it as an error. So a spec would still not guarantee that this error doesn't come back unless it is also tested using Firefox's engine.

voodoorai2000:
Thanks for that spec @xJakub!

No worries about testing it for Firefox specifically. We would have to configure selenium and run it with multiple browsers for this spec, which is probably not worth it.

Looks great as it is!

PR #735, opened 2015-11-27
Use `en` as default locale on tests

josepjaume:
This ensures the language in which the tests will run is the english language, even if a previous test changes it (it's a global setting!).

Running the tests locally, I've detected sometimes `spec/features/emails_spec.rb` would fail depending on the order of execution (tests are executed in random order).

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -21,6 +21,7 @@ RSpec.configure do |config|
   config.before(:each) do |example|
     DatabaseCleaner.strategy = example.metadata[:js] ? :truncation : :transaction
     DatabaseCleaner.start
+    I18n.locale = :en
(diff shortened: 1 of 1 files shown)


xuanxu:
:+1: Thanks @josepjaume!

josepjaume:
Just out of curiosity - did you have that random failure in the past?

xuanxu:
I don't recall to have that failure in local. There's a couple random failures appearing in Travis from time to time, I'm not sure but maybe one of them was i18n related.

josepjaume:
Cool! Random failures are evil. Let's see if this fixes anything.

xuanxu:
Yeah!
Since we're using Poltergeist and Phantom we expect some random misterious failures :)

PR #740, opened 2015-11-30
Auto-assign official level to users with official emails

xJakub:
This pull requests adds a new setting 'email_domain_for_officials'. When set, any user that confirms an email address under that domain (including subdomains) will automatically get level-1 official status. For example, if email_domain_for_officials is set to madrid.es, example@madrid.es and foo@bar.madrid.es would get official level on email confirmation.

A new rake task, 'users:check_for_official_emails' is also included. This task searches for users whose email addresses belong to the official domain, but don't have any official level. After that, it grants them level-1 official status. The task informs of the changes through the standard output, like in the following example:
```
# rake users:check_for_official_emails
admin (admin@madrid.es) is now a level-1 official.
marta (marta@madrid.es) is now a level-1 official.
```

* This task first uses a sql LIKE condition to filter the users in order to speed up the process. After that, it checks each of them individually.

A spec is also included to check if the official level is granted only when the email_address is really under the domain or its subdomains.


voodoorai2000:
Fantastic 	@xJakub!

Just a couple of details, I write them inline.

app/models/user.rb
```diff
@@ -47,6 +47,8 @@ class User < ActiveRecord::Base
   scope :by_document,    -> (document_type, document_number) { where(document_type: document_type, document_number: document_number) }
 
   before_validation :clean_document_number
+  
+  before_save :check_if_confirmation
```
voodoorai2000: This would probably be clearer calling a set_official_position method.

app/models/user.rb
```diff
@@ -199,6 +201,28 @@ def username_required?
   def email_required?
     !erased?
   end
+  
+  def has_official_email?
+    domain = Setting.value_for 'email_domain_for_officials'
+    return false if !email or !domain or domain.length == 0
```
voodoorai2000: This line seems to belong to the rake task rather than a check that should be done in every save.

Rake tasks are great to manipulate existing data, but should not add logic to the models that is not commonly used. We usually create rake tasks to manipulate the code and then delete the rake task, as it can be confusing to keep one-time rake tasks around.

app/models/user.rb
```diff
@@ -199,6 +201,28 @@ def username_required?
...
+    (email.end_with? "@#{domain}") or (email.end_with? ".#{domain}")
+  end
+  
+  # Check if the user is confirmed and has an official email address
+  # In that case, we assign a level 1 official level
+  def check_if_official_email
```
voodoorai2000: Comments are good used sparsely for very complicated code.
In this case you could split the method in two and convey the same meaning.
For example: set_official_position and official_email?

(Instead of using if in a method it's usually nicer to use a ? sign a the end of the method.)
These two methods: has_official_email? and check_if_official_email imply the same meaning. They could probably be merged into a single method or split into more differentiated methods.

spec/models/user_spec.rb
```diff
@@ -239,6 +239,29 @@
...
+      
+      expect(user1.official_level).to eq(1)
+      expect(user2.official_level).to eq(1)
+      expect(user3.official?).to_not eq(true)
+      expect(user4.official?).to_not eq(true)
+      
+      [user1, user2, user3, user4].each { |user| user.destroy }
+      
+      # We reset the officials' domain setting
+      Setting.find_by(key: 'email_domain_for_officials').update(value: '')
```
voodoorai2000: No need for this line, before each tests the database is wiped clean.

app/models/user.rb
xuanxu: Please use the symbolic operator `||` instead of `or` to avoid confusion and unexpected behaviour here.

app/models/user.rb
xuanxu: I'm not sure we need this check on every user's save. It's probably enough to do it only after first registration, or email verification and email address change.

voodoorai2000: +1 to moving the before_save call to the controller.

xJakub:
Alright, I moved the official code to the controller and cleaned the code with your suggestions. Should has_official_email? be also moved outside of the model? I left it there since it seems to be a property to me.

voodoorai2000:
Fantastic! Looks great as it is, thank you very much for those changes :ok_hand:

PR #749, opened 2015-12-06
how to use moved to locale files and translated

Ana06:
**Use it in your local government** page moved to locale files using markdown and translated into English.


decabeza:
Thanks @Ana06! :+1:

PR #757, opened 2015-12-09
Use libsass for a faster stylesheet compilation

josepjaume:
This uses libsass for a faster stylesheet compilation (up to 4x faster!):

https://github.com/sass/libsass


decabeza:
Thanks @josepjaume! :tada:

PR #759, opened 2015-12-10
Simplify Settings API so it better resembles a key-value store

josepjaume:
This simplifies the settings interface in a way that makes it friendlier to read and write, and is also compatible with a key-value store.

The goal is to improve its usability as feature flags and reducing the boilerplate when setting values (decouple it from the `ActiveRecord` interface).


josepjaume:
:boom: :boom: :boom:

josepjaume:
Looks like a random failure, but I can't force rebuilding on Travis :crying_cat_face:

josepjaume:
Superseded by https://github.com/consul/consul/pull/763

Issue #730, opened 2015-11-25
Configurable dictionary for full text searches 


voodoorai2000:
We are currently using Postgres's spanish dictionary for full text search.
We would like to use different dictionaries depending on the configuration of a particular Consul instance.

Suggestion:
Use a setting's key to define the desired dictionary.

javierm:
@katafrakt has offered his help on this one. Thank you very much! :smile:

Maybe we could start by basing it on I18n.default_locale. It that doesn't seem the most simple solution to you, let us know!

katafrakt:
Makes sense to start with that :+1:

PR #762, opened 2015-12-10
Configurable dictionary for full text searches #730

gootyfer:


app/models/concerns/search_cache.rb
```diff
@@ -17,7 +17,7 @@ def searchable_values_sql
   end
 
   def set_tsvector(value, weight)
-    "setweight(to_tsvector('spanish', coalesce(#{quote(value)}, '')), #{quote(weight)})"
+    "setweight(to_tsvector('#{dictionary_for_searches}', coalesce(#{quote(value)}, '')), #{quote(weight)})"
```
xuanxu: this call to `dictionary_for_searches` is erroring and causing most of the [specs to fail](https://travis-ci.org/consul/consul/builds/96036595)

xuanxu:
We have added a lot of features recently and this code is now really behind master branch, so it is better to start from a new master-based branch. If you are still on it please issue a new PR. Thanks!

PR #763, opened 2015-12-10
Parameterize brand and social links

josepjaume:
This is an effort to remove references to any particular city or location from consul's codebase and add everything that I can as a setting.

This PR tackles:

- [x] Brand's name.
- [x] Social links

### DIFF
--- a/app/controllers/users/confirmations_controller.rb
+++ b/app/controllers/users/confirmations_controller.rb
@@ -53,7 +53,7 @@ class Users::ConfirmationsController < Devise::ConfirmationsController
   private
 
     def set_official_position
-      resource.add_official_position! (Setting.value_for 'official_level_1_name'), 1
+      resource.add_official_position! (Setting['official_level_1_name']), 1
     end
 
 end
--- a/app/helpers/admin_helper.rb
+++ b/app/helpers/admin_helper.rb
@@ -7,7 +7,7 @@ module AdminHelper
   def official_level_options
     options = [["",0]]
     (1..5).each do |i|
-      options << [[t("admin.officials.level_#{i}"), Setting.value_for("official_level_#{i}_name")].compact.join(': '), i]
+      options << [[t("admin.officials.level_#{i}"), Setting["official_level_#{i}_name"]].compact.join(': '), i]
     end
     options
   end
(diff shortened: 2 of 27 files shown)


kikito:
Hi, happy new year, thanks a lot for your PR!

I finally had time to review it and I have some comments:

* Could you remove the `setting` helper? I don't think it adds much to the scenario and it makes things confusing by offering two very similar ways of doing the same thing (as an example, in your own pull request there are some views using `Setting[...]` and other using `setting(...)`).
* Could you rename `brand_name` to `org_name`? Consul might be installed in places which are not a brand (Madrid isn't, for instance).
* ¿Can you fix the git merge conflicts, too?

Thanks again!

josepjaume:
Hi @kikito, of course! Everything you said makes a lot of sense. Will get to it when I get some time! :+1:

josepjaume:
@kikito I've applied your changes and rebased to latest master :)

After this, I'd like to work on feature flags. I've been thinking a bit about how to implement that, and although `rollout` does its job, I think we could manage to do the same with `Settings` (but we'd loose the A/B testing facilities which I don't think we need anyway).

That said, I don't feel comfortable in hitting the database every time we check for a setting. Redis might do a better job here, right?

So my proposal would be:

* Use redis for storing all the settings
* Move feature flagging to settings under a `feature.X` namespace (`Settings['feature.debates']`).
* Use feature flags to hide views and within `cancan` to disable entire features.

What do you think?

xuanxu:
:+1: 
One last detail: adding comments to the code in setting.rb ( https://github.com/consul/consul/pull/763/files#diff-8c3f2a547e7f7d66e1756971a1ef20c2R7 ) brakes the norm we are using, could you delete them?

josepjaume:
Of course! Is that written somewhere so I can adhere to it in further PRs?

josepjaume:
Done!

xuanxu:
Thanks!
The general styleguide and best practices policy is just linked in the Contributing.md file but I don't think it mentions explicitly anything about comments. In this case is just about following the format of the existing code. Right now there's no comments in methods, so adding comments only to some of them would make the repo inconsistent.

PR #766, opened 2015-12-12
Normalize page paths

josepjaume:
Normalize page paths so they can be accessed by `page_path(:id)`.

The benefit of this is that forks can add any arbitrary amount of pages without affecting the routing layer.

It also results in a nicer API, IMO.


xuanxu:
:+1: Gracias @josepjaume!

PR #772, opened 2015-12-16
Annotator plus

kikito:
Fixes several things in the current annotator-based implementation.

Missing:

* Permission tests
* Unregistered users still see the “create annotation widget” (fails on save, with proper error message)


app/models/annotation.rb
```diff
@@ -3,4 +3,8 @@ class Annotation < ActiveRecord::Base
 
   belongs_to :legislation
   belongs_to :user
-end
\ No newline at end of file
+
+  def permissions
+    { update: [user_id], delete: [user_id] }
```
kikito: Not sure if it's delete or destroy here

PR #779, opened 2015-12-18
uses proposal total_votes for % helper methods

xuanxu:
There was a bug in the supports % shown for Proposals when physical votes where included


kikito:
:+1:

PR #806, opened 2016-01-05
[WIP] Proposal filters using React

beagleknight:
I'll started my adventure with React and the experience is good so far :D.

I have started implementing a bunch of components and playing with props and events. I'm not 100% happy with the current implementation.

The next step is connecting this simple component with our backend.


beagleknight:
I'm sorry! Wrong fork! :(

PR #807, opened 2016-01-06
Run tests in parallel

josepjaume:
This is an attempt to speed up tests by parallelizing them on CI.


josepjaume:
After checking this out: https://github.com/grosser/parallel_tests/wiki I've decided this adds too much complexity. End of the show!

PR #809, opened 2016-01-07
Selects to links (needs style)

kikito:
Transforms selects into links in debates/index and proposals/index

Related with #680


decabeza:
Styles added! :tada:

PR #827, opened 2016-01-14
Adds Tolk gem for i18n web interface

### LINKED ISSUE #793: Create an interface for traslations
Create an interface for translations to allow other contributors (not programmers) to translate the application in other languages.

We can use something like [tolk](https://github.com/tolk/tolk).

atd:
Just the gem install, migrations and initializers

Closes #793


xuanxu:
👍:tada:

PR #834, opened 2016-01-19
Make search cache reusable

beagleknight:
I wanted to use the `SearchCache` concern in a new model but I detected it was hardcoded to `proposals` table.
I included a reference to the model `table_name` making it reusable.


xuanxu:
:+1: nice catch!

PR #847, opened 2016-01-21
Performance/use passenger

josepjaume:
Phusion passenger is way, way faster than puma at serving static assets.


josepjaume:
Oh man, WRONG PR AGAIN. I'm so sorry :(

PR #851, opened 2016-01-22
Store locale in a user's field and switch locale on mailers

josepjaume:
This will ensure all emails will be sent in the user's locale (the last one she used in in the application).

Tests were behaving as this was already happening, as jobs were executed synchronously. In production though, they were asynchronous and thus were getting the default locale (spanish).


voodoorai2000:
Thanks @josepjaume :+1:

PR #855, opened 2016-01-26
Categories and Geozones

voodoorai2000:
Uses geozones instead of district tags.
Makes proposals filterable by category and geozone.
Adds generic seeds and map for geozones.


xuanxu:
nice!

PR #872, opened 2016-02-03
refactors proposals search

kikito:
This refactors the way searches are done in Proposals.

* `SearchCache` was not applying unaccent when calculating the tsvector, but the search was applying it to the search terms. It is applied everywhere now.
* `SearchCache` automatically discards any blank strings (this was done on each model before)
* `SearchCache` automatically cleans up html before calculating the tsvector
* The option `against:` is ignored in `pg_search_scope` when using a tsvector - so it is removed from `Proposal`'s scope.
* Similarly, the `associated_against:` can be removed because the tsvector already contains the tags inside of it. Removing this option simplifies the generated search sql
* The `prefix: true` option allows looking for `dolor`, but typing `dolo` and `dol` will also bring you results

### DIFF
--- a/app/models/concerns/search_cache.rb
+++ b/app/models/concerns/search_cache.rb
@@ -13,11 +13,14 @@ module SearchCache
   private
 
   def searchable_values_sql
-    searchable_values.collect { |value, weight| set_tsvector(value, weight) }.join(" || ")
+    searchable_values
+      .select{ |k,_| k.present? }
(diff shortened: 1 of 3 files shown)


kikito:
Notice that we will have to run `rake proposals:save` in the servers in order to recalculate the tsvector there.

voodoorai2000:
:ok_hand:

xuanxu:
Niice!

PR #892, opened 2016-02-08
Public stats

xuanxu:
adds /stats (and /stats.json) route to display basic public stats


IvanDuque:
How to access to stats in home page.. for example just total visit and total comments?

PR #983, opened 2016-03-10
Check if spending_proposals is enabled to show 'valuation' link

andion:
The link was broken when Spending Proposals is disabled from administration.

The /valuation link points to `Valuation::SpendingProposalsController` which has a `feature_flag :spending_proposals` flag so clicking the link lead to an error page.


xuanxu:
Thanks @andion!

Issue #945, opened 2016-02-29
Tag links


voodoorai2000:
Due to duplications in tag names (capital letters, spelling mistakes, etc) some tag links don't work as expected. 

Suggestion:
Use search in tag links, for example, instead of https://decide.madrid.es/proposals?tag=movilidad we could use https://decide.madrid.es/proposals?search=movilidad

pablogfb:
We are on it!

PR #984, opened 2016-03-10
fixes tags links through searches to avoid misspellings

pablogfb:


xuanxu:
Thanks @pablogfb!
Closes #945

PR #989, opened 2016-03-11
Fixes valuators name

decabeza:


decabeza:
Done on https://github.com/consul/consul/pull/990

Issue #982, opened 2016-03-09
Add Valuator description 


xuanxu:
So admin can describe briefly each valuator rol when creating them.
The description should be visible in the admin/spending_proposals section when assigning valuators.

PR #997, opened 2016-03-15
Validator description

xuanxu:
Added description to validators.
The description added in the valuators creation form is show as a label title in admin/spending_proposals edit form.


xuanxu:
Closes #982

Issue #922, opened 2016-02-22
Add stats for signups


voodoorai2000:
In the admin/stats section we should have a graph displaying the number of signups per day.

inobCR:
Voy a por ella

voodoorai2000:
Partial work done [here](https://github.com/consul/consul/pull/1003), temporarily on hold, anyone feel free to continue

voodoorai2000:
This has already been implemented using Piwik/Google Analytics

PR #1003, opened 2016-03-17
Add stats for signups #922

inobCR:


voodoorai2000:
Thank you @inobCR and @nachogomez for you work here, as we talked privately you don't have time to wrap this up right now and it's not a priority, so we are closing it for now.

Issue #918, opened 2016-02-22
Add stats for Investment Projects


voodoorai2000:
In the admin/stats section we should have a graph displaying the number of Investment Projects (a.k.a SpendingProposals) completed per day.

inobCR:
Voy

voodoorai2000:
Closed via #1009

PR #1009, opened 2016-03-21
Add stats for Investment Proyect (spending proposals) #918

inobCR:
in the admin/stats


voodoorai2000:
Muchas gracias @inobCR!!

PR #1010, opened 2016-03-21
Select debates which display as featured

MaiteHdezRivas:
Admin users can mark or remove debates as featured.


decabeza:
@MaiteHdezRivas and @palomahnhp continue from #1041 😉

PR #1043, opened 2016-04-06
Management spending proposals

xuanxu:
Adds features to manage spending proposal:
Managers can list/search/view spending proposals and vote in behalf of verified users

Extra: added to dev_seeds verified and level_2 users


xuanxu:
lineamadrid-index-spending

lineamadrid-show-spending

PR #1044, opened 2016-04-06
Spending proposal responsible

kikito:
Saves the responsible name in the spending proposal, in case the user is later eliminated.

Notice that this pr requires a rake task to be executed on the server.


kikito:
Must change to save the document_number instead of the username, closing until fixed

PR #1046, opened 2016-04-07
Valuator extra links

kikito:
Before:
screen shot 2016-04-06 at 12 40 43

After:
screen shot 2016-04-07 at 10 56 30

Much better, I think you'll agree.


xuanxu:
captura de pantalla 2016-04-07 a las 10 58 05

xuanxu:
Not sure I agree :) maybe would be nice to remove the ones with 0 count, doesn't it?

Issue #885, opened 2016-02-06
Expire admin passwords once a year


voodoorai2000:
After one year of creating the account, admin passwords should expire and the admin should be asked to set a new and different password.

gitmarcia:
Me pongo con ella

voodoorai2000:
:clap: :+1:

palomahnhp:
Hemos visto que en Devise Security Extension hay un módulo :password_expirable ¿podemos usarlo o preferís otra forma?

voodoorai2000:
Closed via #1106

PR #1106, opened 2016-05-04
 Expire admin passwords once a year #885 

gitmarcia:


config/initializers/devise_security_extension.rb
```diff
@@ -0,0 +1,127 @@
...
+
+  # Time period for account expiry from last_activity_at
+  # config.expire_after = 90.days
+end
+
+module Devise
+  module Models
+    module PasswordExpirable
+      def need_change_password?
+        if self.administrator?
```
voodoorai2000: how about something like this?

def needs_password_change? 
  self.administrator? && password_expired?
end

def password_expired?
  self.password_changed_at < self.expire_password_after.ago
end

I'm not sure if we need these conditions (self.expire_password_after.is_a? Fixnum or self.expire_password_after.is_a? Float or self.password_changed_at.nil?) 
Let's see if any tests fail ;)

voodoorai2000:
👍

Issue #946, opened 2016-02-29
Automatically collapse big comment threads


voodoorai2000:
Some comment threads are very deep and make it hard to skip to the next root comment. Loading them can also take a long time. We should automatically collapse large comment threads and display a link to see the hidden levels.

Another option, proposed by kikito, would be to load the comments in a secondary request. That way the resource could be visible for the user in less time, meanwhile the comments would start be loaded. This would make a huge impact while using a mobile device (usually with have slower internet connection)

nachogomez:
Vamos a por esta @inobCR @nachogomez

voodoorai2000:
👍

voodoorai2000:
Partial work done [here](https://github.com/consul/consul/pull/1113), temporarily on hold, anyone feel free to continue

lucialuzuriaga:
Closing due to inactivity. Anybody is welcome to pick this up.

PR #1113, opened 2016-05-12
Automatically collapse big comment threads #946

inobCR:
Realizada junto con @nachogomez


inobCR:
Realizada junto con @nachogomez

voodoorai2000:
Thank you @inobCR and @nachogomez for you work here, as we talked privately you don't have time to wrap this up right now and it's not a priority, so we are closing it for now.

PR #1137, opened 2016-05-31
[WIP] Custom folder structure

### LINKED ISSUE #953: Custom folder structure
In order to improve the collaboration between forks and upstream, there should be a place to put custom code and a place for generic code. There are a number of possibilities to solve this issue, such as gems and engines. Whilst we think these are good options, we would like to keep things as simple as possible and start with a custom folder structure.

Some of the custom folders could be:
- [ ] MVC
- [ ] Assets
- [ ] Configuration

andreslucena:
This tries to fix #953 or at least be a good starting point :) 

The general idea is that for every Rails mayor feature we have a custom folder that overrides or extends Rails default flow, so that each fork can have what it needs (images, style sheets, javascripts, views, models and so on). 

There are some things that could be a problem with this approach: 

* db/schema.rb 
* Gemfile.lock 

But at least it could serve as a fix to the changes made on http://www.consultaoviedo.es/ and https://plaza.podemos.info/ so that it's easier for them to upgrade (after moving the forks specific changes and checking out so it'll be easier to merge with consul upstream).

### DIFF
--- a/CUSTOMIZE_ES.md
+++ b/CUSTOMIZE_ES.md
@@ -1,0 +1,214 @@
+# Personalización
+
+Puedes modificar consul y ponerle tu propia imagen, para esto debes primero hacer un fork de https://github.com/consul/consul creando un repositorio nuevo en Github. Puedes usar otro servicio como Gitlab, pero no te olvides de poner el enlace en el footer a tu repositorio en cumplimiento con la licencia de este proyecto (GPL Affero 3). 
+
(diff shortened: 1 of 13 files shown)


voodoorai2000:
Excelente, gracias @andreslucena!

voodoorai2000:
Cheers @andreslucena 🎉 
We have fixed the conflicts with master and added a couple of links in the Readme

andreslucena:
OK, it's a Work In Progress and somethings are missing (for instance public/ folder) but it's a good starting point for other cities.

Issue #836, opened 2016-01-19
Advanced Search Polish


voodoorai2000:
Critical:
- [x] Fix cross-browser issue with Firefox when selecting a custom date (date picker does not appear) #838 

Nice to haves:
- [ ] Add two more user types: 'user' and 'collective user'
- [ ] Remove one user type: 'municipal organization'

inobCR:
voy a por esta

voodoorai2000:
👍

voodoorai2000:
Partial work done [here](https://github.com/consul/consul/pull/1162), temporarily on hold, anyone feel free to continue

javierm:
This one is obsolete since we replaced user types with official levels.

The last trace of user types disappeared with pull request #1336.

PR #1162, opened 2016-06-22
Advanced search polish #836

nachogomez:


voodoorai2000:
Thank you @inobCR and @nachogomez for you work here, as we talked privately you don't have time to wrap this up right now and it's not a priority, so we are closing it for now. Come back soon ;)

PR #1201, opened 2016-09-06
be coherent in the use of bin/rake in the READMEs

jaimeiniesta:


jaimeiniesta:
Let me redo this.

xuanxu:
first time run should not use `bin`, otherwise it fails: #1202 :)
Closing!

PR #1203, opened 2016-09-07
no need to db:create here

jaimeiniesta:


voodoorai2000:
Thank you @jaimeiniesta!!

PR #1204, opened 2016-09-07
Fixes failing spec for badges

jaimeiniesta:
There's a failing spec at `spec/features/official_positions_spec.rb:56` - this fixes it by including the expected badge.

Not sure if this is the correct fix though, maybe badges shouldn't be shown for featured proposals for some reason?

### DIFF
--- a/app/views/proposals/_featured_proposal.html.erb
+++ b/app/views/proposals/_featured_proposal.html.erb
@@ -7,7 +7,14 @@
       <% else %>
         <%= proposal.author.name %>
       <% end %>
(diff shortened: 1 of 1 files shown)


jaimeiniesta:
This is how the featured proposals look now:

captura de pantalla 2016-09-07 a las 13 20 21

And this is how they would look if we include the badge:

captura de pantalla 2016-09-07 a las 13 22 16

decabeza:
Thanks @jaimeiniesta! 👍

PR #1229, opened 2016-09-18
added linux config instructions

gentooza:
my english it's not perfect at all xd, and I'm not completely sure about postgresql and ruby instructions, but they work for me.
It can be a good starting


voodoorai2000:
Thank you @gentooza!

PR #1230, opened 2016-09-19
Polling voters

voodoorai2000:
- Validates that voters are in census
- Validates that voters have not already voted

### DIFF
--- a/app/models/poll/voter.rb
+++ b/app/models/poll/voter.rb
@@ -1,15 +1,26 @@
 class Poll
   class Voter < ActiveRecord::Base
     belongs_to :booth
+    delegate :poll, to: :booth
 
     validate :in_census
+    validate :has_not_voted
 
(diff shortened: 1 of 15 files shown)


app/models/poll/voter.rb
```diff
@@ -0,0 +1,30 @@
...
+    def has_not_voted
+      errors.add(:document_number, :has_voted, name: name) if has_voted?
+    end
+
+    def census_api_response
+      @census ||= CensusApi.new.call(document_type, document_number)
+    end
+
+    def has_voted?
+      poll.voters.where(document_number: document_number, document_type: document_type).present?
```
xuanxu: What about using `.exists?` instead of `.present?` as we don't need the voters data

voodoorai2000: Sounds good!

PR #1237, opened 2016-09-29
French translation

mosleymos:
During the democracy os hackathon we've made a french translation of consul application
You can use it.


xuanxu:
Thanks for the PR @mosleymos. 
We have a FR translation in the contributed locales folder: https://github.com/consul/consul/blob/master/doc/locales/fr.yaml

PR #1265, opened 2016-11-17
Rename legacy Legislation model

amaia:
Renames the model and related stuff from Legislation to LegacyLegislation to avoid name clashes with the future Legislation module.


voodoorai2000:
👍

PR #1282, opened 2016-12-04
[Finished] Allow admins to manage geozones

### LINKED ISSUE #1275: Allow admins to manage geozones
When configuring forks, it would be useful to have an admin section with CRUD actions for geozones.

amiedes:
Resolves #1275

### DIFF
--- a/app/controllers/admin/geozones_controller.rb
+++ b/app/controllers/admin/geozones_controller.rb
@@ -29,6 +29,30 @@ class Admin::GeozonesController < Admin::BaseController
       render :edit
     end
   end
+
+  def destroy
+    # Check that in none of the other associated models talbes a record exists
+    # referencing this geozone
+    safe_to_destroy = true
+
+    # safe_to_destroy &= Proposal.where(geozone: @geozone).empty?
+    # safe_to_destroy &= Debate.where(geozone: @geozone).empty?
+    # safe_to_destroy &= SpendingProposal.where(geozone: @geozone).empty?
+    # safe_to_destroy &= User.where(geozone: @geozone).empty?
+
+    Geozone.reflect_on_all_associations.each do |association|
+      attached_model = association.klass
+      safe_to_destroy &= attached_model.where(geozone: @geozone).empty?
+    end
+
+    if safe_to_destroy
+      @geozone.destroy
+      redirect_to admin_geozones_url, notice: t('admin.geozones.delete.success')
+    else
+      redirect_to admin_geozones_path, flash: { error: t('admin.geozones.delete.error') }
+    end
+  end
+
   private
 
     def geozone_params
--- a/app/controllers/admin/geozones_controller.rb
+++ b/app/controllers/admin/geozones_controller.rb
@@ -33,18 +33,9 @@ class Admin::GeozonesController < Admin::BaseController
   end
 
   def destroy
-    # Check that in none of the other associated models tables a record exists
-    # referencing this geozone
-    safe_to_destroy = true
-
-    Geozone.reflect_on_all_associations.each do |association|
-      attached_model = association.klass
-      safe_to_destroy &= attached_model.where(geozone: @geozone).empty?
-    end
-
-    if safe_to_destroy
+    if @geozone.safe_to_destroy?
       @geozone.destroy
-      redirect_to admin_geozones_url, notice: t('admin.geozones.delete.success')
+      redirect_to admin_geozones_path, notice: t('admin.geozones.delete.success')
     else
       redirect_to admin_geozones_path, flash: { error: t('admin.geozones.delete.error') }
     end
--- a/app/controllers/admin/geozones_controller.rb
+++ b/app/controllers/admin/geozones_controller.rb
@@ -1,4 +1,4 @@
-class Admin::GeozonesController < ApplicationController
+class Admin::GeozonesController < Admin::BaseController
 
   respond_to :html
   
--- a/app/controllers/admin/geozones_controller.rb
+++ b/app/controllers/admin/geozones_controller.rb
@@ -9,6 +9,8 @@ class Admin::GeozonesController < Admin::BaseController
 
   def new
   end
+  def edit
+  end
 
   def create
     @geozone = Geozone.new(geozone_params)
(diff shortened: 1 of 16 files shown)


app/controllers/admin/geozones_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+      safe_to_destroy &= attached_model.where(geozone: @geozone).empty?
+    end
+
+    if safe_to_destroy
+      @geozone.destroy
+      redirect_to admin_geozones_url, notice: t('admin.geozones.delete.success')
+    else
+      redirect_to admin_geozones_path, flash: { error: t('admin.geozones.delete.error') }
+    end
+  end
```
voodoorai2000: When you find yourself writting comments to explain the code, there is usually a refactoring to do.

How about something like this?:

```
def destroy
  if @geozone.safe_to_destroy? 
    @geozone.destroy
    redirect_to admin_geozones_path, notice: t('admin.geozones.delete.success')
  else
    redirect_to admin_geozones_path, flash: { error: t('admin.geozones.delete.error') }
  end
end

class Geozone < ActiveRecord::Base
  
  def safe_to_destroy?
    Geozone.reflect_on_all_associations(:has_many).all? do |association|
      association.klass.where(geozone: self).empty?
    end
  end

end
```

db/dev_seeds.rb
```diff
@@ -34,7 +34,10 @@
 Setting.create(key: 'comments_body_max_length', value: '1000')
 
 puts "Creating Geozones"
-('A'..'Z').each{ |i| Geozone.create(name: "District #{i}") }
+('A'..'Z').each do |i|
+  code = "#{i.each_byte.to_a.first}"
```
voodoorai2000: Here you could use `code = i.ord`

spec/features/admin/geozones_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+
+    expect(page).to have_content 'Fancy District'
+
+    visit admin_geozones_path
+
+    expect(page).to have_content 'Fancy District'
+  end
+
+  scenario 'Edit geozone with no associated elements' do
+    target_geozone = create(:geozone, name: 'Edit me!', census_code: '012')
```
voodoorai2000: Here you could just use geozone instead of target_geozone

spec/features/admin/geozones_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+
+    within("#geozone_#{target_geozone.id}") do
+      click_link "Edit"
+    end
+
+    fill_in 'geozone_name', with: 'New geozone name'
+
+    click_button 'Save changes'
+
+    expect(proposal.reload.geozone.name).to eq('New geozone name')
```
voodoorai2000: In integration tests it's usually better to set expectations on what we can see instead of what is in the database. Here I would set the same expectation as in the previous test, it's also a little more consistent.

spec/features/admin/geozones_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+  end
+
+  scenario 'Delete geozone with no associated elements' do
+    target_geozone = create(:geozone, name: 'Delete me!')
+
+    visit admin_geozones_path
+
+    within("#geozone_#{target_geozone.id}") { click_link 'Delete' }
+
```
voodoorai2000: It's also a good practice for legibility to set an expectation for the flash. Here it would be 'Geozone successfully deleted'

spec/features/admin/geozones_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+    expect(page).not_to have_content('Delete me!')
+    expect(Geozone.find_by_id(target_geozone.id)).to be_nil
```
voodoorai2000: find_by_id is deprecated, try Geozone.where(id: geozone.id).first

spec/features/admin/geozones_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+  end
+
+  scenario 'Delete geozone with associated proposal' do
+    target_geozone = create(:geozone, name: 'Delete me!')
+    proposal = create(:proposal, geozone: target_geozone)
+
+    visit admin_geozones_path
+
+    within("#geozone_#{target_geozone.id}") { click_link 'Delete' }
+
```
voodoorai2000: Here would be the perfect place to put an expectation for the flash error 'This geozone can't be deleted since there are elements attached to it'

amiedes:
Thanks a lot for the detailed review, Rai!

I made the requested changes, but I’m not sure about how to test the flash messages. I was trying to do something like this:

```ruby
expect(flash[:error]).to eq(t('admin.geozones.delete.error'))
```

But I get a `NameError: undefined local variable or method 'flash' for #<RSpec::ExampleGroups::AdminGeozones:…>`

From [here](http://stackoverflow.com/questions/30144272/rspec-flash-integration-test-error-in-rails-4-nameerror-undefined-local-variab) and [here](http://stackoverflow.com/questions/25919458/how-to-stub-a-flash-in-a-rspec-spec) I read that the flash is not accessible from feature specs, just controller ones.

I can change the feature specs into controller ones, but the other specs related to the admin panel are all written as feature specs under the `specs/features/admin` directory, so maybe it’s better to leave them as a feature to remain coherent with the other ones.

app/controllers/admin/geozones_controller.rb
amiedes: Good idea!, the controller looks much clearer like that.

voodoorai2000:
Excellent, my pleasure :)

To test the flash, you can just check for the content, something like this:
`expect(page).to have_content 'This geozone can't be deleted since there are elements attached to it'`

amiedes:
Done :)

amiedes:
Don't merge yet, I just realised now I can write tests in a better way to test the `safe_to_destroy` thing, since it's placed in the model.

amiedes:
Ok, now I think the specs are where they belong.

voodoorai2000:
Awesome! Thank you @amiedes 🎉

PR #1285, opened 2016-12-07
Accesibility

decabeza:


xuanxu:
failing specs seem legit, caused by removed links.

PR #1297, opened 2016-12-20
Fixes bug in menu icon in mobile browsers.

zjuanma:
Menu icon in mobile browsers don't work.


decabeza:
Thanks @zjuanma! 👍 

(We thought that the missing icon was an issue of the foundation-rails gem)

PR #1301, opened 2016-12-22
Remove duplicated code in Admin::BannersController

amiedes:
There's no need to explicitly write a `find_banner` method since cancancan already provides this functionality automatically when invoking the `load_and_authorize_resource` method.

Source: [cancancan](https://github.com/CanCanCommunity/cancancan/wiki/authorizing-controller-actions#load_resource)

Also, I think there's no need no call `assign_attributes(banner_params)` since that is also being done when calling `@banner.update(banner_params)`

The failed spec apparently is not related to the changes made by this commit.


xuanxu:
thanks @amiedes!

Issue #1146, opened 2016-06-11
Devise configuration


anaibol:
Hi, I am deploying the tool for the team in France on Heroku.
I wanted to advice that I had to add the secret_key config option in config/initializers/devise.rb.

Cheers!

ferblape:
Hi @anaibol, take a look to my PR https://github.com/consul/consul/issues/1303, maybe it´s interesting to you.

xuanxu:
solved vía #1303

PR #1303, opened 2016-12-23
Extract mailer from name and address to a setting

ferblape:
This PR proposes to extract to a setting the mailer from name and email address. This would be really helpful to parameterize a new Consul fork.

BTW, it also takes care of #1146 

Please, review and provide feedback.

### DIFF
--- a/app/mailers/application_mailer.rb
+++ b/app/mailers/application_mailer.rb
@@ -1,5 +1,5 @@
 class ApplicationMailer < ActionMailer::Base
   helper :settings
-  default from: "Consul <no-reply@consul.es>"
+  default from: "#{Setting['mailer_from_name']} <#{Setting['mailer_from_address']}>"
   layout 'mailer'
 end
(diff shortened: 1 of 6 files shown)


ferblape:
Hmmm, there's a problem using a Setting in an initializer when running the tests, because the table doesn't exist (it fales in the `rake db:setup` task).

I suggest something like:

```
# config/initializers/devise.rb

  if Rails.env.test?
    config.mailer_sender = "noreply@example.org"
  else
    config.mailer_sender = "#{Setting['mailer_from_name']} <#{Setting['mailer_from_address']}>"
  end

```

xuanxu:
:+1: Thanks Blat!
Can you add that change to the devise.rb initializer? 
(and now that you are at it, it would be better if the default emails in seed files are @consul.dev instead of consul.es)

ferblape:
Of course!

ferblape:
This is green in my travis! https://travis-ci.org/ferblape/consul

xuanxu:
Yay!

Issue #1315, opened 2016-12-28
Manage offices_url as a setting, instead of a locale key


ferblape:
I'm thinking that `offices_url` should be managed as a setting instead of a locale key. If you agree with that I'll send you a PR with the changes.

xuanxu:
+1

decabeza:
Done on #1318 👏

PR #1318, opened 2016-12-29
Manage offices url as a setting

ferblape:
This PR connects with #1315, and it moves the I18n key `offices_url` to a setting that can be stablished by an admin.

### DIFF
--- a/app/views/verification/letter/new.html.erb
+++ b/app/views/verification/letter/new.html.erb
@@ -50,7 +50,7 @@
       <ul>
         <li>
           <%= t("verification.letter.new.office",
-          office: link_to(t("verification.letter.new.offices"), t("verification.letter.new.offices_url"),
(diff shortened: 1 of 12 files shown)


app/views/verification/letter/new.html.erb
```diff
@@ -50,7 +50,7 @@
       <ul>
         <li>
           <%= t("verification.letter.new.office",
-          office: link_to(t("verification.letter.new.offices"), t("verification.letter.new.offices_url"),
+          office: link_to(t("verification.letter.new.offices"), Setting['verification_offices_url'],
```
xuanxu: Please use `setting['verification_offices_url']` instead. 
There is a `setting` helper method to use in views and have the settings memoized per request.

ferblape: Ok!

PR #1323, opened 2017-01-04
Don't fetch setting in dev mode

### LINKED ISSUE #1322: Can't run `rake db:setup`
This problem was introduced in #1303, using a `Setting` instance in an initializer raises an error when the table `settings` doesn't exist.

I'll send a PR.

ferblape:
This PR fixes #1322

### DIFF
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -12,8 +12,8 @@ Devise.setup do |config|
   # Configure the e-mail address which will be shown in Devise::Mailer,
   # note that it will be overwritten if you use your own mailer class
   # with default "from" parameter.
-  if Rails.env.test?
-    config.mailer_sender = "noreply@example.org"
+  if Rails.env.test? || !ActiveRecord::Base.connection.table_exists?('settings')
+    config.mailer_sender = "noreply@consul.dev"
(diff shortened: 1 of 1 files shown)


amiedes:
I was precisely researching about this because I was having problems with `rake db:schema:load` yesterday.

I'm just curious: how is that `Setting` can be accessed inside an initializer in production but not in development or test environments? Because production is supposed to never have to setup the database again from scratch?

kikito:
Hi there, I too had problems with this just last week (making the email dependent on a Setting was a recent change). The error I was getting was "the table 'settings' does not exist".

@ferblape , I think your fix will still fail on certain cases - In production, when someone has created an empty database, but the tables are still not created (i.e. first install on a new server). Can you change the condition to actually check that the settings table exists?

```
if Rails.env.test? || !ActiveRecord::Base.connection.table_exists?('settings')
```

Granted, it is "uglier" than checking for dev, but as I said before, I think this can affect production machines too. I don't see other way to do this.

ferblape:
Thank you for the feedback, and sorry for the problem :)

I didn´t think about production, but you are totally right! This should be fixed now.

PR #1325, opened 2017-01-04
Budgets - several fixes

kikito:
* Enables valuation during the whole process (not just the Valuation phase). Only finished budgets are closed for evaluation.
* Adds more checks when creating a ballot line (the heading must be a valid heading)
* Removes geozone from budget heading (it as a presentation-only field)
* Improves presentation in the admin and valuation UX thanks to @decabeza


decabeza:
Now the valuators only see the budget investments assigned to themselves, I think this spec is not necessary:
```
(Valuation budget investments Index shows assignments info)
rspec ./spec/features/valuation/budget_investments_spec.rb:57
```

PR #1329, opened 2017-01-05
Track email digests

### LINKED ISSUE #1192: Track email digests
We have a rake task to send email digests about author notifications. Users receive an email with the notifications of proposals that they have supported.

We should setup some tracking code to know how many email digests have been sent, and in case there is an exception sending a digest, record it and know which users have not received their email digest.

amiedes:
Resolves #1192


amiedes:
Is this what you had in mind? Maybe it'll be useful also to create another task to periodically try to send again the failed digests.

BTW, is it necessary to test rake tasks?

voodoorai2000:
Thank you @amiedes :)

Sure this great to run in a staging environment and make sure all emails will be send correctly 👌 

It's usually a little cumbersome and slow to test rake tasks. A good approach is to make the rake task call a class/instance method and test that method instead ;)

PR #1335, opened 2017-01-09
Minimum age setting

xuanxu:
This PR removes current (16 years) hardcoded age restriction and moves minimum age to participate to a setting accesible via `User.minimum_required_age` method that defaults to 16.


voodoorai2000:
👌

PR #1344, opened 2017-01-11
More flexibility on tags administration, tags can be checked as featu…

fandrod:
More flexibility on tags administration, tags can be checked as featured on debates or/and as categories on proposals.
Featured tags are proposed during debates creation/edition.


voodoorai2000:
Thank you guys!

We're sorry this feature has become obsolete and needs some cleaning up. Tags created in this interface are not really displayed in Debate and Proposal's forms. What is really displayed are Category Tags created manually or through [seeds](https://github.com/consul/consul/blob/master/db/dev_seeds.rb#L118).

PR #1346, opened 2017-01-13
Use PhantomJS version 2.1.1 in TravisCI

amaia:
As described here https://github.com/travis-ci/travis-ci/issues/3225#issuecomment-200965782

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,19 @@ language: ruby
 addons:
   postgresql: "9.4"
 rvm:
- - "2.3.2"
-cache: bundler
+  - "2.3.2"
+cache:
+  - bundler
+  - directories:
+    - "travis_phantomjs"
+before_install:
+  - "phantomjs --version"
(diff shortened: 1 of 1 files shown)


amaia:
@kikito as we talked yesterday, it still fails, but maybe in a different way that could be a starting point to fix the flaky tests

xuanxu:
(Travis builds relunched to check if those failing specs are flaky)

kikito:
I have found an [issue which seems to be related](https://github.com/teampoltergeist/poltergeist/issues/677#issuecomment-222919584). Apparently, the "can't reach server" error happens when your asset pipeline compilation is not fast enough - rails is trying to do it "on the fly" on the first test, and capybara times out. So the "server unreachable" error might be fixed if we compile the assets before launching the specs.

amaia:
I'm closing this one, as it doesn't solve the errors, and #1364 already improved the flaky specs thing a lot.

PR #1350, opened 2017-01-13
fixes autoloading problem by not autoloading

kikito:


kikito:
This isn't needed in Consul as there's no Ballot class here. Closing

PR #1364, opened 2017-01-19
Increase Poltergeist timeout in specs

amaia:
This is to try to avoid flaky js specs in Travis caused by timeouts


amaia:
It doesn't fix javascript errors but I think it fixes timeout errors, the ones that start with `Capybara::Poltergeist::StatusFailError`

xuanxu:
👍

PR #1384, opened 2017-02-01
Poll rake on steroids

kikito:
Adds the following rake tasks:

* `polls:danger:destroy_all` (since questions have the paranoia gem it is not as simple as one would have imagined to remove all polls)
* `polls:generate_2017_nvotes_tsv` to generate the tsv for nvotes
* `polls:add_2017_nvotes_poll_id` to set the correct tsv poll id on each generated poll

It also modifies the descriptions of the generated polls for Barajas & San Blas, standardizing the way the options are shown. And it removes all mentions of "Ford".


kikito:
closing, wrong repo for the pull request

PR #1412, opened 2017-02-09
Allow erased verified users to come back

xuanxu:
When a user verifies a document in use by an old erased user, votes from the latter are reassigned to the former.
I also allows users that voted in a poll via tablet or booth to register with their correct voting data assigned :tada:


voodoorai2000:
👏

PR #1418, opened 2017-02-14
Break admin's poll show view 

xuanxu:
Breaks the admin poll view into smaller separate pieces/requests: questions, booths, officers, recounts and specs


voodoorai2000:
👏

PR #1429, opened 2017-02-27
fix close div and close thead in admin budgets groups edit partial

struebano:
fix close div and close thead in admin budgets groups edit partial


decabeza:
Hi @struebano!

Thank you bery much for the PR! 👍

PR #1454, opened 2017-03-21
Accessibility badges

jaimeiniesta:
This PR fixes the consul logo in the README, and adds accessibility / HTML issues badges so we can be aware of issues as soon as they're detected :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,4 @@
-![Logo of Consul]
-(https://raw.githubusercontent.com/consul/consul/master/public/consul_logo.png)
+![Logo of Consul](https://raw.githubusercontent.com/consul/consul/master/public/consul_logo.png)
 
 # Consul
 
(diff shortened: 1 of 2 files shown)


xuanxu:
I vote for reverting this changes.
-The code being tested is not Consul, is the madrid fork, so at least it belongs there, not in the Consul repo.
-And having badges in red with "3000" errors I think is misleading, when the site pass AA and is being auditored for AAA certification.

jaimeiniesta:
:+1:

I can submit the PR on https://github.com/AyuntamientoMadrid/consul better.

Is there also a public website that corresponds to https://github.com/consul/consul ?

These are the accessibility rules that Rocket currently checks, using the axe-core engine:

https://rocketvalidator.com/articles/accessibility

jaimeiniesta:
@xuanxu the badges only show 14 accessibility issues, the "3000" errors are HTML issues - mostly unclosed tags etc.

PR #1479, opened 2017-04-03
adds spec

kikito:


spec/models/tag_spec.rb
```diff
@@ -0,0 +1,31 @@
+require 'rails_helper'
+
+describe Tag do
+
+  it "decreases tag_count when a debate is hidden" do
+    debate = create(:debate)
+    tag = create(:tag)
```
Ana06: why not creating the variables in `let`? It is something I generally like as then you find all the variables definitions together and always use lazy evaluation, but in the case of `tag` would also avoid having repeated code, as this variables is created twice, also for the second test. :confounded:

spec/models/tag_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    expect(tag.taggings_count).to eq(1)
+
+    debate.update(hidden_at: Time.now)
+
+    tag.reload
+    expect(tag.taggings_count).to eq(0)
+  end
+
+  it "decreases tag_count when a proposal is hidden" do
```
Ana06: This is exactly the same test as above but for proposal instead debate, what about using [shared examples](https://www.relishapp.com/rspec/rspec-core/docs/example-groups/shared-examples)? :thinking:

spec/models/tag_spec.rb
kikito: Hi! This code has been translated to consul from the Madrid branch, "as it was". Yes, using `let` here would make the code less repetitive. Do you want to send a pr :)?

Ana06: ok, then maybe adding this information to the commit message? :thinking: 

Yes, when this get merge I can send a PR to improve it. :wink:

spec/models/tag_spec.rb
kikito: We don't use shared examples in the whole project. Right now we just leave with repetition in specs (It's surprisingly easy to leave with it).

We're not however closed to adding them. We just never considered doing so.

I would however not include them just to improve these two specs alone - I'm not sure the code repetition it would save would be worth it, compared to the extra complexity of having shared examples here .

That said, if shared examples could be used in to clean the worst offenders, then that would change my mind. The specs where proposals, debates and/or comments are together are usually the worst offenders, repetition-wise. For example:

https://github.com/consul/consul/blob/master/spec/features/comments/debates_spec.rb
https://github.com/consul/consul/blob/master/spec/features/comments/proposals_spec.rb

If you find a way to share code between those two and halve that code, I would gladly accept a pr for that.

voodoorai2000:
Thank you @Ana06 !

Indeed let can be dangerous... https://robots.thoughtbot.com/lets-not
However a PR for shared examples regarding polymorphic associations like comments would be awesome!

Issue #1428, opened 2017-02-23
To add Telegram to the social sharing services


Akronix:
In addition to Facebook, Twitter and Google+, I would expect to have Telegram as well.
For this, the url to point to should be in this form:
`https://telegram.me/share/url?text=<texto>&url=[url]`
Como por ejemplo:
https://telegram.me/share/url?text=Creaci%C3%B3n%20de%20huertos%20urbanos%20en%20espacios%20municipales%20sin%20uso&url=https://decide.madrid.es/proposals/9951-creacion-de-huertos-urbanos-en-espacios-municipales-sin-uso

voodoorai2000:
Hello @Akronix!

Thank you very much for your suggestion, indeed it would be great!
At the moment we are quite busy with other features, but if you send us a PR we will happily merge it!

Cheers!

Akronix:
I've opened an issue regarding this in social-share-button:
https://github.com/huacnlee/social-share-button/issues/120

huacnlee:
Use version 0.9.0

voodoorai2000:
Done!

PR #1481, opened 2017-04-05
Add telegram

amaia:
Adds Telegram to the social share buttons
Also adds Telegram to the footer if the telegram handle is defined as a setting

Connects to #1428


decabeza:
Thanks @amaia! 🎉

Issue #1501, opened 2017-04-26
Add Administrators menu on admin panel


decabeza:
We need add a new section on `Admin > Profiles > Administrators` to allow searching for an existing user and mark/unmark the user as an administrator.

![administrators](https://cloud.githubusercontent.com/assets/631897/25428532/7062fa76-2a76-11e7-8a01-bc9efec4e818.png)

bertocq:
I'm on to this one too!

decabeza:
Closed by #1560

KrzysztofMadejski:
I still can't find an option to add a given user Administrator / Moderator / other role... Wasn't this implemented?

CC: @WojtekSanko

decabeza:
Hi @KrzysztofMadejski on admin section `/admin` navigate to `Admin > Profiles` there are some subsections where you can find any registered user and assign a new role 😌 

![screen shot 2018-07-25 at 12 16 29](https://user-images.githubusercontent.com/631897/43195232-af393dec-9004-11e8-8f2a-673c7e785870.png)

![screen shot 2018-07-25 at 12 18 20](https://user-images.githubusercontent.com/631897/43195273-d760bd4a-9004-11e8-99f7-10efc544f6ba.png)

aitbw:
To keep track of this: Whitespace trimming is fixed via #2161 but hasn't been merged into `master` yet, @KrzysztofMadejski

decabeza:
Thanks @KrzysztofMadejski I take note of your suggestions, we continue to work continuously to improve the admin panel ☺️

PR #1560, opened 2017-05-23
Administrators management in admin area

bertocq:
This PR implements https://github.com/consul/consul/issues/1501

I've just reused existing similar feature for Moderators since it seemed to be equal as requested for Administrators and inventing myself a new way to manage them didn't seem reasonable. 

Anyway I would improve this by allowing to search also by name like on the Officials manage panel. I tried to add that feature but ended up messing too much. I'm still new to this codebase and not sure what's new way and what's the old. If you consider it a nice improvement I'll give it another try.


app/controllers/admin/administrators_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  def create
+    @administrator.user_id = params[:user_id]
+    @administrator.save
+
+    redirect_to admin_administrators_path
+  end
+
+  def destroy
```
bertocq: - Should a admin be able to remove himself from the Administrator list?
- Even in the scenario where he's the only admin in the list?

I guess no on both, but could only be no for the second one...

bertocq:
As stated at https://github.com/consul/consul/pull/1562 I think broken tests (or at least most of them) was due to the time where the build was done

app/controllers/admin/administrators_controller.rb
bertocq: @decabeza Should I go on assuming both actions should be restricted? I was trying not to assume complete specs based on my opinion.

decabeza: @bertocq both sounds perfect, you can restrict them!👌

bertocq: Okey added that scenario to the spec https://github.com/consul/consul/pull/1560/commits/e9ea72113b30968e46225315e5e42ac97caea46b and simple current user check on the controller https://github.com/consul/consul/pull/1560/commits/fe5ddd019bc66fc95dffa8233b31b4552bcdf5a6

Now if the current user tries to remove himself from the administrators list he will get:
screen shot 2017-05-25 at 11 31 08

Any more details that could be polished? :)

bertocq:
Hum that [failing spec](https://travis-ci.org/consul/consul/jobs/235952055) seems like a flaky test to me :(. Created a separate PR https://github.com/consul/consul/pull/1566 to fix it and avoid polluting this one. If you agree that may be the cause... can someone retry this branch travis build? :(

app/controllers/admin/administrators_controller.rb
decabeza: @bertocq it looks perfect! :wink:

xuanxu:
🎉

PR #1562, opened 2017-05-24
Always use Time with zone

bertocq:
#### Where
Yesterday close to midnight in Spain I was getting again and again some test failing and [Travis also](https://travis-ci.org/consul/consul/jobs/235414586). This was bugging me big time because my PR changes wasn't anywhere near close to related code:
 
```
  1) voter save sets demographic info
     Failure/Error: expect(voter.age).to eq(30)

       expected: 30
            got: 29

       (compared using ==)
     # ./spec/models/poll/voter_spec.rb:101:in `block (3 levels) in <top (required)>'
```

#### What

This morning I gave it a try again, and same spec was always passing, so I dig up a big and found `Time.now.utc.to_date` at the Age module to calculate the a User's age in years. Since the test was creating the user age with `30.years.ago` (that uses Time.zone.now and my zone is not exactly UTC but UTC+2) the difference was enough to calculate the age badly.

#### How

Using always `Time.current` that [aliases to Time.zone.now](https://apidock.com/rails/Time/current/class), would fix this issues as well as make Time usage consistent across the codebase.


bertocq:
The failing test passes individually in localhost :(, I'll take a look later anyway just in case

Issue #1474, opened 2017-03-28
Thanks page after creating a proposal


decabeza:
Now when user create a new proposal only appears a simple alert and the view: `proposal / show`

![new_proposal_alert](https://cloud.githubusercontent.com/assets/631897/24415408/ad0cb5e4-13e1-11e7-837d-2c99dcbb160a.png)

The idea is to create a new page before, inviting the user to share the new proposal, and for example add a link to a more information page to improve the campaign:

![new_proposal_page](https://cloud.githubusercontent.com/assets/631897/24415542/112daae2-13e2-11e7-877a-ed7a5226c09a.png)

bertocq:
I think I'm going to adventure with this one as well.

Just to be clear upfront, I imagine this to be implemented by creating a `/proposals/:id/share` route/page that could be reusable in other scenarios (instead of a thank you one). And will hold that "¡Has creado una nueva propuesta!" only if you come from the creation action. If that doesn't sound like the idea just ping me :)

bertocq:
@decabeza so far I've been able to get:

![screen shot 2017-05-30 at 14 40 14](https://cloud.githubusercontent.com/assets/983242/26585792/840c6ffc-454d-11e7-86a9-f2b0b30976d8.jpg)

Althought the `También puedes ver más información de como mejorar tu campaña.` text is not present because I'm not sure which url/path should be used on the link.. its a new page or an existing one?

Code is ready but I've to tackle 9 spec scenarios that are broken because of the change in behaviour after proposal is created, plus create new ones for this share page.

What do you think? :)

decabeza:
Hi @bertocq! I like the idea of reuse the `/proposals/:id/share` url 🎉 

Don't worry about the link of "ver más información" it's only for Madrid's fork (https://decide.madrid.es/mas-informacion/kit-decide)

I think it's better maintain the text `Not now, go to my proposal` like a simple link instead a button. The important CTA in this page should be share the proposal. If you can change this detail and... send a PR! 👏

bertocq:
@decabeza you're right, reverted from button to just a link 👍 

Also I've made the "ver más información" link a configurable Settings option using something like `Setting['proposal_improvement_url'] = 'more-information/proposal-improvement'`. On both seeds.rb and dev_seeds.rb files its will be defaulted to `nil`, so it won't be visible... only at those forks that give a valid value. I'll add specs to both scenarios.

So right now with those changes (and the `proposal_improvement_url` present) looks like:

[...]

PR #1578, opened 2017-05-30
Extract social share buttons to shared partial view

bertocq:
## What
I'm almost done with a PR for https://github.com/consul/consul/issues/1474 but I realized the social share buttons where on many views with exactly the same html code, just changing 3 variables.

In order to make smaller and more comprehensive PR's I've extracted the partial refactor to this PR. And will open the PR that will tackle the issue after this one is merged (assuming there is no drawback to this refactor 😄  )

## How
Just creating a shared partial, and using it on every view that had the same code, passing as variables:
- Share Title: (the title used over the social buttons), seems to be the same on translation files, but I didn't want to go that further refactoring this time.
- Item Title
- Item url

### DIFF
--- a/app/views/budgets/investments/_investment_show.html.erb
+++ b/app/views/budgets/investments/_investment_show.html.erb
@@ -107,16 +107,11 @@
           </p>
         </div>
       <% end %>
-      <div id="social-share" class="sidebar-divider"></div>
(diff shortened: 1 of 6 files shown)


xuanxu:
👌

Issue #1472, opened 2017-03-27
Add rake task to update per page code


voodoorai2000:
To solve accessibility issues we have split the Setting's variable per_page_code into per_page_code_head and per_page_code_body.

We should add a rake task to migrate the data from per_page_code to per_page_code_head

bertocq:
I'm not 100% sure about the tasks to be done, does it mean modifying all files with pattern `locales/settings.**.yml`? There is no `per_page_code` in the codebase. I think the context I'm missing is how Settings work.

bertocq:
Okey after a week in this codebase I've learned enough to be able to understand this issue :D. Doing it in a few!

bertocq:
Github is funky today.. but I got the PR created anyway :D https://github.com/consul/consul/pull/1584

PR #1584, opened 2017-05-31
Rake task to migrate Setting per_page_code to per_page_code_header

bertocq:
This resolves https://github.com/consul/consul/issues/1472

Since I'm not fully aware of all the Setting "scenarios" that any consul installation could have, I've just thought about all the 4 posibilities:

1. **Neither `per_page_code_head` or `per_page_code` Settings exist:** Just create `per_page_code_head` with an empty value.

2. **Both `per_page_code_head` or `per_page_code` Settings exist:** Just delete `per_page_code`

3. **`per_page_code_head` exist, but `per_page_code` doesn't:** 👌  nothing to be done.

4. **`per_page_code_head` doesn't exist, but `per_page_code` does:** The scenario from the issue... just rename `Setting['per_page_code']` key to `per_page_code_head` to keep the same value and remove the old key at the same time.

And reflectet them on 4 scenarios in a new settings rake spec. Maybe bit overkill just for a rake task that would be used once? 😄

### DIFF
--- a/lib/tasks/settings.rake
+++ b/lib/tasks/settings.rake
@@ -1,0 +1,11 @@
+namespace :settings do
+
+  desc "Changes Setting key per_page_code for per_page_code_head"
+  task per_page_code_migration: :environment do
+    per_page_code_setting = Setting.where(key: 'per_page_code').first
+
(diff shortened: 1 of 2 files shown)


voodoorai2000:
👏 👏

PR #1617, opened 2017-06-07
Remove deceptive link in Gemfile_custom

artofhuman:
CUSTOMIZE_EN.md not exists


bertocq:
Nice catch @artofhuman! I can create that missing file by translating the Spanish version. Maybe that would be a better solution, what do you think?

artofhuman:
Yes, its more better solution, i close this pr

bertocq:
Ok :) you got it at https://github.com/consul/consul/pull/1618, keep in mind my english was learnt at spain entirely so I might have typos ^_^

Issue #1489, opened 2017-04-11
Admin section for existing users


voodoorai2000:
We should add an admin section to find and visualize existing users.

Users should be searchable by username, email and DNI.

The following information should be displayed about users:
- Username
- Email
- DNI
- Verification Level
- Roles (admin/moderator/valuator/officer)
- State (bloqued/not bloqued)

voodoorai2000:
Closed via #1560 💃

PR #1631, opened 2017-06-09
Admin section for users

amaia:
This PR implements #1489 

It only shows non-blocked users


voodoorai2000:
Awesome 😌
Thank you @amaia!

PR #1635, opened 2017-06-11
Issue#1354 suggest on investments creation

entantoencuanto:
## Summary

* Some changes has been made on `CommentableActions` and `Polymorphic` concerns to allow filter items to be suggested. With debates and proposals the search is sent directly to the model, but investments are nested resources of budgets, so suggested investments should be selected among those who belong to the budget.

* The original link creation in the partial `shared/suggest` failed with investments due to routes nesting. Also there is a helper method `namespaced_budget_investment_path` to build the path to investments that is used in lists. I've decided to allow the use of custom methods to build the path to resources in the partial using an instance variable and I've used the mentioned helper method.

* The same custom filter used in `index` controller action to show investments is applied for suggestions. In this way there is no difference between what the suggestions counter shows when it exceeds 5 results and the index to which the associated link leads.

* I've only included i18n labels for es and en locales.

### DIFF
--- a/app/controllers/budgets/investments_controller.rb
+++ b/app/controllers/budgets/investments_controller.rb
@@ -24,6 +24,7 @@ module Budgets
 
     invisible_captcha only: [:create, :update], honeypot: :subtitle, scope: :budget_investment
 
+    helper_method :resource_model, :resource_name
     respond_to :html, :js
 
     def index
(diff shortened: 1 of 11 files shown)


voodoorai2000:
Great! Thank you @entantoencuanto :))

PR #1636, opened 2017-06-11
Use table name from model instead custom method

artofhuman:
When we use custom method we do not consider resource model may be
with namespace, for example for model
`Consul::Conversations::Conversation` method table_name returns
wrong result `consul/conversations/conversation`


bertocq:
@artofhuman you closed PR by mistake or you discover a way around without changing code?

artofhuman:
I have troubles because I created column in `tags` table as `conversations_count`, i rename column to all namespace `consul/conversations/conversations_count` and all works, but its so strange behavior to have column named with slashed :)

PR #1637, opened 2017-06-11
Set repo_url and branch for deploy from deploy secrets

artofhuman:


artofhuman:
This change needed to deploy application from another git repo without change main deploy config

javierm:
@artofhuman Sorry for the late reply! :pray: It looks like this change would break existing deployment configurations, since in these cases `deploysecret(:repo_url)` will be `nil` :thinking:.

Regarding the branch, you could also use what we use for the staging environment: `set :branch, ENV["branch"] || :master` and then deploy with `brach=my_branch cap production deploy`. We haven't done this change ourselves because forcing the `master` branch is safer, and we think default options in CONSUL should be as safe as possible; power users like you can easily change their `config/deploy/production.rb` to suit their needs :wink:.

Regarding the repo, we'd rather encourage everyone to include a reference to their repo in version control, in the spirit of promoting AGPL compliance (although it's not directly related). Moving it to the secrets file could give the idea that the repo location should be confidential :thinking:.

PR #1638, opened 2017-06-11
Issue#1595 results url permissions

entantoencuanto:
## Summary
Since Result model doesn't inherits from `ActiveRecord`, the budget cannot be accessed from the `Ability` class, so something like:
```ruby
can [:create, :show], Budget::Result, budget: { phase: "finished" }
```
or
```ruby
can [:create, :show], Budget::Result do |result|
  result.budget.finished?
end
```
won't work. Instead we can ask permissions to instanciated `@budget` in actions of results controller.


voodoorai2000:
👌

Issue #1575, opened 2017-05-29
Tag administration


voodoorai2000:
## What
Redo the admin tag's section to be able to CRUD tags that will be displayed when creating a debate, proposal or investment.

## Why

The current implementation has become obsolete and the tags created there have no effect on the tags displayed when creating a debate, proposal or investment

## How

Only displaying "tags" with the value "category" for the attribute "kind" in the admin section.
Storing this attribute when creating new tags from the admin section.

voodoorai2000:
Closed via #1640

PR #1640, opened 2017-06-12
Issue #1575 tag administration

entantoencuanto:
Solves #1575 

After this, some occurrences of `ActsAsTaggableOn::Tag.where(kind: "category")` and `kind`  attribute assignations can be replaced with the most compact form `ActsAsTaggableOn::Tag.category`


spec/features/admin/tags_spec.rb
```diff
@@ -3,16 +3,21 @@
...
   end
 
   scenario 'Index' do
     create(:debate, tag_list: 'supertag')
+    debate_tag = ActsAsTaggableOn::Tag.find_by(name: "supertag")
     visit admin_tags_path
 
     expect(page).to have_content @tag1.name
-    expect(page).to have_content 'supertag'
+    if debate_tag.category?
```
bertocq: @entantoencuanto I can't understand the usage of this if, could you please explain a bit? Thanks! :)

entantoencuanto: Well, I must admit that I've made a weird test. In background of this spec `@tag1` is created, a tag of kind "category". This tag is expected to be managed in admin/tags controller. In the "Index" scenario another tag named "supertag" is created, associated with a debate. Before the changes of this branch, the test was expecting to find also "supertag" in the index list. But now, the "supertag" won't be listed since the debate's tags are created with `kind` attribute as `nil`, not "category". I'm not sure if in the futute the creation of tags associated with debates or other taggable models should be changed to include the "category" kind by default, so I've tried to prepare the test to both cases: when the tags of a new debate are categories and when not. 

I think I made things more complicated. I'll simplify it and make it clearer.

entantoencuanto:
@voodoorai2000, I have some questions about the featured obsolete trait. I can see that featured attribute is used in admin/tags controller to order tags and can be set in the edit form, but doesn't seem to be used in other parts of the applications, like in proposals or budget/investments categories partial. Maybe I should remove the functionality related with featured in the admin/tags part? Should the featured column in the tags table be removed?

voodoorai2000:
Awesome @entantoencuanto 

Yes, the _featured_ attribute was initially used to define _category_ tags.
With your PR, the _featured_ attribute has become obsolete. 

As far as we know all forks are already using the _category_ tags and not the _featured_ tags so it shouldn't be a problem to remove this attribute and it's associated functionality in admin/tags.

entantoencuanto:
I have removed the strange conditional from the test, it is totally unnecessary.

Also I've eliminated the featured attribute on tags table and its management in admin/tags. The update action in the controller and link in view has been removed since it only served to edit the featured field.
 
Sorry for the long ae9030e commit message! I've added it by mistake, I can amend the message and push again the branch if you wish.

voodoorai2000:
Thank you @entantoencuanto 👌

Issue #1588, opened 2017-06-02
Display only active dates


voodoorai2000:
In the list of processes, display for each process only the dates of the parts of the process that are marked as active.

User interface:
![legislation dates](https://cloud.githubusercontent.com/assets/4169/26720569/188bade0-4789-11e7-8b00-cf119bfcb098.png)

Admin section:
![active dates](https://cloud.githubusercontent.com/assets/4169/26720576/1ce243ea-4789-11e7-9336-8849827f3559.png)

bertocq:
@voodoorai2000 either I don't get the specs or I see this working:

![screen shot 2017-06-02 at 17 44 56](https://cloud.githubusercontent.com/assets/983242/26733496/4972865e-47bb-11e7-814e-81398a63f051.jpg)
![screen shot 2017-06-02 at 17 45 07](https://cloud.githubusercontent.com/assets/983242/26733497/4975c544-47bb-11e7-9f86-14960c7d1a48.jpg)
![screen shot 2017-06-02 at 17 45 15](https://cloud.githubusercontent.com/assets/983242/26733495/497267dc-47bb-11e7-9e53-dbce16997f83.jpg)

Do you mean that the blocks like "Draft publication" or "Allegations" shouldn't even be shown? Dates are being already hidden when not active.

bertocq:
BTW Not sure if bug or feature, but after clicking the "Active" button on the admin panel to disable a process, the dates are removed, and clicking again to enable

PR #1649, opened 2017-06-13
Refactor legislative process phases and publications

bertocq:
## What
While doing a PR for the ticket https://github.com/consul/consul/issues/1588 I learned about the current status of the legislative process phases and publications and seemed to me that [this three open/active/show methods](https://github.com/consul/consul/blob/master/app/models/legislation/process.rb#L22-L64) where pretty similar

## How
Extracted the similar logic to two simple classes `Legislation::Process::Phase` and `Legislation::Process::Publication` just so they are the ones that know if they are:
- **open?** (instead of open_phase?)
- **started?** (instead of show_phase?): The word "show" seems like a view logic, the fact that the view needs is just if the phase has started or if the publication date have arrived. But its a subjective matter open to discussion 😃 
- **enabled?** (instead of active_phase?): For me active could be interpreted like started.. but enabled fully describes the fact that an admin could enable/disable that phase/publication. Again more subjective matter 😄 

Also the `final_publication` concept didn't seem very descriptive to me, so I replaced it for `result_publication`. But that's subjective so I'm open to change it if no-one else finds that as an improvement.

## Screenshots
Not much really to see, only visually test that any of the process phase/publications have gone away :).

## Tests
- Deleted the deprecated process_spec tests for old methods
- Added new tests (reusing deleted ones) for new Phase and Publication classes

## Deployment
As usual. This PR has a migration that renames a column, that should not generate any issue.

## Warnings
I'm based my work for issue https://github.com/consul/consul/issues/1588 on this branch resulting in this PR https://github.com/consul/consul/pull/1652. Maybe too soon ? :D I can always apply on both branches any changes/suggestions though.


voodoorai2000:
Beautiful!

PR #1650, opened 2017-06-13
Cleanup dev_seeds.rb file

bertocq:
What & How
====
Doing some work for another issue, I was reading the db/dev_seeds.rb file and I just started to cleanup some extra spaces here, some extra lines there... 🙊 and endedup doing a whole file cleaning with a bit of help of rubocop 🙈🤖.

The most common cleanup was for example `(1..10).each do` => `10.times do`. That I think reads quicker and easier.

Screenshots
===========
Really nothing to see, no need.

Test
====
Well I ran some times the `rake db:dev_seed` rake task and everything was 👍

Deployment
==========
As usual

Warnings
========
This is a bit of rubocop and a bit of mine own take on code legibility, so for obviously totally open to other opinions and suggestions 😃

### DIFF
--- a/db/dev_seeds.rb
+++ b/db/dev_seeds.rb
@@ -79,7 +79,7 @@ poll_officer = create_user('poll_officer@consul.dev', 'Paul O. Fisher')
 poll_officer.create_poll_officer
 
 level_2 = create_user('leveltwo@consul.dev', 'level 2')
-level_2.update(residence_verified_at: Time.current, confirmed_phone: Faker::PhoneNumber.phone_number, document_number: "2222222222", document_type: "1" )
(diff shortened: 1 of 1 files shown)


xuanxu:
All the ranges `(1..X).each` are used instead of `X.times`to avoid index 0. 
With this change some things will not be affected but others will not work (when creating officials there is not _Level 0_) and some would be confusing: user0 to user39 instead of user1 to user40. 
I'd keep them all in ranges  for consistency.

bertocq:
You're right @xuanxu ! I didn't realize that when the index was used :(. I've fixed the commit and force pushed it again :)

PR #1634, opened 2017-06-11
Fix Flaky test related with Time/Date in Search concern

bertocq:
## Where

I've been getting the following flaky tests failures at night for weeks:

screen shot 2017-06-11 at 10 41 24
screen shot 2017-06-11 at 10 41 17

<details>
  <summary>Click to expand text version</summary>
<code>
   1) Proposals Search Advanced search Search by date Search by multiple filters
     Failure/Error: expect(page).to have_content("There is 1 citizen proposal")
       expected to find text "There is 1 citizen proposal" in "Consul Language: Transparency Open data Consul Sign in Register Debates You are in Proposals Voting Legislation processes Participatory budgeting More information Searcher Search Search Proposals Search results citizen proposals cannot be found Advanced search With the text By author category By date most active highest rated newest relevance Archived Create a proposal CATEGORIES TRENDING DISTRICTS TOP WEEKLY The most supported proposals by category RETIRED PROPOSALS Proposals retired by the author Open government This portal uses the Consul application which is open-source software. For technical assistance enters technical assistance Participation Decide how to shape the city you want to live in. Transparency Find out anything about the city. Open data Every detail about the City Council is yours to access. Consul, 2017 | Privacy Policy | Terms and conditions of use | Accessibility"
     # ./spec/features/proposals_spec.rb:1079:in `block (5 levels) in <top (required)>'
  2) Proposals Search Advanced search Search by date Predefined date ranges Last day
     Failure/Error: expect(page).to have_content("There are 2 citizen proposals")
       expected to find text "There are 2 citizen proposals" in "Consul Language: Transparency Open data Consul Sign in Register Debates You are in Proposals Voting Legislation processes Participatory budgeting More information Searcher Search Search Proposals Search results There is 1 citizen proposal Advanced search With the text By author category By date most active highest rated newest Archived Proposal 410 title No comments • 2017-06-10 • Manuela2283 In summary, what we want is... 473 0% / 100% No supports 53,726 supports needed Support Create a proposal CATEGORIES TRENDING DISTRICTS TOP WEEKLY The most supported proposals by category RETIRED PROPOSALS Proposals retired by the author Open government This portal uses the Consul application which is open-source software. For technical assistance enters technical assistance Participation Decide how to shape the city you want to live in. Transparency Find out anything about the city. Open data Every detail about the City Council is yours to access. Consul, 2017 | Privacy Policy | Terms and conditions of use | Accessibility"
[...]

PR #1651, opened 2017-06-13
Fix DirectMessage today scope

bertocq:
## What
DirectMessage `today` scope was comparing dates without taking into account that `created_at` is a timestamp with timezone.

## How
Database stores created_at as timestamp with the timezone, so when comparing DATE(created_at) to something we have to convert it to DATE as well with the postresql native function, but using Time.current instead of Date.current to take into account the user timezone

## Screenshots
No need

## Tests

Modified to use utc timezone when directly changing the `created_at` attribute.

## Deployment

As usual

## Warnings

This is related with my last try to fix flaky tests https://github.com/consul/consul/pull/1634 but on that PR I just tried to use timezone, but still kept using Date instead of Time

### DIFF
--- a/app/models/direct_message.rb
+++ b/app/models/direct_message.rb
@@ -8,7 +8,7 @@ class DirectMessage < ActiveRecord::Base
   validates :receiver, presence: true
   validate  :max_per_day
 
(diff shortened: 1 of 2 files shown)


voodoorai2000:
👍

PR #1652, opened 2017-06-13
Legislation Process Date display & Enabling/Disabling

bertocq:
Where
=====
* **Related Issue:** https://github.com/consul/consul/issues/1588
* **Related PR's:** https://github.com/consul/consul/pull/1649

What
====
Display for each process only the dates of the parts of the process that are marked as active.
+1 to maintain dates when clicking on the checkbox

How
===
- Firsly I did this refactor https://github.com/consul/consul/pull/1649 to cleanup a bit the phase/publication concepts and usage

- I created a `*_enabled` boolean column for each phase and publication dates https://github.com/consul/consul/commit/bae9d105fa4c22869afbe9b1d9996c1b97ec018d

- With the new `*_enabled` attributes I changed the `enabled?` logic for both Phase and Publication classes to no longer check for date existence, but just for the enabled flag https://github.com/consul/consul/commit/cb7358bb974f1009be7035b0256db0e13c92fd73. This will allow the user to keep the phase/publication dates even when its disabled.

- Finally added the new enabled flags to the admin panel form and used the new `enabled?` on the process list to display only those dates enabled.

- Yes, I renamed from "active" to "enabled" as for me it communicates better to both dev and UI user the intent/feature.. to avoid confusion with the concept of "started" or "we're on the date range of the phase/publication"

Screenshots
===========
**User facing views for processes with some disabled phases / publications**:
screen shot 2017-06-13 at 23 53 17

**Admin panel with `Enabled` checbox**:
screen shot 2017-06-13 at 23 56 59

Test
====
- Increased coverage for the scenario when a date is disabled and should not be shown to user
- Fixed specs based on old enabled logic

Deployment
==========
As usual, only new columns on migrations, and defaulted to false. I assume no other consul installations are working with Legislation Process right now (its a feature just announced 12h ago). But just manually checking the enabled checkboxes for the dates of any already created process would fix any issue.

Warnings
========
This PR is based on the refactor done at https://github.com/consul/consul/pull/1649 ... that's why the diff is so big :(, maybe better to check that one first?


voodoorai2000:
:ok_hand:

Issue #1647, opened 2017-06-13
Locale site customization


decabeza:
## Background

In Admin site customization section `/admin/site_customization/pages` you can create custom info pages. There is an option to show them on `/more-information` page (inside the **Other information of interest** section).

<img width="277" alt="show_on_more_info" src="https://user-images.githubusercontent.com/631897/27090210-e18bc71c-505c-11e7-93ca-37d47f59b410.png">

Now you can create pages in any language using the WYSIWYG editor but you can't select the locale used in the page.

## What we need

- [x] Add a select input for available locales.
- [x] On `/more-information` show only the pages written on selected locale.

With this you can create for example a proposals extra info page and write it in different languages using differents slugs:

English: `proposal-extra-info`
Spanish: `propuestas-informacion-adicional`

Maybe also you want some info pages are showed only in a specific language.

## Extra info links

- [Site customization PR](https://github.com/consul/consul/pull/1484)
- Some `/admin/site_customization/content_blocks` examples:
    - [site_customization/content_block.rb](https://github.com/consul/consul/pull/1484/files#diff-fad3cc5c8309d80c10f2ad32836b1c7cR4)
    - [site_customization/content_blocks/_form.html.erb](https://github.com/consul/consul/pull/1484/files#diff-f1bc7d3446224099f062ffed77f4c342R23)

MariaCheca:
I'm taking this ticket! :) cc @voodoorai2000

voodoorai2000:
Closed via #1716

PR #1657, opened 2017-06-14
Allow overriding faq and how-to-use pages with custom pages

amaia:
This PR changes the views files for `more-information/faq` and `more-information/how-to-use` to allow overriding its content with a custom page created from the admin panel. 

If there are no custom pages with these slugs created from the admin panel everything should work as before.

### DIFF
--- a/app/models/site_customization/page.rb
+++ b/app/models/site_customization/page.rb
@@ -3,7 +3,7 @@ class SiteCustomization::Page < ActiveRecord::Base
 
   validates :slug, presence: true,
                    uniqueness: { case_sensitive: false },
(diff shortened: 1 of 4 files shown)


xuanxu:
@amaia thanks for the PR!, but we are solving this differently: We want to [add locale to the custom pages](https://github.com/consul/consul/issues/1647), that way we can have /faq and /preguntas-frecuentes and show the apropiate one.
Because with the changes in this PR we lose the i18n of the faq/how-to-use pages.

amaia:
@xuanxu ok! 👍

amaia:
@xuanxu would you accept a PR with just the bit in `app/models/site_customization/page.rb` that allows slashes in slugs?

xuanxu:
@amaia sure!

PR #1660, opened 2017-06-15
Fix some Rubocop issues

bertocq:
## What

I'm going to slowly trim down the [rubocop_todo](https://github.com/consul/consul/blob/master/.rubocop_todo.yml) issues list

## How

Using `rubocop -a --only NAME_OF_THE_ISSUE` on the console and manually reviewing any autocorrection, or doing it myself

## Warnings

Maybe this PR has gone too big (27 files changed). I've done a commit for each rule, to make it easier to see each rule changes and even remove one of them in case it causes some debate

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -14,6 +14,7 @@ Lint/InheritException:
   Exclude:
     - 'app/controllers/concerns/feature_flags.rb'
 
+<<<<<<< HEAD
 # Offense count: 13
 Lint/ParenthesesAsGroupedExpression:
   Exclude:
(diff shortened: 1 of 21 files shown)


.rubocop_todo.yml
```diff
@@ -45,11 +14,7 @@ Lint/InheritException:
   Exclude:
     - 'app/controllers/concerns/feature_flags.rb'
 
-# Offense count: 7
-Lint/NestedMethodDefinition:
-  Exclude:
-    - 'lib/acts_as_paranoid_aliases.rb'
-
+<<<<<<< HEAD
```
voodoorai2000: There is little conflict here

bertocq: Oh crap, fixing!

voodoorai2000:
👏

PR #1664, opened 2017-06-18
Change log level in production to warn

artofhuman:


bertocq:
Hello @artofhuman , as I said on your other PR, Raimond is on holiday leave :) so no answer from him for a while. And in this particular PR, I think you're totally right, but as my experience on the codebase is not longer than some months, I would rather wait for Raimond to comeback and share any knowledge that might be behind that production debug logging level before merging this PR.... just in case :). Is that ok for you? 

Thanks again for your contributions 👍 😃

artofhuman:
> Is that ok for you?

Yes ;)

voodoorai2000:
Yes, this is great @artofhuman, appreciate it!

PR #1670, opened 2017-06-21
Add Budget Investments filter by winners

bertocq:
What
====
- Budget Investments list needed a filter to see the Winner ones, as well as a column with that value for other filters

How
===
- Adding a new winners scope on Budget::Investment model
- Adding a new filter on both controller and view for admin budget investments list
- Adding a Winner column on admin budget investments list with yes/no values

- Adding translations for all new labels

Screenshots
===========
### Budget Investments Winners filter:
screen shot 2017-06-21 at 00 16 57

Test
====
- Increased the Budget spec test to cover the new 'Winners' filter in the Investments list

Deployment
==========
As usual

Warnings
========
None

### DIFF
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -291,6 +291,12 @@ FactoryGirl.define do
       selected true
       feasibility "feasible"
       valuation_finished true
+
+    end
+
+    trait :winner do
+      selected
+      winner true
     end
 
     trait :unselected do
--- a/spec/features/admin/budget_investments_spec.rb
+++ b/spec/features/admin/budget_investments_spec.rb
@@ -463,6 +463,7 @@ feature 'Admin budget investments' do
     let!(:feasible_bi)    { create(:budget_investment, :feasible, budget: @budget, title: "Feasible project") }
     let!(:feasible_vf_bi) { create(:budget_investment, :feasible, :finished, budget: @budget, title: "Feasible, VF project") }
     let!(:selected_bi)    { create(:budget_investment, :selected, budget: @budget, title: "Selected project") }
+    let!(:winner_bi)      { create(:budget_investment, :winner, budget: @budget, title: "Winner project") }
 
     scenario "Filtering by valuation and selection" do
       visit admin_budget_budget_investments_path(@budget)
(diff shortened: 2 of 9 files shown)


spec/features/admin/budget_investments_spec.rb
```diff
@@ -532,14 +543,14 @@
       visit admin_budget_budget_investments_path(@budget)
       within('#filter-subnav') { click_link 'Selected' }
 
-      expect(page).to have_content('There is 1 investment')
+      expect(page).to have_content('There are 2 investments')
```
bertocq: This change and the next one its because the whole context uses a @budget reference :(. Some day I'll go down the specs folder and make each scenario as self-sufficient and isolated as possible.

spec/features/admin/budgets_spec.rb
```diff
@@ -170,5 +170,27 @@
       end
     end
 
+    scenario 'Calculate Heading winners', :js, :winners do
+      budget = create(:budget, name: 'Yearly participatory budget')
+      group  = create(:budget_group, budget: budget, name: 'Districts improvments')
+      heading  = create(:budget_heading, group: group, price: 5)
+      unselected_investment = create(:budget_investment, :unselected, heading: heading, ballot_lines_count: 3, price: 4)
```
bertocq: Winner selection is simple: Take all selected investments, in ballot_lines_count order, until we run out of price :D (classic [Knapsack problem](https://en.wikipedia.org/wiki/Knapsack_problem))

spec/factories.rb
```diff
@@ -291,6 +291,12 @@
       selected true
       feasibility "feasible"
       valuation_finished true
+
+    end
+
+    trait :winner do
+      selected
```
bertocq: This got me a bit confused, but `selected` is both an attribute of budget investment, and a trait of its factory :) So yes it works!

app/views/admin/budgets/_group.html.erb
```diff
@@ -53,13 +54,19 @@
     <!-- /. new heading form -->
     <!-- headings list -->
     <% headings.each do |heading| %>
-    <tr>
+    <tr id="group-<%= group.id %>-heading-<%= heading.id %>">
```
bertocq: Just for testing purposes >_>

voodoorai2000:
👍

Issue #1601, opened 2017-06-02
Generate budget results from admin


xuanxu:
Budget results are calculated when running the `Result#calculate_winners`method, but right now it has to be executed via terminal or cron task.

We need to add a button so admins can re/generate results from the admin panel once a budget is closed.

bertocq:
Well I know enough now to tackle this one in a few hours, mine! :)

bertocq:
Done with  #1678

PR #1678, opened 2017-06-21
Add button to Calculate Budget Winner investments

bertocq:
Where
=====
* **Related Issue:** https://github.com/consul/consul/issues/1601

What
====
We need a button to call the Budget::Result#calculate_winners on each Budget's Heading.

Only admins can use it, and only if the Budget is reviewing ballots.

How
===
Adding a new `calculate_winners` method on the Admin Budget Controller that calls the method for each heading, and then redirects to the Budget's Investment list, filtering by "winner" status.

Screenshots
===========
#### The button is at the bottom of the page, aligned to the right to prevent miss clicks:
screen shot 2017-06-21 at 19 46 29

Test
====
Added new scenario on adming budget spec feature to check both reviewing ballots and finished budgets behaviour.

Maybe there is a bit of redundancy for the reviewing ballots budget scenario since I'm checking the winner investments in the list... and the Budget::Result#calculate_winners [already has a unit test](https://github.com/consul/consul/blob/master/spec/models/budget/result_spec.rb). But removing those checks the test would just see if the button was available... no other way to check if the method was called or not.

Deployment
==========
As usual

Warnings
========
The button placement its a bit hidden, oncoming documentation will have to reflect this in order to make it usable.

### DIFF
--- a/app/controllers/admin/budgets_controller.rb
+++ b/app/controllers/admin/budgets_controller.rb
@@ -14,10 +14,15 @@ class Admin::BudgetsController < Admin::BaseController
     @budget = Budget.includes(groups: :headings).find(params[:id])
   end
 
-  def new
-  end
+  def new; end
 
-  def edit
+  def edit; end
+
+  def calculate_winners
+    return unless @budget.balloting_process?
+    @budget.headings.each { |heading| Budget::Result.new(@budget, heading).calculate_winners }
+    redirect_to admin_budget_budget_investments_path(budget_id: @budget.id, filter: 'winners'),
+                notice: I18n.t("admin.budgets.winners.calculated")
   end
 
   def update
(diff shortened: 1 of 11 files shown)


app/controllers/admin/budgets_controller.rb
```diff
@@ -14,10 +14,15 @@ def show
...
   end
 
-  def new
-  end
+  def new; end
+
+  def edit; end
+
+  def calculate_winners
+    @budget.headings.each { |heading| Budget::Result.new(@budget, heading).delay.calculate_winners }
```
bertocq: We're using DelayedJob here, since the `calculate_winners` hasn't been tested for its performance on production.... And its easy to imagine that a ton of Headings with a ton of Investments could draw some power from the server if not run in the background :)

bertocq:
@xuanxu something I'm still not clear about.... Should `calculate_winners` be called when the Budget status changes to "finished" automatically? Or the first time should always be a manual action?

bertocq:
We have to swithc the condition budget.finished? for budget.reviewing_ballots?

PR #1679, opened 2017-06-21
Add Population field to Budget Headings

bertocq:
Where
=====
* **Related Issue:** https://github.com/AyuntamientoMadrid/consul/issues/812

What
====
For the Issue https://github.com/AyuntamientoMadrid/consul/issues/812 we'll need an optional population value for each Budget::Heading in order to be able to show stats about Census participation percentage.

How
===
- Adding a population integer column, defaulted to 0, to budget_headings table
- Adding the population field to the form, list and controller allowed params on Admin Budget Headings
- Adding texts and translations

Screenshots
===========
#### Form:
screen shot 2017-06-21 at 20 25 44

#### List after creation:
screen shot 2017-06-21 at 20 25 57

Test
====
Increased the admin budget spec feature to also check the population field on Heading creation

Deployment
==========
As usual

Warnings
========
This is a small PR in order to make stats PR smaller to review


db/migrate/20170621180611_add_population_to_budget_headings.rb
```diff
@@ -0,0 +1,5 @@
+class AddPopulationToBudgetHeadings < ActiveRecord::Migration
+  def change
+    add_column :budget_headings, :population, :integer, default: 0
```
xuanxu: I think a default to nil would be better and less problem prone, so only if present population stats are calculated. WDYT @voodoorai2000

bertocq: Problem is people may not understand on the admin panel difference between empty field and 0. So if they input a 0 assuming that's going to mean "No population" the rule may not work correctly, right?

xuanxu: yep but that can be solved with a model validation where population must be > 0

bertocq:
I've still pending:
- Add validation on Heading creation form for population field
- Add explanation on heading creation form for population field usage and restriction (either empty or >0)

bertocq:
Can't really get my head around the validation :( I think its not imprescindible for the feature to work (I can check `population&.positive?` on the stats. But for sure this need to be done eventually, issue created and moving on https://github.com/consul/consul/issues/1710

PR #1680, opened 2017-06-21
Chore/small rubocop cleanup

bertocq:
Some rubocop cleanups

What
====
Trimming the list of issues at `.rubocop_todo.yml`, bit by bit it will dissapear :)

How
===
- Removing the issue and files list from `.rubocop_todo.yml`
- Running`rubocop -a --only NAME_OF_COP`
- Reviewing manually autocorrections and warnings. Discarding those false positives and fixing those that need manual intervention

There is a commit for each Fixed cop in order to better see the changes and review them.


voodoorai2000:
👏

PR #1692, opened 2017-06-24
[WIP] Testing Crowdin translations

bertocq:
A lot of work to configure, results are bad, really impossible to get right with no error messages or support


config/locales/es-ES/general.yml
```diff
@@ -45,12 +44,10 @@ es:
       responses:
         one: 1 Respuesta
         other: "%{count} Respuestas"
-        zero: Sin respuestas
```
bertocq: WTF Crowdin? WHAT THE FUCK?

bertocq:
Crowdin still doing many things wrong.

config/locales/es-ES/general.yml
Andrulko: @bertocq Commonly we follow CLDR and English has only 2 forms: one and other. The zero form is trimmed on import and won't be included into exported files. Zero belongs to "other" plural form too, so your app will show "0 Respuestas" in Spanish. Anyway, checking with team how we can improve the current behavior

config/locales/es-ES/kaminari.yml
```diff
@@ -1,23 +1,20 @@
...
         one: entrada
         other: entradas
-        zero: entradas
       more_pages:
         display_entries: Mostrando <strong>%{first}&nbsp;-&nbsp;%{last}</strong> de un total de <strong>%{total} %{entry_name}</strong>
       one_page:
         display_entries:
           one: Hay <strong>1 %{entry_name}</strong>
           other: Hay <strong>%{count} %{entry_name}</strong>
-          zero: No se han encontrado %{entry_name}
```
bertocq: @Andrulko this is another example of `zero` that would read strangly.

Instead of "Entry %{entry_name} wasn't found" I don't know what would get the user :( But I understand it must be something configurable somehow

PR #1694, opened 2017-06-25
[WIP] Localization files refactor

bertocq:
Where
=====
* **Related Issue:** https://github.com/AyuntamientoMadrid/consul/issues/815

What
====
- We need to get translations rolling with [Crowdin project](https://crowdin.com/project/consul) for non-tech people to be able to easily contribute without going through github or code.

- We also need to re-estructure the `/config/locales` directory to hold all language files in a subdirectory for itself. We'll use [ISO 639-1 codes](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) when possible, but still support for 4 letter locales (pt-BR, es-PE, etc...)

- Custom localization files should keep working

- Remove Spanish from i18n-tasks config file, so no longer test will fail if spanish translation is not present. This file should be overwritten on each fork based on the languages it specifically supports.

How
===
- Switching from structure pattern `config/locales/activemodel.en.yml` on to `config/locales/en/activemodel.yml` for all languages. And reflecting that change on the `config/i18n-tasks.yml` file that keeps an eye on missing and unused translations.

- Adding a `bin/migrate_locales` executable to help on that matter for any remaining custom language any fork may have. Execute as for example `bin/migrate_locales it bg` if you have Italian and Bulgarian localization files. 

- Adding to config.i18n.load_path any nested folder files, afterwards the custom folder to let them overwrite existing ones. 

- Completing the `crowdin.yml` config file to correctly go from 4 letter locale codes on to two letter locale codes as Rails expects. This was a mayor pain when pt-BR came and overwrote pt-PT (pt) or es-PE (Peru) with es.

- Editing both Customization docs up to date.

Screenshots
===========
No need

Test
====
i18n task tests pass 👌

Deployment
==========
As usual

Warnings
========

This file refactor may be an issue for some forks, at least for those that didn't get their custom translations into the custom folder and modified the core ones. Also for those that added a new language. We may want to make an notice for everyone once this goes to master.


bertocq:
Closing because its outdated, newer version here https://github.com/consul/consul/pull/1718

PR #1706, opened 2017-06-26
Refactor comments partial

bertocq:
# What

We had two similar comment tree partials, but one just didn't need the comment count display

# How

*   Pass a variable to the partial to show/hide the comments count
*   Reuse the comments comment_tree original partial on budget investments show page
*   Remove no longer needd budget investments comment partial

# Screenshots

No need

# Test

Passing :D

# Deployment

As usual

# Warnings

None


voodoorai2000:
👏 🚿

PR #1708, opened 2017-06-27
Help texts on forms

decabeza:
What
====

Sometimes to provide **more information on forms**, you want include a note between `label` and `input field`.

For this I create the custom class `.note`

How
===
Removes custom class `.note` and change to Foundation `.help-text` class (maintain same style).

Also I included the `aria-describedby` attribute to improve accessibility.

More information about [aria-describedby on Foundation docs](http://foundation.zurb.com/sites/docs/forms.html)

Screenshots
===========
`Example of help text on form`
form-note

Warnings
========

Update the class `.note` to `.help-text`, if you don't do it there is only a minor change, the note text would be bigger.

### DIFF
--- a/app/assets/stylesheets/layout.scss
+++ b/app/assets/stylesheets/layout.scss
@@ -825,23 +825,6 @@ form {
     margin: $line-height / 2 0 $line-height / 2 $line-height / 4;
   }
 
-  .note,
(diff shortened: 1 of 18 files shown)


voodoorai2000:
👌

PR #1709, opened 2017-06-27
Rubocop fixes

bertocq:
# What

- Trimming the list of issues at `.rubocop_todo.yml`, bit by bit it will disappear :)

- Disable [Style/EmptyMethod](http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/EmptyMethod) after agreement

# How

*   Removing each cop issue and files list from `.rubocop_todo.yml`
*   Running`rubocop -a --only NAME_OF_COP`
*   Reviewing manually autocorrections and warnings. Discarding those false positives and fixing those that need manual intervention

There is a commit for each Fixed cop in order to better see the changes and review them.

### DIFF
--- a/app/models/concerns/searchable.rb
+++ b/app/models/concerns/searchable.rb
@@ -5,14 +5,15 @@ module Searchable
     include PgSearch
     include SearchCache
 
-    pg_search_scope :pg_search,
-                    against: :ignored, # not used since using a tsvector_column
-                    using: {
-                      tsearch: { tsvector_column: 'tsv', dictionary: "spanish", prefix: true }
-                    },
-                    ignoring: :accents,
-                    ranked_by: '(:tsearch)',
-                    order_within_rank: (self.column_names.include?('cached_votes_up') ? "#{self.table_name}.cached_votes_up DESC" : nil)
+    pg_search_scope :pg_search, {
+      against: :ignored, # not used since using a tsvector_column
+      using: {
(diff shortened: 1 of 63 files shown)


app/models/concerns/searchable.rb
```diff
@@ -5,15 +5,14 @@ module Searchable
...
-    pg_search_scope :pg_search, {
-      against: :ignored, # not used since using a tsvector_column
-      using: {
-        tsearch: { tsvector_column: 'tsv', dictionary: "spanish", prefix: true }
-      },
-      ignoring: :accents,
-      ranked_by: '(:tsearch)',
-      order_within_rank: (self.column_names.include?('cached_votes_up') ? "#{self.table_name}.cached_votes_up DESC" : nil)
-    }
+    pg_search_scope :pg_search,
```
xuanxu: I'd keep the format here, easier to read and understand IMO.

bertocq: Agree, many auto-correct indentations from rubocop doesn't look more than "hardcoded rules". But in this case what's happening is just removing braces from a hash parameter ({...}) that are not neccesary. So.. @xuanxu what should I do?

A- Revert completely the changes

B- Revert the formatting, but keep the removal of the curly braces?

xuanxu: I find searcheable.rb clearer as it was (both indentation and curly braces help readability here I think)

bertocq: Fixed! :)

Issue #1585, opened 2017-06-02
Annotation's color


voodoorai2000:
The color of highlighted text should be according to a score type: 
1 point for each comment in that word, 1 point for each positive vote for each comment in that word
![color highlight legislation](https://cloud.githubusercontent.com/assets/4169/26720367/3f888c7a-4788-11e7-9696-7bbf5cc4c9e5.png)

Please send PR against the [medialab-legislation branch](https://github.com/consul/consul/compare/medialab-legislation?expand=1)

bertocq:
@voodoorai2000 I've been trying to come up with a solution for this, but my CSS-fu or googling skills aren't enough to just determine if its 100% possible to have a simple solution or we've to go for the complex one 😞 .

## Current status:
- 👍  Each annotation has its own `.weight` property, and that's used to choose one [`weight-*` style](https://github.com/consul/consul/blob/master/app/assets/stylesheets/annotator_overrides.scss#L106-L120)

- **Scenario 1 Nested Annotations**: Outer annotation `Minima aut totam voluptatem` with 3 comments, and inner one `aut totam` with 1 comment) it looks like:
![screen shot 2017-06-11 at 22 41 07](https://user-images.githubusercontent.com/983242/27014327-1f4c9e9e-4ef7-11e7-96b5-ddba37887921.jpg)❌ 
Instead of :
![screen shot 2017-06-11 at 22 40 44](https://user-images.githubusercontent.com/983242/27014326-1f389944-4ef7-11e7-907e-3210f7dfe415.jpg)✅
Because the inner `aut totam` doesn't know it's inside another annotation with weight 3, it just knows about its own weight 😔 . 
Html is:
```html
<span class="annotator-hl weight-3" data-annotation-id="10">
   Minima
   <span class="annotator-hl weight-1" data-annotation-id="11">aut totam</span> 
   voluptatem
</span>
```

- **Scenario 2 Overlapping Annotations**: Given the text `dolorem odio laboriosam` with one annotation `dolorem odio` with 1 comment, and second annotation `odio laboriosam` with 1 comment.. the result looks like:
![screen shot 2017-06-11 at 22 13 32](https://user-images.githubusercontent.com/983242/27014164-6ec2641c-4ef3-11e7-896d-4dfb6fc67192.jpg)❌ 
But should be:
[...]

PR #1711, opened 2017-06-28
Annotation on legislation

decabeza:
Where
=====
* **Related Issue:** #1585 

What
====
- This PR improves the color palette commented on #1585. Also fixes a publish comment button text.

How
===
- I create a new palette of colors. Finally I used HEX colors instead of HSLA because every browser interprets the colors a little bit different:

color_palette

Pending to do
===========
- The main problem are not the colors. Is the implementation of JS. The problem is the nesting of `weight-N` classes, for example: 

**You annotate over a text already annotated:**

screen shot 2017-06-28 at 12 38 09

**This is the HTML code generated:**

screen shot 2017-06-28 at 12 38 24

**A `weight-2` inside of another `weight-2` doesn't work at all.** 😔
**The correct situation would be if the new annotation maintain `weight-2` class and is deleted from his parent:**

screen shot 2017-06-28 at 12 39 02

**So the result would be OK:** 👍🏻

screen shot 2017-06-28 at 12 39 08

### DIFF
--- a/app/assets/stylesheets/annotator_overrides.scss
+++ b/app/assets/stylesheets/annotator_overrides.scss
@@ -105,23 +105,43 @@
   background: rgba(255, 249, 218, 0.75);
 
   &.weight-1 {
-    background: #fff9da;
+    background: hsla(50, 100, 93, 0.5);
+
+    .weight-1 {
+      background: hsla(50, 100, 93, 0.75);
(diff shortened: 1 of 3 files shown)


voodoorai2000:
Marvellous 🎉 
Let's try it out! 💃

Issue #1674, opened 2017-06-21
Enabled legislation process


decabeza:
## Background

Now when creating a new legislation process in the admin section `admin/legislation/processes` you must specify the process dates, and if not is the current date appears on "Next processes" `/legislation/processes?filter=next`

## What

Sometimes the admins don't want show the process in "Next processes" menu.

## How

Add a "Enabled" checkbox on process dates (similar to Debate and Allegations phases) to allow hides the whole process.

![enabled_processes](https://user-images.githubusercontent.com/631897/27380025-ef014850-567d-11e7-8981-952ab6f1e716.jpg)

bertocq:
I thought about this when doing the other "Enabled" buttons independently. But didn't took the time to ask @voodoorai2000 and @xuanxu about it.

What I did was to enclose in two simple ruby classes [`Legislation::Process::Phase`](https://github.com/consul/consul/blob/master/app/models/legislation/process/phase.rb) and [`Legislation::Process::Publication`](https://github.com/consul/consul/blob/master/app/models/legislation/process/publication.rb) the logic around the [four phases and publications](https://github.com/consul/consul/blob/master/app/models/legislation/process.rb#L24-L38)

So maybe treating the "main Process" as another `Legislation::Process:Phase` and moving  the [`status`](https://github.com/consul/consul/blob/master/app/models/legislation/process.rb#L48-L58) logic on to that class, we could get this clean and tidy. Also a boolean attribute `process_enabled` would be needed, like the ones used for other phase/publications (https://github.com/consul/consul/blob/master/db/schema.rb#L419-L422)

What do you think?

voodoorai2000:
👌 +1

xuanxu:
+1, and +1 to change attribute name to `visible`or `published`

bertocq:
Yes @xuanxu it was previously "active" but I didn't quite seemed to me a good drescription of the action as enabling/disabling was easier to undestand for me. But its true that its mon-verb, and Published its way more descriptive.

So "Enabled" => "Published" +1 (maybe the database columns could keep the `_enabled` sufix, but adding another migration to rename them should be easy :) )

voodoorai2000:
+1 to `published`

entantoencuanto:
Hi, I'm taking a look to this issue, and I'm thinking about making some changes in the `legislation\processes_controller`, but I have some doubts:

[...]

PR #1714, opened 2017-06-30
Issue#1674 enabled legislation process

entantoencuanto:
Where
=====
* **Related Issue:** #1674

What
====
- Add a new column `published` to `Legislation::Process` associated table.
- Allow administrators set this attribute from `admin/legislation/processes` section.
- A not published process should be hidden from lists in `legislation/processes` for all filters.
- Also a not published process should be inaccessible to any user which is not administrator, through any of the actions of the processes controller.

How
===
- I've added a column `published` to `legislation_processes` table, and a scope `published` to associated model. By default this value is `true`. In this way, visibility of previously created legislation processes will be preserved.
- This scope has been used in the query to set `@processes` in the `#index` action of  `legislation/processes`.
- A checkbox has been added in the admin part to set published status.
- `debate`,  `draft_publication`, `allegations` and `result_publication` has been defined in routes as member actions of processes.
- Last actions and read permissions has been changed in `Everyone` abilities to permit only published processes.

Screenshots
===========
`admin/legislation/processes/new`
image

Test
====
- Some test where added to cover visibility of not published processes and verify the presence of published checkbox in admin edition part. 

Deployment
==========
- Don't forget to run migrations!

Warnings
========
none

### DIFF
--- a/app/controllers/admin/legislation/processes_controller.rb
+++ b/app/controllers/admin/legislation/processes_controller.rb
@@ -49,7 +49,8 @@ class Admin::Legislation::ProcessesController < Admin::Legislation::BaseControll
         :debate_phase_enabled,
         :allegations_phase_enabled,
         :draft_publication_enabled,
-        :result_publication_enabled
+        :result_publication_enabled,
+        :published
       )
     end
 end
--- a/app/controllers/legislation/processes_controller.rb
+++ b/app/controllers/legislation/processes_controller.rb
@@ -13,14 +13,13 @@ class Legislation::ProcessesController < Legislation::BaseController
     if @process.allegations_phase.enabled? && @process.allegations_phase.started? && draft_version.present?
       redirect_to legislation_process_draft_version_path(@process, draft_version)
     elsif @process.debate_phase.enabled?
-      redirect_to legislation_process_debate_path(@process)
+      redirect_to debate_legislation_process_path(@process)
     else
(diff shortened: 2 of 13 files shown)


entantoencuanto:
I'd like to go further with this PR using the comments in #1674:
- A I've talked with @bertocq, I'd like to move logic in `Legislation::Process::Phase` and `Legislation::Process::Publication to a concern, merging both models (after all, a Publication can be considered as a Phase without end date).
- Some stuff in `Legislation::Process` could be added to the concern, like [status](https://github.com/consul/consul/blob/master/app/models/legislation/process.rb#L48-L58), validations related with date ranges and especially scopes associated with status.
- The concern can use `define_method` to generate some methods as [currently defined](https://github.com/consul/consul/blob/master/app/models/legislation/process.rb#L24-L26), based in the phase name
- The phases concern can be used in many other models.

`Legislation::Process` model would look like this:

```ruby
class Legislation::Process < ActiveRecord::Base
  # ...
  include HasPhases
  
  has_phase :debate, :allegations
  has_phase :draft, :result, publication: true
  has_phase :global, start_date: :start_date,
                     end_date: :end_date,
                     phase_enabled: :published
  delegate :status, to: :global_phase

  # ...
end
```

voodoorai2000:
Impecable 😌 
+1 to the Phase concern

PR #1716, opened 2017-07-02
Locale site customization

MariaCheca:
Where
=====
* **Related Issue:** #1647 

What
====
This changes add a new feature to the informations pages to classify them by locale. This way there can be different pages with the same content but in multiple languages which will be only showed if the page locale is the same one.

How
===
I implemented it adding a new attribute `locale` to the Page model and allowing it to be edited from the form. The `more-information` page now uses a scope to list only the custom pages with the same locale.

Screenshots
===========
screen shot 2017-07-02 at 23 09 01
screen shot 2017-07-02 at 22 58 45

Test
====
No need to test it manually, just extended the current tests to perform with the changes.


bertocq:
Hey @MariaCheca thank you very much for your contribution 👏  😄 

Sorry about the test suite being broken right now 😭  I hope you didn't get much bothered about it. We where on fire last week for Madrid's Budget closing and got Legislation process controller broken in between. I think we've a fix here https://github.com/consul/consul/pull/1715 but tomorrow we'll take a good look a it.

After master is fixed I'll ping you again so you can just rebase and push your branch again 👍

MariaCheca:
Thanks @bertocq! No problem, this things happen :)

I'll stay tuned!

voodoorai2000:
Fantastic @MariaCheca 

Great idea to display only the pages available for the current locale.
Would you mind adding a nice feature spec for that too 😌

app/views/admin/site_customization/pages/index.html.erb
```diff
@@ -27,6 +28,7 @@
         <td><%= I18n.l page.created_at, format: :short %></td>
         <td><%= I18n.l page.created_at, format: :short %></td>
         <td><%= t("admin.site_customization.pages.page.status_#{page.status}") %></td>
+        <td><%= name_for_locale(page.locale) rescue nil %></td>
```
bertocq: What do you think about moving that `rescue nil` on to the [method definition](https://github.com/MariaCheca/consul/blob/master/app/helpers/locales_helper.rb)?

config/locales/activerecord.nl.yml
```diff
@@ -135,7 +135,8 @@ nl:
         title: Titel
         updated_at: Bijgewerkt op
         more_info_flag: Toon in extra informatie pagina
-        print_content_flag: Print inhoud 
+        print_content_flag: Print inhoud
+        locale: Taal
```
bertocq: 👏  for the extra effort in translations! (BTW: We're removing from specs the need to add translations for any language but English, new strings will be translated at https://crowdin.com/project/consul/)

app/views/admin/site_customization/pages/index.html.erb
MariaCheca: That makes more sense, on it!

MariaCheca:
Sorry I forgot to comment last changes are applied!

voodoorai2000:
Great 😌

Issue #1594, opened 2017-06-02
Incompatible investments


voodoorai2000:
### What
Allow admins to mark investments as incompatible and display incompatibles investments in the budget's results page.

Once an investment has been marked as incompatible, automatically recalculate [winners](https://github.com/consul/consul/pull/1555/files#diff-152c45b655078e6bccdda7600adfbfeaR11)

### Why
Some investments become incompatible, even though they have many votes, due to other investments having more votes and implying the same idea.

### Screenshots
![incompatible investments](https://user-images.githubusercontent.com/4169/27130903-a0bc56d2-5108-11e7-94d7-3c014e114b73.png)

### How

Implemented in this [PR](https://github.com/AyuntamientoMadrid/consul/pull/299) for Madrid's fork
Note that spending_proposals are deprecated and have been substituted by budget_investments. This feature is only meant for budget_investments.

bertocq:
I understand this is just a "cherry-pick" job, right?

voodoorai2000:
Yes, also that PR was meant for _spending proposals_. In Consul now _spending_proposals_ are obsolete and have been substituted for a cleaner implementation of _budget_investments_. This feature is only meant for _budget_investments_

bertocq:
Makes sense to allow only on the "Balloting finished" and "Finished" phases... although it could be done at any time in the Bugdet life process.

bertocq:
Ok, I think I'm going with:
1. Add a boolean "incompatible" column on budget_investments table
2. Add a scope "compatibles" to Budget::Investment
3. Add the "compatibles" scope to Budget::Result#calculate_winners at https://github.com/AyuntamientoMadrid/consul/blob/master/app/models/budget/result.rb#L21
4. Adding a checkbox on the investment edit page (only for admins). And checking on model save if the value has changed on to "true", if so... we'll recalculate the winners for the investment's Heading:`Budget::Result.new(budget, heading).calculate_winners`
5. Removing the "Valuators" column from the Admin's Budget Investments list, and adding an 
 "Incompatible" one in between the "Selected" and "Winner" ones. I think table should reflect important info and status... not massive data.
6. Adding a "Incompatible" filter on the Admin's Budget Investments list. A "Compatible" filter doesn't make much sense to me... but it could be a nice improvement to have a double filter.
7. Adding Incompatible list at the bottom of the Budget Results list

voodoorai2000:
Sounds perfect 👌

voodoorai2000:
🎉

PR #1720, opened 2017-07-03
Incompatible investments

bertocq:
Where
=====
* **Related Issue:** https://github.com/consul/consul/issues/1594

What
====
- Admins need a way to flag Budget Investments as "Incompatible". Incompatible investments are those that have been selected, but can't be winners because of internal conflicts with another investment (same area, etc...)

- Users need to be able to see which investments where discarded because there wasn't enough budget money for them, and those that where discarded because incompatibilities.

How
===
- Adding a boolean `incompatible` column to Budget::Investment defaulted to false, and both `.compatible` and `.incompatible` scopes to make use of it.

- Adding a "Mark as incompatible" checkbox on the Admin Budget Investment edit page. If checked and saved... it will trigger the `Budget::Result#calculate_winners` function for the Investment's heading.

- `Budget::Result#calculate_winners` now discards incompatible investments, so those can't be winners.

- Adding both discarded and incompatible investments to the budget results list (by heading), and a toggle button to just show winners or unhide all.

- On the side, refactoring both `app/views/budgets/results/_results_table.html.erb` and `app/views/budgets/results/show.html.erb` to be able to reuse that partial correctly.

Screenshots
===========
## Switching between Winners <=> Winners&Discarded and Incompatibles as a gif 🎉 
out

## Incompatible flag on Selected investments list: 
screen shot 2017-07-03 at 23 45 04

## Mark as incompatible Checkbox on Admin Investment edit page:
screen shot 2017-07-04 at 01 14 33

## Showing compatibility status on Admin Investment detail page:
screen shot 2017-07-04 at 01 14 06

Test
====
- Increased the Budget Result model test

- Increased the Admin Budget Investment edit feature

- Increased the User Budget Result feature 

Deployment
==========
- As usual, and asap! Madrid is waiting :)

Warnings
========
Some good taste needs to be applied to styles 💃  but our guy is on holidays so lets close our eyes 🙈

### DIFF
--- a/app/assets/javascripts/forms.js.coffee
+++ b/app/assets/javascripts/forms.js.coffee
@@ -15,6 +15,11 @@ App.Forms =
   toggleLink: ->
     $('.js-toggle-link').unbind('click').on('click', ->
       $($(this).data('toggle-selector')).toggle("down")
+      if $(this).data('toggle-text') isnt undefined
+        toggle_txt = $(this).text()
+        $(this).text( $(this).data('toggle-text')  )
+        $(this).data('toggle-text', toggle_txt)
+
       false
     )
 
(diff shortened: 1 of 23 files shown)


voodoorai2000:
👌

PR #1721, opened 2017-07-04
Add slugs to budget related

bertocq:
What
====
- Budget / Budget::Heading and Budget::Group need slugs in order to make many urls (like the Budget::Results related ones that I aim to fix) prettier and easier to use.

- Slugs must be present, and be unique:
     1. Budget slug: among other budgets
     2. Group slug: among other groups from its budget
     3. Heading slug: among other headings from all the groups from its budget

- Forks with existing objects of those classes won't have slugs, we need a rake task to backfill them

How
===
- Adding `slug` columns to the 3 models

- Adding a shared example spec for a Sluggable concern to test slug generation, making use of it on the 3 class model specs. Also increasing model specs to check uniqueness for each scenario.

- Adding a Sluggable concern that generates slug based on `name` attr. 

- Creating a rake task usable as `rake slugs:generate` that looks for all models including the concern and generates a slug for each object

Screenshots
===========
No need

Test
====
- Covered with a shared concern used on the 3 classes model specs.
- Increased each model spec to cover specific slug uniqueness rules.

Deployment
==========
## Remember to run rake task after deployment!! Next PR's will come with slug usage and may brake things

Warnings
========
Check deployment warning... this may get you some cool 500 internal server errors if you don't pay attention :)

### DIFF
--- a/app/models/concerns/sluggable.rb
+++ b/app/models/concerns/sluggable.rb
@@ -1,0 +1,11 @@
+module Sluggable
+  extend ActiveSupport::Concern
+
+  included do
+    before_validation :generate_slug
+  end
+
+  def generate_slug
+    self.slug = name.to_s.parameterize
+  end
+end
--- a/lib/tasks/slugs.rake
+++ b/lib/tasks/slugs.rake
@@ -1,0 +1,8 @@
+namespace :slugs do
+  desc "Generate slug attribute for objects from classes that use Sluggable concern"
+  task generate: :environment do
+    %w(Budget Budget::Heading Budget::Group).each do |class_name|
+      class_name.constantize.all.each(&:generate_slug)
+    end
+  end
+end
--- a/spec/models/concerns/sluggable.rb
+++ b/spec/models/concerns/sluggable.rb
@@ -1,0 +1,16 @@
+require 'spec_helper'
+
+shared_examples_for 'sluggable' do
+
+  describe 'generate_slug' do
+    before do
+      create(described_class.name.parameterize.tr('-', '_').to_sym, name: "Marlo Brañido Carlo")
+    end
+
+    context "when a new sluggable is created" do
+      it "gets a slug string" do
+        expect(described_class.last.slug).to eq("marlo-branido-carlo")
+      end
+    end
+  end
+end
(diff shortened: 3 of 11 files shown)


app/models/concerns/sluggable.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  included do |base|
+    base.validates :slug, presence: true, uniqueness: true, format: /\A[a-z0-9\-_]+\z/
+    base.before_validation :generate_slug
+  end
+
+  def generate_slug
+    return true if slug.present?
+    attrs = Array slug_attrs
+    slug = attrs.shift.to_s.parameterize
```
bertocq: Picks the first value from the attributes... this is prepared for a scenario where in a class we may specify `[title, created_at.month, created_at.year]` for example. 

So it's not going to use all of the attributes... just the first if there is none other object with the same title! But if there is.. it will try to create a title-month and so on

lib/tasks/slugs.rake
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+namespace :slugs do
+  desc "Generate any missing slug string for objects created before the Sluggable concern"
+  task generate_missing: :environment do
+    Rails.application.eager_load!
+    class_names = ActiveRecord::Base.descendants.select{|c| c.included_modules.include?(Sluggable)}.map(&:name)
```
bertocq: I really enjoyed finding out this one :), without `Rails.application.eager_load!` it doesn't work, but I'm worried it won't be there for Rails 5. Anyway forks that just get this along with Rails 5 update.. may have to find their way manually backfilling slugs. This commodity is just for the cool ones that are updated :)

spec/models/concerns/sluggable.rb
```diff
@@ -0,0 +1,25 @@
+require 'spec_helper'
+
+shared_examples_for 'sluggable' do
+  let(:factory_name) { described_class.name.parameterize.tr('-', '_').to_sym }
+  let(:attr_name) { "#{described_class.name} Second Third" }
+
+  describe 'generate_slug' do
+    before do
+      create(factory_name, name: attr_name)
```
bertocq: I'm using the `name` default attr here :( seemed overkill to find my way around the possible overwritting of the attr used for the slugs... so when the time comes we'll figure it out, or if it's a fork.. they'll find their way for sure :)

PR #1712, opened 2017-06-28
[WIP] Budget Stats generation & display

bertocq:
Where
=====
* **Related Issues:** 
- https://github.com/AyuntamientoMadrid/consul/issues/812
- https://github.com/consul/consul/issues/1737

What
====
Madrid's fork got some nice Budget Stats view, on deprecated Spending Proposal model. We want them for the new Budget model and usage.

Getting current specific code from https://github.com/AyuntamientoMadrid/consul/blob/master/app/controllers/spending_proposals_controller.rb#L71 and specific view. And moving on from Geozones usage on to Headings usage for population/district specifics.

How
===
- Adding a missing ballot_lines_count counter_cache to Budget::Ballot::Line relationship with Budget::Ballot for `balloters` stats

- Adding votes and ballots to Budget::Investments on dev_seeds so local development can try this things. Maybe its not done quite nice or good... need to invest time in dev_seeds to completly get all scenarios done right.

- Adding a Budget Stats controller, route and ability for anyone to see stats when Budget is either on ballots reviewing or finished phases.

- Adding a Budget::Stats class and its view (although I need some help with this, since styles are not my thing :D, also many stat names need to be fixed and maybe some style class names as well) @decabeza you MVP!!

- Adding link to results page on tabs under header

Screenshots
===========
localhost-3000-budgets-7-stats

Test
====
Incoming from Madrid.

Deployment
==========
As usual

Warnings
========
Still some bits to fix on the Results side... new PR will follow

PR #1727, opened 2017-07-05
Followable

taitus:
Where
=====
* **Related PR's:** https://github.com/rockandror/consul/issues/6

What
====
> Allow logged users to follow/unfollow proposals or budgets investments
> Show in the user profile their interests according to the proposals or budgets investments they are following.
> User can mark his interests as public.
> ~~Show current followers at proposal and budget investment pages.~~
> Followers of a proposal or budget investment will receive related notifications.

How
===
> Create polymorphic association on model concern “Followable” and adding to Proposal and Budget investments.
> Adding check “public_interest” to user account.
> Adding interests to user profile.
> ~~Adding new tab to proposals and budget investment for display Followers and notifications~~
> Create new reusable partials to display followable buttons, followable followers

Screenshots
===========
**1. Follow button - Mobile**
captura de pantalla 2017-07-11 a las 10 37 43

**2. Follow button - Tablet**
captura de pantalla 2017-07-11 a las 10 37 29

**3. Follow button - Desktop**
captura de pantalla 2017-07-11 a las 10 37 15

**4. Public interest check on user account**
captura de pantalla 2017-07-05 a las 17 49 25

**5. Interest list on user profile**
captura de pantalla 2017-07-05 a las 17 50 18

Test
====
> Adding `Followable` shared feature specs, and apply to Proposals and Budget Investment
> Increased the user spec feature to display public interest
> Increased the notification and proposal notification spec feature to check notification sending to new followers

Deployment
==========
> Run migrations

Warnings
========
> None


db/schema.rb
```diff
@@ -95,9 +95,8 @@
   create_table "budget_ballots", force: :cascade do |t|
     t.integer  "user_id"
     t.integer  "budget_id"
-    t.datetime "created_at",                     null: false
-    t.datetime "updated_at",                     null: false
-    t.integer  "ballot_lines_count", default: 0
```
bertocq: This is 👍 , I messed up somehow on a PR and didn't realize it. I added this on a feature branch that is still on PR stage https://github.com/consul/consul/pull/1712/files#diff-ddcb0cf5e4dfc6ac0c06b85bac771760 My apologies >_< 🙇

app/controllers/account_controller.rb
```diff
@@ -25,8 +25,8 @@ def account_params
       if @account.organization?
         params.require(:account).permit(:phone_number, :email_on_comment, :email_on_comment_reply, :newsletter, organization_attributes: [:name, :responsible_name])
       else
-        params.require(:account).permit(:username, :public_activity, :email_on_comment, :email_on_comment_reply, :email_on_direct_message, :email_digest, :newsletter, :official_position_badge)
+        params.require(:account).permit(:username, :public_activity, :public_interest, :email_on_comment, :email_on_comment_reply, :email_on_direct_message, :email_digest, :newsletter, :official_position_badge)
```
Senen: @bertocq maybe public_interest should be public_interests?

app/models/follow.rb
```diff
@@ -0,0 +1,31 @@
+class Follow < ActiveRecord::Base
+  belongs_to :user
+  #TODO Rock&RoR: Check touch usage on cache system
```
Senen: @taitus We have to do the check, fix if necessary and remove the comment.

app/controllers/account_controller.rb
bertocq: +1 Yes it sounds more descriptive, right? Feature is "User can mark his interests as public." so plural makes sense.

app/models/follow.rb
```diff
@@ -0,0 +1,31 @@
+class Follow < ActiveRecord::Base
+  belongs_to :user
```
bertocq: I think a cool improvement here (even if there is no actual use on the code right now) would have been to use a [counter_cache](http://guides.rubyonrails.org/association_basics.html#options-for-belongs-to-counter-cache)

Something like `belongs_to :user, counter_cache: :following_count`

Without forgetting the migration:
```
    add_column :users, :following_count, :integer, default: 0
    add_index :users, :following_count
```

app/models/follow.rb
```diff
@@ -0,0 +1,31 @@
...
+  #TODO Rock&RoR: Check touch usage on cache system
+  belongs_to :followable, polymorphic: true
```
bertocq: I think here the `counter_cache` usage I suggested for the `belongs_to :user` would actually be used, so its more like a performance improvement for any `@investment.follows.count` or `@proposal.follows.count` around. What do you think about naming it "followers_count" twitter-style? 😄

app/models/follow.rb
bertocq: Don't worry about adding this, seems like there is no actual usage not even reason to. On the mockups there is nowhere to use it (only individual Followable counts) so when that time comes we'll add it if @decabeza thinks it makes sense to put a global following count 👍 .

app/controllers/follows_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def create
+    @followable = find_followable
+    @follow = Follow.create(user: current_user, followable: @followable)
+    render :refresh_follow_button
+  end
+
+  def destroy
+    @follow = Follow.find(params[:id])
+    @followable = @follow.followable
```
bertocq: I think I'm missing something, and it may be my ignorance about destroy usage. Where would be used the `@followable` variable?

app/models/follow.rb
Senen: Because of that we do not have implemented it. Yes, maybe is a YAGNI thing for now.

app/models/follow.rb
```diff
@@ -0,0 +1,31 @@
...
+          followable_type: followable.class.to_s,
+          followable_id: followable.id)
+  end)
+
+  def self.followed?(user, followable)
+    return false unless user
+    !! by_user_and_followable(user, followable).try(:first)
+  end
+
+  def self.interests_by(user)
```
bertocq: What do you think about using a symbol proc and { |item| .. } syntax to make function body smaller? It would be something like `user.follows.map { |follow| follow.followable.tags.map(&:name) }.flatten.uniq` (the entire function body)

app/views/follows/_followers.html.erb
```diff
@@ -0,0 +1,15 @@
+<div class="row">
+  <div class="small-12 column">
+
+    <% if followable.follows.blank? %>
+      <div class="callout primary text-center">
+        <%= t('follows.no_followers') %>
+      </div>
+    <% end %>
```
bertocq: Adding here an `else` and including inside it the following `.each` loop could maybe be safer?

app/models/proposal.rb
```diff
@@ -170,6 +171,10 @@ def notifications
     proposal_notifications
   end
 
+  def users_to_notify
```
bertocq: Maybe increasing proposal model spec to check this new method, just reusing existing spec for `voters` could be a nice test improvement https://github.com/rockandror/consul/blob/followable/spec/models/proposal_spec.rb#L382-L425

app/models/follow.rb
bertocq: As we commented, this method could make more sense in the User model, to be used as user.interests

spec/factories.rb
```diff
@@ -166,8 +166,8 @@
     end
 
     trait :flagged do
-      after :create do |debate|
```
bertocq: Nice catch! 👍

spec/shared/features/followable.rb
```diff
@@ -0,0 +1,182 @@
+shared_examples "followable" do |followable_class_name, followable_path, followable_path_arguments|
```
bertocq: Really like the usage of shared_examples 👏  😊

spec/spec_helper.rb
```diff
@@ -5,6 +5,7 @@
 require 'knapsack'
 Dir["./spec/models/concerns/*.rb"].each { |f| require f }
 Dir["./spec/support/**/*.rb"].sort.each { |f| require f }
+Dir["./spec/shared/**/*.rb"].sort.each { |f| require f }
```
bertocq: Seems like we have 4 shared_examples for concerns, but they are across 3 different subfolder 🙈 . I'll correct this afterwards don't worry :). IMHO `/spec/models/concerns/` should be the path for a concern shared_example

PR #1734, opened 2017-07-09
Fix date localization

bertocq:
What
====
- Dates/Times should always use the locale format

How
===
- Using i18n http://guides.rubyonrails.org/i18n.html#adding-date-time-formats
- Fixing the date used on the test and expectation.

Screenshots
===========
screen shot 2017-07-09 at 16 43 29
screen shot 2017-07-09 at 16 43 43

Test
====
Updated and improved with fixed creation date.

Deployment
==========
As usual

Warnings
========
None


voodoorai2000:
Nice

PR #1746, opened 2017-07-10
Rubocop fixes 🤖: Rails edition 💎🛤

bertocq:
What
====
- Cleanup some rubocop_todo Rails/* issues

How
===
*   Removing each cop issue and files list from `.rubocop_todo.yml`
*   Running`rubocop -a --only NAME_OF_COP`
*   Reviewing manually autocorrections and warnings. Discarding those false positives and fixing those that need manual intervention


voodoorai2000:
👌

Issue #1655, opened 2017-06-14
SEO for budget's results page 


voodoorai2000:
### What
Add social meta tags and SEO information to the budget's results page

### How
The results page can be found [here](https://github.com/consul/consul/blob/master/app/controllers/budgets/results_controller.rb)

And a nice example of SEO for proposals [here](https://github.com/consul/consul/blob/master/app/views/proposals/show.html.erb#L1)

Include these elements:
- Title
- Meta description
- Social media meta tags
- Canonical url

voodoorai2000:
Closed via #1749

PR #1749, opened 2017-07-13
Budget results meta tags

voodoorai2000:
Where
=====
* **Related Issue:** #1655 

What
====
- Add social meta tags for budget results

How
===
- Adding results url to meta tags
- Adding budget's finished description to meta tags

Bonus
===
- Displays results for first heading if no heading_id param is specified

Warnings
===
- If using custom urls for budget results, change `social_url`and `canonical_url` appropriately

### DIFF
--- a/app/controllers/budgets/results_controller.rb
+++ b/app/controllers/budgets/results_controller.rb
@@ -1,22 +1,31 @@
 module Budgets
   class ResultsController < ApplicationController
+    before_action :load_budget
+    before_action :load_heading
 
     load_and_authorize_resource :budget
 
     def show
       authorize! :read_results, @budget
       @investments = load_result.investments
-      @heading = heading
     end
 
     private
 
       def load_result
-        Budget::Result.new(@budget, heading)
+        Budget::Result.new(@budget, @heading)
       end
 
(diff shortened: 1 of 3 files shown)


app/controllers/budgets/results_controller.rb
```diff
@@ -1,22 +1,31 @@
...
       end
 
-      def heading
-        @budget.headings.find(params[:heading_id])
+      def load_budget
+        @budget = Budget.find_by(id: params[:budget_id])
+      end
+
+      def load_heading
+        if params[:heading_id].present?
```
bertocq: What do you think about replacing the whole function body with:
```
 @heading = @budget.headings.find_by(id: params[:heading_id]) || @budget.headings.first
```
When slugs come just replacing the `id:` for `slug:` will do

bertocq: 🤔  maybe the downside of me one-liner is that  we'll do database query when we can be sure theres not going to be a result (when no params doesn't exist)

voodoorai2000: Sure, one liner sounds good

PR #1750, opened 2017-07-13
User page: new following section

Senen:
What
====
Show user followable activity on public user page.

**Refactor**

* Remove popup on follow and unfollow buttons
* Add AJAX notice on follows controller actions to keep user better informed

How
===
Add new tab to show followed objects by user.

Screenshots
===========
captura de pantalla 2017-07-13 a las 18 47 13
captura de pantalla 2017-07-13 a las 18 46 58

**Refactor screenshots**
captura de pantalla 2017-07-19 a las 19 54 50
captura de pantalla 2017-07-19 a las 19 54 38

Test
====
Increased test coverage to:
- Check automatic following tab activation when filter "follows" defined
- Check "Following" tabs UI behaviour
- Check following accordion behavior
- Check followed object presence in correct accordion tab
- Check action buttons presence,  they should be shown when current user has enough rights
- Check following and unfollowing notices. 

Deployment
==========
As usual.


Senen:
@voodoorai2000, @bertocq, @decabeza ahí va otro pull-request desde la playa! 😄 Esperamos vuestros comentarios.

**Falta la verificación de la funcionalidad por parte de Yago Bermejo.**

app/controllers/users_controller.rb
```diff
@@ -1,13 +1,12 @@
 class UsersController < ApplicationController
-  has_filters %w{proposals debates budget_investments comments}, only: :show
+  has_filters %w{proposals debates budget_investments comments follows}, only: :show
 
   load_and_authorize_resource
   helper_method :author?
-  helper_method :author_or_admin?
```
bertocq: I can't really understand the reason to remove `author_or_admin?` method and its usage, can you explain a bit further? :)

app/helpers/follows_helper.rb
```diff
@@ -52,6 +52,18 @@ def unfollow_drop_id(followable)
     "unfollow-drop-#{entity_full_name(followable)}-#{followable.id}"
   end
 
+  def entity_title(title)
+    entity = title.gsub('::', '/').downcase
+    t("activerecord.models.#{entity}.other")
+  end
+
+  def entity_icon(entity)
+    case entity
```
bertocq: Have you considered the option to use a hash here instead of a case statement? Something like:

```ruby
def entity_icon(entity)
  {
    'Proposal' => 'proposals',
    'Budget::Investment' => 'budget'
  }[entity]
end
```

It could be improved by moving the hash to a constant variable, or even switching keys for values to use new hash syntax:
```
def entity_icon(entity)
  {
    proposals: 'Proposal',
    budget: 'Budget::Investment'
  }.invert[entity]
end
```

Last but not least, it seems odd that "proposals" is plural bug "budget" is singular 🤔

app/models/abilities/common.rb
```diff
@@ -50,7 +50,7 @@ def initialize(user)
         can :create, Budget::Investment,               budget: { phase: "accepting" }
         can :suggest, Budget::Investment,              budget: { phase: "accepting" }
         can :destroy, Budget::Investment,              budget: { phase: ["accepting", "reviewing"] }, author_id: user.id
-        can :vote, Budget::Investment, budget: { phase: "selecting" }
+        can :vote, Budget::Investment,                 budget: { phase: "selecting" }
```
bertocq: ^_^ /bertocq likes this

app/views/users/_proposal.html.erb
```diff
@@ -0,0 +1,28 @@
+<tr id="proposal_<%= proposal.id %>">
+  <td>
+    <%= link_to proposal.title, proposal, proposal.retired? ? { class: 'retired' } : {} %>
```
bertocq: I like that `? { class: 'retired' } : {}` :) I think I have added some `class=""` around :( couldn't figure out this was the way 👏

app/views/users/_proposal.html.erb
```diff
@@ -0,0 +1,28 @@
...
+
+  <% if author?(proposal) %>
+    <td class="text-center">
+      <%= link_to t("users.proposals.send_notification"),
+                  new_proposal_notification_path(proposal_id: proposal.id),
+                  class: 'button hollow' %>
+    </td>
+  <% end %>
+
+  <% if author?(proposal) || proposal.retired? %>
```
bertocq: It took me a bit to understand all the conditionals around this block and realize there is only 2 scenarios. Maybe this could be easier to read:
```
<% if proposal.retired? %>
 <td class="text-center">
  <span class="label alert"><%= t('users.proposals.retired') %></span>
 </td>
<% elsif author?(proposal) %>
 <td class="text-center">
   <%= link_to t('users.proposals.retire'),
               retire_form_proposal_path(proposal),
               class: 'button hollow alert' %>
 </td>
<% end %>
```

app/views/users/show.html.erb
```diff
@@ -46,7 +46,7 @@
         <p><%= t('users.show.private_activity') %></p>
       <% end %>
 
-      <% if @user.public_interests || @authorized_current_user %>
+      <% if valid_interests_access? %>
```
bertocq: Nice catch!

app/helpers/follows_helper.rb
```diff
@@ -52,6 +52,18 @@ def unfollow_drop_id(followable)
     "unfollow-drop-#{entity_full_name(followable)}-#{followable.id}"
   end
 
+  def entity_title(title)
+    entity = title.gsub('::', '/').downcase
```
bertocq: I think you'll love this page http://inflector.me/Budget::Investment :)

This two lines could be just one `t("activerecord.models.#{title.underscore}.other")`

app/views/users/_following.html.erb
```diff
@@ -0,0 +1,34 @@
...
+        <span class="title">
+          <strong><%= entity_title(followable_type) %></strong>
+        </span>
+      </a>
+
+      <div class="accordion-content" data-tab-content>
+        <table>
+          <tbody>
+            <% @follows.where(followable_type: followable_type).each do |follow| %>
+              <% if followable_type == "Proposal" %>
```
bertocq: After seeing both `entity_title` and `entity_icon` helper methods. I think a third one could be nice to avoid spilling `== "Proposal"` and `== "Budget::Investment"` on a view :D, seems tricky to remember if another entity becomes as well "followable"

So the entire if-else could be replace for something like:
```
<%= render entity_partial(followable_type), entity_partial(followable_type).to_sym => follow.followable %>
```

Being the helper function at `app/helpers/follows_helper.rb`:
```
def entity_partial(title)
  title.parameterize.gsub('-','_')
end
```

What do you think?

app/helpers/follows_helper.rb
Senen: Thanks for this resource. 😄 👏

app/controllers/users_controller.rb
taitus: @bertocq The main reason is that apparently it was not being used anywhere in the application.

bertocq: 👍  you're right, I did a project search and no use, although sometimes looking for the method string is not enough if you do compose de method name with a string interpolation like `send("#{this}_or_#{that}")` but that's not the case I guess based on the specs passing ^_^

Nice catch & cleanup 👏 😄

app/views/users/_following.html.erb
Senen: Yes @bertocq while we were programming exactly your solution we thought that this abstraction forces the developer to follow entity_partial helper method to understand where to put and how to name new followable object partials. Because of that we decide to let these lines explicitly written, but if you prefer, we make this change in a second. I think both solutions has some disadvantage and in this case we prefer explicitness over reducing code lines.

app/views/users/_proposal.html.erb
Senen: Totally agree @bertocq! We will make this change too!

app/views/users/_following.html.erb
bertocq: In the scenario were another entity becomes followable, for both icon and title translation following the trail to the helper will be needed, so I don't see it as a big downside to the proposed refactor. I just imagine 4-5 entities being followable and the if-elsif-elsif.. being too big when just following a convention (that also complains with how rails expects things to be named) could make it easier.

But thats like you say a subjective matter, I prefer conventions over more lines, maybe @voodoorai2000 can take a look and say if he has a hard opinion about it or not.

app/helpers/follows_helper.rb
Senen: I like very much this option! Less words and same clarity always better!

app/helpers/follows_helper.rb
```diff
@@ -9,23 +9,22 @@ def show_unfollow_action?(followable)
   end
 
   def follow_entity_text(followable)
-    entity = followable.class.name.gsub('::', '/').downcase
```
bertocq: 👏  nice cleaning up on the side! 💪

Senen:
@bertocq thanks for all your advices! They are allways welcome! 😃.We are still waiting customer validation. We'll let you know when it's done.

app/views/users/_following.html.erb
```diff
@@ -0,0 +1,31 @@
+<ul class="accordion" data-accordion data-allow-all-closed="true">
+
+  <% @followable_types.each do |followable_type| %>
```
voodoorai2000: Hey, 
I just sent you a small PR regarding this, hope you like it :)

voodoorai2000:
Regarding the concept of `entity`... it's something new that we haven't used yet in Consul... maybe we can postpone the decision of using `entity` by creating a `FollowablesHelper` and chaging the following `helper methods`:

```
FollowsHelper:
def follow_entity_text
def unfollow_entity_text
def entity_full_name
def entity_title
def entity_icon
def entity_partial
```

for something like this:

```
FollowsHelper:
def follow_text
def unfollow_text

FollowablesHelper:
def followable_full_name
def followable_title
def followable_icon
def followable_partial
```

app/views/users/_following.html.erb
```diff
@@ -0,0 +1,30 @@
...
+        <span class="title">
+          <strong><%= entity_title(followable_type) %></strong>
+        </span>
+      </a>
+
+      <div class="accordion-content" data-tab-content>
+        <table>
+          <tbody>
+            <% follows.each do |follow| %>
+              <%= render_follow(follow) %>
```
Senen: @voodoorai2000 nice refactor! 😍

Senen:
@voodoorai2000, @bertocq, @decabeza  i think we finally have final version. I made all suggested changes by @voodoorai2000 and removed unused translation key, also we removed popup on follow buttons and add a notice message to keep user informed. Off course we increased tests coverage with new notice checks and delete a lot of ugly code needed by popup but useless now.

As always, thanks for your advices. 😃

Senen:
@bertocq, @voodoorai2000, @decabeza  this feature was approved by Yago. We can merge it!

Issue #1725, opened 2017-07-05
Allow users to verify their account against a local copy of the Census


voodoorai2000:
Why
===
Some forks do not have an API to verify their users against the population's Census.

What
====
- Add a table to the database to import user data
- Allow users to verify their account against this table

How
===
First we have to understand how the current verification process works.

It's a double verification process:

Census verification 
https://github.com/consul/consul/blob/master/app/models/verification/residence.rb
https://github.com/consul/consul/blob/master/lib/census_api.rb

SMS verification
https://github.com/consul/consul/blob/master/app/models/verification/sms.rb
https://github.com/consul/consul/blob/master/lib/sms_api.rb

After these two verifications a user can carry out any action in the web.
There is an additional letter verification, which is optional.

The screenshots show a little more this process.

Now that we have an idea of what it works, we can try to find the place to add our changes.
Which should probably go here:
https://github.com/consul/consul/blob/master/app/models/verification/residence.rb
https://github.com/consul/consul/blob/master/app/models/concerns/verification.rb

Only verifying the following attributes against this new table is necessary:
- Document number
- Document type
- Date of birth
- Postal code

Screenshots
====
Census Verification

![census verification](https://user-images.githubusercontent.com/4169/27864070-5dc91b54-618d-11e7-9795-df339054ff42.png)

SMS Verification

![sms verification](https://user-images.githubusercontent.com/4169/27864074-60a22244-618d-11e7-8686-c5c50d2f9c69.png)

Verification successful

![verification succesful](https://user-images.githubusercontent.com/4169/27864077-62f0d004-618d-11e7-83c3-534505161199.png)

My account

![my account verified](https://user-images.githubusercontent.com/4169/27864083-67402d08-618d-11e7-988e-b364ce3a6a2b.png)

Warnings
========
- The verification system is a critical part of Consul and is being used by almost all forks. Any change could break the verification system for users that have modified it. Be careful.

voodoorai2000:
Closed via #1752

PR #1752, opened 2017-07-14
Allow users to verify their account against a local copy of the Census

MariaCheca:
Where
=====
* **Related Issue:** #1725

What
====
As described in the [issue](https://github.com/consul/consul/issues/1725), the platform needed another way to verify user's identity using a database table where a copy of the Census would be stored. This resource will be used in case there's no API to provide that information.

How
===
To achieve this purpose I created a new model called `LocalCensusRecords` associated to the mentioned census table, and abstracted the logic to request the census information into a `CensusCaller` class. Inside this class both calls are performed: to the `CensusApi` and `LocalCensus`, and returns one or another depending on which one is available. Please note that the data format returned by `CensusCaller` is _universal_ no matter what request has been successful, so there won't be any trouble once the data is read and stored.

Test
====
I added new tests to cover the new files `LocalCensus` and `CensusCaller` to make sure their behaviour is as expected.

Hope you like this solution :)
cc/ @voodoorai2000


app/models/local_census_record.rb
```diff
@@ -0,0 +1,3 @@
+class LocalCensusRecord < ActiveRecord::Base
+
```
MariaCheca: The same question came to my mind before removing those validations from the model.

I decided to finally do it because I assumed the table would be filled without passing through Rails, and as there shouldn't be any way to modify the records once they're in the database - too sensitive data to make it easy to change it - it isn't necessary to cover those validations because they will never be used.

Anyway, if you think it makes more sense to validate the attributes also in the model to have everything covered I can add them again :)

app/models/officing/residence.rb
```diff
@@ -102,7 +102,7 @@ def date_of_birth
   private
 
     def call_census_api
```
MariaCheca: Sure! That's better, I'll change it.

app/models/local_census_record.rb
bertocq: 🤔  its a feature not defined yet.... Yes, forks may use it in different ways, either with a /admin interface to upload a .csv or .xls file (that would go through Rails), but someone a bit hacky could like to load a sql dump directly into postgresql :D.

Anyway for those that go through Rails the validations may be a good idea. I think on madrid we'll do that, even here in consul, just because this "Local Census" may be used for Demos as well (not only for Associations).

MariaCheca: Ok then, I'll add the validations in Rails asap!

MariaCheca:
@bertocq validations in the `LocalCensusRecord` model are restored!

MariaCheca:
Horray! :D

PR #1757, opened 2017-07-17
Fix/budget recommendation

bertocq:
Where
=====
* **Related Rollbar Issue:** https://rollbar.com/voodoorai2000/Participacion/items/1211/

What
====
- When there is no current budget, code fails with a 500 internal server error. 

How
===
- Simple fix, sending budgets url instead of budget url for now. Needs some thinking and create an issue to describe a possible "current budget" feature.

Screenshots
===========
No need

Test
====
Don't think og metatags deserve a feature test.

Deployment
==========
- Remember to close the rollbar issue afterwards

Warnings
========
None


bertocq:
F# github opening PR to upstream instead of cloned repo!!! NOT COOL

PR #1762, opened 2017-07-21
Legislation process phases refactor

entantoencuanto:
Where
=====
* **Related Issue:** #1674
* **Related PR's:** #1714

What
====
Legislation processes go through some phases with similar validations and scopes. Each phase has a method wich returns an object of class `Legislation::Process::Phase` or `Legislation::Process::Publication`. This object can report about its status and validates the date range. Also some similar scopes and validations are defined for each phase. The idea is to abstract this similar logic with a concern.

How
===
`Legislation::Process::Phase` and `Legislation::Process::Publication` has been merged into one class `HasPhases::Phase`. If a class includes `HasPhases` concern the phases can be defined calling from it `has_phase`.

The arguments of `has_phase` can be 1 or more phase names represented with strings or symbols. By default, with no options, `has_phase` assumes default column names. For example:

```ruby
has_phase 'debate'
```

expects that the table has defined the columns `debate_start_date`, `debate_end_date` and `debate_phase_enabled`. The default column names can be overwritten with the `:start_date`, `:end_date` and `:phase_enabled` options, but these options will be ignored when more than a phase is defined in a `has_phase` statement. For example:

```ruby
has_phase 'global', start_date: :start_date,
                    end_date: :end_date,
                    phase_enabled: :published
```

defines the phase with the columns `start_date`, `end_date` and `published`.

The phase definition with `has_phase` provides:

- An instance method named with the phase which returns an object of class `HasPhases::Phase` initialized with the associated columns values.  Examples:

```ruby
has_phase 'debate' # provides #debate_phase method
has_phase 'draft', phase_keyword: :publication # provides #draft_publication method
```

- A group of validations for phases defined using related columns.
- A group of scopes dependent on dates and enabled status. For example:

```ruby
has_phase 'debate' # provides .debate_published, .debate_next, .debate_started,
                   # .debate_open and .debate_past
```

- The list of phases is stored on class and is accesible with the class method `::phases_info`.
- The last information is used by the instance method `#enabled_phases_and_publications_count` returning the count of enabled phases.

### Options

- `:start_date`: Name of the column for the start date. Ignored if more than one phase name is given. By default is generated appending "_start_date" to phase name.
- `:end_date`: Name of the column for the end date. Ignored if more than one phase name is given. By default is generated appending "_end_date" to phase name.
- `:phase_enabled`: Name of the column for the enabled attribute.  Ignored Ignored if more than one phase name is given. By default is generated appending "_phase_enabled" to phase name.
- `:phase_keyword`: Use this option to use a different name for the method which returns the phase. For example:

```ruby
    has_phase :draft, phase_keyword: :publication
    # defines #draft_publication instead of #draft_phase
```

- `:publication` - Can be `true` or `false`. Default is `false`. When `true`, the phase only consider the start date and there is no end date. A phase defined as publication doesn't generate validations and only provides `::published`, `::next` and `::started` scopes.

The concern is very flexible to allow refactor the `Legislation::Process` making changes only in the model but can be simpler removing some options and unifying some name conventions:

- Using only the term phase to name the method which returns the `Phase` object (for example replace `draft_publication` with `draft_phase`), the `phase_keyword` option can be removed.
- In the original code the columns related with publications doesn't have validations. The validations can be modified to be used by any kind of phase.

Test
====
I've defined some shared examples to avoid repeating examples when a new phase is defined in a model. Providing the column names used to define the phase a group of examples defined in spec/models/concerns/has_phases.rb can be included in the model spec. For example:

```ruby
  it_behaves_like "a class with phase", :debate, :debate_start_date, :debate_end_date, :debate_phase_enabled
```

tests the debate phase. There is a different group of examples for publications, but can be removed if the changes about naming conventions discussed in the previous section are made.


bertocq:
Sorry @entantoencuanto for the review delay 😞 , we're still tight on time and catching up with tasks. Hopefully in a week or less we're going to be making a new release and this will be reviewed and included. Right now there's a bit of a jam from Consul on to Madrid's fork ^_^ and we're giving higher priority to requested features 🙇 . Thanks again for your contributions!! ❤️

javierm:
@entantoencuanto Thank you very much, and sorry it took us four years to finally have a look at this pull request :pray: and that we're actually closing it :grimacing:.

We've been trying to steadily catch up with the pull requests pending review (two years and a half ago we had about 120), and are only now catching up with the last ones :relieved:.

I think given the fact that so much time has passed and that you're now working on other projects, asking you to update this pull request to the current master branch with no guarantees that we'll merge it in the end would be even more impolite than just closing it. Correct me if I'm wrong!

In the last few years, the code in legislation processes hasn't seen much activity other than a few bugfixes. Due to that, we haven't needed to refactor this code except for what we did in pull request #4521.

So we don't plan to do refactorings like this one until there are new developments in legislation processes or we find cases where refactoring the code like this can make it easier to implement custom changes in forks.

Again, really sorry for the whole situation!

PR #1768, opened 2017-07-25
Rubocop autocorrections 🤖

bertocq:
# What

Automatic rubocop cleanup:

*   Redundant spaces
*   Redundant curly braces around a hash parameter.
*   tr => gsub (http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Performance/StringReplacement)

# How

`rubocop -a` as we have already existent issue at `.rubocop_todo.yml`

# Screenshots

No need

# Test

Hope they'll pass :D

# Deployment

As usual

# Warnings

None

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -440,6 +440,8 @@ Style/AccessorMethodName:
 Style/BracesAroundHashParameters:
   Exclude:
     - 'app/models/concerns/searchable.rb'
+    - 'spec/features/budgets/investments_spec.rb'
+    - 'spec/features/proposals_spec.rb'
 
 # Offense count: 119
 # Configuration parameters: EnforcedStyle, SupportedStyles.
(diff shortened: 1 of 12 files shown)


.rubocop_todo.yml
```diff
@@ -440,6 +440,8 @@ Style/AccessorMethodName:
 Style/BracesAroundHashParameters:
   Exclude:
     - 'app/models/concerns/searchable.rb'
+    - 'spec/features/budgets/investments_spec.rb'
```
bertocq: This two where added since we didn't agree on the usage of curly braces on those two specs. Will re-evaluate later as we tackle again rubocop_todo list

app/helpers/followables_helper.rb
```diff
@@ -20,7 +20,7 @@ def render_follow(follow)
   end
 
   def followable_class_name(followable)
-    followable.class.to_s.parameterize.gsub('-', '_')
+    followable.class.to_s.parameterize('_')
```
bertocq: I didn't knew about this :) but `parameterize` accepts a param for the substitution http://api.rubyonrails.org/classes/ActiveSupport/Inflector.html#method-i-parameterize

PR #1774, opened 2017-08-01
User activity

decabeza:
What
====
- Now when the user comment an element and this is hidden, appears a `<abbr>` tag with the title `This X has been deleted`. 

How
===
- Improves styles and replace `abbr title` with an explicit text.

Screenshots
===========
`BEFORE`
activity_before

`AFTER`
activity_after

### DIFF
--- a/app/helpers/users_helper.rb
+++ b/app/helpers/users_helper.rb
@@ -16,10 +16,8 @@ module UsersHelper
     if commentable.nil?
       deleted_commentable_text(comment)
     elsif commentable.hidden?
-      "<del>".html_safe + commentable.title.html_safe + "</del>".html_safe + ' ' +
-      '<span class="small">'.html_safe +
-      '(' + deleted_commentable_text(comment) + ')' +
-      "</span>".html_safe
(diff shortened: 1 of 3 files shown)


app/helpers/users_helper.rb
```diff
@@ -16,9 +16,10 @@ def comment_commentable_title(comment)
     if commentable.nil?
       deleted_commentable_text(comment)
     elsif commentable.hidden?
-      "<abbr title='#{deleted_commentable_text(comment)}'>".html_safe +
-      commentable.title +
-      "</abbr>".html_safe
+      "<del>".html_safe + commentable.title.html_safe + "</del>".html_safe + ' ' +
```
bertocq: What do you think about using [content_tag](http://api.rubyonrails.org/classes/ActionView/Helpers/TagHelper.html#method-i-content_tag)? Like:
```
content_tag(:del, commentable.title) + ' ' +
  content_tag(:span, deleted_commentable_text(comment), class: 'small')
```

decabeza: done! 👌🏻😉

PR #1798, opened 2017-08-10
Show bottom proposals button only if proposals exists

artofhuman:
What
====
- if there are no any proposals in responsive design renders two buttons for create proposal too close.
 

How
===
- Render bottom button only if proposals exists

Screenshots
===========

Before
====

screen shot 2017-08-10 at 23 47 02

After
====
screen shot 2017-08-10 at 23 47 15

Deployment
==========
- As usual


bertocq:
Nice catch @artofhuman 👏

PR #1805, opened 2017-08-25
Users proposal list

decabeza:
What
====
- This PR fixes table on users proposal list, sometimes appears different number of `<td>` on each `<tr>` row generating a HTML warning validation.

warning

For example, the proposal lis for a user if have some retired proposal:
user_wrong

For another user viewing the same list:
other_wrong

This PR, adds a `colspan="2"` for the first case:
user_ok

And a new `See proposal` button for the second case:
other_ok

### DIFF
--- a/app/views/users/_proposal.html.erb
+++ b/app/views/users/_proposal.html.erb
@@ -7,26 +7,32 @@
 
   <% if proposal.retired? %>
 
-    <td class="text-center">
+    <td class="text-center" colspan="2">
       <span class="label alert"><%= t('users.proposals.retired') %></span>
     </td>
 
   <% elsif author?(proposal) %>
 
-    <td class="text-center">
+    <td>
       <%= link_to t("users.proposals.send_notification"),
                   new_proposal_notification_path(proposal_id: proposal.id),
-                  class: 'button hollow' %>
+                  class: 'button hollow expanded' %>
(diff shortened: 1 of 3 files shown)


app/views/users/_proposal.html.erb
```diff
@@ -7,28 +7,34 @@
 
   <% if proposal.retired? %>
 
-    <td class="text-center">
+    <td class="text-center" colspan="2">
```
MariaCheca: As I've seen in the screenshots (thanks!) the buttons appear inside the `<td>` aligned to the center and in cases where the element has margin around it looks weird, but what if we use the [`.text-right`](https://github.com/consul/consul/blob/master/app/assets/stylesheets/admin.scss#L64) selector to make it a bit more _clean_? This is just a silly question, you can perfectly discard it 🙃

decabeza: Thanks for the feedback!

I try with the `text-right`but i don't like so much (for the other user I think is better centered). Anyway all the users views needs a good redesign 😅. I add the "proposal retired" to make the label bigger and clarify better to users. 

On the right a little space to the border... 
right

I think is better in the center...
all_users

MariaCheca: So perfect then!

PR #1806, opened 2017-08-25
External links on proposal show

decabeza:
What
====
- Now if the external links are too much long appears a little broken 😅. This PR improves styles for the external links (and video links) on proposal show.

How
===

Also adds a background to highlight the content and a title with the icon to clarify this links to users.

**BEFORE**
external_link_wrong

**AFTER**
external_link_ok


MariaCheca:
P.D. Also I like the design changes you applied, now it says more than before :)

PR #1691, opened 2017-06-24
Budget investment associated image

Senen:
Add image and image_title to budget investments

Where
=====
* **Related Issue:** https://github.com/consul/consul/issues/1600

What
====
Allow author to add, edit or remove an image and image title to own investment projects during new investment creation or later on edit image page. Admisnistrator are able to edit or remove images also.

How
===
Adding image file field and image title field to new investment form.
Creating a new page called "edit_image" to allow author and administrators to change or destroy investment image.

Screenshots
===========

**Budget investment show with image and caption**
budget-investment-show-with-image-and-caption

**Budget investment edit image page**
busget-investment-edit-image-page

Budget investment index with images and placeholders
budget-investments-index-with-images

Budget investment index with images and placeholders (Mobile version)
budget-investments-index-with-images-mobile-version

Test
====
1. Increase investments specs to check validations error for new image fields.
2. Added edit page feature specs. 
3. Added model specs for new validations
4. Increase common and administrator abilities specs to cover permissions

Deployment
==========
Run migrations

Warnings
========
None

Issue #1677, opened 2017-06-21
Images & videos on polls/questions


decabeza:
## Background

When create a new poll/question on Admin section `admin/questions/new` you can select:

- Poll
- Question
- Posibles answers
- Description

## What 

We need include **two more fields** on form:

- [ ] Image(s)
- [x] External video link (same behavior as `proposal/new` video input) (https://github.com/consul/consul/pull/1835)
- [x] 1 PDF (https://github.com/consul/consul/pull/1835)

Something similar to:

`Admin section`
![create_question](https://user-images.githubusercontent.com/631897/27395548-71a34e20-56b0-11e7-91dd-8f3eb232894d.jpg)

And the front for users:

`Polls show`
![question_index](https://user-images.githubusercontent.com/631897/27395556-759a6036-56b0-11e7-848e-3aaf861cefe6.jpg)

`Question show`
![question_show](https://user-images.githubusercontent.com/631897/27395579-86945f86-56b0-11e7-989f-8851c1e58788.jpg)

decabeza:
Closes in favor to #1950 and #1951 🎉 

All about polls on --> #1855

PR #1835, opened 2017-09-06
Poll questions document and video

MariaCheca:
Where
=====
* **Related Issue:** #1677 
* **Related PR:** https://github.com/consul/consul/pull/1691 

What
====
New feature to attach poll questions:
- 1 video link.
- 1 PDF document.

How
===
- Created migration to add poll questions a new field called `video_url`. The poll question create/edit form has a new input to add the video URL and show it in the question public page.

- Using `Documentable`, added the feature to attach a PDF document to the poll question and make it appear in the question public page, same as the video link.

Screenshots
===========
localhost-3000-admin-questions-new
localhost-3000-questions-63 1
localhost-3000-questions-63
localhost-3000-admin-questions-63 1

Warnings
========
- There are some bugs in the screenshots pending to be fixed.
- **The feature is not completed, adding images to poll questions is still left.**


db/schema.rb
```diff
@@ -662,13 +662,28 @@
...
     t.tsvector "tsv"
+    t.string   "video_url"
   end
 
   add_index "poll_questions", ["author_id"], name: "index_poll_questions_on_author_id", using: :btree
   add_index "poll_questions", ["poll_id"], name: "index_poll_questions_on_poll_id", using: :btree
   add_index "poll_questions", ["proposal_id"], name: "index_poll_questions_on_proposal_id", using: :btree
   add_index "poll_questions", ["tsv"], name: "index_poll_questions_on_tsv", using: :gin
 
+  create_table "poll_recounts", force: :cascade do |t|
```
bertocq: This table, as well as the tolk ones, should not have been added. Old table schemas on your local database may have the blame, just:

```bash
rake db:reset
rake db:dev_seed
rake db:migrate
```
And you'll see how it goes back to normal (maybe the tipical change for `t.datetime "password_changed_at"` default.. but that must be ignored until we fix it :D )

app/views/polls/questions/show.html.erb
```diff
@@ -55,11 +55,36 @@
   </div>
 </div>
 
+<% if @question.video_url.present? %>
+  <div class='row margin-top poll-question-show'>
```
decabeza: We always use double quotes `" "` for classes  🤓

app/views/polls/questions/show.html.erb
```diff
@@ -55,11 +55,36 @@
...
+    <div class="small-12 medium-9 column video-link">
```
decabeza: You must move the `<div class="video-link">` inside the `small-12 medium-9 column` to align correctly this block. 👌 

grid_freak

MariaCheca: Nice one! I didn't notice 😅

PR #1854, opened 2017-09-12
Allow only YouTube/Vimeo URLs on 'video_url' attribute

### LINKED ISSUE #1841: Disallow user from introduce no-video URLs in the proposals' video field
Currently the proposals form accepts any string in the `video_url` field. This lets users introduce anything that the platform is not able to process as it always expects a URL to a Youtube or Vimeo video.

So in order to avoid strange behaviours in the UI we should:
- Restrict the string format to URLs matching the video platforms we support.
- In case the user submits an invalid format the form will alert about the error.
- Also and just in case, improve https://github.com/consul/consul/blob/master/app/views/proposals/show.html.erb#L75-L81 to avoid showing `flex-video` box if the `embeeded_video_code` is not a video.

Thanks @bertocq for helping me redacting this issue :)

**BONUS: If you increase the supported platforms to Twitter/Facebook/Instagram/etc... extra kudos for you!**

aitbw:
Where
=====
Fixes #1841

What
====
This PR solves issue #1841, by allowing only valid YouTube/Vimeo URLs if the `video_url` attribute is present.

How
===
A custom validator called `valid_video_url?` was defined on the `Proposal` model to check whether the submitted `video_url` is valid. The method is only executed when the aforementioned attribute is not blank. A constant titled `VALID_DOMAINS` was declared to include all the possible hosts an URL could match.

Screenshots
===========
screenshot-2017-9-12 consul
Here's how the form looks when an invalid URL is submitted

Test
====
Test coverage was slightly increased in order to solve this issue.


app/models/proposal.rb
```diff
@@ -15,6 +15,7 @@ class Proposal < ActiveRecord::Base
   include ActsAsParanoidAliases
 
   RETIRE_OPTIONS = %w(duplicated started unfeasible done other)
+  VALID_DOMAINS = %w(www.youtube.com youtube.com youtu.be www.vimeo.com vimeo.com)
```
bertocq: `VALID_VIDEO_DOMAINS` would be more descriptive as a constant name, what do you think?

app/models/proposal.rb
```diff
@@ -118,6 +121,14 @@ def votable_by?(user)
     user && user.level_two_or_three_verified?
   end
 
+  def valid_video_url?
+    return if video_url.blank?
```
bertocq: I think the constant holding the valid video domains and the body of this function should be shared on a common helper for the near future (other models are for sure getting video urls this next week :) ). Like at https://github.com/consul/consul/blob/master/app/helpers/embed_videos_helper.rb. And this `valid_video_url?` could just be using it.

spec/factories.rb
```diff
@@ -155,7 +155,7 @@
     description          'Proposal description'
     question             'Proposal question'
     external_url         'http://external_documention.es'
-    video_url            'http://video_link.com'
+    video_url            'https://youtu.be/nhuNb0XtRhQ'
```
bertocq: 😸  I was expecting "Gandalf Sax Guy 10 Hours HD" or good old RickRolling ^_^ But cool 👏

spec/models/proposal_spec.rb
```diff
@@ -67,6 +67,13 @@
     end
   end
 
+  describe "#video_url" do
+    it "should not be valid when URL is not from Youtube or Vimeo" do
```
bertocq: Really great to have test for it 👍 💯 . What do you think about testing valid scenarios? Maybe looping through the valid video domains array checking the helper `valid_video_url?` method validates them all?

aitbw: Just for clarification: You mean replacing this test with one where the `valid_video_url?` method tests each domain in the `valid_video_domains` array?

bertocq: Not replacing, but increasing. I mean a test should both check valid and invalid scenarios, right? And in this particular case, the valid scenarios are only a few so checking all of them may not be too slow I think (but I may be wrong and just checking one could be enough)

aitbw: I think one example should do. I created a `.txt` file under `/fixtures/files` with 5 valid URLs to test all the use cases, but according to Knapsack, this is too slow.

Creating an array on the test could work, but looks ugly since we have all these Youtube/Vimeo links.

What do you think?

Issue #1775, opened 2017-08-01
Comments empty HTML


decabeza:
Background
=

When a user create a new comment this also generate an empty HTML code:

```html
<li>
  <ul id="comment_001_children" class="no-bullet comment-children">
  </ul>
</li>
```

Screenshot
=
<img width="1650" alt="children comments" src="https://user-images.githubusercontent.com/631897/28834614-01d7d518-76e3-11e7-8109-2951b272217b.png">

What we need
=

Avoid creating empty HTML code. 

- Only create the new "container" when necessary (a parent comment have children comments).

aitbw:
Since this is already fixed on #1862 and merged into master, we should close this issue.

Cc/ @bertocq @voodoorai2000

PR #1862, opened 2017-09-13
Iagirre comments empty html

iagirre:
Where
=====
* **Related Issue:** #1775

What
====
- Whats the objective of this changes ?

Erase the extra empty HTML code that was generated for comments without any reply.

How
===
- How you implemented/achieved the objective ?

The HTML for comments' replies is only rendered if there are any replies for them. When a new reply is made, it first checks if the container exists. If so, JS adds the reply as normal. If not, the container is created and the reply is added.

Screenshots
===========
- If changes affect UI just show them drag&dropping images here

Not needed

Test
====
- Is manual test needed or you just increased/fixed coverage?

Increased the test coverage for comments#show.

Deployment
==========
- Any details to remember when this feature is deployed?
No

Warnings
========
- Some caveats or important things to notice?
No

### DIFF
--- a/app/views/comments/_comment.html.erb
+++ b/app/views/comments/_comment.html.erb
@@ -93,7 +93,7 @@
         </div>
       <% end %>
     </li>
-    <%unless child_comments_of(comment).empty?%>
+    <% unless child_comments_of(comment).empty? %>
     <li>
       <ul id="<%= dom_id(comment) %>_children" class="no-bullet comment-children">
         <% child_comments_of(comment).each do |child| %>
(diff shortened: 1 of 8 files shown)


app/views/comments/_comment.html.erb
```diff
@@ -93,6 +93,7 @@
         </div>
       <% end %>
     </li>
+    <%unless child_comments_of(comment).empty?%>
```
bertocq: @iagirre to be consistent with the rest of the erb code, spaces before `<%` and after `%>` would be nice 👍

app/views/comments/_comment.html.erb
```diff
@@ -102,5 +103,6 @@
         <% end %>
       </ul>
     </li>
+    <%end%>
```
bertocq: Same here about the erb formatting :)

PR #1868, opened 2017-09-14
Validate presence 'poll_id' attribute on Poll::Question model

### LINKED ISSUE #1831: Validate poll is selected when a question is created
In the admin section, creating a question (`http://localhost:3000/admin/questions/new`) the form allows to submit it and creates a new record without a poll associated. This should be changed and verify a poll is selected to avoid questions without any association, as it doesn't make sense.

aitbw:
Where
=====
Fixes #1831

What
====
Validates presence of `poll_id` attribute on `Poll::Question` model, which previously allowed a question to be created with no association to a poll

How
===
Added the necessary validation on the `Poll::Question` model

Screenshots
===========
screenshot-2017-9-14 admin
Here's how the form looks when a poll is not selected

Test
====
Added tests for both valid and invalid scenarios, also did a small refactor to improve DRYness for the related spec

### DIFF
--- a/app/models/poll/question.rb
+++ b/app/models/poll/question.rb
@@ -20,6 +20,7 @@ class Poll::Question < ActiveRecord::Base
 
   validates :title, presence: true
   validates :author, presence: true
+  validates :poll_id, presence: true
 
   validates :title, length: { minimum: 4 }
   validates :description, length: { maximum: Poll::Question.description_max_length }
--- a/spec/features/admin/poll/polls_spec.rb
+++ b/spec/features/admin/poll/polls_spec.rb
@@ -183,7 +183,7 @@ feature 'Admin polls' do
 
       scenario 'Add question to poll', :js do
         poll = create(:poll)
-        question = create(:poll_question, poll: nil, title: 'Should we rebuild the city?')
+        question = create(:poll_question, title: 'Should we rebuild the city?')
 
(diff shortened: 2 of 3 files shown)


bertocq:
@aitbw only one problem with tests, I've run three times locally `spec/features/admin/poll/polls_spec.rb` and it always fails at:

```
  1) Admin polls Questions Poll show Add question to poll
     Failure/Error: question = create(:poll_question, poll: nil, title: 'Should we rebuild the city?')

     ActiveRecord::RecordInvalid:
       Validation failed: Poll can't be blank
     # ./spec/features/admin/poll/polls_spec.rb:186:in `block (4 levels) in <top (required)>'
```
Because of the `poll: nil` at https://github.com/wairbut-m2c/consul/blob/aperez-validate-poll-question-is-selected/spec/features/admin/poll/polls_spec.rb#L186

aitbw:
I'll fix it right away. Should I fix the issue, rebase and update the PR or a new commit would be better, @bertocq?

bertocq:
@aitbw I think in the same commit that you were expecting poll not to be nil that fix could be as well in context of the rest of the commit, right? "Same commit" I mean :)

aitbw:
The failing spec is fixed, so you shouldn't have any problems with it, @bertocq :D

Issue #1851, opened 2017-09-11
Display Datepicker arrows


bertocq:
At least on the Polls edit page when you popup the datepicker trying to edit a date, the next/prev month arrows icons aren't displaying

![screen shot 2017-09-11 at 22 42 52](https://user-images.githubusercontent.com/983242/30295939-bbac4706-9742-11e7-9a2a-7f5318ff0e64.jpg)

iagirre:
I'll take this one.

decabeza:
Closed by #1869

PR #1869, opened 2017-09-14
Iagirre display datepicker arrows

iagirre:
Where
=====
* **Related Issue:** #1851 

What
====
Make the arrows of the datepicker for changing the month appear. There was a link and it was working, but there were no icons, so the link was invisible.

How
===
There was a CSS file (`jquery-ui/theme` from `jquery-ui`) that wasn't been loaded. It was supposed to be loaded automatically when using datepicker widget, but the include wasn't done using Sass, therefore, there was no file. I included it in application.scss, and I also included a custom class to make the icons white (by default they were dark grey).
This issue explains it: https://github.com/jquery-ui-rails/jquery-ui-rails/issues/89

Screenshots
===========
calendario

Test
====
Not apply

Deployment
==========
Not apply

Warnings
========
Not apply


db/schema.rb
```diff
@@ -974,7 +974,7 @@
     t.boolean  "email_digest",                              default: true
     t.boolean  "email_on_direct_message",                   default: true
     t.boolean  "official_position_badge",                   default: false
-    t.datetime "password_changed_at",                       default: '2017-09-06 18:19:39', null: false
+    t.datetime "password_changed_at",                       default: '2017-06-22 11:21:30', null: false
```
bertocq: Hey @iagirre this shouldn't go into any PR, it's a problem we've because of https://github.com/consul/consul/blob/master/db/migrate/20161102133838_default_password_changed_at.rb#L3 but haven't stopped yet to fix it

bertocq: Is one of those things that we just dodge normally instead of fix... but now it is fixed forever :) https://github.com/consul/consul/commit/e3a45baa37b60e49a0bc349f1609d7931248832a

decabeza:
Hi @iagirre! thanks for the PR! some comments by file:

**`application.scss`**

```
@import 'jquery-ui/theme';
@import 'jquery-ui-theme_overrides';
````
This imports are not necessary, already there are `@import 'jquery-ui/datepicker'`  for original styles and `@import 'datepicker_overrides'` to override them.

**`datepicker_overrides.scss`**

You write:
```sass
line-height: $line-height;
top: 0;
position: absolute;
top: 2px;
width: 1.8em;
height: 1.8em;
```
Beware to repeat attributes (first `top:0` and after `top:2px`). We use absolute units (px) for absolute elements or borders, for others such as width & height it's better use `rem-calc` mixin included on foundation (i.e. `width: rem-calc(30)` where 30 is the number of pixels) 😉 

**`jquery-ui-theme_overrides.scss.erb`**

We don't like `scss.erb` and personally I think is a dangerous and unnecessary mix 😱 After a review, I think it's better to reuse consul icons (`icons.scss`) instead of external images... so a little help (this mini PR has a tricky part) 😅

You can simply **replace and add** to `datepicker_overrides.scss:43`

```sass
.ui-datepicker-prev,
.ui-datepicker-next {
    color: #fff;
    cursor: pointer;
    font-family: "icons" !important;
    font-size: rem-calc(24);
    font-weight: normal;
    height: rem-calc(30);
    line-height: $line-height;
    position: absolute;
    top: 4px;
    width: rem-calc(30);

    &:hover {
      text-decoration: none;
    }
  }

.ui-datepicker-prev::after {
  content: '\62';
}

.ui-datepicker-next::after {
  content: '\63';
}
```

Also you can **delete L36 - L42** to improve margins:

```sass
.ui-datepicker-prev {
  left: 12px;
}

.ui-datepicker-next {
  right: 12px;
}
```

Sorry for the long comment, I try to explain each point in the better way 😄

app/assets/stylesheets/datepicker_overrides.scss
```diff
@@ -32,22 +32,30 @@
...
     cursor: pointer;
+    font-family: "icons" !important;
+    font-size: rem-calc(24);
     font-weight: normal;
-    font-size: $small-font-size;
+    height: rem-calc(30);
     line-height: $line-height;
-    top: 0;
+    position: absolute;
+    width: 4px;
```
decabeza: I think this line is `top: 4px` ;)

PR #1874, opened 2017-09-15
When "Profiles" submenus are selected, the "Site custom" menu doesn't…

iagirre:
… open. Moreover, when "Site custom" submenus are selected, the menu remains opened.

Where
=====
* **Related Issue:** #1871 

What
====
When, in admin menu, one of the profile submenus where selected, the "Site custom" menu was opened. Also, when one of the customization submenus was clicked, the "Site custom" menu was closed, even if the submenu was selected.

How
===
I created a new helper method to check if the user clicked one of the "Site customization" menu's category. If so, the menu remains active. If them clicked another option, the customization menu closes.

Screenshots
===========
First: The menu remains opened when clicked a submenu.

menu_custom

Second: The menu remains closed when clicked a profile submenu.

menu_profiles

Test
====
Not apply

Deployment
==========
Not apply

Warnings
========
Not apply

### DIFF
--- a/app/views/admin/_menu.html.erb
+++ b/app/views/admin/_menu.html.erb
@@ -158,8 +158,8 @@
         <span class="icon-settings"></span>
         <strong><%= t("admin.menu.title_site_customization") %></strong>
       </a>
-      <ul <%= "class=is-active" if menu_profiles? %>>
-        <li <%= "class=active" if controller_name == "pages" %>>
+        <ul <%= "class=is-active" if menu_customization? %>>
+            <li <%= "class=active" if controller_name == "pages" %>>
           <%= link_to t("admin.menu.site_customization.pages"), admin_site_customization_pages_path %>
         </li>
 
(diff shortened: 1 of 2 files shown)


app/views/admin/_menu.html.erb
```diff
@@ -158,8 +158,8 @@
         <span class="icon-settings"></span>
         <strong><%= t("admin.menu.title_site_customization") %></strong>
       </a>
-      <ul <%= "class=is-active" if menu_profiles? %>>
-        <li <%= "class=active" if controller_name == "pages" %>>
+        <ul <%= "class=is-active" if menu_customization? %>>
```
bertocq: @iagirre have you configured your Netbeans with editorconfig plugin? https://github.com/welovecoding/editorconfig-netbeans#readme that <ul> and <li> indentation seems off compared to the rest of the html

app/views/admin/_menu.html.erb
```diff
@@ -158,8 +158,8 @@
...
+            <li <%= "class=active" if controller_name == "pages" %>>
```
decabeza: I don't know if these blank spaces before (L161-162) are necessary for a correct indenting ☺️

app/views/admin/_menu.html.erb
iagirre: Yes, I configured it, but there must be something wrong with it. I think it's OK now.

PR #1877, opened 2017-09-15
Improve A11y when using tab key

### LINKED ISSUE #1850: Accessibility focus message
Background
==

When users not logged try to support a proposal appears a message to sign in or sign up to continue.

The problem is if **the user navigate with the keyboard using tab key**, the message appears on _focus_ but when _blur_ this message keep showing, so... the user can't view the supports info.

Screenshot
==
participation not allowed

What we need
==

The message has to appear on `mouseover/focus` with tab, allow navigate on `sign in/ sign up links` and hide again on `mouseleave/blur`.

The desired behaviour is: 😌 
(The user should be navigate over the links with the keyboard and after tab over a proposal and tab over next one keep seeing the number supports info)

tabs_desired

aitbw:
Where
=====
Fixes #1850

What
====
Before this change, when using the `tab` key on either Proposals or Debates pages **and** the user wasn't logged in, the "You need to sign up or log in to continue" message box wouldn't disappear unless the user hovered over said message box with the mouse, affecting the UX. This commit makes the aforementioned box to hide when focus is lost, whether the mouse or the `tab` key is used.

How
===
Added the missing `blur`[1] event on the `app/assets/javascripts/votes.js.coffee` file.

Test
====
Since this is a UX-related fix, manual testing is needed in order to see the changes.

Notes
====
[1] [blur (event)](https://developer.mozilla.org/en-US/docs/Web/Events/blur) on MDN

### DIFF
--- a/app/assets/javascripts/votes.js.coffee
+++ b/app/assets/javascripts/votes.js.coffee
@@ -5,7 +5,7 @@ App.Votes =
       'mouseenter focus': ->
         $("div.participation-not-allowed", this).show()
         $("div.participation-allowed", this).hide()
-      mouseleave: ->
+      'mouseleave blur': ->
         $("div.participation-not-allowed", this).hide()
         $("div.participation-allowed", this).show()
     }, votes


decabeza:
Hi @aitbw! thanks for the PR! 

The message disappear 👍  but the user can't tab over "sign in/sign up" links 😞

With this changes the behaviour is:

tabs_now

But the desired behaviour is: 😌 
(The user should be navigate over the links with the keyboard and after tab over a proposal and tab over next one keep seeing the number supports info)

tabs_desired

aitbw:
Yeah, I noticed that right after I submitted the Pull Request. I'll fix it and update the PR accordingly.

aitbw:
This is going to take longer than I expected 😅 The `blur` event added for this PR is what's causing the new (incorrect) behavior, but removing it takes us to the old (also incorrect) behavior.

From what I've researched, it seems that we're approaching this in the wrong way. Apparently, we're focusing on 2 elements **at the same time** and obviously we can't do that.

Any ideas, @decabeza?

decabeza:
Hi @aitbw! thanks again for the PR! I updated the issue https://github.com/consul/consul/issues/1850 with more descriptive behaviour, I close this PR at the moment.

PR #1882, opened 2017-09-18
Proposal show

decabeza:
What
====
- Now the actions for user on their own proposals shows before the title without any space.

How
===
- Moves this buttons to the proposal/show sidebar.
- Creates a helper to show this actions only to users/admins with permissions.
- Also improves margins for tags and geozone tags on proposal show.

Screenshots
===========
**Buttons before** _OMG!_
ugly_buttons

**Buttons after** 😌 
buttons_sidebar

**Tags before** 😬 
tags before

**Tags after** 😄 
tags_after

### DIFF
--- a/app/helpers/proposals_helper.rb
+++ b/app/helpers/proposals_helper.rb
@@ -32,16 +32,16 @@ module ProposalsHelper
     Proposal::RETIRE_OPTIONS.collect { |option| [ t("proposals.retire_options.#{option}"), option ] }
   end
 
-  def can_create_document?
-    can?(:create, @document) && @proposal.documents.size < Proposal.max_documents_allowed
+  def can_create_document?(document, proposal)
+    can?(:create, document) && proposal.documents.size < Proposal.max_documents_allowed
   end
 
-  def author_of_proposal?
-    author_of?(@proposal, current_user)
+  def author_of_proposal?(proposal)
+    author_of?(proposal, current_user)
   end
 
-  def current_editable?
-    current_user && @proposal.editable_by?(current_user)
+  def current_editable?(proposal)
+    current_user && proposal.editable_by?(current_user)
   end
 
 end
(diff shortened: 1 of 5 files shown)


app/helpers/proposals_helper.rb
```diff
@@ -32,4 +32,20 @@ def retire_proposals_options
     Proposal::RETIRE_OPTIONS.collect { |option| [ t("proposals.retire_options.#{option}"), option ] }
   end
 
+  def can_create_document?
+    can?(:create, @document) && @proposal.documents.size < Proposal.max_documents_allowed
```
bertocq: The use of `@document` and `@proposal` at lines 40 and 44 seems strange to me. I mean, this is a helper that share functions across multiple files/views/etc... and this functions assume that the context where they are called will have a variable with that exact name and the right content inside... 

Yes it makes everything quite clean at the view :), but not sure its a good idea from a code point of view (neither I've seen this as a pattern on other helpers of this codebase)

I would go with good old parameters on each function for those 3.

app/helpers/proposals_helper.rb
```diff
@@ -32,4 +32,20 @@ def retire_proposals_options
...
+
+  def author_of_proposal?
+    author_of?(@proposal, current_user)
+  end
+
+  def current_editable?
+    current_user && @proposal.editable_by?(current_user)
+  end
+
+  def show_actions_menu?
```
bertocq: I get why you refactored this but thinking again that this is a Helper common to other Proposal stuff, and that `show_actions_menu` is not very specific about (and doesn't seem to be candidate to be reused on other parts... just on that proposal show view) I think leaving that 3 OR's on the view would be better to easily check there when the html will be rendered. It may be a very subjective opinion but for me if its not reusable, or complex logic then there is no need to go on a helper.

bertocq:
Flaky failures, no worries 👏

Issue #1797, opened 2017-08-09
Switch flag/unflag buttons on use via ajax


bertocq:
Right now when you flag a proposal, to get the "unflag" button you need to refresh the page.

https://github.com/consul/consul/blob/master/app/views/proposals/_flag_actions.html.erb has the condition, but is not being refreshed after use of the actions.

Ideally after using a button, the opposite action button should be available replacing it, via ajax like the follow/unfollow button does with https://github.com/consul/consul/blob/master/app/views/follows/refresh_follow_button.js.erb

xarlybovi:
I'll be glad to help you with this one

bertocq:
Awesome @xarlybovi go for it 👏 😄

aitbw:
@xarlybovi submitted #1883 which was merged successfully. Let's close this one, shall we? ;-)

PR #1883, opened 2017-09-18
Switch flag/unflag buttons on use via ajax

xarlybovi:
Where
=====
* **Related Issue:** #1797

What
====
- Refresh proposal flag/unflag button via AJAX

How
===
- Implemented like follow/unfollow button and reloaded foundation in order to fix dropdown on AJAX partial render

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -60,6 +60,7 @@
 //= require legislation_annotatable
 //= require watch_form_changes
 //= require followable
+//= require flagable
 //= require documentable
 //= require tree_navigator
 //= require custom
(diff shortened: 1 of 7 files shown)


voodoorai2000:
This is great @xarlybovi!

A small detail, the flaggable concern is used by another two models, `debates` and `comments`.

Could you use `resource_id` instead of `proposal_id` and add a couple of specs to make sure these changes fix the problem?

xarlybovi:
Yeah, sure!

xarlybovi:
@voodoorai2000 Done.

voodoorai2000:
Great! Appreciate it @xarlybovi

PR #1884, opened 2017-09-18
Added .gitignore rules for Jetbrains's IDE

xarlybovi:
Where
=====
* **Related Issue:** None

What
====
- Environment setup for Jetbrains's IDE (Rubymine...)

How
===
- Add .gitignore rules for .idea folder

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,6 @@ public/system/
 
 #Netbeans projects files
 /nbproject
+
+# Jetbrains project files
+.idea


aitbw:
I think IDE/text editor-specific files shouldn't be added into Consul's .gitignore. You should handle this on your own by adding these files under `$HOME/.gitignore_global`.

bertocq:
@aitbw we talked about this this afternoon and agree (after also checking the .gitignore from otherOSS projects) that its not something that should be added.

But I understand @xarlybovi reasons, seeing that two lines up @iagirre added the netbeans folder and it was my fault to ignore that change and focus on the rest of [the PR](https://github.com/consul/consul/pull/1862). So to be consistent [I've removed it](https://github.com/consul/consul/commit/ba198726f25c266d9aa4a0d54064a2416033aa2e)

About the global gitignore file here is simple github guide to create it: https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

And an example https://github.com/bertocq/dotfiles/blob/master/.gitignore_global and its usage at `.gitconfig` file https://github.com/bertocq/dotfiles/blob/master/.gitconfig#L13

Again I'm sorry @xarlybovi about the confusion, I hope you agree with the solution.

voodoorai2000:
+1 to using `.gitignore_global` for local environment files

xarlybovi:
Thanks for the explanation. I didn't know the global .gitignore config, neat trick!

bertocq:
Thanks again @xarlybovi for your contribution :) And great to see it helped you learn something about git :)

Issue #1838, opened 2017-09-07
Refactor Residence Verification


bertocq:
## Context

We've a custom model [Verification::Residence](https://github.com/consul/consul/blob/master/app/models/custom/verification/residence.rb#L13) that's being used to always check the user's zip_code matching Madrid's zip code pattern.

This is not good because:
- Forks have to overwrite that `280` leading numbers on the zip_code pattern (as well as `madrid` name on the consul codebase)
- ZipCode check doesn't make sense for CONSUL installations which audience is not just a city but organizations, political parties, etc... but other types of custom verifications.

## What

We need to make CONSUL easier to customize on those features that aren't mandatory but should be optional. The proposed changes to make this better are:

- Code should not have `madrid` name hardcoded.
- The `280` zip code pattern shouldn't be hardcoded, but configurable as a setting value.
- The custom Residence Verification should be possible to be turned off completely using a Settings value (instead of having to totally remove the file and specs).

zamuro:
I'll take this one.

zamuro:
Hi, guys

I've been thinking on using a yml file in which users will check against. Of course, that would need for the devs to create the yml file with the zipcodes they will be using. Something like this gem [1]

What do you think?

[1]: https://github.com/monterail/zip-codes

javierm:
Reopening since this one definitely needs to be implemented at some point, and we've been (intermittenly) working on issues related to verification in the last year and a half.

PR #1896, opened 2017-09-20
Mlucena refactor residence verification

zamuro:
Where
=====
* **Related Issue:** #1838

What
====
- To enable/disable residential verification process for forks

How
===
- Created configuration option on _seeds.rb_ that can be set to true or nil whether feature is needed or not
- Created validation on _verification_controller.rb_ that checks said configuration and act accordingly by redirecting to account page if set to false or initiating verification process if set to true
- Disabled accesing via address bar to residence verification, redirecting to account show and displaying an alert.

Screenshots
===========
captura de pantalla_2017-09-21_15-31-31

Test
====
- no manual test is needed. Feature spec was changed so it does not run residence verification if setting is not present, then it only verifies that it redirects to SMS verification instead.

Deployment
==========
- By default is set to true.

Warnings
========
- If set to nil, manual rake db:seed need to be performed.


voodoorai2000:
Thank you @zamuro, but this issue has become obsolete 🙏
We already have a way to skip user verification[1]

[1] https://github.com/consul/consul/pull/2399

PR #1904, opened 2017-09-21
Community design

decabeza:
What
====
- Improves HTML/CSS for community views.

How
===
- Adds consistency on headings tags, links and buttons.
- Move author actions to topic/show.
- Improves i18n.

Screenshots
===========
**Community show BEFORE**
community_show_before

**Topic show BEFORE**
topic_show_before

**Community show AFTER**
community_show_after

**Topic show AFTER**
topic_show_after


bertocq:
I've got pending task at https://github.com/consul/consul/compare/community-design-improvement?expand=1 currently checking if solution would break too much specs hoping for the "bad behaviour" on other features 🙈  we'll go with if-else link generation if so.. and cleanup things in a separate PR with that "more elegant" fix

Issue #1703, opened 2017-06-26
Autocomplete tags


decabeza:
## Background

Now when create a new proposal appear an alert box of proposals with the same content (for avoid duplicating)

`This alert prevent duplicated content.`
<img width="889" alt="proposals_alert" src="https://user-images.githubusercontent.com/631897/27549043-0d7b296a-5a9b-11e7-9e54-9c5369c39814.png">

## What

The idea is include a similar behavior to the tags input field (the users can write whatever they want), but if they use the tags already used by others improve the searches for the rest of users. For example, instead of a new tag `motorcycle` it's better if anyone already used `motorbike`

This is needed for tags input field on:

- New debate
- New proposal
- New investment projects

`Example`
<img width="910" alt="tag_suggest" src="https://user-images.githubusercontent.com/631897/27549294-c96ad9ea-5a9b-11e7-9dbf-0421a5a00d8b.png">

iagirre:
I'll take this one.

bertocq:
Closed via #1905

PR #1905, opened 2017-09-21
Iagirre autocomplete tags

iagirre:
Where
=====
* **Related Issue:** #1703 

What
====
When the user is writting the tags for a proposal, a debate or an investment project, the page should suggest other tags that already exist so that users can select those tags instead of creating new and similiar ones (e.g. motorbike and motorcycle).

How
===
I used the autocomplete widget from `jquery-ui` to do it. When the user is writing, the widget looks for similar words in the DB using the `pg_search` gem. The gem retrieves the results and they are shown in a list under the text field. The user can select one of the results using the mouse or the down and up arrows on the keyboard. When an option is selected, it is automatically added to the list.

Screenshots
===========
autocomplete_tags

Test
====
I increased the test coverage for the tags' search function.

Deployment
==========
Not to apply

Warnings
========
Not to apply

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -14,6 +14,7 @@
 //= require jquery_ujs
 //= require jquery-ui/widgets/datepicker
 //= require jquery-ui/i18n/datepicker-es
+//= require jquery-ui/widgets/autocomplete
 //= require jquery-fileupload/basic
 //= require foundation
 //= require turbolinks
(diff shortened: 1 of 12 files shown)


decabeza:
Hi @iagirre! Thanks for the PR! 👍 

For the scss file `autocomplete_overrides.scss` please run `scss-lint` on console to check some minor erros (don't care about other scss, only included on this PR) 😉

Issue #1901, opened 2017-09-21
Budget related entries in Management shouldn't appear if budgets are globally disabled


periflo90:
We've found something interesting in the Management panel. There are many entries in the left sidebar which reference budgets. For the time being, we have globally disabled the budgets feature.

Having disabled it, when accessing a budget-related feature, it gives the following Rails error:

```
FeatureFlags::FeatureDisabled in Management::BudgetsController#create_investments
Feature disabled: budgets
```

I suggest those sidebar entries not to appear if budgets are globally disabled, because the product may seem weak to the end user.

periflo90:
I could try. I'll keep you posted.

periflo90:
Do you mean disabling them as in checking other entries which should throw that error and making them throw it, or checking if the feature is enabled and if false not rendering the menu items?

aitbw:
From what I can understand, you should check if the feature is enabled and if `false`, don't render "Budgets" on the menu, @periflo90.

Feel free to see #1903 to see how this can be achieved in full.

periflo90:
Okay, done. I just made a pull request. Hope everything's okay.

bertocq:
Closed via #1941 👏

PR #1941, opened 2017-09-29
Issue #1901: Budget related entries in Management shouldn't appear if budgets are globally disabled

periflo90:
Where
=====
* **Related Issue:** #1901

What
====
Checks if the Budget setting is enabled, and if not, does not render its related entries in the Management menu.

How
===
Three if clauses which check against the Budgets setting.

Screenshots
===========
screenshot


decabeza:
Hi @periflo90! Thanks for the PR! 👍 

I think you can use only one `<% if Setting['feature.budgets'] %>` to wrap the three links related to budgets. Feel free to move them 😉

periflo90:
@decabeza I could wrap the first two list items under the same if clause, but the last one had two other items in between, so I left that one under a different if clause.

decabeza:
As I indicated, you can move al links inside only one if 😉 

```ruby
<% if Setting['feature.budgets'] %>
      <li <%= "class=active" if (controller_name == "budget_investments" && action_name == "new") ||
                                (controller_name == "budget" && action_name == 'create_investments') %>>
        <%= link_to create_investments_management_budgets_path do %>
          <span class="icon-budget"></span>
          <%= t("management.menu.create_budget_investment") %>
        <% end %>
      </li>

      <li <%= "class=active" if (controller_name == "budget_investments" && action_name == "index") ||
                                (controller_name == "budget" && action_name == "support_investments")%>>
        <%= link_to support_investments_management_budgets_path do %>
          <span class="icon-like"></span>
          <%= t("management.menu.support_budget_investments") %>
        <% end %>
      </li>

      <li <%= "class=active" if (controller_name == "budget_investments" && action_name == "print") ||
                                (controller_name == "budgets" && action_name == "print_investments") %>>
        <%= link_to print_investments_management_budgets_path do %>
          <span class="icon-print"></span>
          <%= t("management.menu.print_budget_investments") %>
        <% end %>
      </li>
    <% end %>
```

periflo90:
But, wouldn't that affect the order of the menu items? Would that be OK with you? If so, I could do it.

decabeza:
It's better follow DRY principle than the order menu 😛 moreover I think also it's better than all the related items are together.

So yes, you can change the order 😉 

Thanks! 🙏

periflo90:
Okay, done. I hope everything's okay.

PR #1960, opened 2017-10-02
Requested polls admin section improvements

MariaCheca:
What
====
- Removed link leading to poll question in the web in admin section. https://github.com/consul/consul/commit/1e0aec37abbeafdb074290dbc93f31aab4c16df8
- Fixed officer search action. Before, when a user searched for an officer already assigned the interface was blocked because the officer path rendered wasn't correct. https://github.com/consul/consul/commit/14fe771fa55296f57e765653d28a358cafb781a5
- Booths and shifts index pages minor fixes. Showing only actions to assign and edit shifts in shifts index, and option to create a new booth in booths index. https://github.com/consul/consul/commit/fd1d09c8c23efda9082057d02a5a10c41b85ad06
- Removed searchbox from poll questions tab and unassign button. https://github.com/consul/consul/commit/adf18ee75635928fce0b8f029b024aa9b160829f
- Removed actions offered in poll officers search results. https://github.com/consul/consul/commit/2365ec96fa4c74e7a1dfeec9cf729f522b203f95
- Removed total results table from officer assignment as now there are only final results. https://github.com/consul/consul/commit/2c72a054acc2e9f27994b1ecc181c75b22298f6c
- Poll booth preview now renders recounts tab by default instead of the officers one. https://github.com/consul/consul/commit/799e2475da66dfe90d93b2528572d01c1480e512

Test
====
As usual.

Deployment
==========
As usual.


app/views/admin/poll/polls/_questions.html.erb
```diff
@@ -20,10 +20,6 @@
           </strong>
         </td>
         <td class="text-right">
```
decabeza: I think in this file you should delete this empty `<td class="text-right"></td>` and the unnecessary `<th>` on L12: `<th class="text-right"><%= t('admin.polls.show.table_assignment') %></th>` 😉

app/views/admin/poll/polls/show.html.erb
```diff
@@ -3,6 +3,5 @@
 <div id="poll-resources">
   <%= render "subnav" %>
 
-  <%= render "search_questions" %>
```
bertocq: I think this was the only usage of `search_questions`, if we're not planning on using it in the near future there is things pending to be cleaned up:

/app/views/admin/poll/polls/search_questions.js.erb
/app/views/admin/poll/polls/_search_questions.html.erb
/app/views/admin/poll/polls/_search_questions_results.html.erb

https://github.com/consul/consul/blob/polls_minor_changes/config/routes.rb#L276
https://github.com/consul/consul/blob/polls_minor_changes/app/models/abilities/administrator.rb#L59
https://github.com/consul/consul/blob/polls_minor_changes/app/controllers/admin/poll/polls_controller.rb#L4
https://github.com/consul/consul/blob/polls_minor_changes/app/controllers/admin/poll/polls_controller.rb#L50-L55

Issue #1948, opened 2017-09-29
Polls: Admin section


decabeza:
What we need
==

- [ ] Add a `summary` input for poll.
- [ ] Add a `description` input for poll.
- [ ] Add a `main image` (with `alt text`) input for poll.

Screenshot
==
![ok_polls_admin_1](https://user-images.githubusercontent.com/631897/31026608-6762a452-a547-11e7-9892-1a87977874c9.jpg)

PR #1962, opened 2017-10-02
Iagirre admin polls section

iagirre:
Where
=====
* **Related Issue:** #1948 

What
====
Add summary, description and a main image for polls in the admin section.

How
===
I created a migration for adding summary and description columns to polls table. I also added those fields in the create/edit form, so that they could be filled when creating/editing.
I made the polls model `Imageable` and added the form fields to add an image to the poll. The partial for adding the image is slightly different to mach the mockup given in the issue.

Screenshots
===========
edit_create_poll

edit_create_poll_2

Test
====
Tests added for polls creation to check if the poll is created when filling summary and description.

Deployment
==========
There is a new migration file, to create summary and description columns in polls table.

Warnings
========
Not apply.


spec/features/admin/poll/polls_spec.rb
```diff
@@ -58,6 +58,8 @@
     fill_in "poll_name", with: "Upcoming poll"
     fill_in 'poll_starts_at', with: start_date.strftime("%d/%m/%Y")
     fill_in 'poll_ends_at', with: end_date.strftime("%d/%m/%Y")
+    fill_in 'poll_summary', with: "Upcoming poll's summary. This poll..."
+    fill_in 'poll_description', with: "Upcomming poll's description. This poll..."
```
voodoorai2000: Missing the corresponding expectations for this input fields. 
The spec related to what a non-admin user sees, should also take these new fields into account.

voodoorai2000: Missing image specs

app/views/admin/poll/polls/_form.html.erb
```diff
@@ -20,6 +20,22 @@
   </div>
 
   <div class="row">
+    <div class="small-12 column">
+      <%=f.text_area :summary, rows: 4%>
```
voodoorai2000: Space after `<%=`

app/views/images/_admin_image.html.erb
```diff
@@ -0,0 +1,45 @@
+<div class="images small-12 column">
+    <div>
+      <%= f.label :image, t("images.form.admin_title") %>
+
```
voodoorai2000: Extra newline

app/views/admin/poll/polls/_form.html.erb
```diff
@@ -20,6 +20,22 @@
...
+    </div>
+  </div>
+
+  <div class="row">
+    <div class="small-12 column">
+      <%=f.text_area :description, rows: 8%>
```
bertocq: Same on this one, I fear we don't have a erb style checker to automate this corrections :( but a space after `<%=` and before `%>` (on previous comment as well) is always welcomed to make it easier to read 👍

config/locales/en/images.yml
```diff
@@ -9,6 +9,9 @@ en:
       delete_button: Remove image
       note: "You can upload one image of following content types: %{accepted_content_types}, up to %{max_file_size} MB."
       add_new_image: Add image
+      title_placeholder: Add a descriptive title for the image
```
bertocq: This `title_placeholder` is already present at line 7 of this file, maybe the fact that [the `es` version](https://github.com/wairbut-m2c/consul/blob/master/config/locales/es/images.yml) has a different order tricked you into adding it on the same position. I would go with the same order that the `es` version has to unify both 👍

voodoorai2000:
Awesome! Thank you @iagirre

Issue #1954, opened 2017-09-29
Polls: questions


decabeza:
Background
==

With the new Poll views now all the info is on `poll/show` view, so... some old views are no needed anymore.

What we need
==

- Remove or hide `questions/show` (maybe this urls can redirect to polls/show)
- Remove or hide `question description` (now we use new answer description)

MariaCheca:
Closed here: https://github.com/consul/consul/pull/1972

PR #1972, opened 2017-10-03
Removed poll question show view and description attribute

MariaCheca:
Where
=====
* **Closes:** https://github.com/consul/consul/issues/1954

What
====
- Removed Poll Question show view and references to its path in the code.
- Removed Poll Question `description` attribute.

Test
====
As usual.

Deployment
==========
As usual


spec/features/polls/questions_spec.rb
```diff
@@ -11,121 +11,4 @@
 
     expect(proposal_question.title).to appear_before(normal_question.title)
   end
-
```
voodoorai2000: Well noticed 👌

Issue #1983, opened 2017-10-04
Polls - Remove documents from questions


voodoorai2000:
What
===
- Remove documents from questions
- Doble check question documents are not displayed in a poll's show view

Why
===
- Documents now belong to the Answers model, making it easier to group documents for each Answer

Screenshots
===
<img width="1070" alt="remove documents from questions" src="https://user-images.githubusercontent.com/4169/31202004-12ec2c4a-a961-11e7-80e2-3f56c12cc77c.png">

voodoorai2000:
👍

PR #1986, opened 2017-10-05
Iagirre remove documents questions

iagirre:
Where
=====
* **Related Issue:** #1983

What
====
Remove the document attachment from admins questions creation/edition.

How
===
- Removing the file uploader from the form.
- Making the question model not documentable (removing the `include Documentable` and the rest of the configuration involved).
- Removing the html that showed the documents in admins questions.
- Modifying the specs to make them pass.
- The i18n keys related to documents and questions where also deleted.

Screenshots
===========
remove question documentable

Test
====
The test were modified to remove the documentable part from questions, because question model is no longer documentable.

Deployment
==========
Nothing to apply.

Warnings
========
Nothing to apply.

### DIFF
--- a/spec/features/admin/poll/questions_spec.rb
+++ b/spec/features/admin/poll/questions_spec.rb
@@ -111,22 +111,4 @@ feature 'Admin poll questions' do
 
   pending "Mark all city by default when creating a poll question from a successful proposal"
 
-  it_behaves_like "nested documentable",
-                  "administrator",
-                  "poll_question",
-                  "new_admin_question_path",
-                  { },
-                  "documentable_fill_new_valid_poll_question",
-                  "Save",
-                  "Star Wars: Episode IV - A New Hope"
-
-  it_behaves_like "nested documentable",
-                  "administrator",
-                  "poll_question",
-                  "edit_admin_question_path",
-                  { "id": "id" },
-                  nil,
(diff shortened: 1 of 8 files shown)


spec/features/admin/poll/questions_spec.rb
```diff
@@ -111,22 +111,4 @@
 
   pending "Mark all city by default when creating a poll question from a successful proposal"
 
-  it_behaves_like "nested documentable",
```
MariaCheca: Hi! I think the methods that after this deletion are not used anywhere else should be removed:
https://github.com/consul/consul/blob/master/spec/shared/features/nested_documentable.rb#L322

spec/features/admin/poll/questions_spec.rb
```diff
@@ -111,22 +111,4 @@
...
-                  "administrator",
-                  "poll_question",
-                  "new_admin_question_path",
-                  { },
-                  "documentable_fill_new_valid_poll_question",
```
MariaCheca: Hi! I think the methods that after this deletion are not used anywhere else should be removed:
https://github.com/consul/consul/blob/master/spec/shared/features/nested_documentable.rb#L322

voodoorai2000:
This is great, thanks @iagirre 👏

Issue #1985, opened 2017-10-04
Polls - Voting identifier


voodoorai2000:
What
===
- When voting (as soon as one of the options is marked, even if you do not answer the other questions in a poll) we must generated **in the client side** a random identifier (a string of letters and numbers long enough to ensure that no two are repeated with millions of votes). 

- The following message should be displayed in small format:
"You can write down this vote identifier, to check your vote on the final results."
("Puedes apuntar este identificador de voto, para comprobar tu votación en el resultado final".)

- This string must be saved in db associated to the vote in that poll. 
- We should generate different identifiers for each poll in which the user participates.

Why
===
So that at the end of the voting we can publish a csv with a field with votes, and another field with the identifiers, so that people can see that their vote has been properly collected.

bertocq:
```
var rand = function() {
    return Math.random().toString(36).substr(2); // remove `0.`
};

var token = function() {
    return rand() + rand(); // to make it longer
};

token();
```
Seems good enough? each `Math.random().toString(36).substr(2);` returns 11 chars 

https://stackoverflow.com/a/8532436

But I'm concerned about the fact that we'll be shoving into the user a string too long (like a WIFI password, but without upcase letters :D )

PR #1994, opened 2017-10-06
Voter Answer unique Token

bertocq:
Where
=====
* **Related Issue:** https://github.com/consul/consul/issues/1985

What
====
We need to generate a unique voting token for each time a Voter registers an answer so he can be sure in the future it was taken into the total scrutiny.

Also `rake db:dev_seed` was broken because:
- It was trying to set a description to the Poll::Question model that doesn't have it any longer
- Wasn't creating Poll::Question::Answer's so on the UI we couldn't get options to vote

How
===
- We needed a new string `token` attribute on Poll::Answer always present (check migration and model changes) https://github.com/consul/consul/commit/7e2fd754e110855207c270b7e96dc728fdb80c68
- We needed a helper function to generate a 64 char string that also checked it wasn't already present on the Poll::Answer table https://github.com/consul/consul/commit/9e2acba5be1a343805566bf67f97e274bc3c8882
- We needed to set a new unique token on each answer button and once used display that value to the user so he can register it https://github.com/consul/consul/commit/74a25cc64e0c2f9579d91f66b3c3c29025f12243

- Rake: Just getting the answers on both the "almost deprecated" `valid_answers` Poll::Questions attribute https://github.com/consul/consul/commit/6e29b6adf5909f613752f9e6d174567163a9a91e

Screenshots
===========
token generation

Test
====
⚠️  **PENDING** (need to double check with team solution is 👌  first)

Deployment
==========
- As usual. The presence validation for token on Poll::Answer shouldn't create any problems since Poll::Answers should not be modified for polls already existing/closed

Warnings
========
Concerns about token generation security will arise. My take is this:

- **We need to guarantee the token is not fixed (its not always returning the same value from the server)**: If you change your answer multiple times between two options you won't see same value twice. Even just by refreshing the page you could see the tokens change randomly on the html... the fact that its backend instead of frontend (js) the generator doesn't change a thing ✅ 

- **We need to guarantee the token is unique**. Done ✅ 

- User still could edit html and set a token value of any sort but it will be valid (only problem would be to try to introduce same value as on another answer... but that's why the unique validation on Poll::Answer#token is set)

### DIFF
--- a/app/assets/javascripts/polls.js.coffee
+++ b/app/assets/javascripts/polls.js.coffee
@@ -17,6 +17,7 @@ App.Polls =
     $(".js-question-answer").on
       click: =>
         token_message = $(".js-token-message")
-        token_message.html(token_message.html() + "<br><strong>" + @token + "</strong>");
-        token_message.show()
+        if !token_message.is(':visible')
+          token_message.html(token_message.html() + "<br><strong>" + @token + "</strong>");
+          token_message.show()
     false
--- a/app/assets/javascripts/polls.js.coffee
+++ b/app/assets/javascripts/polls.js.coffee
@@ -1,7 +1,11 @@
 App.Polls =
   generateToken: ->
-    rand = Math.random().toString(36).substr(2) # remove `0.`
-    token = rand + rand # to make it longer
+    token = ''
+    for n in [0..5]
+      rand = Math.random().toString(36).substr(2) # remove `0.`
+      token = token + rand;
+
+    token = token.substring(0, 64)
     return token
 
   replaceToken: ->
--- a/app/assets/javascripts/polls.js.coffee
+++ b/app/assets/javascripts/polls.js.coffee
@@ -1,0 +1,21 @@
+App.Polls =
+  generateToken: ->
+    rand = Math.random().toString(36).substr(2) # remove `0.`
+    token = rand + rand # to make it longer
+    return token
+
+  replaceToken: ->
+    for link in $('.js-question-answer')
+      token_param = link.search.slice(-6)
+      if token_param == "token="
+        link.href = link.href + @token
+
+  initialize: ->
+    @token = App.Polls.generateToken()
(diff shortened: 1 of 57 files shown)


app/assets/javascripts/polls.js.coffee
```diff
@@ -0,0 +1,27 @@
+App.Polls =
+  generateToken: ->
+    token = ''
+    for n in [0..5]
+      rand = Math.random().toString(36).substr(2) # remove `0.`
```
bertocq: Could `rand` definition be outside the loop?

app/views/polls/show.html.erb
```diff
@@ -38,13 +38,21 @@
           <%= t("polls.show.already_voted_in_booth") %>
         </div>
       <% else %>
+
+      <% if !@poll.voters.find_by_user_id(current_user).present? %>
```
bertocq: Check `poll_voter_token` function, we should be checking as well `origin: 'web'` I think. Since Poll::Voter could have been created through booth. 

Also `!QUERY.present?` == `QUERY.blank?` and its more readable I think

I would go with `poll_voter_token(@poll, current_user).empty?` because if that method returns `''` means that there is no token already generated and we could be making this div visible nearly. That way logic still is kept on one single place.

app/assets/javascripts/polls.js.coffee
bertocq: nah its cool its not a function but a string my bad 😅

ruescasd: Calling Math.random repeatedly is not a good way to collect entropy. Instead you may want to use crypto.getRandomValues. Here's an example  

```
// provides 128 bits of entropy
window.crypto.getRandomValues(new Uint32Array(4)).reduce(function(a, v) { return a.toString(36) + v.toString(36); })
```

You can calculate collision probabilities using the birthday attack model.

Issue #1992, opened 2017-10-06
Polls - Answers edit


decabeza:
What
==

We need to include a `/edit` view for poll question answers to be able to change the **title** and **description**.

We can include a link in the title of the response to `/admin/answers/N/edit`

Screenshot
==

<img width="1233" alt="edit_answer" src="https://user-images.githubusercontent.com/631897/31275314-0de5d9bc-aa97-11e7-8455-cdcf50ca1066.png">

PR #1995, opened 2017-10-06
Allow poll question answers to be editable

aitbw:
Where
=====
* **Related Issue:** #1992
* **Related PR:** #1977

What
====
- Allow poll question answers to be editable

How
===
* Adapted/created views/forms according to the new design
* Adapted the controller to fit the new feature
* Appended the necessary routes
* Added translated labels in English, Spanish and French
* Wrote tests for the feature

Screenshots
===========
* See mockups on #1992

Test
====
* Increased coverage for Create and Update scenarios

### DIFF
--- a/app/controllers/admin/poll/questions/answers_controller.rb
+++ b/app/controllers/admin/poll/questions/answers_controller.rb
@@ -50,4 +50,8 @@ class Admin::Poll::Questions::AnswersController < Admin::Poll::BaseController
       @answer = ::Poll::Question::Answer.find(params[:id] || params[:answer_id])
     end
 
+    def load_question
+      @question = ::Poll::Question.find(params[:question_id])
+    end
+
 end
--- a/app/controllers/admin/poll/questions/answers_controller.rb
+++ b/app/controllers/admin/poll/questions/answers_controller.rb
@@ -1,5 +1,6 @@
 class Admin::Poll::Questions::AnswersController < Admin::Poll::BaseController
-  before_action :load_question
+  before_action :load_question, except: [:show, :edit, :update]
+  before_action :load_answer, only: [:show, :edit, :update]
 
   load_and_authorize_resource :question, class: "::Poll::Question"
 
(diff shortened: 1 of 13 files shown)


config/routes.rb
```diff
@@ -300,7 +300,7 @@
         end
       end
 
-      resources :questions do
+      resources :questions, shallow: true do
         resources :answers, except: [:index, :destroy, :delete], controller: 'questions/answers', shallow: true do
```
aitbw: I'm pretty sure we don't need `shallow: true` here if we have it specified on `resources :questions`

voodoorai2000: 👍

app/controllers/admin/poll/questions/answers_controller.rb
```diff
@@ -44,7 +50,4 @@ def load_answer
       @answer = ::Poll::Question::Answer.find(params[:id] || params[:answer_id])
     end
 
-    def load_answer
```
aitbw: Why is this being deleted but the `before_action` stays? We need this to show, edit and update the answer when requested.

voodoorai2000: It's still there, it was duplicated
https://github.com/wairbut-m2c/consul/blob/23ef9ded1d02dcbf4304bd8eb89cd304d0f29156/app/controllers/admin/poll/questions/answers_controller.rb#L49

config/routes.rb
```diff
@@ -301,7 +301,7 @@
       end
 
       resources :questions, shallow: true do
-        resources :answers, except: [:index, :destroy, :delete], controller: 'questions/answers', shallow: true do
+        resources :answers, except: [:index, :destroy], controller: 'questions/answers', shallow: true do
```
aitbw: I don't think we need the `shallow: true` here if we have it declared on `resources :questions`

app/controllers/admin/poll/questions/answers_controller.rb
aitbw: Oh, sorry about that.

voodoorai2000: no problem, appreciated it!

config/routes.rb
voodoorai2000: 👍

Issue #2003, opened 2017-10-06
Polls - Display answer documents 


voodoorai2000:
What
===
- Display the documents of an answer in the standard user's poll's show view

Right now, documents can be created in the admin section for an answer, but they are not displayed to the user when he goes to vote a poll

aitbw:
I'm on it

Issue #1998, opened 2017-10-06
Polls - Empty answers


decabeza:
Background
==

Maybe some times there is a poll with questions and answers but without answers description or images.

What
==

Don't show empty `<div class="expanded poll-more-info-answers">` if there is no answers description or images.

**Note**: some similar (for answer title is made on https://github.com/consul/consul/pull/1996)

Screenshot
==

**Avoid this empty block**
<img width="1679" alt="screen shot 2017-10-06 at 17 15 13" src="https://user-images.githubusercontent.com/631897/31284810-f1776c5a-aab9-11e7-96b5-8f2930f3bc2f.png">

aitbw:
I'm currently working on this one.

Issue #1997, opened 2017-10-06
Polls - Questions limit


decabeza:
What
==

Remove **question** length limit on `/admin/questions/new` (this form is used only for admin users, there in no need to limit)

Screenshot
==

<img width="574" alt="screen shot 2017-10-06 at 17 11 26" src="https://user-images.githubusercontent.com/631897/31284667-8d8d37a6-aab9-11e7-918a-d9d46ba1b177.png">

aitbw:
I already took care of this, PR is pending along with fix for #1998 as well.

voodoorai2000:
Awesome @aitbw 🙏

PR #2006, opened 2017-10-06
Adjuments for Poll-related features

aitbw:
Where
=====
* **Related Issues:** #1997, #1998, #2003

What
====
* Adjustments for various Poll-related features

How
===
* Removed length limit for `title` attribute when creating a Poll::Question (commit d62c0d5)
* `poll-more-info-answers` div for Poll won't be rendered into the markup if no description is available or the answer does not have any associated records (commit 7fad604)
* Answer's associated documents are shown when users are voting on a Poll (commit a1907c6)
* Answer's title is shown if the answer has a description or any associated records (commit 8b8e8af)

Test
====
- Manual testing is needed

### DIFF
--- a/app/views/admin/poll/questions/_form.html.erb
+++ b/app/views/admin/poll/questions/_form.html.erb
@@ -12,13 +12,19 @@
                     label: t("admin.questions.new.poll_label") %>
     </div>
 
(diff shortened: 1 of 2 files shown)


voodoorai2000:
Awesome!

PR #2013, opened 2017-10-07
Removed questions anchor from participate poll button

MariaCheca:
What
====
Removed questions anchor from "participate in poll" button in polls list.

Test
====
As usual.

Deployment
==========
As usual.


MariaCheca:
This is not the PR you're looking for.

Issue #2018, opened 2017-10-08
Poll::Question::Answer should be orderable


bertocq:
Answers are displayed in creation order right now at https://github.com/consul/consul/blob/master/app/views/polls/questions/_answers.html.erb#L3

We need to add a "order" attribute that is used to list them on that "client" view and as well on the backend for the admin.

We'll need a way on the backend panel (admin) to change the order of the Answers, without much hassle :)

iagirre:
I'll take this one.

bertocq:
Here are the parts where we're using order (by id) right now on madrid https://github.com/AyuntamientoMadrid/consul/commit/14a35a2aeadfdb3e9f98dfe32ab9fbdc6fcb756d although the app/views/polls/show.html.erb is not correct, and should be revisited carefully with a spec

bertocq:
Closed via #936

PR #2037, opened 2017-10-10
Iagirre question ansers ordenable

iagirre:
Where
=====
* **Related Issue:** #2018 

What
====
Make the answers sortable from the admin page, so that they appear in the given order when user enters the poll.

How
===
- I added a column to the poll_question_answers table named `given_order`
- In order to make the answers sortable, I added the jquery-ui widget sortable to the table in the admins panel (where admin can add new questions). Admins drag and drop questions inside the table to sort them.
- I added an `order` to the relation between question and answers, so that every time we use `question.question_answers` they appear ordered by `given_order` field.

Screenshots
===========
iagirre-answer-order_1
Drag and drop to change the order

iagirre-answer-order_2
The order has been changed

iagirre-answer-order_3

User sees the new order

iagirre-answer-order_4

Test
====
I increased the test coverage to test if:
- the new answers are created with an order.
- the answers appear in the right order when updated.
- the admin sees the questions answers in the right order.
- the answers of a question appear in the right order.
- the more info answers appear in the right order.

Deployment
==========
Not needed.

Warnings
========
- When creating new answers, the order of the new one will be the last one created + 1. I mean:
There is already an answer with given_order = 1. The next will be given_order = 2, and so on. If it's the first one, it will be given_order = 1.
- I'm not sure if some more style will be needed in the admins page so that they know how to drag and drop.

### DIFF
--- a/app/controllers/admin/poll/questions/answers_controller.rb
+++ b/app/controllers/admin/poll/questions/answers_controller.rb
@@ -42,7 +42,6 @@ class Admin::Poll::Questions::AnswersController < Admin::Poll::BaseController
 
   def order_answers
     ::Poll::Question::Answer.order_answers(params[:ordered_list])
-    # redirect_to admin_question_path(params[:question_id])
     render nothing: true
   end
 
--- a/app/controllers/admin/poll/questions/answers_controller.rb
+++ b/app/controllers/admin/poll/questions/answers_controller.rb
@@ -37,6 +37,12 @@ class Admin::Poll::Questions::AnswersController < Admin::Poll::BaseController
     @documents = @answer.documents
 
     render 'admin/poll/questions/answers/documents'
+  end  
+  
+  def order_answers
+    ::Poll::Question::Answer.order_answers(params[:ordered_list])
+    #redirect_to admin_question_path(params[:question_id])
+    render :nothing => true
   end
 
   private
--- a/app/controllers/admin/poll/questions/answers_controller.rb
+++ b/app/controllers/admin/poll/questions/answers_controller.rb
@@ -9,7 +9,6 @@ class Admin::Poll::Questions::AnswersController < Admin::Poll::BaseController
 
   def create
     @answer = ::Poll::Question::Answer.new(answer_params)
-    @answer.set_order
 
     if @answer.save
       redirect_to admin_question_path(@answer.question),
--- a/app/controllers/admin/poll/questions/answers_controller.rb
+++ b/app/controllers/admin/poll/questions/answers_controller.rb
@@ -9,6 +9,7 @@ class Admin::Poll::Questions::AnswersController < Admin::Poll::BaseController
 
   def create
     @answer = ::Poll::Question::Answer.new(answer_params)
+    @answer.set_order
 
     if @answer.save
       redirect_to admin_question_path(@answer.question),
@@ -37,12 +38,12 @@ class Admin::Poll::Questions::AnswersController < Admin::Poll::BaseController
     @documents = @answer.documents
 
     render 'admin/poll/questions/answers/documents'
-  end  
-  
+  end
+
   def order_answers
     ::Poll::Question::Answer.order_answers(params[:ordered_list])
-    #redirect_to admin_question_path(params[:question_id])
-    render :nothing => true
+    # redirect_to admin_question_path(params[:question_id])
+    render nothing: true
   end
 
   private
--- a/app/models/poll/question/answer.rb
+++ b/app/models/poll/question/answer.rb
@@ -1,5 +1,5 @@
 class Poll::Question::Answer < ActiveRecord::Base
-	include Galleryable
+  include Galleryable
   include Documentable
   documentable max_documents_allowed: 3,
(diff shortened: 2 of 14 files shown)


app/models/poll/question/answer.rb
```diff
@@ -14,4 +14,18 @@ class Poll::Question::Answer < ActiveRecord::Base
   def description
     super.try :html_safe
   end
+
+  def self.order_answers(ordered_array)
+    ordered_array.each_with_index do |answer_id, order|
+      answer = find(answer_id)
+      answer.update_attribute(:given_order, (order + 1))
+      answer.save
```
bertocq: [`update_attribute`](https://apidock.com/rails/ActiveRecord/Persistence/update_attribute) saves the record already. I would refactor this 3 lines into just `find(answer_id).update_attribute(:given_order, (order + 1))`

app/models/poll/question/answer.rb
```diff
@@ -14,4 +14,18 @@ class Poll::Question::Answer < ActiveRecord::Base
...
+    end
+  end
+
+  def set_order
+    last_position = Poll::Question::Answer.where(question_id: question_id).maximum("given_order") || 0
```
bertocq: I think we don't need `Poll::Question::Answer` self reference

db/migrate/20171009151031_add_given_order_to_poll_question_answers.rb
```diff
@@ -0,0 +1,5 @@
+class AddGivenOrderToPollQuestionAnswers < ActiveRecord::Migration
+  def change
+    add_column :poll_question_answers, :given_order, :integer
```
bertocq: Nice one! I started a quick refactor to add this but found myself in the pitfall of creating an column named `order` that raised many issues 🙈  this name is better

Although by reading all code seems to me that the `given_order` values will always start from `1`, right? Would it make sense to set a `default: 1` just as a reminder on the code and as well as a security measure?

Also I think if `given_order` is an attribute needed there should be a presence validation on the model (as well as one that checks its unique for the question_id (only one Answer on a particular order for its Question)

spec/features/admin/poll/questions/answers/answers_spec.rb
```diff
@@ -24,9 +24,27 @@
     expect(page).to have_content(description)
   end
 
+  scenario 'Create second answer and place after the first one' do
```
bertocq: 👍

app/controllers/admin/poll/questions/answers_controller.rb
```diff
@@ -9,6 +9,7 @@ def new
 
   def create
     @answer = ::Poll::Question::Answer.new(answer_params)
+    @answer.set_order
```
bertocq: I think since there might be other ways to create an answer (think through rake task, console, etc...) and a value at the `given_order` attribute is always needed. This `set_order` method shouldn't be called here, but on a `before_create` on the Poll Question Answer model

bertocq:
Really nice the default order in the relation between questions and answers 👏  cool trick :)

There's an additional place where we'll need answers ordered where that default order won't help sadly :( https://github.com/consul/consul/blob/master/app/views/polls/show.html.erb#L82 on Madrid's fork we did this "hack" sorting in Ruby but that's not really a nice way to do it either :D https://github.com/ayuntamientomadrid/consul/blob/master/app/views/polls/show.html.erb#L98. 

I think the trick here would be to instead of `@poll.questions.map(&:question_answers).flatten.each` use `Poll::Question::Answer.where(question: @poll.question).each` but we'll be needing an spec to check that (we hotfixed it on madrid not even time to add spec to it there)

We can fix this in a second PR if you think it will be too much for this one (tell me and I'll open an Issue)

app/models/poll/question/answer.rb
```diff
@@ -10,8 +10,26 @@ class Poll::Question::Answer < ActiveRecord::Base
   has_many :videos, class_name: 'Poll::Question::Answer::Video'
 
   validates :title, presence: true
+  validates :given_order, presence: true, uniqueness: { scope: :question_id }
+
+  before_validation :set_order, on: :create
```
iagirre: I finally used `before_validation on: :create` to set the order, because if it was set `before_create`, the validation run before all of that (so, before `set_order` method) and it failled.
For the 1st there was no problem, but for the 2nd it couldn't pass the validation and it couldn't create the record.

spec/factories.rb
```diff
@@ -503,8 +503,8 @@
 
   factory :poll_question_answer, class: 'Poll::Question::Answer' do
     association :question, factory: :poll_question
-    sequence(:title) { |n| "Question title #{n}" }
-    sequence(:description) { |n| "Question description #{n}" }
+    sequence(:title) { |n| "Answer title #{n}" }
+    sequence(:description) { |n| "Answer description #{n}" }
```
iagirre: I changed the name for the `poll_question_answer` because it made conflicts with the `poll_question` name when running specs (they had the same name, but, as I understand, the answers should be named 'Answer title X' instead of 'Question title X').

Issue #2045, opened 2017-10-11
Remove Author and usage from Poll


bertocq:
On the Poll Comments PR https://github.com/consul/consul/pull/1961 some things got introduced that shouldnt:
- `author_id` on the polls table
- `belongs_to :author, -> { with_hidden }, class_name: 'User', foreign_key: 'author_id'` on the poll model

In order to remove this without issues, we have to trace all `commentable.author` usage, and add something like `unless  @comment.commentable.is_a?(Poll)` to avoid the usage... as far as I've researched its always code related to email notifications about comments and since Poll comments should not be notificable to author (there is NO author on the specs) that's not neccesary.

zamuro:
I'll take this, since I've worked on the previous one.

javierm:
Closing this one since now there are dashboard polls where having the author makes sense.

PR #2048, opened 2017-10-11
20171011 - Removed author from polls comments

zamuro:
Where
=====
* **Related Issue:** #2045 

What
====
- To remove `author_id` from polls and to refactor related views and controllers to respond accordingly

How
===
- By setting several conditions into
  - comments' show view
  - poll and mailer models
  - comments controller

Screenshots
===========
- n/a

Test
====
- As usual.

Deployment
==========
- You might need to run a migration if author_id is still in your poll table.

Warnings
========
- Check if author_id is not present in your schema.

### DIFF
--- a/app/views/comments/_comment.html.erb
+++ b/app/views/comments/_comment.html.erb
@@ -48,26 +48,19 @@
                 <%= t("shared.collective") %>
               </span>
             <% end %>
-            <% unless comment.commentable.class.name == 'Poll' %>
-              <% if comment.user_id == comment.commentable.author_id %>
-                &nbsp;&bull;&nbsp;
-                <span class="label round is-author">
-                  <%= t("comments.comment.author") %>
-                </span>
-              <% end %>
+            <% comment.commentable.is_a?(Poll) ? nil : if comment.user_id == comment.commentable.author_id %>
+              &nbsp;&bull;&nbsp;
+              <span class="label round is-author">
+                <%= t("comments.comment.author") %>
+              </span>
             <% end %>
           <% end %>
 
           &nbsp;&bull;&nbsp;<span><%= l comment.created_at.to_datetime, format: :datetime %></span>
         </div>
-
         <div class="comment-user
                   <%= user_level_class comment %>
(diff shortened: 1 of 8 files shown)


app/views/comments/_comment.html.erb
```diff
@@ -48,21 +48,23 @@
                 <%= t("shared.collective") %>
               </span>
             <% end %>
-            <% if comment.user_id == comment.commentable.author_id %>
+            <% comment.commentable.is_a?(Poll) ? nil : if comment.user_id == comment.commentable.author_id %>
```
zamuro: Not entirely sure if a ternary operator is the way to go here. Comments welcome.

app/views/comments/_comment.html.erb
```diff
@@ -48,21 +48,23 @@
...
         </div>
-
         <div class="comment-user
                   <%= user_level_class comment %>
-                  <%= comment_author_class comment, comment.commentable.author_id %>">
+                  <% if comment.commentable.is_a?(Poll) %>
+                    <%= comment_author_class comment, nil %>
+                  <% else %>
+                    <%= comment_author_class comment, comment.commentable.author_id %>
+                  <% end %>">
```
zamuro: Since it's just a CSS class, would it be better if it goes into a helper method?

app/views/comments/_comment.html.erb
```diff
@@ -48,21 +48,19 @@
...
-
           <% end %>
 
           &nbsp;&bull;&nbsp;<span><%= l comment.created_at.to_datetime, format: :datetime %></span>
         </div>
-
         <div class="comment-user
                   <%= user_level_class comment %>
-                  <%= comment_author_class comment, comment.commentable.author_id %>">
+                  <%= commentable_author_id_for_polls(comment, comment.commentable) %>">
```
zamuro: I replaced the logic here with a helper method, as I commented in a prior commit. It seems somewhat cleaner to me.

voodoorai2000:
Thanks Manuel, this feature has become obsolete, we'll review again soon!

Issue #2027, opened 2017-10-09
Right answer resize should reorder


bertocq:
# Current Behaviour
Right now when you click expand the gallery of the second answer on a Poll you'll get it expanded but below the first answer:
![zoom_right_answer](https://user-images.githubusercontent.com/983242/31355082-2601b8d2-ad39-11e7-800a-8c337b0a27d5.gif)

# Desired Behaviour
Instead of expanding below the first answer, it should be on top of it. With the same effect as expanding the first answer gallery does.

iagirre:
I'll take this.

decabeza:
Hi @iagirre! For this one I use a Foundation "little trick" with [Foundation toggler](https://foundation.zurb.com/sites/docs/toggler.html) to avoid create new js. Feel free to change it if it's necessary 😉

PR #2054, opened 2017-10-13
Iagirre answer resize reorder

iagirre:
Where
=====
* **Related Issue:** #2027

What
====
Make the answers more info expand on the top of the other answer, not below it.

How
===
- Using javascript, I made a function that reorders the answer elements IF the expand button belongs to the second answer.
- I let the toggler function like Foundation Toggler does, but I made it manually to avoid bugs when expanding (if one is executed before another it didn't work).

Screenshots
===========
reorder_answers

Test
====
Nothing to apply.

Deployment
==========
Nothing to apply.

Warnings
========
Nothing to apply.


bertocq:
Nice and clean @iagirre !

Issue #2031, opened 2017-10-09
Trim Shift date range with current officer shifts


bertocq:
When creating a Shift for an Officer on a given Booth, for each Shift type (vote or recount&scrutiny) we display a different date range to pick from.

We should remove from that date range the dates on which the Officer has already a Shift of that type on that Booth, so the list is smaller and easier to see which dates still need to be assigned to him (if any)

MariaCheca:
Closed here: #2062

PR #2062, opened 2017-10-16
Excluded officer shift dates from each task date range

MariaCheca:
Where
=====
* **Closes** #2031

What
====
In the admin `#new` form to assign shifts to officers, excluded from task dates lists **already assigned dates for that officer in the booth** selected.

Screenshots
===========
screen shot 2017-10-16 at 12 20 50
screen shot 2017-10-16 at 12 20 59
screen shot 2017-10-16 at 12 21 11

Test
====
As usual.

Deployment
==========
As usual.

### DIFF
--- a/app/helpers/shifts_helper.rb
+++ b/app/helpers/shifts_helper.rb
@@ -1,15 +1,15 @@
 module ShiftsHelper
 
   def shift_vote_collection_dates(polls)
-    date_options((start_date(polls)..end_date(polls)))
+    date_options((start_date(polls)..end_date(polls)), 0)
   end
 
   def shift_recount_scrutiny_dates(polls)
-    date_options(polls.map(&:ends_at).map(&:to_date).sort.inject([]) { |total, date| total << (date..date + 1.week).to_a }.flatten.uniq)
+    date_options(polls.map(&:ends_at).map(&:to_date).sort.inject([]) { |total, date| total << (date..date + 1.week).to_a }.flatten.uniq, 1)
   end
 
-  def date_options(dates)
-    dates.map { |date| [l(date, format: :long), l(date)] }
+  def date_options(dates, task_id)
+    dates.reject { |date| officer_shifts(task_id).include?(date) }.map { |date| [l(date, format: :long), l(date)] }
   end
 
   def start_date(polls)
@@ -24,4 +24,9 @@ module ShiftsHelper
     officers.collect { |officer| [officer.name, officer.id] }
   end
 
+  private
+
+  def officer_shifts(task_id)
+    @officer.shifts.where(task: task_id).map(&:date)
+  end
 end
--- a/app/helpers/shifts_helper.rb
+++ b/app/helpers/shifts_helper.rb
@@ -1,15 +1,19 @@
 module ShiftsHelper
 
   def shift_vote_collection_dates(polls)
-    date_options((start_date(polls)..end_date(polls)), 0)
+    date_options((start_date(polls)..end_date(polls)), Poll::Shift.tasks[:vote_collection])
   end
 
   def shift_recount_scrutiny_dates(polls)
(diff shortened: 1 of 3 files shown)


app/helpers/shifts_helper.rb
```diff
@@ -1,15 +1,15 @@
 module ShiftsHelper
 
   def shift_vote_collection_dates(polls)
-    date_options((start_date(polls)..end_date(polls)))
+    date_options((start_date(polls)..end_date(polls)), 0)
```
bertocq: Instead of using a harcoded `0` here (or `1` at line 8)  I would give more context to those values as it could be very hard to understand for someone that doesn't know about Poll Shift tasks being a enumeration.

Maybe since:
```
irb(main):002:0> Poll::Shift.tasks
=> {"vote_collection"=>0, "recount_scrutiny"=>1}
```

Using ` Poll::Shift.tasks[:vote_collection]` and `Poll::Shift.tasks[:recount_scrutiny]` could be way better to make code easier to read.

app/helpers/shifts_helper.rb
```diff
@@ -1,15 +1,15 @@
...
   end
 
   def shift_recount_scrutiny_dates(polls)
-    date_options(polls.map(&:ends_at).map(&:to_date).sort.inject([]) { |total, date| total << (date..date + 1.week).to_a }.flatten.uniq)
+    date_options(polls.map(&:ends_at).map(&:to_date).sort.inject([]) { |total, date| total << (date..date + 1.week).to_a }.flatten.uniq, 1)
   end
 
-  def date_options(dates)
-    dates.map { |date| [l(date, format: :long), l(date)] }
+  def date_options(dates, task_id)
```
bertocq: It bugs me (just a little bit) that a method does two things (both trimming down shift dates from task date range, and creating the date options for the select box)

app/helpers/shifts_helper.rb
```diff
@@ -24,4 +24,9 @@ def officer_select_options(officers)
     officers.collect { |officer| [officer.name, officer.id] }
   end
 
+  private
+
+  def officer_shifts(task_id)
+    @officer.shifts.where(task: task_id).map(&:date)
```
bertocq: Thanks to rails Enumeration magic you could go `@officer.shifts.recount_scrutiny.map(&:date)` So if instead of passing around a `0` or `1` ... I would rethink this method to accept a symbol (`:recount_scrutiny` or `: vote_collection`) and transform this into `@officer.shifts.send(task.to_s).map(&:date)`

app/helpers/shifts_helper.rb
```diff
@@ -1,15 +1,15 @@
...
 
   def shift_recount_scrutiny_dates(polls)
-    date_options(polls.map(&:ends_at).map(&:to_date).sort.inject([]) { |total, date| total << (date..date + 1.week).to_a }.flatten.uniq)
+    date_options(polls.map(&:ends_at).map(&:to_date).sort.inject([]) { |total, date| total << (date..date + 1.week).to_a }.flatten.uniq, 1)
   end
 
-  def date_options(dates)
-    dates.map { |date| [l(date, format: :long), l(date)] }
+  def date_options(dates, task_id)
+    dates.reject { |date| officer_shifts(task_id).include?(date) }.map { |date| [l(date, format: :long), l(date)] }
```
bertocq: Nice usage of reject 👍  😃

app/helpers/shifts_helper.rb
bertocq: It's ok, as we talked the scopes aren't working correctly and too much refactoring for this small task. We'll clean it up later 👍

spec/features/admin/poll/shifts_spec.rb
```diff
@@ -91,6 +91,37 @@
     end
   end
 
+  scenario "Vote Collection Shift and Recount & Scrutiny Shift don't include already assigned dates to officer", :js do
```
bertocq: Nice!!

Issue #1946, opened 2017-09-29
Add investment's user generated tags to admin interface


voodoorai2000:
What
===
- Allow admins to see and edit user generated tags

Why
===
There are two kind of tags related to budget investments:
- User tags
- Valuation tags

User tags are the ones added by users when they create an investment.

Valuation tags, are used to internally tag investments.

Case example:
- The Health area of the Government, deals with internal tags for “hospitals and ambulances”
- The Culture area of the Government, deals with internal tags for “theatres and street music”

Right now, only valuation tags are being displayed in the admin section.

How
===
This work has already been done in [Madrid's fork](https://github.com/AyuntamientoMadrid/consul/pull/710/files)

Screenshots:
===
Current state:

- Form without tags
![form without tags](https://user-images.githubusercontent.com/4169/31015115-c6d5fb60-a51e-11e7-8909-fb110e51bf2a.png)

- Admin show without tags
![show without tags](https://user-images.githubusercontent.com/4169/31015118-c997156e-a51e-11e7-9a54-dc6589df8396.png)

Desired state:

- Form with tags
<img width="1096" alt="form with tags" src="https://user-images.githubusercontent.com/4169/31015121-cc827098-a51e-11e7-8229-0fa0843cc91a.png">

- Admin show with tags
![show with tags](https://user-images.githubusercontent.com/4169/31015124-cee34b1e-a51e-11e7-80be-2cb929e76198.jpg)

ortegacmanuel:
If not one else is working this, I could work on this. I find this a great first contribution to the project.

bertocq:
Awesome @ortegacmanuel !! go for it 😃 👍

ortegacmanuel:
@bertocq Done! :)

bertocq:
Closed thanks to #2068

PR #2068, opened 2017-10-17
Add investment user tags admin interface

ortegacmanuel:
Where
=====
* **Related Issue:** #1946

What
====
- Adds investment's user generated tags to admin interface

How
===
- Following the implementation done in [Madrid's fork](http://github.com)

Why
====

There are two kind of tags related to budget investments:

- User tags
- Valuation tags

User tags are the ones added by users when they create an investment.

Valuation tags, are used to internally tag investments.

Case example:

- The Health area of the Government, deals with internal tags for “hospitals and ambulances”
- The Culture area of the Government, deals with internal tags for “theatres and street music”

### DIFF
--- a/app/controllers/admin/budget_investments_controller.rb
+++ b/app/controllers/admin/budget_investments_controller.rb
@@ -51,7 +51,7 @@ class Admin::BudgetInvestmentsController < Admin::BaseController
 
     def budget_investment_params
       params.require(:budget_investment)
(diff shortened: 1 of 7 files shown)


ortegacmanuel:
@bertocq  My pleasure :) Thanks for the comment about cherrypicking the commit from madrid's fork...I'll keep this in mind for next one ;)

PR #2083, opened 2017-10-19
Poll slugs

### LINKED ISSUE #2019: Poll Slugs
Backport from madrid https://github.com/AyuntamientoMadrid/consul/pull/919#pullrequestreview-67866805

But adding:
- Generate slug for Polls
- Allow to manually overwrite the slug from the rails console so it won't be overwritten later on (I suspect using Sluggable would do that wrong)

Maybe make Sluggable comply with those specs and use it 👌

iagirre:
Where
=====
* **Related Issue:** #2019 
This PR closes #2019 

What
====
- Add a slug attribute to polls so that users can navigate to them using the slug instead of the id.
- If this slug was modified from the rails console, it should not be updated when updating the poll.

How
===
- I added the Sluggable concern to polls model to make polls have the slug attribute.
- I modified the Sluggable concern so that it updates the slug only if there wasn't modified by hand (e.g. from rails console). If it was modified, it won't change it anymore.
- I also updated the `rake slugs:generate` task to add `Poll` model.

Screenshots
===========
iagirre-poll-slug

Test
====
- I updated the sluggable concern shared_example to test the update of the slug for all the sluggable models.
- I also added specs to test that users can navigate to poll using either the slug or the id.

Deployment
==========
I created a rake task to generate a slug for all the existing polls in the DB, `rake temp:poll_slug`, and it should be run in order to make things work.

Warnings
========
Nothing to apply.

### DIFF
--- a/app/controllers/polls_controller.rb
+++ b/app/controllers/polls_controller.rb
@@ -1,6 +1,8 @@
 class PollsController < ApplicationController
   include PollsHelper
 
+  before_action :load_poll, except: [:index]
+  
   load_and_authorize_resource
 
   has_filters %w{current expired incoming}
@@ -26,5 +28,11 @@ class PollsController < ApplicationController
     @commentable = @poll
     @comment_tree = CommentTree.new(@commentable, params[:page], @current_order)
   end
+  
+  private
+  
+    def load_poll
+      @poll = ::Poll.find_by(slug: params[:id]) || ::Poll.find_by(id: params[:id])
+    end
 
 end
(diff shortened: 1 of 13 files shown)


bertocq:
🙏  to whomever reviews and merges this PR... please add slug attributes to Polls created in dev seeds as this removed TODO comments & lines where trying to remind https://github.com/consul/consul/commit/baa3c5280f8d67c2b943eb65e555e4d7582abc63

javierm:
Closed in favor of #3504.

PR #2100, opened 2017-10-23
Fix `direct_messages_max_per_day` set to nil

deivid-rodriguez:
When set to nil, it should mean not zero, but "infinite".

Where
=====
_None_.

What
====
- Whats the objective of this changes ?

The objective is that there's a way to have unlimited direct messages per day, and also that messages can be sent by default on a development setup, since it currently gives a "max messages per day reached" error that prevents every message from being sent.

How
===
- How you implemented/achieved the objective ?

I checked for the existance of the setting and skip validation in that case. This way we skip the automatic `nil.to_i` (`== 0`) conversion leading to the current behavior. 

Screenshots
===========
_None_.

Test
====
- Is manual test needed or you just increased/fixed coverage?

I added coverage.

Deployment
==========
- Any details to remember when this feature is deployed?

I don't think so.

Warnings
========
- Some caveats or important things to notice?

~Nope~. Well, maybe yes. If some installation is intentionally using this (buggy, in my opinion) behavior, they'll have to start being explicit about it and set `Setting['direct_max_messages_per_day'] = 0`.

### DIFF
--- a/app/models/direct_message.rb
+++ b/app/models/direct_message.rb
@@ -13,6 +13,7 @@ class DirectMessage < ActiveRecord::Base
   def max_per_day
     return if errors.any?
     max = Setting[:direct_message_max_per_day]
+    return unless max
 
     if sender.direct_messages_sent.today.count >= max.to_i
       errors.add(:title, I18n.t('activerecord.errors.models.direct_message.attributes.max_per_day.invalid'))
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
I'm assuming the failure is unrelated, let me know if I'm wrong though.

voodoorai2000:
Thank you so much @deivid-rodriguez!
Sorry for the delay we are little overwhelmed 🙏 🙇

deivid-rodriguez:
No problem, thanks to you! :)

Issue #1881, opened 2017-09-18
Postal Vote on Officers panel


bertocq:
Currently at Madrid's fork there is a "Letter Officer" feature that allows volunteers to help with the Postal Voting and this feature needs to be ported back to Consul with some improvements.

1. We need to have a `origin` value `letter`. (I would go with `postal` but its just a suggestion of refactor)
2. We need to have a new Poll::Shift task `postal_vote`
3. *We need to clarify with UI/UX* what would be the best way to avoid confusion between "Booth Recount&Scrutiny" and "Postal Recount&Scrutiny"... maybe separate but identical menus on the Officing panel... or just a "Postal Vote" option on the booth selector on that "Recount & Scrutiny" form.
4. We need to set the origin value to `letter` on the relative PartialResults and Recounts to be able to distinguish them on stats and results & recounts lists.
5. We need to increase Results & Recounts panels on both admin and public interfaces to display this new vote category.

PR #2107, opened 2017-10-26
[Backport] Token error

### LINKED ISSUE #2088: Backport Token problems fix from madrid's fork
Backport Token problems fix from madrid's fork, some parts done at https://github.com/AyuntamientoMadrid/consul/pull/957 but need to compare it fully

iagirre:
## References

* Backports AyuntamientoMadrid#957 
* Closes #2088 

## Objectives

Backport problems generated when the voter token wasn't received correctly in the client side.

I took the code related to tokens in Madrid's fork and adapted to make it work in consul.

## Visual changes

iagirre-token-problem-from-madrid

## Notes

- I added the test from Madrid's fork, but some of them needed some modifications because the features they test didn't exist in consul.
- I removed some test code from features/polls/voter_spec.rb which checks the presence of token in postal votes. In consul there isn't postal vote yet (there is an issue about it #1881). Just to be aware of this when doing that issue.
- In Madrid's fork there was a [call to a method](https://github.com/AyuntamientoMadrid/consul/blob/18550eae40bc0e9534e922e80d88dfe004800b04/app/controllers/polls/questions_controller.rb#L20) to log the vote event. That method was part of a concern call [`analytics.rb`](https://github.com/AyuntamientoMadrid/consul/blob/master/app/controllers/concerns/analytics.rb) that doesn't exists in consul, and I'm not sure if it should be in consul repo. I didn't include neither that `log_event` call nor the concern, but I will if it's needed.
- This PR fixes the problem generated when, by some reason, the answer links don't have the token attribute in the `href` property, BUT it doesn't fix the problem generated when the user has the JS disabled from the browser. If that happens, when the user tries to answer, the `link_to` behaves like a GET, not a POST (event though there is an explicit `method: :post`), and there isn't any route GET that fulfills the path, so the app raises a routing error.

> Note that if the user has JavaScript disabled, the request will fall back to using GET.
> https://apidock.com/rails/ActionView/Helpers/UrlHelper/link_to

### DIFF
--- a/app/assets/javascripts/polls.js.coffee
+++ b/app/assets/javascripts/polls.js.coffee
@@ -15,16 +15,16 @@ App.Polls =
       if token_param == "token="
         link.href = link.href + @token
 
+  showTokenMessage: ->
+    token_message = $(".js-token-message")
+    if !token_message.is(':visible') && $('.js-marked-answer').length > 0
+      token_message.html(token_message.html() + "<br><strong>" + @token + "</strong>");
+      token_message.show()
+
   initialize: ->
     @token = App.Polls.generateToken()
     App.Polls.replaceToken()
 
-    $(".js-question-answer").on
-      click: =>
-        token_message = $(".js-token-message")
-        if !token_message.is(':visible')
-          token_message.html(token_message.html() + "<br><strong>" + @token + "</strong>");
-          token_message.show()
     false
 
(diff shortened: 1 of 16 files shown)


javierm:
Closed in favor of #3501, which brings more code this one depends on and gives credit to the original author.

PR #1989, opened 2017-10-05
Add Poll::Admin answer videos

aitbw:
Where
=====
* **Related Issues:** #1855, #1950, #1951, #1952
* **Related PRs:** #1977, #1988 

What
====
* Adapt the Admin::Poll questions UI to the new design discussed in #1855

How
===
* Did the necessary database migrations
* Adapted the views/forms according to the new design
* Created the necessary controllers/models
* Appended the necessary routes
* Added translated labels in English, Spanish and French

Screenshots
===========
* See mockups on #1951 

Test
====
- Added test for the new feature

Notes
========
* The model created on this PR takes the custom validation for `video_url` implemented in #1854, which it's not DRY at all. @iagirre created a concern to solve #1865 and the PR is pending. As soon as that PR is done and merged, I'll handle the refactor to maximize DRYness.

* Some errors were made on #1977, but were fixed on this PR. See commits' description for further information.

* Since #1988 is done and waiting for approval/merge, we should be careful when merging since both PRs handle the `routes.rb` file and English/Spanish locales, as well as the Questions 'show' view. This is important since we want to keep the markup clean and tidy

PR #2108, opened 2017-10-26
Create a Video concern

### LINKED ISSUE #1865: Create a Video entity to abstract `video_url` attributes
As there are going to be multiple models with a `video_url` attribute that needs more logic to be managed across the platform, it could be a good idea to create a new Video/_Videable_ entity to treat those URLs as models and use the logic inside it to:

- Validate the value (it's really a video URL and only that).
- Help when it needs to be used in views.
- Whatever we want.

Something as we have done with documents to, what do you think?

cc/ @bertocq @voodoorai2000 @decabeza

iagirre:
Where
=====
* **Related Issue:** #1865 
This PR closes #1865 

What
====
Create a concern for all the things related to video_urls so that the management of this logic would be easier to maintain.

How
===
- I created a concern called `Videable` where I put the regex to check video origins and the `valid_video_url?` method.
- I moved all that code from the `EmbedVideosHelper`, and used the regex from Videable in the function `embedded_video_code`. 

Screenshots
===========
There is nothing to show

Test
====
Tests were added to check that the functions moved to `Videable` concern continue working.

Deployment
==========
Nothing to apply.

Warnings
========
This issue was mentioned in #1989 PR. Some of the code added there should be moved to this concern.

### DIFF
--- a/app/helpers/embed_videos_helper.rb
+++ b/app/helpers/embed_videos_helper.rb
@@ -1,8 +1,5 @@
 module EmbedVideosHelper
 
-  VIMEO_REGEX = /vimeo.*(staffpicks\/|channels\/|videos\/|video\/|\/)([^#\&\?]*).*/
-  YOUTUBE_REGEX = /youtu.*(be\/|v\/|u\/\w\/|embed\/|watch\?v=|\&v=)([^#\&\?]*).*/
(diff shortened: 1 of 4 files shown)


javierm:

Issue #2070, opened 2017-10-17
Use timestamps at Poll Recount & Partial Result


bertocq:
As stated at https://github.com/consul/consul/pull/2069 we're currently setting "manually" the valeu for `date` attribute on both Poll::Recount and Poll::PartialResult to `Date.current`, when we could just rely on the automatic `created_at` timestamp. Check `/app/controllers/officing/results_controller.rb` lines 58 and 74

aitbw:
I'll take this one.

bertocq:
As we talked, we'll refactor this at higher level

PR #2111, opened 2017-10-27
Use timestamps on Poll::Recount and Poll::PartialResult

aitbw:
Where
=====
* **Related Issue:** #2069, #2070

What
====
* Replace `date` attribute with `created_at` timestamps for `Poll::Recount` and `Poll::PartialResult` models

How
===
* Ran the necessary migrations to remove the `date` attribute on the aforementioned models and add the new `created_at` timestamp
* Replaced matches of the deprecated `date` attribute on the `dev_seeds.rb` file with the new `created_at` timestamps
* Replaced matches of the deprecated `date` attribute on Poll Results-related views, controller and specs

Test
====
* Fixed coverage for affected Poll Results specs


db/migrate/20171027124746_add_timestamps_and_remove_date_to_poll_recount.rb
```diff
@@ -0,0 +1,8 @@
+class AddTimestampsAndRemoveDateToPollRecount < ActiveRecord::Migration
+  def change
+    change_table :poll_recounts do |t|
+      t.remove :date
+      t.date :created_at
```
bertocq: It would be better to use a time column to store the time as well as date, since many times the admins want to debug some problem and having the time of creation can be helpful (as well at the PartialResult object).

aitbw:
PR updated accordingly, @bertocq :)

bertocq:
I've tried to fix the issues and got to a solution but it is in my opinion worse than we have just because it would require database querys searching for records created today based on a datetime ... not sure we'll really get an improvement for this refactor. I'm prone to close both PR and issue, assuming in the refactors that will come on Polls this will be solved at a higher lever of abstraction (solving the main requirement here that is no more than one Recount or PartialResult per day, booth_assignment_id etc..)

bertocq:
As we talked, we're closing this, we'll refactor at a higher level, thank you @aitbw for your effort! 👏

PR #2113, opened 2017-10-27
Updated Gemfile_custom

pedroleon:
Updated Rest-client gem


pedroleon:
Sorry again! At this hour I think is better to stop working to prevent more worng PR

Issue #2061, opened 2017-10-16
Booth assignment removal


bertocq:
When a Booth is unassigned from a Poll, all existing Shifts on that Booth-Poll relationship should be destroyed, showing a warning accept/reject message to the user (only if there are existing Shifts, if there aren't no need for the alert)

Also it should not be possible to unassign a Booth from an expired Poll (after the final voting day, check & use the Poll `expired` scope).

# Warning:
Please check https://github.com/consul/consul/issues/2034 is completed already before taking this

bertocq:
closed via #2114  👏

PR #2114, opened 2017-10-30
Iagirre booth assignment removal

iagirre:
Where
=====
* **Related Issue:** #2061 

What
====
Delete the shifts associated to a `booth_assigment` when that booth is unassigned.

How
===
- I make a callback `before_destroy` on the `booth_assignment` model that deletes the shifts related to that `booth_assignment`.
- When there are shifts related to a booth_assignment, an alert is shown before unassinging it so that the user can cancel the action.
- For expired polls there are no links to Assing/Unassing booths (to reach de booth management for an expired poll, the user needs to modify the URL directly, but, just in case it happens, there is no option to change the assignments).

Screenshots
===========
booth_removal

Test
====
Test added to:
- check that the booth unassings correctly event if there are shifts for it.
- check that there is not link to unassing/assing booths when poll is expired
- test the booth_assignment models `shifts? `and `destroy_shifts` methods.

Deployment
==========
Nothing to apply.

Warnings
========
Nothing to apply.

### DIFF
--- a/app/models/poll.rb
+++ b/app/models/poll.rb
@@ -90,8 +90,4 @@ class Poll < ActiveRecord::Base
     end
   end
 
-  def shifts_in_booth(booth_id)
-    officer_assignments.where(booth_assignment_id: booth_assignments.where(booth_id: booth_id).pluck(:id)).pluck(:officer_id).uniq
-  end
-
 end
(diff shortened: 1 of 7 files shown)


voodoorai2000:
This is great! Thank you @iagirre 🎉

PR #2117, opened 2017-11-06
Date of birth on management verification email

### LINKED ISSUE #1452: Adds date of birth on management verification email
When a manager check if the user's document is valid (and the user is already registered), there is the option to send an email and verify the account:

management

This is the email sent:

management_email

The idea is the link on email should redirect to a new page where the user introduce their date of birth before verify their account, something like this:

date_of_birth

It's important save the user's date of birth (is checked in some processes whit the setting `Setting['min_age_to_participate']`)

iagirre:
Where
=====
* **Related Issue:** #1452
This PR close #1452 

What
====
When a user receives a verification email from the admin, the page should ask their date of birth before verifying the account.

How
===
I modified the flow of how a verification is done from an email. The link sent  to the user has been changed, and now it points to a URL where the date of birth is asked. When the user selects its date of birth and sends the information the page verifies the account. If the user tries to send a nil date the page asks for it again.

Screenshots
===========
birth01

birth02

Test
====
I added/modified the tests that check the verification via email. For the test that checks the verification flow modified, I added the selection of the date. For the other two, I modified the regex that takes the token from the email link.

Deployment
==========
Nothing to apply.

Warnings
========
Nothing to apply.

### DIFF
--- a/app/controllers/management/email_verifications_controller.rb
+++ b/app/controllers/management/email_verifications_controller.rb
@@ -13,11 +13,24 @@ class Management::EmailVerificationsController < Management::BaseController
       render :new
     end
   end
+  
+  def date_of_birth    
+  end
+  
+  def save_date_of_birth
+    user = User.where(email_verification_token: params[:email_verification_token]).first
+    user.date_of_birth = date_of_birth
+  end
 
   private
 
     def email_verification_params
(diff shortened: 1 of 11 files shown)


javierm:

PR #2125, opened 2017-11-13
Redefine postal code verification methods to use setting config parameter

LauraConcepcion:
## Where

* **Related Issue:** #1838 

## What

Generalize postal code verification methods.

## How

Rename methods to generalize them from Madrid to location. Create new setting parameter to insert the postal code valid values for a city (as a range or array). Also added dev_seed and seed entries with initial values.

## Test

Removed former custom test and added new scenarios to main verification residence model test.

## Deployment

Created dev_seed and seed with an example configuration, it must be corrected in Administration -> Global Configuration (view).

## Release notes

:warning: We've changed the way to define valid postal codes for the users verification. In the past, you had to change the code in `app/models/custom/verification/residence.rb`. Now CONSUL administrators can define postal codes in the settings section of the administration. If you had already defined custom postal codes in `app/models/custom/verification/residence.rb`, when you get a conflict when merging these changes, keep your changes; they'll most probably still work but, depending on your implementation, the postal codes defined by administrators might be ignored. In this case, you might want to double-check your custom changes still work as expected and consider removing this custom code and use the setting instead. In order to ease the transition, the setting allows regular expressions. When using regular expressions, omit the `^` and `$` characters; we're already adding them in the code. For example, if your code was `postal_code =~ /^280[0-9]{2}$/`, use `280[0-9]{2}` in the setting.


PierreMesure:
I just thought of the same improvement. I would add that you should change the error message that appears when the postal code doesn't match the Regex. Today, it says: "In order to be verified, you must be registered.". It should just say something like "Invalid postcode".

javierm:
@LauraConcepcion Thank you for this pull request! And sorry it took four years to finally have a look at it :pray: :grimacing:.

I've updated the pull request to use the `local_postal_code` and `local_residence` methods introduced in #4381 and the way to define settings introduced in pull request #3564.

After releasing version 1.4.0, we'll have a closer look and check for compatibility with the way other countries define postal codes.

taitus (approved):

javierm:
@LauraConcepcion Thank you so much! Really appreciate it.

PR #2134, opened 2017-11-16
Add a previous question link to debates

### LINKED ISSUE #1589: Add a previous question link to debates
In the discussion of a legislations, there is a "Next question" link but there is no "Previous question" link
next question

iagirre:
Where
=====
This PR close #1589 

What
====
- Add a "Previous question" link in the legislation debate's header.
- Change the "Share" `h3` to `p` in order to match the styles.

How
===
I added the HTML needed for rendering the button and the functions in the model to get the correct ids. When the first question is reached, the link changes to "Last question", as happens when the last question is reached (the link goes to the first question).

Screenshots
===========
back_question01
back_question02

Test
====
Test coverage increased to check:
- the UI navigation through questions.
- the functions that get the previous and last question id.

Deployment
==========
Nothing to apply.

Warnings
========
There is some kind of conflic with the `icon-angle-left` class that makes the arrow icon invisible on hover.
back_question03_warning
The conflict comes from the `layout.scss`, where the icon class is styled:
```
.back,
.icon-angle-left {
  clear: both;
  color: $text-medium;
  float: left;
}
```

### DIFF
--- a/app/assets/stylesheets/legislation_process.scss
+++ b/app/assets/stylesheets/legislation_process.scss
@@ -203,29 +203,38 @@
       }
     }
 
-    .quiz-next,
-    .quiz-previous {
+    %quiz {
       background: #ccdbe5;
       color: $brand;
       font-size: $small-font-size;
       font-weight: bold;
-      text-align: right;
       text-transform: uppercase;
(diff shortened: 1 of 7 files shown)


javierm:

Issue #1625, opened 2017-06-09
Investment's CSV Exporter


voodoorai2000:
In Public Institutions it's common to use excel files to share work in progress.
In the transition to substitute excel files for web interfaces these type of files are still useful.

Add a csv generation of the current selection in the admin investment's section.

rfuentesm2c:
I'll take this one. :)

rfuentesm2c:
maybe we can use https://github.com/randym/axlsx to export directly in xlsx

bertocq:
Closed by #1106

PR #2147, opened 2017-11-21
added csv download link to budget_investments (issue 1625)

raul-fuentes:
Where
=====
* **Related Issue:**  https://github.com/consul/consul/issues/1625

What
====
- Added download csv link to budget investments

How
===
- it just replicates the content of the html table into CSV

Screenshots
===========
screenshot from 2017-11-21 13 35 26

Test
====
- works with current tests

Deployment
==========
- No

Warnings
========
- No

### DIFF
--- a/app/controllers/admin/budget_investments_controller.rb
+++ b/app/controllers/admin/budget_investments_controller.rb
@@ -65,19 +65,19 @@ class Admin::BudgetInvestmentsController < Admin::BaseController
       csv_string = CSV.generate(headers: true) do |csv|
         csv << attributes
         @investments.each do |investment|
-          id    = investment.id.to_s
+          id = investment.id.to_s
           title = investment.title
           total_votes = investment.total_votes.to_s
-          if investment.administrator.present?
-            administrator = investment.administrator.name
-          else
-            administrator = t("admin.budget_investments.index.no_admin_assigned")
-          end
-          if investment.valuators.empty?
-            valuators = t("admin.budget_investments.index.no_valuators_assigned")
-          else
-            valuators = investment.valuators.collect(&:description_or_name).join(', ')
-          end
+          administrator = if investment.administrator.present?
+                            investment.administrator.name
+                          else
+                            t("admin.budget_investments.index.no_admin_assigned")
+                          end
+          valuators = if investment.valuators.empty?
+                        t("admin.budget_investments.index.no_valuators_assigned")
+                      else
+                        investment.valuators.collect(&:description_or_name).join(', ')
+                      end
           heading_name = investment.heading.name
           price = t("admin.budget_investments.index.feasibility.#{investment.feasibility}", price: investment.formatted_price)
           valuation_finished = investment.valuation_finished? ? t('shared.yes') : t('shared.no')
--- a/app/controllers/admin/budget_investments_controller.rb
+++ b/app/controllers/admin/budget_investments_controller.rb
@@ -79,7 +79,8 @@ class Admin::BudgetInvestmentsController < Admin::BaseController
                         investment.valuators.collect(&:description_or_name).join(', ')
                       end
           heading_name = investment.heading.name
-          price = t("admin.budget_investments.index.feasibility.#{investment.feasibility}", price: investment.formatted_price)
+          feasibility_string = "admin.budget_investments.index.feasibility.#{investment.feasibility}"
+          price = t(feasibility_string, price: investment.formatted_price)
           valuation_finished = investment.valuation_finished? ? t('shared.yes') : t('shared.no')
           csv << [id, title, total_votes, administrator, valuators, heading_name, price, valuation_finished]
         end
--- a/app/controllers/admin/budget_investments_controller.rb
+++ b/app/controllers/admin/budget_investments_controller.rb
@@ -1,7 +1,5 @@
 class Admin::BudgetInvestmentsController < Admin::BaseController
 
-  require 'csv'
-
   include FeatureFlags
   feature_flag :budgets
 
@@ -18,7 +16,8 @@ class Admin::BudgetInvestmentsController < Admin::BaseController
     respond_to do |format|
       format.html
       format.csv do
-        send_data set_csv, filename: 'budget_investments.csv'
+        send_data Budget::Investment.to_csv(@investments, {headers: true}),
+                  filename: 'budget_investments.csv'
       end
     end
   end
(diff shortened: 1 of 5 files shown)


app/controllers/admin/budget_investments_controller.rb
```diff
@@ -43,6 +52,41 @@ def toggle_selection
...
+                          else
+                            t("admin.budget_investments.index.no_admin_assigned")
+                          end
+          valuators = if investment.valuators.empty?
+                        t("admin.budget_investments.index.no_valuators_assigned")
+                      else
+                        investment.valuators.collect(&:description_or_name).join(', ')
+                      end
+          heading_name = investment.heading.name
+          price = t("admin.budget_investments.index.feasibility.#{investment.feasibility}", price: investment.formatted_price)
```
bertocq: Could we do the interpolation outside of the `i18n.t` method call? So i18n-tasks will be able to check this string is in use?

app/views/admin/budget_investments/_investments.html.erb
```diff
@@ -55,26 +58,9 @@
...
-          <%= link_to toggle_selection_admin_budget_budget_investment_path(@budget,
-                                                                           investment,
-                                                                           filter: params[:filter],
-                                                                           page: params[:page]),
-                      method: :patch,
-                      remote: true,
-                      class: "button small hollow expanded" do %>
-            <%= t("admin.budget_investments.index.select") %>
-          <% end %>
+        <%= options = investment_selected_link_options(investment) %>
```
bertocq: This `<%=` should be `<%` to avoid printing the result of the function. We normally require screenshots to both improve the review and also to "force" the coder to manually verify interface changes just in case something happens...

app/views/admin/valuators/search.html.erb
```diff
@@ -29,7 +29,12 @@
                 <%= t("admin.valuators.index.no_description") %>
               <% end %>
             <td>
-              <% unless user.valuator? %>
+              <% if user.valuator? %>
+                <%= link_to t("admin.valuators.valuator.delete"),
```
bertocq: Not sure why this changes appear here, seems like your branch isn't rebased with master, and some commits from angel and xuanxu came along on the PR making it a bit more difficult to review

spec/features/admin/budget_investments_spec.rb
```diff
@@ -588,7 +588,23 @@
...
+      investment2 = create(:budget_investment, :feasible, budget: @budget)
 
+      visit admin_budget_budget_investments_path(@budget, format: :csv)
+
+      header = page.response_headers['Content-Disposition']
+      header.should match(/^attachment/)
+      header.should match(/filename="budget_investments.csv"$/)
+
+      expect(page).to have_content investment2.title
+      expect(page).to have_content investment1.title
```
bertocq: I like this way to check de csv download 👍  😃  I'm only missing a negative scenario, I mean: would it be possible to check that filtering by feasible only budget investments, the unfeasible ones doesn't appear at the file? Or any other scenario that you see easier to reproducer, where we can verify that only the investments that the user see are at the html view the ones exported in the file :)

app/controllers/admin/budget_investments_controller.rb
rfuentesm2c: In admin.yml, the translation admin.budget_investments.index.feasibility includes at the end (%{price})".
Should I change the yml file?

app/views/admin/valuators/search.html.erb
rfuentesm2c: Im working on it

app/controllers/admin/budget_investments_controller.rb
bertocq: No, what I mean is that instead of:

spec/features/admin/budget_investments_spec.rb
rfuentesm2c: Ok, It will be in a moment

app/controllers/admin/budget_investments_controller.rb
rfuentesm2c: Oh, you're right. I'll change it

app/controllers/admin/budget_investments_controller.rb
```diff
@@ -43,6 +52,42 @@ def toggle_selection
 
   private
 
+    def set_csv
```
voodoorai2000: This method is a bit too large and does not belong in a controller
Could you extract this logic to the investment's model?
Something on these lines http://railscasts.com/episodes/362-exporting-csv-and-excel

app/helpers/budgets_helper.rb
```diff
@@ -1,5 +1,23 @@
 module BudgetsHelper
 
+  def csv_params
+    csv_params = params.clone.merge(format: :csv)
+    csv_params.delete(:page)
+    csv_params
+  end
+
+  def investment_selected_link_options(investment)
```
voodoorai2000: Great refactoring 👍 
Keep in mind that if you run into a refactoring to do, it's nicer to do it in a separate PR

app/views/admin/budget_investments/_investments.html.erb
```diff
@@ -55,26 +58,9 @@
...
-                                                                           investment,
-                                                                           filter: params[:filter],
-                                                                           page: params[:page]),
-                      method: :patch,
-                      remote: true,
-                      class: "button small hollow expanded" do %>
-            <%= t("admin.budget_investments.index.select") %>
-          <% end %>
+        <% options = investment_selected_link_options(investment) %>
+        <%= link_to toggle_selection_admin_budget_budget_investment_path(@budget,  investment, filter: params[:filter], page: params[:page]), method: :patch, remote: true, class: options[:link_class] do %>
```
voodoorai2000: This line is too long
Try to keep the length of lines to 80 characters or less :ok_hand:
A helper method would be great here or just splitting attributes in separate lines

spec/features/admin/budget_investments_spec.rb
```diff
@@ -588,7 +588,36 @@
         expect(page).to_not have_link('Selected')
       end
     end
+  end
+
+  context "Selecting csv" do
```
voodoorai2000: Great 👌 
Please also test for the presence of each attribute in the csv

raul-fuentes:
hello @voodoorai2000  the changes that you requested are already made, I hope that now everything is correct.

voodoorai2000:
Excellent @raul-fuentes!

PR #2152, opened 2017-11-24
Deleted items

### LINKED ISSUE #1696: Deleted items
## Background

Now when a comment, debate, proposal,...  is deleted, if you try to go to their link the app redirect to a **404 status page**. 

This is a problem for SEO if this url was indexed previously.

## What 

We need show the deleted item (comment, debate, proposal...) with a `This item was deleted` message instead of a 404 page.

It's already done in the show views for deleted comments (if they have children comments) and should be equal in the single views:

deleted_comment

ortegacmanuel:
## References

* Closes #1696

## Objectives

Show a «item deleted message» when user gets to CONSUL by the url of a deleted/hidden item

## Visual changes

itemdeleted


app/views/shared/item_deleted.html.erb
```diff
@@ -0,0 +1,14 @@
+<div class="item-deleted-show">
+  <div id="<%= dom_id(@resource) %>" class="row">
+    <div class="small-12 medium-9 column">
+      <div class="callout primary">
+        <%= t("shared.deleted_item",
+              resource: link_to(t("#{@resource.class.to_s.downcase.pluralize}.show.#{@resource.class.to_s.downcase.pluralize}"), proposals_path, class: "probe-message"),
+              home_page: link_to(t("shared.homepage"), root_path, class: "probe-message"),
+              org: setting['org_name']).html_safe %>
+      </div>
+    <aside class="small-12 medium-3 column">
```
decabeza: Hi @ortegacmanuel! We have to avoid empty HTML, could you  please delete this empty `<aside>` 🙏 

Thanks!

config/locales/es/general.yml
```diff
@@ -449,6 +449,7 @@ es:
       title_external_url: "Documentación adicional"
       title_video_url: "Vídeo externo"
       author: Autor
+      proposals: Propuestas
```
bertocq: Do we need this proposals string? If so it should be present as well on the english translation file, or crowdin will delete this key

app/views/shared/item_deleted.html.erb
ortegacmanuel: Done! I also added a missing `</div>`. Doing this change I noticed that the file's name `item_deleted` does not start with an `_`. Should I rename it in order to follow that name formate. All other files's names in shared starts with an `_` 

If is needed to rename the file name, just tell me ;)

config/locales/es/general.yml
ortegacmanuel: That string is not needed. I deleted it and instead I use those one in the layous header section.

ortegacmanuel:
@bertocq @decabeza Working on your reviews I noticed that the link to the index of the resource was always pointing to `proposals_path`. I just added a new commit in order to solve this. Now the link to the index path of the resource is generated dynamically.

javierm:
@ortegacmanuel Thank you very much for this pull request! :smile:

Sorry it took us so long to review it! :pray: The pull request itself looks fine; however, the requirements in the original issue are not :sweat_smile:.

I've updated the original issue with sources mentioning the right thing is to return a 404 code (maybe 410). Returning a 200 code would make search engines index the page.

However, it's true our 404 page could be improved a lot :thinking:, so I've updated the issue title. Would you still like to have a go at it? It's perfectly understandable if you don't :wink:. After all, it took us 2 years to review this pull request.

ortegacmanuel:
@javierm no problem. I take a look at this later this week. :+1:

javierm:
@ortegacmanuel How are you doing? :smile:

Are you still working on this pull request?

javierm:
Closing since the original author doesn't seem to be working on this pull request anymore.

PR #2165, opened 2017-12-04
Knapsack Pro gem

ArturT:
Where
=====
* travis.yml configuration

What
====
* added knapsack_pro gem https://knapsackpro.com and configure it to use Queue Mode which allocates tests across CI nodes in a dynamic way to save time execution for each CI build.
* The forked version of the https://github.com/consul/consul repo can still run tests without exposing the secret API key `KNAPSACK_PRO_TEST_SUITE_TOKEN_RSPEC` thanks to fallback mode as described here: https://github.com/KnapsackPro/knapsack_pro-ruby#how-to-make-knapsack_pro-works-for-forked-repositories-of-my-project

How
===
* added `bin/knapsack_pro_rspec` that can handle forked version repo. Based on https://github.com/KnapsackPro/knapsack_pro-ruby#how-to-make-knapsack_pro-works-for-forked-repositories-of-my-project

Test
====
- ensure the CI build is green (I rebased my branch with the current master which was red, unfortunately)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ before_script:
   - bundle exec rake db:setup
 script:
   - "bundle exec rake assets:precompile RAILS_ENV=test"
-  - "bundle exec rake knapsack_pro:rspec"
+  - "bundle exec rake knapsack_pro:queue:rspec"
 env:
   global:
     - KNAPSACK_PRO_CI_NODE_TOTAL=2
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,10 +10,10 @@ before_script:
   - bundle exec rake db:setup
 script:
   - "bundle exec rake assets:precompile RAILS_ENV=test"
-  - "bundle exec rake knapsack:rspec"
+  - "bundle exec rake knapsack_pro:rspec"
 env:
   global:
-    - CI_NODE_TOTAL=2
+    - KNAPSACK_PRO_CI_NODE_TOTAL=2
   matrix:
-    - CI_NODE_INDEX=0
-    - CI_NODE_INDEX=1
+    - KNAPSACK_PRO_CI_NODE_INDEX=0
+    - KNAPSACK_PRO_CI_NODE_INDEX=1
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ before_script:
   - bundle exec rake db:setup
 script:
   - "bundle exec rake assets:precompile RAILS_ENV=test"
-  - "bundle exec rake knapsack_pro:queue:rspec"
+  - "bin/knapsack_pro_rspec"
 env:
   global:
     - KNAPSACK_PRO_CI_NODE_TOTAL=2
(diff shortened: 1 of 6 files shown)


ArturT:
@voodoorai2000 I've rebased my PR with the latest master and the test suite is green now.

voodoorai2000:
Awesome! Thank you very much @ArturT 😃🎉

ArturT:
@voodoorai2000 cool :)

bertocq:
Hey @ArturT we're not getting specs run on Travis :( `No test files were executed on this CI node.` Check https://travis-ci.org/consul/consul/builds/315338295 its ✅  but 2mins on each node was too impresive (took about 10-15min each) so once I opened I realized its not running any tests :(

I see the explanation:
```
D, [2017-12-12T13:25:14.118580 #6201] DEBUG -- : [knapsack_pro] When you use knapsack_pro queue mode then probably reason might be that CI node was started after the test files from the queue were already executed by other CI nodes. That is why this CI node has no test files to execute.

D, [2017-12-12T13:25:14.118595 #6201] DEBUG -- : [knapsack_pro] Another reason might be when your CI node failed in a way that prevented knapsack_pro to save time execution data to Knapsack Pro API and you have just tried to retry failed CI node but instead you got no test files to execute. In that case knapsack_pro don't know what testes should be executed here.

```

But not sure how to fix it... As I understand the described scenario its not what's happening as both nodes have no specs run.. and I can't find any crashes in the build that may have prevented knapsack from executing them :(. Any ideas? 🙏  thank you in advance!

ArturT:
@bertocq Could you set `KNAPSACK_PRO_FIXED_QUEUE_SPLIT=true` env variable. This way you will be able to retry CI build correctly. 

Explanation here:
https://github.com/KnapsackPro/knapsack_pro-ruby/#knapsack_pro_fixed_queue_split-remember-queue-split-on-retry-ci-node

bertocq:
Thanks @ArturT ! I just added the variable at travis via web settings, but I assume it would be better at the [`.travis.yml` file](https://github.com/consul/consul/blob/master/.travis.yml#L14-L20) right? or maybe it won't be a good idea for the forks that inherit that configuration?

bertocq:
Anyway its still behaving the same :( https://travis-ci.org/consul/consul/jobs/315338297 this is the env variables that are set on travis:

screen shot 2017-12-12 at 15 35 13

ArturT:
@bertocq The `KNAPSACK_PRO_FIXED_QUEUE_SPLIT=true` does not matter for forked repos.

It will start working for new CI builds so Knapsack Pro API will cache them. You can create a new commit for the particular branch you want to run.

bertocq:
Awesome @ArturT !! I just [added the variable to travis.yml file](https://github.com/consul/consul/commit/26437029a610b5c347ba4ba5def15e08cb6adf8d) and pushed the commit... the build this time ran all the specs as expected! 👏  :)

Thanks for your support and neat trick for having KnapsackPro at Consul without affecting the forks :)

PR #2180, opened 2017-12-13
Adds timestamps to polls

voodoorai2000:
What
====
- Adds timestamps to polls

Why
===
- Generally a good practice to have them in every table

Test
====
- Not needed

Deployment
==========
- As usual

Warnings
========
- Existing Polls will still not have a created_at value. 
  A manual update or a rake task would be needed for this.

Update
===
Added rake task for the happiness of forks

Release Notes
===
Run `rake polls:initialize_timestamps ` to initialize attributes `created_at` and `updated_at` with the current time for all existing polls


bertocq:
Just to remember things, we talked about setting `null: false` on migration and creating a rake task to set Time.current for existing Polls (and some notes to copy to the v0.12 release Changelog to remember/explain usage)

PR #2181, opened 2017-12-13
Updates factory traits

voodoorai2000:
What
====
- Add `residence_verified_at` to the `verified` trait
- Makes a legislation process's debate phase open by default 

Why
===
- In most places we can just use the `verified_at` attribute to know if a user is verified.
  [Other places](https://github.com/consul/consul/blob/master/app/helpers/comments_helper.rb#L80) rely on the residece_verified_at attribute to be set.
- Having all legislation phases open by default makes testing easier

Test
====
- Not needed

Deployment
==========
- As usual

Warnings
========
- None


voodoorai2000:
🎉

PR #2182, opened 2017-12-13
Adds model validations for legislation proposals

voodoorai2000:
What
====
- Adds validation model specs for legislation proposals

Why
===
- Always good to add some tests
- Needed for an upcoming PR which requires having the spec file present

Deployment
==========
- As usual

Warnings
========
- None


voodoorai2000:
💃

Issue #944, opened 2016-02-29
Admin notifications


voodoorai2000:
To inform users of updates, there should be an admin section where notifications can be created. This notifications would appear for all users in their section '/notifications'.

palomahnhp:
Lo hacemos @MaiteHdezRivas y @palomahnhp

ManjiChiba:
Precisamente iba crear un hilo sobre este mismo tema.
- Aviso por mensaje de que una Propuesta ha pasado a 3ª fase. (53 apoyos)
     https://decide.madrid.es/proposals/8704

iagirre:
I'll do this one.

decabeza:
Closed in favor to https://github.com/consul/consul/issues/2145

PR #2187, opened 2017-12-13
Fix notifications' exceptions 

voodoorai2000:
Relate PR
===
- Notifications for hidden resources https://github.com/consul/consul/pull/2172

Why
===
- Some users are experiencing exceptions when viewing their notifications

What
====
- Refactors notification system
- Tests every combination of in-app notifications with the different notifiable models
- Gotchas solved:
  - Notifications for resources that have been hidden by a moderator
  - Notifications for resources that have been retired by the author
  - Notifications for resources that did not have a created_at column

How
===
- Using [a concern](https://github.com/consul/consul/compare/notifications?expand=1#diff-9c1ce30d792136b7759e6318cc62972fR1) to extract the logic common to every notifiable resource
- Using shared_examples to test [model specs](https://github.com/consul/consul/compare/notifications?expand=1#diff-e709b99b823bfca535554a5bc72c71dcR1) and [feature specs](https://github.com/consul/consul/compare/notifications?expand=1#diff-e1a307428131a510750b459ce98ac822R1) related to notifications for each resource that is notifiable through in-app notifications (🔔 )

Test
====
- Staging environment testing will be required to make sure all users can now see their notifications

Deployment
==========
- As usual

Warnings
========
- [Proposal Notifications](https://github.com/consul/consul/pull/1140) are a special case. Tests are [duplicated](https://github.com/consul/consul/compare/notifications?expand=1#diff-8c85dc8fdaf2a587e2b2ef04eccd7becR311). It should be refactored in a separate PR. Probably best once we continue working on [Admin Notifications](https://github.com/consul/consul/issues/944) and seeing how these two types would fit in nicely with the exiting Notifications about comments and comment replies.


app/models/concerns/notifiable.rb
```diff
@@ -0,0 +1,35 @@
+module Notifiable
+  extend ActiveSupport::Concern
+
+  def notifiable_title
```
bertocq: I think just `title` could be good enough, like `notification.title` at [the only usage of `notificable_title`](https://github.com/consul/consul/blob/master/app/views/notifications/_notification.html.erb#L9), as it was a column in the Notifications table.

Issue #2168, opened 2017-12-11
Proposal Recommendations


voodoorai2000:
What
===
Proposal recommendations should not include archived nor already supported proposals

aitbw:
I'm going in with this one.

bertocq:
Closed via #2200

PR #2200, opened 2017-12-15
New scopes for proposal recommendations

aitbw:
Where
=====
* **Related Issue:** #2168 

What
====
* Proposal recommendations for users on `WelcomeController#index` won't include archived or already supported records

How
===
* Usage of chained scopes (A new scope called `not_supported_by_user` was defined)
* Fixed rubocop warnings

Test
====
* Increased coverage to make sure proposals recommendations won't return `archived` or `already supported` records. Also, manual testing was done in order to make sure ActiveRecord was generating the expected SQL query

### DIFF
--- a/app/models/proposal.rb
+++ b/app/models/proposal.rb
@@ -28,8 +28,8 @@ class Proposal < ActiveRecord::Base
 
   belongs_to :author, -> { with_hidden }, class_name: 'User', foreign_key: 'author_id'
   belongs_to :geozone
-  has_many :comments, as: :commentable
-  has_many :proposal_notifications
+  has_many :comments, as: :commentable, dependent: :destroy
+  has_many :proposal_notifications, dependent: :destroy
 
   validates :title, presence: true
   validates :question, presence: true
@@ -41,7 +41,7 @@ class Proposal < ActiveRecord::Base
   validates :description, length: { maximum: Proposal.description_max_length }
   validates :question, length: { in: 10..Proposal.question_max_length }
   validates :responsible_name, length: { in: 6..Proposal.responsible_name_max_length }
-  validates :retired_reason, inclusion: {in: RETIRE_OPTIONS, allow_nil: true}
+  validates :retired_reason, inclusion: { in: RETIRE_OPTIONS, allow_nil: true }
 
   validates :terms_of_service, acceptance: { allow_nil: false }, on: :create
 
@@ -51,8 +51,8 @@ class Proposal < ActiveRecord::Base
 
   before_save :calculate_hot_score, :calculate_confidence_score
 
-  scope :for_render, -> { includes(:tags) }
-  scope :sort_by_hot_score, -> { reorder(hot_score: :desc) }
+  scope :for_render,               -> { includes(:tags) }
+  scope :sort_by_hot_score,        -> { reorder(hot_score: :desc) }
   scope :sort_by_confidence_score, -> { reorder(confidence_score: :desc) }
   scope :sort_by_created_at,       -> { reorder(created_at: :desc) }
   scope :sort_by_most_commented,   -> { reorder(comments_count: :desc) }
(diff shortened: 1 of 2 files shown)


spec/models/proposal_spec.rb
```diff
@@ -952,6 +952,18 @@
       expect(result).to eq [proposal2]
     end
 
+    it "should not return archived proposals" do
+      archived_proposal = create(:proposal, :archived)
+      result = Proposal.recommendations(user)
+      expect(result.size).to eq(0)
```
bertocq: I think the test scenario would be more robust if there was a non-archived proposal so we know the filter is working (letting through only the ones that should). Same at the other spec at line 961

app/models/proposal.rb
```diff
@@ -75,12 +75,20 @@ def self.recommendations(user)
...
+      .not_archived
+      .not_supported_by_user(user)
   end
 
   def self.not_followed_by_user(user)
     where.not(id: followed_by_user(user).pluck(:id))
   end
 
+  def self.not_supported_by_user(user)
+    @supported_by_user = user.find_voted_items(votable_type: "Proposal")
```
bertocq: Not sure why we need a class variable `@supported_by_user` instead of just a variable `supported_by_user`

app/models/proposal.rb
```diff
@@ -75,12 +75,20 @@ def self.recommendations(user)
...
+    @supported_by_user = @supported_by_user.reject { |p| p.nil? } if @supported_by_user.include?(nil)
```
bertocq: I tried to reproduce the scenario where the `user.find_voted_items(votable_type: "Proposal")` query would return empty/nil results but couldn't (I tried softdeleting as you said a voted Proposal with `.destroy`)

app/models/proposal.rb
aitbw: Not needed, actually, but it feels kinda weird using local variables.

bertocq: But weird in what way? I mean the scope of the variable (global, class, local, constant, instance...) it

app/models/proposal.rb
aitbw: With a freshly seeded database, the admin user has 3 votes: 2 for debates and 1 for a proposal, but that proposal is hidden/soft-deleted. The `find_voted_items` method fetches this record (Apparently, it completely ignores the `acts_as_paranoid` behavior since `user.find_voted_items(votable_type: 'Proposal').count` outputs `1`) as `nil`, which returns the following array:

`[nil, Debate#1, Debate#2]`

If we use the aforementioned method only to fetch an user's upvoted proposals, we would receive the following array: `[nil]`

And then, when the `recommendations` method is chained to the `sort_by_recommendations` scope, it will raise an exception since Ruby's `NilClass` doesn't know anything about this, hence we need to **exclude** those `nil` elements to avoid the error.

app/models/proposal.rb
aitbw: I know, I know. I'm just so accustomed to instance variables (because of how common they are) that it's weird for me to use local variables 😅 But don't worry, I'll fix that.

bertocq: But the point of instance variables is to be able to use them in other places, if its not needed then 🤔 , well there's also [memoization](https://www.justinweiss.com/articles/4-simple-memoization-patterns-in-ruby-and-one-gem/) but still this scenario doesn't ask for it. Problem with using something not needed is that it distracts the code reader by making he guess what's going on and where's that instance variable used as well.

bertocq: And I'm talking about it like this is totally wrong or it could lead to bad performance... we could totally let it there forever. But since you seem concerned as well with code consistency & readability (because I see you normally check with rubocop the files you touch) maybe you saw it the same way as me

aitbw:
PR updated accordingly, @bertocq :D

Issue #2194, opened 2017-12-15
Dev Seeds spec


bertocq:
Sometimes we break dev seeds rake task without noticing, a simple spec that runs it and checks everything goes by as expected would be great

voodoorai2000:
+1

aitbw:
I'm gonna go with this one :)

bertocq:
closed via #2201  👏

PR #2201, opened 2017-12-18
Add spec for 'rake db:dev_seed' task

aitbw:
Where
=====
* **Related Issue:** #2194

What
====
* Adds a spec for `rake db:dev_seed` to make sure the DB is properly seeded (useful when a table attribute is removed/deprecated)

How
===
* Wrote a rake task spec under `spec/lib/tasks`

Test
====
* Increased coverage for the aforementioned spec


bertocq:
👏  @aitbw !

Issue #2171, opened 2017-12-11
Failure when attaching a second document to a Proposal


bertocq:
## Current Behaviour
If you create a Proposal with more than one document you'll get the following failure:

![screen shot 2017-12-11 at 11 26 24](https://user-images.githubusercontent.com/983242/33826708-2e306cb6-de66-11e7-898e-80621b1a5708.jpg)

## Expected Behaviour
Obviously not a crash 😅 , but a valid proposal with more than one document attached.

## Reasons behind crash
As far as I've researched, the problem is that all attached document  hidden `cached_attachment `  inputs on the Proposal creation form have the same value (from the first document) so as the form submission is processed the temp files are removed and when it tries to create the second attachment with the first attachment temp file... it doesn't exist anymore.

## Proposed solution path
I would:
- Find all involved specs (not sure if only `/spec/shared/features/nested_documentable.rb`) and increase coverage with an scenario that tries to add 3 attached documents
- Solve the issue at hand (not sure if js only or there's ruby involved)
- Try both specs and manual test

Update
===
We have received some complaints from users about this.
Remember to inform the appropriate people when it's fixes

iagirre:
I'll go with this.

iagirre:
As the issue sais, the problem starts with the name of the attached documents: they all have the exact same name, so, when you upload a second one, the first one is overwriten. I realized that this isn't only happening when you upload 2 files in the same documentable object; it also happens if you upload 1 file in one of them and this same user uploads another one in a different object (e.g. upload a file to Proposal 1, open Proposal 2 in another tab, upload another file and save proposal. Go to Proposal 1 tab, save it and get this error).

I've managed to solve the problem changing the hash_data that generates the file name:
```
  has_attached_file :attachment, url: "/system/:class/:prefix/:style/:hash.:extension",
                                 hash_data: ":class/:attachment/:id/:style/:updated_at",
                                             # insetad of ":class/:style"
                                 use_timestamp: false,
                                 hash_secret: Rails.application.secrets.secret_key_base
```

Note: It works with any hash_data that differs one from another.

The problem is that now I'm facing another error when dealing with cached elements (it doesn't delete them), so I'm trying to figure out if that solutions is the correct one.

iagirre:
### The problem
As I said in the previous comment, if you generate a `hash_data` that is unique for each file uploaded the error disappears; the files are uploaded and saved for the documentable. So far, so good.

At this moment, the paperclip configuration is this:
```
[...]

PR #2218, opened 2017-12-22
[WIP] Failure when attaching second document to documentable

iagirre:
Where
=====
* **Related Issue:** #2171 

What
====
When a user tries to upload two documents to one documentable, the app crashes. Documents can only be uploaded one by one.

How
===
As explained in the issue, there is an error when generating the hash for a uploaded document. The solution provided in this PR works as long as the title of the document remains the same and the user doesn't upload 2 or more documents with the same title (and at the same time).

Screenshots
===========
Not yet.

Test
====
A spec has been added to test the functionality, but the second document is not been uploaded.

Deployment
==========
Nothing.

Warnings
========
Nothing.


bertocq:
Thanks @iagirre for your great research effort! This was fixed at master branch

PR #2264, opened 2018-01-06
Enables Capybara/FeatureMethods cop and fixes all issues

bertocq:
What
====
We use before/it on rspec tests, and background/scenario on feature specs
that use capybara, but many times we don't follow those rules and mix
things up. Choosing one way or the other would make things more consistent
and easy to read.

Read about the cop at http://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/Capybara/FeatureMethods

How
===
Enabling the cop at .rubocop.yml file and running `rubocop -a --only Capybara/FeatureMethods`

Screenshots
===========
No need

Test
====
Passing

Deployment
==========
As usual

Warnings
========
This is a suggestion to improve code readability and consistency, but some very used to background/scenario from years of experience with capybara may not feel comfortable with this.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -45,3 +45,6 @@ RSpec/BeforeAfterAll:
 
 Capybara/CurrentPathExpectation:
   Enabled: true
+
+Capybara/FeatureMethods:
+  Enabled: true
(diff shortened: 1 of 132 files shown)


bertocq:
After re-reading myself the changes, it doesn't even make sense, closing

Issue #2277, opened 2018-01-08
Improve Budget investment form


decabeza:
What we need
===

- Delete "Link to additional documentation" input field (we already have a new feature to upload documents)
- Change label "Location" to "Additional info of location"

MariaCheca:
Closed here: https://github.com/consul/consul/pull/2280

PR #2280, opened 2018-01-08
Budgets new form

MariaCheca:
Where
=====
* **Related Issue:** https://github.com/consul/consul/issues/2277

What
====
- Removed `external_link` input in budget investment form. The database column will remain for previous budgets where the attribute was used, as it could be useful at some point, or just to keep all the original data as it was introduced.
- Changed `location` input label.

Screenshots
====
localhost_3000_budgets_1_investments_new laptop with mdpi screen

Test
====
As usual.


config/locales/es/budgets.yml
```diff
@@ -58,6 +58,7 @@ es:
         map_location: "Ubicación en el mapa"
         map_location_instructions: "Navega por el mapa hasta la ubicación y coloca el marcador."
         map_remove_marker: "Eliminar el marcador"
+        location: "Información adicional de la localización"
```
bertocq: Maybe to be consistent with other spanish texts around locations the word **localización** should be changed for **ubicación** as that's what the map uses 3 times on title/subtitle/checkbox

MariaCheca: Good catch!

spec/features/budgets/investments_spec.rb
```diff
@@ -311,7 +311,6 @@ def investments_order
       select  'Health: More hospitals', from: 'budget_investment_heading_id'
       fill_in 'budget_investment_title', with: 'Build a skyscraper'
       fill_in 'budget_investment_description', with: 'I want to live in a high tower over the clouds'
-      fill_in 'budget_investment_external_url', with: 'http://http://skyscraperpage.com/'
```
bertocq: I can see 3 more `budget_investment_external_url` occurences in the codebase 🤔  in two more tests:

bertocq:
Maybe
https://github.com/consul/consul/blob/master/db/dev_seeds.rb#L441
and
https://github.com/consul/consul/blob/master/spec/factories.rb#L282
should be removed as well

To completly deprecate the usage of the attribute

app/views/budgets/investments/_investment_show.html.erb
```diff
@@ -51,12 +51,6 @@
         </div>
       <% end %>
 
-      <% if investment.external_url.present? %>
```
bertocq: But... we want existing budgets with external_url's but no documents to still show those values right?

decabeza: Yes, we want to disallow this input field, but maintain the existing links 😌

spec/features/budgets/investments_spec.rb
decabeza: Yes, in management panel also there is a new budget investment form with this field ☺️

PR #2319, opened 2018-01-15
Improves admin banners section

### LINKED ISSUE #1626: Improves admin banners section
During a Participatory Budget some users get confused and create Citizen Proposals instead of Investments.

Ponder on how to make clear the process of creating an investment.

iagirre:
Where
=====
* **Related Issue:** #1626 (this PR closes #1626)

What
====
Improve the generation of a banner adding two color-pickers to the form so that the admin can choose the background color and the font color separately.
Some checkboxes were added also to let the admin choose the web sections where each banner will appear.

How
===
- Add two HTML5 color picker form input, one for the background color and another one for the font color. Some text_fields were added also to each picker, because it wont work correctly in IE.
- Add the checkboxes to select the web sections where the banner will appear. For this, a new DB table has been added to store the sections.
- Now, banners show the choosen color for the background and for the font.

Screenshots
===========
banner01

Test
====
- Feature specs to test the color changing when selecting from the color-picker or when writting in the text-field.
- The color selection has been added to the generation of the banner. 

Deployment
==========
- Run rake task `rake web_sections:generate` to fill the web_sections table.
- Run rake task `rake banners:migrate_style` to migrate the banner `style` column to `background_color` and `font_color`, so that the banners won't appear all white.

Warnings
========
- `dev_seeds` where updated to include the generation of the web_sections.

Deprecation warnings
=
The Changelog of v0.13 should include two warnings:
- Columns `style` and `image` from `banners` table will be removed in v0.14. Run `rake banners:migrate_style` to migrate the colors to `background_color` and `font_color`.
- Banner images (app/assets/images/banners) will be removed in v0.14, as they won't be used anymore. If you want to keep them, backup them manually.


app/assets/javascripts/banners.js.coffee
```diff
@@ -23,3 +31,21 @@ App.Banners =
     $("#banner_image").on
```
MariaCheca: I would remove all code related to banner images, as they won't be used anymore (also from the Banners ddbb table).

app/controllers/admin/banners_controller.rb
```diff
@@ -38,7 +39,10 @@ def destroy
   private
 
     def banner_params
-      params.require(:banner).permit(:title, :description, :target_url, :style, :image, :post_started_at, :post_ended_at)
+      params.require(:banner).permit(:title, :description, :target_url, :style, :image,
```
MariaCheca: Also `:image` should be removed.

app/controllers/admin/banners_controller.rb
```diff
@@ -49,4 +53,8 @@ def banner_imgs
       @banner_imgs = Setting.all.banner_img.map { |banner_img| [banner_img.value, banner_img.key.split('.')[1]] }
```
MariaCheca: Check this out too 👆

spec/features/admin/banners_spec.rb
```diff
@@ -102,6 +102,16 @@
     expect(page).to have_link 'Such banner many text wow link', href: 'https://www.url.com'
   end
 
+  scenario "Update banner color when changing from color picker or text_field", :js do
```
MariaCheca: Also I would add a feature test to check if the banner created in admin is correctly shown in the web.

MariaCheca:
Also @iagirre looks like the images are not being rendered properly:

localhost_3002_admin_banners_4_edit_locale en laptop with hidpi screen
screen shot 2018-01-24 at 13 24 09

Could you check it please?

iagirre:
Update
=
Appart from the suggestions on the PR, I updated the dev_seed file to remove the `image` and `style` of the banner and set a random hex color for background and font, instead. The same with spec factories.
There is also a migration to remove both columns from DB.

## UI updates

### Admin
banner01

### User
banner02

bertocq:
@iagirre whenever we're going to remove columns from the database, we need to be aware that consul is being used on many forks that may loose data on those columns "by surprise". If we're not using anymore those columns the correct procedure is to:

1- Totally avoid using them anymore on this PR (I suppose that's the case so 👍 )
2- Add in the "Warning" section of this PR that the Changelog for next release (v0.13) should include a *Deprecation* warning about the usage of those two columns. (Best would be to directly include yourself that reference but @MariaCheca will do it if you just add the notice) The "There are also migrations to run." is really not as informative as it should, the important thing is that we're potentially *destroying data*
3. This PR should include a rake task to migrate (if needed) the values, and state that on the Changelog at the deprecation section as well.

4. Create a *separate PR*, and clearly state that it should not be merged until 2 releases in the future (that's relesae v0.14) and on it we'll just include that migration that removes the columns. At the changelog it will be included as a simple change with a reference the first PR for extra contents and "reminder" that if the rake task hasn't been ran yet, before updating up to v0.14 its a good idea to do it

# Why all this procedures
Think that few forks update their codebase with consul#master (and in fact NONE should do it as master is often broken and we can guarantee otherwise). So forks that update release by release... get a "big chunk" of changes at once. Many will not read release notes (that's their problem) but for sure almost none will read the code diff :) and then won't notice the migration that removes columns (along with data).

So for those great maintainers that both read changelog and code diff... they'll face a hard situation as deploying that "chunk" of changes will run the migration and destroy the data before they're even able to access the rails console to run the rake task 🤔 .

Yes... some people fix this by "integrating" the rake task code inside the migration but in my experience that's not something nice as those migrations are often hard to revert and you're mixing to intentions (database changes and data changes).... but I've done it in the past in different projects (this is the most special one we'll ever work on because of the forks)

The point is ... will it hurt anyone if those no-longer-used columns live along with the data inside for the time that will take those maintainers to upgrade to the next version release? Answer is no in almost all cases, so the really easy and safe path is to do things in two steps, with a deprecation warning, and giving enough time to run (or even modify if there are customizations in the fork's side) the rake task.

aitbw:
Hi! :wave: I'd like to shed some light on your latest comment, @iagirre: the problem is not `selenium-webdriver` _per se_ but rather PhantomJS which made us work with hacks thinking it was the "correct" way to write tests.

The problem lies in the fact that, most of the times, PhantomJS (because of poor CSS support) interacts **directly** with the DOM rather than with the UI (which is with what the user **actually** works) —this approach has a clear advantage that everything we do front-end wise will work but it makes our tests fragile as it doesn't actually simulate user interaction, hence the need to write stuff like `page.find("body").click`

I faced a similar problem during the migration from PhantomJS to Headless Chrome (see 4th point under the **To do** section in #2534) and fixed it in [`765d273`](https://github.com/consul/consul/pull/2534/commits/765d273bdf498165d81824e62168e143095e021a)

GitLab faced similar issues when they migrated from PhantomJS to Headless Chrome as well, as stated [here](https://about.gitlab.com/2017/12/19/moving-to-headless-chrome/)

Issue #2317, opened 2018-01-15
Add "Current Budget" concept


bertocq:
We need a way to know which Budget is currently in active phase. For now we don't contemplate as valid the scenario where more than one budget can be in an active phase.

Active phases are all but "Drafting" phase. That means the "Current Budget" should be the last budget created that's not still in "Drafting".

This should be a helper method that will replace any `Budget.last` appearance in the codebase.

bertocq:
Closed via #2322 🙌

PR #2322, opened 2018-01-15
Change concept of current budget to account for multiple budgets

voodoorai2000:
Where
=====
* **Related Issue:** https://github.com/consul/consul/issues/2317

What
====
- Add `currrent_budget` helper for controllers and views
- The `current_budget` helper takes into account that the current budget is the last budget created
- It also takes into account, that it's possible to have a budget in drafting phase, in which case, the `current_budget` is really the previous budget if it exists or nil otherwise
- Refactor Management and Valuator controllers and views to display only the `current_budget` 
- Add scope `Budget.open` to represent, budgets that are not finished. This substitutes the old `Budget.current` implementation and it's used only in the admin budget's index view, which has two tabs _open budgets_ and _finished_budgets_ (see screenshot)

How
===
- Adding a private method `current_budget` to the ApplicationController and making it accessible to views
- Refactoring existing concept of Budget.current

Screenshots
===========
- Admin budget's index
admin budget s index

Test
====
- Added unit and controller specs for `current_budget` and `Budget.open`

Deployment
==========
- As usual

Warnings
========
- None

### DIFF
--- a/app/models/budget.rb
+++ b/app/models/budget.rb
@@ -31,7 +31,9 @@ class Budget < ActiveRecord::Base
   scope :reviewing_ballots, -> { where(phase: "reviewing_ballots") }
   scope :finished, -> { where(phase: "finished") }
 
-  scope :current, -> { where.not(phase: "finished") }
+  def self.current
+    where.not(phase: "drafting").last
+  end
 
   def description
     send("description_#{phase}").try(:html_safe)
@@ -91,10 +93,6 @@ class Budget < ActiveRecord::Base
 
   def balloting_or_later?
     balloting_process? || finished?
-  end
-
-  def current?
-    !finished?
   end
 
   def heading_price(heading)
--- a/app/models/budget.rb
+++ b/app/models/budget.rb
@@ -30,6 +30,7 @@ class Budget < ActiveRecord::Base
   scope :balloting, -> { where(phase: "balloting") }
   scope :reviewing_ballots, -> { where(phase: "reviewing_ballots") }
   scope :finished, -> { where(phase: "finished") }
+  scope :open, -> { where.not(phase: "finished") }
 
   def self.current
     where.not(phase: "drafting").last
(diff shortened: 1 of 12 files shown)


app/models/budget.rb
```diff
@@ -30,8 +30,11 @@ class Budget < ActiveRecord::Base
   scope :balloting, -> { where(phase: "balloting") }
   scope :reviewing_ballots, -> { where(phase: "reviewing_ballots") }
   scope :finished, -> { where(phase: "finished") }
+  scope :open, -> { where.not(phase: "finished") }
```
bertocq: I wonder if we should exclude as well 'drafting' budgets, as those should be totally hidden to the public. I know 'drafting' phase wasn't included in this PR (but in one that I created) but just realized/discovered about this filter and that we may not be doing it right.

app/views/management/budgets/print_investments.html.erb
```diff
@@ -1,12 +1,10 @@
 <table>
-<% @budgets.each do |budget| %>
-  <tr id="<%= dom_id(budget) %>">
-    <td><%= budget.name %></td>
-    <td><%= budget.translated_phase %></td>
+  <tr id="<%= dom_id(@budget) %>">
```
bertocq: I think we're not covering the scenario where @budget may not be present (for new clean installations without a proper database seed, or that have cleaned up "dummy data" from seeds)

app/models/budget.rb
voodoorai2000: Interesting, it makes sense at a code level, however at the interface level, it's not yet aplicable. As this scope is only used in the admin interface. Where there are only 2 tabs: open and close, which I think fits better with the current definition of open. 

Maybe we can review this when we have a need to extend the budget's admin interface or the user facing interface due to having more budgets in different phases?

app/views/management/budgets/print_investments.html.erb
voodoorai2000: Yes :) +1 to an issue for this 👌 
These are the tests that where failing due to not having a budget, it would be a great starting point to implementa that issue 👍 
https://github.com/consul/consul/pull/2322/commits/34e0c23bb37bbd0c7b9b17d26a71d6adec9802e2

voodoorai2000:
💃

PR #2342, opened 2018-01-18
Budget: add column show to valuators

### LINKED ISSUE #2328: Budget: add column show to valuators
Background
==

Now on `/admin/budgets/N/budget_investments/X/edit#classification` admin users can assign valuators to a budget investment. Once assigned, the valuators can see their assigned budget investments to valuate.

What we need
==
- When the tab `Under valuation` is active, the last column should show **"Show to valuator"** (instead of "Selected") with a checkbox (_view screenshot_).

- The valuators on `/valuation/budgets/N/budget_investments` only should see their assigned budget investments with this checkbox active.

Notes
==

This feature is created on Madrid's fork on [this commit](https://github.com/AyuntamientoMadrid/consul/commit/0fe631dcdb56b57da44cce757f7600c9ee06aa44), use for reference:

- [`admin/budget_investments/_investments.html.erb (I)`](https://github.com/AyuntamientoMadrid/consul/blob/master/app/views/admin/budget_investments/_investments.html.erb#L18-L24)

- [`admin/budget_investments/_investments.html.erb (II)`](https://github.com/AyuntamientoMadrid/consul/blob/master/app/views/admin/budget_investments/_investments.html.erb#L75-L81)

- [`controllers/admin/budget_investments_controller.rb`](https://github.com/AyuntamientoMadrid/consul/blob/master/app/controllers/admin/budget_investments_controller.rb#L63)

- [`controllers/valuation/budget_investments_controller.rb`](https://github.com/AyuntamientoMadrid/consul/blob/master/app/controllers/valuation/budget_investments_controller.rb#L17)

- [`models/budget/investment.rb`](https://github.com/AyuntamientoMadrid/consul/blob/master/app/models/budget/investment.rb#L55)

- [`spec/factories.rb`](https://github.com/AyuntamientoMadrid/consul/blob/master/spec/factories.rb#L355)

- [`db/schema.rb`](https://github.com/AyuntamientoMadrid/consul/blob/master/db/schema.rb#L193)

🙏  Don't forget specs to this features (references on):

[`features/admin/budget_investments_spec.rb`](https://github.com/AyuntamientoMadrid/consul/blob/master/spec/features/admin/budget_investments_spec.rb) and [`features/valuation/budget_investments_spec.rb`](https://github.com/AyuntamientoMadrid/consul/blob/master/spec/features/valuation/budget_investments_spec.rb)

Also don't forget to include this on [`dev_seeds.rb`](https://github.com/AyuntamientoMadrid/consul/blob/master/db/dev_seeds.rb#L562-L566)

Screenshot
==
show_to_valuators

iagirre:
Where
=====
* **Related Issue:** #2328 (This PR closes #2328)

What
====
- Add a column called "Show to valuator" to the table that appears when an admin clicks "Under valuation" in the `budget_investment` management page.
- Valuators should only see the investments that are marked.

How
===
As said in the issue, this feature was completed in [this commit](https://github.com/AyuntamientoMadrid/consul/commit/0fe631dcdb56b57da44cce757f7600c9ee06aa44) from Madrid's fork, so, I took the changes from there and adapt them to the code of consul.
Basically, the changes are:
DB:
- Add a migration to generate the 'visible_to_valuators' column.

Admin:
- Change the last column depending of the selected tab. If it's "Under valuation", show the checkbox.
- Update the controller to save the new value 'visible_to_valuators'.

Valuators:
- Just take the investments that are marked as visible.

Screenshots
===========
### Only one investment marked
valuator01

valuator02

### Both investments marked
valuator03

valuator04

Test
====
The test were taken from Madrid's fork, but they were slightly modified to work with the new changes in the valuators index.
Moreover, a function was added to make Capybara wait for AJAX, because the tests failed randomly due to this (sometimes the test continue before the AJAX call was finished and the 'visible_to_valuators' wasn't saved when the valuator access the page). More info [here](https://robots.thoughtbot.com/automatically-wait-for-ajax-with-capybara)

### UPDATE
To avoid the flakyness of the spec, `wait_for_ajax` was used in the first instance. This, however, is not ideal, so a new aproach has been used. It is explained deailedly in this PR https://github.com/AyuntamientoMadrid/consul/pull/1256, but, to sum up, the original spec was too fast sometimes and that's the reason to fail. This new spec divides the test in two parts, one only for testing what valuators see and the other one to check if admins can mark as visible the investments. The spec to check that checkbox is slow enough to let the AJAX call finish before checking the checkbox status.

Deployment
==========
There is a migration to run

Warnings
========
I noticed when a `create(:budget_investment, budget: @budget)` was done, a new budget object was created, and I ended up having 3 budgets, `@budget` object and other 2, where the last one was different from `@budget`. Thats why I changed the way of creating the investmentes, passing a heading instead of a budget. See [budget_investments_spec.rb:681](https://github.com/consul/consul/compare/master...wairbut-m2c:iagirre-budgets-column-show-valuators#diff-366210ce87db4d5a9d74bc535dfba01fR681)

### DIFF
--- a/spec/support/common_actions.rb
+++ b/spec/support/common_actions.rb
@@ -362,12 +362,6 @@ module CommonActions
     fill_in "newsletter_from", with: (options[:from] || "no-reply@consul.dev")
     fill_in "newsletter_body", with: (options[:body] || "This is a different body")
   end
-  def wait_for_ajax
-    Timeout.timeout(Capybara.default_max_wait_time) do
-      loop until page.evaluate_script('jQuery.active').zero?
-    end
-  end
-
 
   def click_notifications_icon
     find("#notifications a").click
--- a/spec/support/common_actions.rb
+++ b/spec/support/common_actions.rb
@@ -362,6 +362,12 @@ module CommonActions
     fill_in "newsletter_from", with: (options[:from] || "no-reply@consul.dev")
     fill_in "newsletter_body", with: (options[:body] || "This is a different body")
   end
+  def wait_for_ajax
+    Timeout.timeout(Capybara.default_max_wait_time) do
+      loop until page.evaluate_script('jQuery.active').zero?
+    end
+  end
+
 
   def click_notifications_icon
     find("#notifications a").click
--- a/app/controllers/admin/budget_investments_controller.rb
+++ b/app/controllers/admin/budget_investments_controller.rb
@@ -96,7 +96,7 @@ class Admin::BudgetInvestmentsController < Admin::BaseController
     def budget_investment_params
       params.require(:budget_investment)
             .permit(:title, :description, :external_url, :heading_id, :administrator_id, :tag_list,
-                    :valuation_tag_list, :incompatible, :selected, valuator_ids: [],
+                    :valuation_tag_list, :incompatible, :visible_to_valuators, :selected, valuator_ids: [],
                     valuator_group_ids: [])
     end
 
--- a/app/controllers/valuation/budget_investments_controller.rb
+++ b/app/controllers/valuation/budget_investments_controller.rb
@@ -17,7 +17,7 @@ class Valuation::BudgetInvestmentsController < Valuation::BaseController
   def index
     @heading_filters = heading_filters
     @investments = if current_user.valuator? && @budget.present?
-                     @budget.investments.scoped_filter(params_for_current_valuator, @current_filter)
+                     @budget.investments.visible_to_valuators.scoped_filter(params_for_current_valuator, @current_filter)
                             .order(cached_votes_up: :desc)
                             .page(params[:page])
                    else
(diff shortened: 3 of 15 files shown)


spec/support/common_actions.rb
```diff
@@ -357,4 +357,10 @@ def nested_path_for(resource)
     end
   end
 
+  def wait_for_ajax
```
bertocq: Just before we forget about it.. (not saying this should be fixed now) But as we talked while dealing with flakies... maybe there's an alternative to the usage of `wait_for_ajax` as you did at https://github.com/AyuntamientoMadrid/consul/pull/1206 🤔

iagirre: This flaky is related to this https://github.com/AyuntamientoMadrid/consul/issues/1194, and I'm looking for a solution in Madrid's fork.

voodoorai2000:
This looks great @iagirre 👍 

Could you fix the conflicts and use the alternative techniques we talked about to avoid the flakiness?

PR #2372, opened 2018-01-24
Budget: move calculate winners button

### LINKED ISSUE #2346: Budget: move calculate winners button
Background
==

Now for calculate budget investments winners there is a button "Calculate winners" on edit budget view `/admin/budgets/N/edit`

What we need
==

- Moves "Calculate winners" button to "Winners" tab on `/admin/budgets/N/budget_investments?filter=winners&page=1`

- The button only should work on "Balloting projects", "Reviewing Ballots" and  "Finished budget" phases (this is already working). When the budget phase are different show disabled button with message "The budget has to stay on phase "Balloting projects", "Reviewing Ballots" or  "Finished budget" in order to calculate winners projects"

- For the text button, if there are no winners projects show "Calculate winner" if there are some project listed show "Recalculate winners" instead.

- Changes tab Spanish translation "Ganadoras" to "Ganadores"

Screenshots
===

winners_1

winners_2

winners_3

iagirre:
Where
=====
* **Related Issue:** #2346 (this PR closes #2346)

What
====
- Move the "Calculate Winner Investments" button to the budget_investments index page.
- Change the name of the button if there are investments generated, from "Calculate [...]" to "Recalculate [...]"
- If the budgets is in a phase different from "Balloting projects", "Reviewing Ballots" or "Finished budget", the button is disabled and a message appears informing about the phases.
- Change the column name (in spanish) from "Ganadoras" to "Ganadores".

How
===
- Move the button to the budgets_investments index and put it in the rigth with CSS.
- Generate a helper function to show the button enabled (with the correct name), or the button disabled with the message.
- Add `clear` class name to the blue callout so that the button is not floating under it.

Screenshots
===========
### Without winner investments
winners01

### With existing winner investments
winners02

### In a phase that doesn't let users calculate winners
winners03

Test
====
- The spec which test the creation of the winner investments was modified to match the new situation of the button. It's no longer in the budget edition, but in the budget_investments index.
- A new scenario was added to test if the button is disabled when a budget is in an incorrect phase.

Deployment
==========
Nothing to apply.

Warnings
========
Nothing to apply.

### DIFF
--- a/app/helpers/budgets_helper.rb
+++ b/app/helpers/budgets_helper.rb
@@ -68,4 +68,16 @@ module BudgetsHelper
     return unless current_budget.present?
     MapLocation.where(investment_id: current_budget.investments).map { |l| l.json_data }
   end
+
+  def display_calculate_winners_button?(budget)
+    budget.balloting_or_later?
+  end
+
+  def calculate_winner_button_text(budget)
+    if budget.investments.winners.empty?
+      t("admin.budgets.winners.calculate")
+    else
+      t("admin.budgets.winners.recalculate")
+    end
+  end
 end
--- a/spec/features/admin/budget_investments_spec.rb
+++ b/spec/features/admin/budget_investments_spec.rb
@@ -366,6 +366,23 @@ feature 'Admin budget investments' do
       expect(page).not_to have_select("tag_name", options: ["All tags", "Accessibility"])
     end
 
+    scenario "Disable 'Calculate winner' button if incorrect phase" do
+      budget.update(phase: 'reviewing_ballots')
+
+      visit admin_budget_budget_investments_path(budget)
+      click_link 'Winners'
+
+      expect(page).to have_link "Calculate Winner Investments"
+
+      budget.update(phase: 'accepting')
+
+      visit admin_budget_budget_investments_path(budget)
+      click_link 'Winners'
+
+      expect(page).not_to have_link "Calculate Winner Investments"
+      expect(page).to have_content 'The budget has to stay on phase "Balloting projects", "Reviewing Ballots" or "Finished budget" in order to calculate winners projects'
+    end
+
     scenario "Limiting by max number of investments per heading", :js do
       group_1 = create(:budget_group, budget: budget)
       group_2 = create(:budget_group, budget: budget)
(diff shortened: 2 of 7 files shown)


app/helpers/budgets_helper.rb
```diff
@@ -76,4 +76,17 @@ def current_budget_map_locations
       }
     end.flatten.compact
   end
+
+  def calculate_winner_investments(budget)
```
voodoorai2000: Helpers are awesome, we try to stick to using html in views and logic without html in the helpers
How about something like this?

#helpers
```
def display_calculate_winners_button?(budget)
  budget.balloting_or_later?
end

def calcuate_winner_button_text(budget)
  if budget.investments.winners.empty? 
    t("admin.budgets.winners.calculate")
  else
    t("admin.budgets.winners.recalculate")
  end
end
```

#view
```
<% if display_calculate_winners_button?(budget) %>
  <%= link_to button_text,
              calcuate_winner_button_text(budget),
              method: :put,
              class: "button hollow float-right" %>
<% else %>
  <span class="button hollow disabled float-right">
    <%= t("admin.budgets.winners.calculate")%>
    <div class="callout warning clear">
      <%= t("admin.budget_investments.index.cannot_calculate_winners") %>
    </div>
  </span>
<% end %>
```

spec/features/admin/budget_investments_spec.rb
```diff
@@ -284,6 +284,23 @@
       expect(page).to have_select("tag_name", options: ["All tags", "Hospitals", "Teachers"])
     end
 
+    scenario "Disable 'Calculate winner' button if incorrect phase" do
+      @budget.update(phase: 'reviewing_ballots')
+
+      visit admin_budget_budget_investments_path(@budget)
+      click_link 'Winners'
+
+      expect(page).to have_link I18n.t("admin.budgets.winners.calculate")
```
voodoorai2000: In specs we try to use the literal english strings instead of translations, it makes the specs more readable 👌

voodoorai2000:
Oh also could you remove my merge commit and rebase with master?
Sorry for having interfered 😀🙏
Merge conflicts make it a little harder to cherry-pick commits for backports and quick fixes in forks

voodoorai2000:
Excellent @iagirre 👏

Issue #2380, opened 2018-01-24
Current Budget investments map not showing markers


MariaCheca:
In the current Budget page, sometimes the investment projects map doesn't show the points.

Observations:
- When the map is rendered some point latitudes or longitudes appear with `***********` instead of the value in the DDBB.
- if it's the longitude what shows the asterisks and the latitude is updated the problem is still there, so the longitude has to be changed (the same happens on the contrary).
- The should be a test to check that the location attributes are only integer values, not strings or anything else.
- Avoid the method who retrieves the locations to pass any location with empty value to prevent the JS to crash.

MariaCheca:
Backport issue: https://github.com/consul/consul/issues/2389

PR #2381, opened 2018-01-24
Fix map markers with asterisks

voodoorai2000:
Where
=====
* **Related Issue:** https://github.com/consul/consul/issues/2380

What
===
- Skips invalid map locations

Why
===
Sometimes even though the latitude and longitud of a map location is stored correctly in DB it is passed to the map as “********”. This prevents all points in the map from being displayed

We should figure out where are these asterisks are coming from, maybe it’s from one of the map libraries. In the meantime this should reduce the problem

Screenshots
===========
- Map locations with asterisks instead of numbers
map asterisks

Test
====
- Added map_locations_helper spec to clean coordinates

Deployment
==========
- As usual

Warnings
========
- Some valid map locations might not be displayed. Figure out what is truly happening here.

Update
===
- Tested in production environment. This does not solve the problem.
   It might have to do with the map libraries


voodoorai2000:
Solved via https://github.com/AyuntamientoMadrid/consul/pull/1162

Issue #2365, opened 2018-01-23
Return 404 instead of 500 error when budget url is invalid


MariaCheca:
If a non existing budget url is requested, return a 404 error instead of the current 500.

For example  accesing `localhost:3000/budgets/r/n`

raul-fuentes:
I'll do this one

raul-fuentes:
The problem is currently solved, I think

- budget not fount 
![screenshot from 2018-01-24 12 24 13](https://user-images.githubusercontent.com/33748390/35329923-d6dc6564-0101-11e8-87cf-e07a5ae0c338.png)

- budget investment not fount
![screenshot from 2018-01-24 12 23 54](https://user-images.githubusercontent.com/33748390/35329922-d6c32b1c-0101-11e8-9ec2-84c532cdf02a.png)

bertocq:
@raul-fuentes please try suggested route in issue description `localhost:3000/budgets/r/n`

bertocq:
![screen shot 2018-01-24 at 12 31 33](https://user-images.githubusercontent.com/983242/35330103-88b433de-0102-11e8-97ab-ed79a1bf657d.jpg)
![screen shot 2018-01-24 at 12 31 29](https://user-images.githubusercontent.com/983242/35330104-88d799be-0102-11e8-9759-b0fcaadfa6fa.jpg)

raul-fuentes:
Also work's

![screenshot from 2018-01-24 12 31 48](https://user-images.githubusercontent.com/33748390
/35330473-de689ecc-0103-11e8-8319-ca0f96d8680c.png)

MariaCheca:
Looks like it's something happening only in https://github.com/ayuntamientomadrid/consul due to some differences with the route paths. @raul-fuentes feel free to investigate there :)

raul-fuentes:
I'm going to check it there

voodoorai2000:
Indeed, this problem is related to customs routes used in Madrid's fork
Probably some bot accessing random urls, closing for now.

PR #2383, opened 2018-01-25
2365 return 404 instead of 500 error when budget url is invalid

raul-fuentes:
Where
=====
* **Related Issue:** https://github.com/consul/consul/issues/2365

What
====
- Changed the error shown on budgets/groups#show from 500 to 404 when the budget is'n found, also adds the possibility to use slugs instead of ids in the url

How
===
- Adding before_action :load_budget and before_action :load_group to budgets/groups_controller and
raising manually ActionController::RoutingError when the budget isn't found

Screenshots
===========
screenshot from 2018-01-25 12 09 32

Test
====
- Created a new  spec file for budgets/group that checks the correct operation of the url with slugs and the correct type of error when the budget is not found

Deployment
==========
- As usual

Warnings
========
- This PR is to correct the behavior of the url that gives an error 500 instead of a 404 in the fork of the city council of Madrid. Given that some of the url are overloaded in this repository it is possible that some link that should point to budget/investments end up in budget/groups and an raises an error 500 instead of the expected 404

### DIFF
--- a/app/controllers/budgets/groups_controller.rb
+++ b/app/controllers/budgets/groups_controller.rb
@@ -1,12 +1,18 @@
 module Budgets
   class GroupsController < ApplicationController
-    load_and_authorize_resource :budget
-    load_and_authorize_resource :group, class: "Budget::Group"
 
     before_action :set_default_budget_filter, only: :show
     has_filters %w{not_unfeasible feasible unfeasible unselected selected}, only: [:show]
 
     def show
+      begin
+        @budget = Budget.find_by(id: params[:budget_id])
+        @group = @budget.groups.find_by(id: params[:id])
+      rescue
+        raise ActionController::RoutingError, 'Not Found'
+      end
+      authorize! :show, @budget
+      authorize! :show, @group
     end
 
   end
--- a/spec/features/budgets/groups_spec.rb
+++ b/spec/features/budgets/groups_spec.rb
@@ -1,0 +1,29 @@
+require 'rails_helper'
+
+feature 'Groups' do
+
+  let(:user)   { create(:user, :level_two) }
+  let(:budget) { create(:budget) }
+  let(:group)  { create(:budget_group, budget: budget) }
+
+  context 'show' do
+
+    scenario "group not found raises 404 error" do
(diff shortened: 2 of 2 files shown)


app/controllers/budgets/groups_controller.rb
```diff
@@ -1,5 +1,7 @@
 module Budgets
   class GroupsController < ApplicationController
+    before_action :load_budget
```
voodoorai2000: These before_actions should not be necessary
How about moving the logic to raise the exception to the `show` action?

spec/features/budgets/groups_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+  let(:budget) { create(:budget) }
+  let(:group)  { create(:budget_group, budget: budget) }
+
+  context 'show' do
+
+    scenario "group not found raises 404 error" do
+      expect { visit budget_group_path('s', 'n') }.to raise_error(ActionController::RoutingError)
+    end
+
+    scenario "Budget correctly found using slug" do
```
voodoorai2000: This is good, but I would use `id` instead of `slug` in the scenario names, as that is what we are really testing 
To test the slug we would just pass as a parameter the budget object, instead of the id

raul-fuentes:
I think that now its correct

bertocq:
@raul-fuentes can you check the test suite locally? The travis builds have been restarted 3 times and always end up with `The job exceeded the maximum time limit for jobs, and has been terminated.`

raul-fuentes:
I don't know what happened to the branch, but it passes all the test locally and I'm going to try on travis in order to finish the PR

bertocq:
Have you closed the PR by accident? Or you mean you're going to create another branch & PR in order to see if its just a problem with travis & that particular branch?

raul-fuentes:
I'm trying to figure out what happened to travis. I think that I messed up something with git. 
So I have created a brand new branch an move the changes carefully by hand and now locally pass all the tests, so when travis gives me the ok I'm going to make a new Pr or reuse this one if its possible

raul-fuentes:
ok, something is wrong with chrome headless, I'm going to ask @aitbw

Issue #2377, opened 2018-01-24
Show investment links only on balloting or later


bertocq:
Show [List of all investment projects](https://github.com/consul/consul/blob/master/app/views/budgets/index.html.erb#L93) and [List of all investment projects not selected for balloting](https://github.com/consul/consul/blob/master/app/views/budgets/index.html.erb#L99) links if the current budget is on Balloting phase or later (reviewing_ballots or finished)

raul-fuentes:
I'll do this one

voodoorai2000:
Closed via https://github.com/consul/consul/pull/2386

PR #2386, opened 2018-01-26
Show investment links only on phase balloting or later 

raul-fuentes:
Where
=====
* **Related Issue:** https://github.com/consul/consul/issues/2377

What
====
- Restricted links to budget_investments_path(@budget.id) and budget_investments_path(budget_id: @budget.id, filter: 'unselected') only for balloting or later phases on budgets#index

How
===
- Adding a method to budget_helper checking current_buget phase and using the result as condition to draw the links

Screenshots
===========

screenshot from 2018-01-26 09 39 34

screenshot from 2018-01-26 09 42 01

Test
====
- Increased index budget spec checking the links phase by phase except the first two in which it does not apply

Deployment
==========
- Not needed

Warnings
========
- None

### DIFF
--- a/app/helpers/budgets_helper.rb
+++ b/app/helpers/budgets_helper.rb
@@ -1,5 +1,9 @@
 module BudgetsHelper
 
+  def show_links_to_budget_investments(budget)
+    ['reviewing_ballots', 'finished'].include? budget.phase
+  end
+
   def heading_name_and_price_html(heading, budget)
     content_tag :div do
       concat(heading.name + ' ')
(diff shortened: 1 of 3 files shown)


app/helpers/budgets_helper.rb
```diff
@@ -1,5 +1,9 @@
 module BudgetsHelper
 
+  def show_links_to_budget_investments(budget)
+    ['reviewing_ballots', 'finished'].include? budget.phase
```
bertocq: Maybe I wasn't explicity enough, but with ` current budget is on Balloting phase or later (reviewing_ballots or finished)` I actually meant balloting phase plus those two next phases as well.

raul-fuentes: I'll fix the misunderstanding in a moment.

Issue #2368, opened 2018-01-23
Show only current budget tags in admin budget investment view


MariaCheca:
Following the route `admin/budgets/[:budget_id]/budget_investments`, in the select "All tags" must be available only the tags belonging to the budget selected, not all of them.

Just in case, we should check if this list appears in other parts of the admin section and apply the same filter if it's necessary.

Screenshots
========

<img width="1049" alt="screen shot 2018-01-23 at 15 33 01" src="https://user-images.githubusercontent.com/2141690/35281283-d18b3a68-0052-11e8-82f4-e2e16097bc17.png">

vicentemendozam:
I'll take this one

voodoorai2000:
Closed via https://github.com/consul/consul/pull/2387

PR #2387, opened 2018-01-26
Show only current budget tags in admin budget

vicentemendozam:
Where
=====
* **Related Issue:** #2368 

What
====
- Show only the tags of the current budget in the admin budget page

How
===
- In the  tag's search modify the method to only find the tags of the current budget.

Screenshots
===========
-  No changes have been added to the interface.

Test
====
- Added a test with two different budgets and different tags. And verify that only the current budget tags are displayed.

Deployment
==========
-N/A

### DIFF
--- a/app/helpers/budgets_helper.rb
+++ b/app/helpers/budgets_helper.rb
@@ -52,8 +52,8 @@ module BudgetsHelper
     Budget::Ballot.where(user: current_user, budget: @budget).first
   end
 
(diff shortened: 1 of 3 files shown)


voodoorai2000:
Awesome!

@vicentemendozam when in a controller or helper, try to avoid using `where`, instead create a `scope` or `class method` in the model 👌

Issue #2291, opened 2018-01-09
Add setting to skip user verification


voodoorai2000:
Context
===
- The verification process consists of three steps:
  - Census verification
  - SMS verification
  - Postal Letter verification

What
===
- Allow a configuration option from admin to disable all user verifications
- When this option is activated, all users should be allowed to carry out verified user actions
- The initial welcome page when you sign up and my account should reflect that the user is verified 

Why
===
- In the first few days of setup, forks do not have access to an API to verify that a user is in the city census nor access to the the API for SMS. Having this option will simplify the initial setup

How
===
- Add the setting at /admin/settings
- The main user verification checks are in [verification.rb](https://github.com/consul/consul/blob/master/app/models/concerns/verification.rb). Add checks for this new setting

Screenshots
===
- Initial welcome page when you sign up
![welcome page after sign up](https://user-images.githubusercontent.com/4169/34736694-2583982a-f574-11e7-8fa5-a7107f93460b.gif)

- My account
![my account](https://user-images.githubusercontent.com/4169/34737135-8a2b8994-f575-11e7-981e-f7e8829f30c9.gif)

- Settings
![settings](https://user-images.githubusercontent.com/4169/34737961-f0908e76-f577-11e7-994c-81a7bcfacd8d.gif)

vicentemendozam:
I'll take this one

iagirre:
I'll do this one.

raul-fuentes:
I'll do this one

lingani:
Hi all, what is the stage of this issue? I would much appreciate if someone can guide me to contribute. I really need this feature. Thanks!

raul-fuentes:
I made a PR and now we should wait for a review from the mantainers of the repo

raul-fuentes:
You can fork the repo and add the feature if you are in a hurry, or even pull the branch an merge into master

voodoorai2000:
Closed via https://github.com/consul/consul/pull/2399

PR #2399, opened 2018-01-30
Added setting on admin to skip user verification

raul-fuentes:
Where
=====
* **Related Issue:** https://github.com/consul/consul/issues/2291

What
====
- It allows to ignore the user verification (postal mail, sms, telephone, email and census, all appear verified) using a new setting. If activated, any user will be considered verified and see the welcome page only in his first login. 

- Using seeds on a new installation sets this setting active (on seeds and dev_seeds)

How
===
- Added a new option to settings (skip_validation) which makes to returns true any methods from models/concerns/validation.rb wich involves any kind of validation

- Also added conditions to force the user to go to welcome_path only in his first login

Screenshots
===========
- As you can see after the activation of the new setting a unverified user can login, with all participation options active and see the welcome page only once

peek 2018-04-13 11-13

Test
====
- Added test to verify the correct activation of the setting
- Added test for any verification method skipped
- Added test to ensure that welcome view is showed one time to any user on his first login if the setting is active and shows correctly the four green checks about active users

Deployment
==========
- No

Warnings
========
- No

### DIFF
--- a/app/models/concerns/verification.rb
+++ b/app/models/concerns/verification.rb
@@ -10,31 +10,42 @@ module Verification
     scope :incomplete_verification, -> { where("(users.residence_verified_at IS NULL AND users.failed_census_calls_count > ?) OR (users.residence_verified_at IS NOT NULL AND (users.unconfirmed_phone IS NULL OR users.confirmed_phone IS NULL))", 0) }
   end
 
+  def skip_verification?
+    Setting["feature.user.skip_verification"].present?
+  end
+
   def verification_email_sent?
+    return true if skip_verification?
     email_verification_token.present?
   end
 
   def verification_sms_sent?
+    return true if skip_verification?
     unconfirmed_phone.present? && sms_confirmation_code.present?
   end
 
   def verification_letter_sent?
+    return true if skip_verification?
     letter_requested_at.present? && letter_verification_code.present?
   end
 
   def residence_verified?
+    return true if skip_verification?
     residence_verified_at.present?
   end
 
   def sms_verified?
+    return true if skip_verification?
     confirmed_phone.present?
   end
 
   def level_two_verified?
+    return true if skip_verification?
     level_two_verified_at.present? || (residence_verified? && sms_verified?)
   end
 
   def level_three_verified?
+    return true if skip_verification?
     verified_at.present?
   end
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -247,7 +247,8 @@ class User < ActiveRecord::Base
   end
 
   def show_welcome_screen?
-    sign_in_count == 1 && unverified? && !organization && !administrator?
+    verification = Setting["feature.user.skip_verification"].present? ? true : unverified?
+    sign_in_count == 1 && verification && !organization && !administrator?
   end
 
   def password_required?
--- a/config/locales/en/settings.yml
+++ b/config/locales/en/settings.yml
@@ -42,6 +42,7 @@ en:
       legislation: Legislation
       user:
         recommendations: Recommendeds
+        skip_verification: Skip user verification
       community: Community on proposals and investments
       map: Proposals and budget investments geolocation
       allow_images: Allow upload and show images
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -123,3 +123,5 @@ Setting['map_zoom'] = 10
 
 # Related content
 Setting['related_content_score_threshold'] = -0.3
+
+Setting["feature.user.skip_verification"] = 'true'
--- a/spec/features/admin/settings_spec.rb
+++ b/spec/features/admin/settings_spec.rb
@@ -89,4 +89,36 @@ feature 'Admin settings' do
 
   end
 
+  describe "Skip verification" do
+
+    scenario "deactivate skip verification", :js do
+      Setting["feature.user.skip_verification"] = 'true'
+      setting = Setting.where(key: "feature.user.skip_verification").first
+
+      visit admin_settings_path
+
+      accept_alert do
+        find("#edit_setting_#{setting.id} .button").click
+      end
+
+      expect(page).to have_content 'Value updated'
+    end
+
+    scenario "activate skip verification", :js do
+      Setting["feature.user.skip_verification"] = nil
+      setting = Setting.where(key: "feature.user.skip_verification").first
+
+      visit admin_settings_path
+
+      accept_alert do
+        find("#edit_setting_#{setting.id} .button").click
+      end
+
+      expect(page).to have_content 'Value updated'
+
+      Setting["feature.user.skip_verification"] = nil
+    end
+
+  end
+
 end
--- a/spec/features/welcome_spec.rb
+++ b/spec/features/welcome_spec.rb
@@ -68,4 +68,22 @@ feature "Welcome screen" do
     expect(page).to have_current_path(root_path)
   end
 
+  scenario 'a regular users sees it the first time he logs in, with all options active
+            if the setting skip_verification is activated' do
+
+    Setting["feature.user.skip_verification"] = 'true'
+
+    user = create(:user)
+
+    login_through_form_as(user)
+
+    4.times do |i|
+      expect(page).to have_css ".user-permissions > ul:nth-child(2) >
+                                li:nth-child(#{i + 1}) > span:nth-child(1)"
+    end
+
+    Setting["feature.user.skip_verification"] = nil
+  end
+
+
 end
--- a/spec/support/verifiable.rb
+++ b/spec/support/verifiable.rb
@@ -179,4 +179,79 @@ shared_examples_for "verifiable" do
     end
   end
 
+  describe "methods modified by Setting user.skip_verification" do
+
+    let(:user) {create(:user)}
+
+    before do
+      Setting["feature.user.skip_verification"] = 'true'
+    end
+
+    after do
+      Setting["feature.user.skip_verification"] = nil
+    end
+
+    describe "#residence_verified?" do
+      it "is true if skipped" do
+        expect(user.residence_verified?).to eq(true)
+      end
+    end
+
+    describe "#sms_verified?" do
+      it "is true if skipped" do
+        expect(user.sms_verified?).to eq(true)
+      end
+    end
+
+    describe "#level_two_verified?" do
+      it "is true if skipped" do
+        expect(user.level_two_verified?).to eq(true)
+
+        user.update(residence_verified_at: Time.current)
+        expect(user.level_two_verified?).to eq(true)
+
+        user.update(confirmed_phone: "123456789", residence_verified_at: false)
+        expect(user.level_two_verified?).to eq(true)
+      end
+    end
+
+    describe "#level_three_verified?" do
+      it "is true if skipped" do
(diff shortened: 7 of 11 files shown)


app/controllers/users/sessions_controller.rb
```diff
@@ -3,13 +3,17 @@ class Users::SessionsController < Devise::SessionsController
   private
 
     def after_sign_in_path_for(resource)
-      if !verifying_via_email? && resource.show_welcome_screen?
+      if !verifying_via_email? && resource.show_welcome_screen? && !override?
```
voodoorai2000: We don't want to skip the welcome page, 
it's nice to show it to all users the first time they login

What we want is to display the green check for all actions
Have a look at the gif "Initial welcome page when you sign up" [here](https://github.com/consul/consul/issues/2291)

app/controllers/users/sessions_controller.rb
```diff
@@ -3,13 +3,17 @@ class Users::SessionsController < Devise::SessionsController
...
         welcome_path
       else
         super
       end
     end
 
+    def override?
+      Setting["feature.user.override_verification"] == 'active'
```
voodoorai2000: How about something like this for the name of the setting?
`Setting["feature.user.skip_verification"] == true`

The true is also important, all other features are either true or nil

db/seeds.rb
```diff
@@ -122,3 +122,5 @@
 
 # Related content
 Setting['related_content_score_threshold'] = -0.3
+
+Setting["feature.user.override_verification"] = nil
```
voodoorai2000: As a default, better to set skip verification to `true` 
People need time to configure their verification system, 
and it's a common request to skip verification in the initial setup

config/locales/en/settings.yml
```diff
@@ -42,6 +42,7 @@ en:
       legislation: Legislation
       user:
         recommendations: Recommendeds
+        override_verification: Override user verification
```
voodoorai2000: This translation should be updated accordingly to the feature name

spec/support/verifiable.rb
```diff
@@ -160,4 +160,66 @@
...
+      user.update(letter_requested_at: Time.current, letter_verification_code: nil)
+      expect(user.verification_letter_sent?).to eq(true)
+
+      user.update(letter_requested_at: nil, letter_verification_code: nil)
+      expect(user.verification_letter_sent?).to eq(true)
+    end
+
+  end
+
 end
```
voodoorai2000: Great!

It be nice to also add an integration spec in [admin settings](https://github.com/consul/consul/blob/master/spec/features/admin/settings_spec.rb) for this issue

Something on the lines of an admin updating this setting 
and making sure a standard user can then perform an action that requires verification

app/models/concerns/verification.rb
```diff
@@ -10,32 +10,37 @@ module Verification
     scope :incomplete_verification, -> { where("(users.residence_verified_at IS NULL AND users.failed_census_calls_count > ?) OR (users.residence_verified_at IS NOT NULL AND (users.unconfirmed_phone IS NULL OR users.confirmed_phone IS NULL))", 0) }
   end
 
+  def override(value)
```
voodoorai2000: The way we usually skip the normal behaviour is with a [Guard Clause](http://www.thechrisoshow.com/2009/02/16/using-guard-clauses-in-your-ruby-code/)
How about something like this?:
```
def skip_verification? 
  Setting["feature.user.skip_verification"].present?
end
```
```
def verification_email_sent?
  return true if skip_verification?
  email_verification_token.present?
end
```

clairezed:
This is a feature we need in a project I'm involved in. 
Is there anything I could do/improve on this PR to help it beeing merged ?

raul-fuentes:
Hi @clairezed, currently I'm finishing a few changes requested by @voodoorai2000 and when he approves my changes, this PR will be done :)

raul-fuentes:
All fixed!

bertocq:
Thanks for expressing your interest and offering help @clairezed ! This PR will be reviewed again soon and you can count it will be merged before next v0.15 release is done

app/controllers/users/sessions_controller.rb
```diff
@@ -3,13 +3,17 @@ class Users::SessionsController < Devise::SessionsController
   private
 
     def after_sign_in_path_for(resource)
-      if !verifying_via_email? && resource.show_welcome_screen?
+      if !verifying_via_email? && (resource.show_welcome_screen? || (skip_verification? && resource.first_login?))
```
bertocq: Why are we adding logic here that should be included inside `show_welcome_screen?` ?

app/controllers/users/sessions_controller.rb
```diff
@@ -3,13 +3,17 @@ class Users::SessionsController < Devise::SessionsController
...
         welcome_path
       else
         super
       end
     end
 
+    def skip_verification?
+      Setting["feature.user.skip_verification"]
```
bertocq: THere's a trailing `.present?` missing here, although this method won't be no longer needed here (and its a dup of the one at the verification concern

app/models/user.rb
```diff
@@ -328,6 +328,10 @@ def interests
     follows.map{|follow| follow.followable.tags.map(&:name)}.flatten.compact.uniq
   end
 
+  def first_login?
```
bertocq: There's already similar logic at  `show_welcome_screen?`, couldn't we just prevent duplicating code?

spec/support/verifiable.rb
```diff
@@ -179,4 +179,66 @@
     end
   end
 
+  describe "#overrided_methods" do
+
+    let(:user) {create(:user)}
+
+    before do
+      Setting["feature.user.skip_verification"] = 'active'
```
bertocq: We're using `'true'` values all along, and here `'active'` why?

spec/support/verifiable.rb
```diff
@@ -179,4 +179,66 @@
     end
   end
 
+  describe "#overrided_methods" do
```
bertocq: This is not consistent with rspec standards or the codebase.. Also "override" is not the verb agreed upon, but `skip`.

spec/support/verifiable.rb
```diff
@@ -179,4 +179,66 @@
...
+
+      user.update(confirmed_phone: "123456789", residence_verified_at: false)
+      expect(user.level_two_verified?).to eq(true)
+    end
+
+    it "level_three_verified? is true  if overrided" do
+      expect(user.level_three_verified?).to eq(true)
+    end
+
+    it "unverified? is true if overrided" do
```
bertocq: This subject is not consistent with the expectation! It says it should be `true` but the expectation wants a `false`... also `unverified?` method is NOT affected but the skip verification :/ Please read carefully code changes before commiting, and after opening the PR. Reviewers time is really important having to read tons of PR's, you only have to read one at a time before moving into the next issue.

spec/features/welcome_spec.rb
```diff
@@ -68,4 +68,25 @@
     expect(page).to have_current_path(root_path)
   end
 
+  scenario 'is not shown to a regular user if the verification its overrided and the user has already logged in 1 time' do
+    user = create(:user, sign_in_count: 1)
+
+    Setting["feature.user.skip_verification"] = 'true'
```
bertocq: Remember every time a spec alters a setting, it should revert the change back afterwards, flakys on random spec order are born because of this

spec/features/welcome_spec.rb
```diff
@@ -68,4 +68,25 @@
...
+
+    login_through_form_as(user)
+
+    expect(page).to have_current_path(root_path)
+  end
+
+  scenario 'is showed to a regular user if the verification its overrided and the user logs in for the first time' do
```
bertocq: Do we really need two scenarios for this? Just merge both in a

spec/features/admin/settings_spec.rb
```diff
@@ -89,4 +89,48 @@
 
   end
 
-end
+  describe "Skip verification" do
+
+    scenario "deactivate verification", :js do
+      Setting["feature.user.skip_verification"] = 'true'
```
bertocq: Again, setting and resetting a `Setting` value in a before & after blocks are the right way as you can see in allmost all  the codebase specs using this kind of behaviour. Please read similar specs to understand how to tackle a particular spec/code, you'll run into this problems again and again if not.

spec/features/admin/settings_spec.rb
```diff
@@ -89,4 +89,48 @@
...
+
+      visit admin_settings_path
+
+      accept_alert do
+        find("#edit_setting_#{setting.id} .button").click
+      end
+
+      logout
+
+      user = create(:user, sign_in_count: 1)
```
bertocq: Shouldn't both scenarios be the same? We're trying to see if the having the setting enabled/disabled makes a difference... but you're changing the conditions on both so ...what are we really checking here?

raul-fuentes:
This is one of my first contributions to consul, give me a moment that I rewrite a few things with what I have been learning

spec/support/verifiable.rb
```diff
@@ -179,4 +179,85 @@
...
+      end
+    end
+
+    describe "#level_three_verified?" do
+      it "is true if skipped" do
+        expect(user.level_three_verified?).to eq(true)
+      end
+    end
+
+    describe "#unverified?" do
```
bertocq: Please take your time to read my comments again, `unverified?` has nothing to do neither its affected with the verification skip feature. Why are we testing that by default it returns false here?

spec/features/admin/settings_spec.rb
```diff
@@ -89,4 +89,49 @@
...
+        find("#edit_setting_#{setting.id} .button").click
+      end
+
+      logout
+
+      user = create(:user)
+
+      expect(user.sms_verified?).to eq(false)
+
+      Setting["feature.user.skip_verification"] = nil
```
bertocq: Both `Setting["feature.user.skip_verification"] = 'true'` can be moved onto a `before` block, and both `Setting["feature.user.skip_verification"] = nil` into an after block, as I said in the comments about this spec

bertocq:
We've lost the changes at `app/controllers/users/sessions_controller.rb` and user model files, aren't those no longer neccesary?

spec/support/verifiable.rb
raul-fuentes: I'll remove that test, but that function it's affected by the skip setting.
Its defined 
```
def unverified?
  !level_two_or_three_verified?
end
```
and level_two_or_three_verified? like 
```
def level_two_or_three_verified?
  level_two_verified? || level_three_verified?
end
```
also affected by the skip setting.

I thought that I should test any function affected by my changes

clairezed:
Hey there, it's me again. As far as I understand, there are still problems with this PR ? 

v0.15 hasn't been released yet, so this PR could still be part of it ? 

Once again, is there anything I could do to help ? Current state of this PR and what remains to be done is not really clear to me. My own need for this feature has become quite pressing, so really don't hesitate share your needs !

voodoorai2000:
This is great @raul-fuentes
Thank you very much for the effort 💪 

@clairezed, @MarkEwert this PR is now merged into the main codebase
Feel free to update your code from the master branch 👍

PR #2411, opened 2018-01-31
Admins can search investments by title or ID

aitbw:
Where
=====
* **Related Issue:** #2401 (Closes #2401 once merged)
* **Related PR's:** #2378

What
====
* Allow admins to search investments by `title` or `ID`

How
===
* Added a new class method under `Budget::Investment` to allow search by `title` or `ID` —since the latter is less common, the query to the database for this only happens when the parameter matches a RegEx (numbers only)
* **Fixed a bug** where investments from other budgets appeared if they matched the query (introduced by yours truly on #2378)
* Fixed Rubocop warnings

Screenshots
===========
peek 2018-01-31 10-07

Test
====
* Increased coverage, all green! ✅

PR #2417, opened 2018-02-02
Add `integer?` method to Investment model for robust input type checking

aitbw:
Where
=====
* **Related PR:** #2411

What
====
* Adds a private `integer?` class method to `Budget::Investment` model for a more performant, robust and safe type checking when searching investments by ID

How
===
* Replaced the RegEx the input was compared against with a `integer?` method that receives the input as an argument and returns `true` if it is an actual `Integer`

Test
====
* All green! ✅

### DIFF
--- a/app/models/budget/investment.rb
+++ b/app/models/budget/investment.rb
@@ -140,9 +140,11 @@ class Budget
 
     def self.search_by_title_or_id(params)
       results = Investment.where(budget_id: params[:budget_id])
-
-      return results.where(id: params[:title_or_id]) if params[:title_or_id] =~ /\A[0-9]+\z/
-      results.where("title ILIKE ?", "%#{params[:title_or_id].strip}%")
+      if integer?(params[:title_or_id])
+        results.where(id: params[:title_or_id])
+      else
+        results.where("title ILIKE ?", "%#{params[:title_or_id].strip}%")
+      end
     end
 
     def searchable_values
@@ -356,6 +358,12 @@ class Budget
       def set_denormalized_ids
         self.group_id = heading.try(:group_id) if heading_id_changed?
         self.budget_id ||= heading.try(:group).try(:budget_id)
+      end
+
+      def self.integer?(input)
+        Integer(input)
+      rescue ArgumentError, TypeError
+        false
       end
 
   end


app/models/budget/investment.rb
```diff
@@ -141,7 +141,7 @@ def self.limit_results(budget, params, results)
     def self.search_by_title_or_id(params)
       results = Investment.where(budget_id: params[:budget_id])
 
-      return results.where(id: params[:title_or_id]) if params[:title_or_id] =~ /\A[0-9]+\z/
+      return results.where(id: params[:title_or_id]) if integer?(params[:title_or_id])
```
bertocq: I think this code would be much clearer if we used a clear if statement

bertocq: Something like:

app/models/budget/investment.rb
```diff
@@ -358,5 +358,11 @@ def set_denormalized_ids
         self.budget_id ||= heading.try(:group).try(:budget_id)
       end
 
+      def self.integer?(input)
```
bertocq: What do you think about:

aitbw: From what I've researched, that method tends to return some unexpected results, such as `'01' == '01'.to_i.to_s` returns `false`, which in both AR and native SQL can return a result. See [this answer](https://stackoverflow.com/a/18113520) on Stack Overflow.

Also, according to Avdi Grimm (of [RubyTapas](https://www.rubytapas.com/) fame), `Integer(input)` is the canonical way to test "integerness" in Ruby (see [here](https://stackoverflow.com/questions/1235863/test-if-a-string-is-basically-an-integer-in-quotes-using-ruby#comment1061943_1235876) and [here](https://stackoverflow.com/questions/1235863/test-if-a-string-is-basically-an-integer-in-quotes-using-ruby#comment1065258_1235876)), that's because of the permissiveness of the `to_i` method.

bertocq: Awesome explanation @aitbw 👏 !! Didn't knew about that at all 🙇  I wasn't taking into account leading zeros as an option because

bertocq: It may endup returning a different record than intended:

bertocq: Maybe we could start by the important (and missing) part here... the test :). And after we agree on all the valid and invalid inputs we can then find the best and most efficient way to solve the test suite.

aitbw: Yes @bertocq, you're right on the octal part but we can pass a second argument to `Integer`, which will be the base of the numbers we want to handle, so we'd have `Integer(input, 10)` which will parse our numbers to base 10 so we would have `Integer('0200', 10)` to return `200`

aitbw: screenshot_2018-02-02_22-17-57

bertocq: Yes but, should a "001" or "0200" be considered as a valid input? This is an internal search function aimed to managers/admins.

bertocq: As you may have seen at https://stackoverflow.com/questions/1235863/test-if-a-string-is-basically-an-integer-in-quotes-using-ruby#comment1061883_1235876 and https://stackoverflow.com/a/1235891/599722 (where I got my code example) The exception raising option has some smells around it not everyone likes. And btw I think you'll know as well Sarah Mei :) (I'm a rubytapas subscriptor too)

bertocq: So my argument goes down to: The scenarios they are talking about on that issue are wider than what we've got here:

aitbw: Yes, it should work as expected since `base-ten` (or decimal) are the numbers we work with on a daily basis.

I completely understand the `exceptions as control flow` caveat because, even though I think it's the best way to test this, still rings a bell (I even discussed it thoroughly with some colleagues) but because of Ruby's type system, this seems like the safe bet.

It's worth noting that the [dry-types](https://github.com/dry-rb/dry-types/blob/master/lib/dry/types/coercions/params.rb#L32) project also takes this approach.

We could also take advantage of Rust's type system and embrace [Helix](https://usehelix.com/) but going low level for such a small requirement is just senseless.

bertocq: Sorry but I'm trying to bring the debate back to a more practical level (and in particular to our use case scenario.. not the general "is this a numeric value"), the theory and flame wars aren't much my thing :) I really don't care who thinks what's the best for general purposes or whats being used where... because we can start finding usages of both and waste ours on that direction, same as with who thinks exceptions are nice and who hates them.

If you want to ellaborate about best things, the rabbit hole here is the user, not the performance or the "safety" of the code. Let me explain: Have you considered how this feature is going to be used? From a test point of view at least (I still can se none)

A user could be expecting "consistency" in that search bar as he is totally unaware of the inner mechanics that divide searches by string (to a partial mach) or number to a individual search... so if he searches for "Pro" he expects to see a list of:
`Proposal for 2020 olympics`
`Proyect for 20 days 20 things`
etc... because all have `pro` in title, right?

So what happens when he searches for `20` ? If there is no investment with id==20 we won't show him a thing? Issues will start being created :).

You want to be the most efficient in finding out if the user is searching by id or title, but the user doesn't know about that neither care about the difference... he wants to find something and we don't know what so we'll show him anything that relates.

If he types 020... he may just did a typo trying to find ID '20', or he's going for the `..2020 olympics` title...

The issue asks about searching for ID and the ID's are clearly integers without leading zeros. If we don't need to support the `020` value to search by number.. why should we go with the "best general" option that really doesnt feel like right (if its not neccesary)?

Exceptions remember me of Java and all the messy things I had to do in my time there, this is Ruby, and Ruby is small, is clean, is poetry not prose.

This is not a contest about performance & most secure options (the most secure code is the code that's tested, even if its shitty code... tests are the best code you'll write in an app, without it the cleanest code will be worth a 4am call to hotfix production)

This is about maintenance, about the lines and lines of code that we need to add to allow new features, about being consistent with the rest of the codebase (rubocop is about that, not about enforcing rules because rules are cool), and although we're still far away in this codebase to a nice % of consistency.. I've a really hard opinion against moving in the opposite direction.

aitbw: I think we can't omit the theory from this discussion as this is, fundamentally, a theory discussion —especially when stuff like `monads` from Haskell (or [dry-monads](http://dry-rb.org/gems/dry-monads/)) or a more robust type system (like Rust's) would be extremely useful in these kind of scenarios because this is an edge case.

From my experience, an user searching records by database-generated identifiers is not very common; if you want to search records by a certain ID, you create and make them user-friendly enough for the users to take advantage of them.

That said, the current approach (talking about `master` here) will execute an ID-based query if the input is an Integer so that wouldn't match all the possible results the user could expect; or maybe a scenario similar to what you mentioned could happen: What if an user inputs a number but because it's `020`, the database returns records that include `020` on the title instead of the record with the ID `20`?

You can't execute `SELECT * FROM budget_investments WHERE id = "input" OR title = 
"%input%";` because it will raise a `RuntimeError` as soon as the user hits `enter`/clicks the `search` button if the input is an actual String.

One possible solution would be to have two (2) separate forms, one for titles and one for IDs but that would add an unnecessary complexity to the UI/UX for the end-user and that's not what we want.

Hence, today I came up with this proposal:

```ruby
if integer?(input)
  results.where('id = ? OR title ILIKE ?', params, "%params.strip%")
else
  results.where('title ILIKE ?', "%params.strip%")
end
```

I think this would cover (for now) our current needs, showing the user all the results he/she could expect from a search query where the input could be directed to either the ID or the title, without (avoidable) runtime errors or unexpected behaviors with `#to_i.to_s`

And once again, I come up with the theory stuff because because of Ruby's type system and the permissiveness of the methods for casting integers, the proposed solution would be the best way to respond accordingly to the unknown input —whether it is a "code smell" is up for debate, but I think I prefer handling an exception I'm aware of than `RuntimeError`s or inputs coerced as octals.

bertocq: Lets get back to coding by:

bertocq: +1 to that search by title || id or search just by title, but using pg_search instead of ILIKE as its more efficient and its how consul usually does searches by text.

javierm:
I found a small bug in this method in the master branch, so I've opened a pull request which fixes it and I think addresses the concerns in this pull request.

Closing in favor of #3782.

Issue #2442, opened 2018-02-06
Use user locale to format currency


bertocq:
## Current behaviour
All currencies are formatted using the default locale language (I18n.default_locale) instead of the user selected language:

![screen shot 2018-02-06 at 11 11 07](https://user-images.githubusercontent.com/983242/35853900-1f12df3e-0b2f-11e8-8264-84e837e1714e.jpg)

## Expected behaviour

Use the user language to format currencies 
![screen shot 2018-02-06 at 11 20 15](https://user-images.githubusercontent.com/983242/35854108-bb59d870-0b2f-11e8-81b3-68a858d6b30c.jpg)

raul-fuentes:
I'll do this one

bertocq:
@raul-fuentes there's a PR open already that only has pending tests to be completed

raul-fuentes:
I'm blind

bertocq:
Closed by #2443

PR #2443, opened 2018-02-06
Use user locale instead of default locale to format currencies

bertocq:
Where
=====
* **Related Issue:** https://github.com/consul/consul/issues/2442

What
====
All currencies are formatted using the default locale language (I18n.default_locale) instead of the user selected language. We need to use the user language to format currencies

How
===
- Changing I18n.default_locale for I18n.locale
- Unifying currency format logic into Budget#formatted_price method

Screenshots
===========
## Before
screen shot 2018-02-06 at 11 11 07

## After
screen shot 2018-02-06 at 11 20 15

Notes
====
None

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -58,7 +58,7 @@ module ApplicationHelper
   end
 
   def format_price(number)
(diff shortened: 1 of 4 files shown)


bertocq:
Checked at Madrid's fork 👌 production server

Issue #2318, opened 2018-01-15
Admin send emails


rzido:
Background
==

It’s very important in participation processes to control communications sent.  Now Consul sends several automatic notifications by email to users along the processes. For example, for budgets/spending proposals: 
-	E-mail to author on creation of spending proposal
-	E-mail to authors of unfeasible and feasible spending proposals
-	E-mail to authors of selected and unselected spending proposals

Format and content of these emails is composed with erb files in consul/app/views/mailer/* and localized mailers.yml files. 
  
But many times it is needed to replace texts for each campaign or phase in several languages to add additional information in e-mail content. It can be error prone. Now only IT staff can do that. There is not a easy option for admins to preview how e-mail will look or to create it. A basic and advanced implementation for functional admins could be developed.  

**1.	Basic implementation. Preview e-mail templates.** At least functional administrator should can preview how will look e-mails, for each language available, that will be sended to users for different communication actions related to the process:  creation of spending proposal, notification to authors of unfeasible and feasible spending proposal, notification to author of selected and unselected investments.

Functional admin users could select language and notification action by e-mail and preview a sample e-mail  that will be send to users. Form could be composed by:   

- _Action:_   [select field with available actions] (i.e.  Investment proposal creation e-mail; selected proposal e-mail; ...)    
- _Language:_    [select field with available languages]   (i.e. English, Spanish,...) 
-  _E-mail preview_

**2. Advanced implementation. Edit and preview e-mail templates option.** Create an admin section where functional admins can edit  and preview e-mail templates, for each language available, that will be send on different communication actions of the process.  Form could be composed by:

- _Action_:    	 [select field for available actions] {budget investment created, budget investment unfeasible, budget investment feasible, budget investment selected, budget investment unselected,...}    
- _Language_:  [select field for available languages]  {English, Spanish,..} 
- _To_:   [text field]  {e-mail user(s) variable: author email, all authors emails}  
- _Email Subject_:   [text field]	 {variable or free text}  
- _Email Content_:   [HTML Editor] {free text with variables embedded} 
- _Help_:  a list of all available variables that could be used in email to, subject or content i.e.   {code}, {title}, {author},{link},{org_name}. ...... 
 And link to usage help or guide 

What we need
==

A new section on Admin panel to allow send emails, notifications and download a CSV file with a specific segment of users to use an external service.

- Send emails (this issue)
- Send notifications (see https://github.com/consul/consul/issues/2145)
[...]

Issue #2120, opened 2017-11-08
Admin download emails list


bertocq:
Background
===
Sometimes admin wants to download a CSV file with a segmented list of users.

For example, once a Budget has finished, it could be very handy to have a way to get the list of emails for authors from:

- Winner investment proposals
- Selected investment proposals (that did not win)
- Proposed investment proposals that weren't selected (and obviously didn't win)

What we need
==

⚠️  Maybe this issue will be easier when https://github.com/consul/consul/issues/2318 it's done.

Improve the "Newsletter" section on admin, and include a select form to allow segment the list the admin wants to download.

Remove blank lines from the csv (for users that have not marked the newsletter checkbox)

Screenshot
==

![6_admin_newsletters_download](https://user-images.githubusercontent.com/631897/35634985-5acfb6ae-06ad-11e8-9ad7-ef3fd9bc8ca4.jpg)

bertocq:
About the 3 emails list, check the work already done at Madrid's fork with some simple rake tasks to quickly just output emails at https://github.com/ayuntamientomadrid/consul/blob/master/lib/tasks/budgets.rake#L33-L59

bertocq:
@decabeza if you have a better idea about where to place this feature 🙏

bertocq:
Already done!

PR #2462, opened 2018-02-12
Admin newsletter emails

MariaCheca:
Where
=====
* **Related Issue:** https://github.com/consul/consul/issues/2318

What
====
Added feature to send email newsletters to users from admin.

Notice that once a newsletter is sent, this action can't be performed anymore, so the "Send" button will disappear.

Screenshots
===========

Test
====
As usual.

Warnings
=====
Removed "Download emails list" button from `/admin/newsletters`, as this functionality will be improved and extended in https://github.com/consul/consul/issues/2120


app/views/admin/newsletters/show.html.erb
```diff
@@ -0,0 +1,51 @@
+<%= back_link_to %>
+
+<h2><%= t("admin.newsletters.show.title") %></h2>
+
+<%- recipients_count = @newsletter.list_of_recipients.count %>
```
bertocq: I just learned about <%- (https://puppet.com/docs/puppet/4.9/lang_template_erb.html#space-trimming-1) but I wonder if it was intentional... and if so ... should we use it as an standard good practice?

app/models/newsletter.rb
```diff
@@ -0,0 +1,23 @@
+class Newsletter < ActiveRecord::Base
+  enum segment_recipient: { all_users: 1,
+                    proposal_authors: 2,
```
bertocq: I think all the hash keys should be aligned, like:

spec/lib/user_segments_spec.rb
```diff
@@ -0,0 +1,176 @@
...
+describe UserSegments do
+
+  describe "#all_users" do
+    it "returns all active users with newsletter enabled" do
+      active_user1 = create(:user, newsletter: true)
+      active_user2 = create(:user, newsletter: true)
+      active_user3 = create(:user, newsletter: false)
+      erased_user  = create(:user, erased_at: Time.current)
+
+      expect(UserSegments.all_users).to include active_user1
```
bertocq: I think its a good practice to use `described_class` instead of the class name like:

spec/lib/user_segments_spec.rb
```diff
@@ -0,0 +1,176 @@
...
+      expect(UserSegments.all_users).to include active_user2
+      expect(UserSegments.all_users).to_not include active_user3
```
bertocq: Another Rubocop-rspec rule we have enabled (as you can see there are no issues for it on master branch) is [RSpec/NotToNot](http://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/NotToNot) as for me " to not" is not something you would say in a normal conversation... "not to ..." is the normal form used. Also the fact that it differs from `.to` makes it easier for an eye to catch on that difference IMHO but we can discuss about it and change the rule enforcing about it in an issue if you want to

spec/features/admin/newsletter/emails_spec.rb
```diff
@@ -0,0 +1,131 @@
...
+    click_link "Send"
+
+    total_users = newsletter.list_of_recipients.count
+    page.accept_confirm("Are you sure you want to send this newsletter to #{total_users} users?")
+
+    expect(page).to have_content "Newsletter sent successfully"
+  end
+
+  scenario "Select list of users to send newsletter" do
+    Newsletter.segment_recipients.each do |user_group, n|
```
bertocq: If you're only going to use the keys of a Hash iterating with .each (as I see no use for `n` variable inside the loop) you can better use `each_key` like `.each_key do |user_group|` . It's also a rubocop performance cop [Performance/HashEachMethods](https://github.com/bbatsov/ruby-style-guide#hash-each) the one backing up this suggestion

app/views/admin/newsletters/index.html.erb
```diff
@@ -1,3 +1,43 @@
-<h2><%= t("admin.newsletters.index.title") %></h2>
+<h2 class="inline-block"><%= t("admin.newsletters.index.title") %></h2>
+<%= link_to t("admin.newsletters.index.new_newsletter"), new_admin_newsletter_path,
+                                                         class: "button float-right" %>
 
-<%= link_to t("admin.newsletters.index.button"), users_admin_newsletters_path, class: "button" %>
\ No newline at end of file
+<% if @newsletters.any? %>
```
bertocq: Every time I see an "If wadus exists" I expect an else that prints "There is no wadus" so the user knows what's going on in that scenario... do you think as well it would improve user experience?

config/locales/en/admin.yml
```diff
@@ -515,9 +516,43 @@ en:
       search:
         title: 'Moderators: User search'
     newsletters:
+      create_success: Newsletter created successfully
+      update_success: Newsletter updated successfully
+      send_success: Newsletter sent successfully
+      delete_success: Newsletter deleted successfully
+      recipients:
```
bertocq: I think we should refactor as well this `newsletters.recipients.*` into `user_segments.*` to make it accesible from other features that will use UserSegments.

lib/user_segments.rb
```diff
@@ -0,0 +1,42 @@
+class UserSegments
+  def self.all_users
+    User.newsletter.active
+  end
+
+  def self.proposal_authors
+    author_ids = Proposal.not_archived.not_retired.pluck(:author_id).uniq
+    all_users.where(id: author_ids)
```
bertocq: This line is the same as line 40 in this file, what makes me think we could delete it and rewrite previous line into just:

spec/lib/user_segments_spec.rb
```diff
@@ -0,0 +1,176 @@
...
+    end
+  end
+
+  describe "#feasible_and_undecided_investment_authors" do
+    it "returns authors of a feasible or an undecided budget investment" do
+      user1 = create(:user)
+      user2 = create(:user)
+      user3 = create(:user)
+
+      investment1 = create(:budget_investment, :feasible, author: user1)
```
bertocq: I see many `create`'s for users and investments in this file repeating on each scenario, I think just with a bunch of `let` at the top we could make scenarios much faster to be read. But at least descriptive naming would be a plus like instead of `investment1` a `feasible_investment` would make much sense

spec/support/common_actions.rb
```diff
@@ -357,4 +357,25 @@ def nested_path_for(resource)
     end
   end
 
+  def deliver_newsletter
```
bertocq: I don't see a point of making two helper functions (this one and fill_in_newsletter_form) that are only used once in the codebase... Instead I would avoid having 5 different calls to ` fill_in "newsletter_subject"` and all the other 4 lines of interaction with the form... and just have a "fill_newsletter_form" helper function that has default parameters and allows overwiting those that need to change (and also doesn't have expectations or other actions like `click_link "send"` that may not be expected when calling it on other places.

app/models/abilities/administrator.rb
```diff
@@ -82,6 +82,8 @@ def initialize(user)
 
       can [:create], Document
       can [:create, :destroy], DirectUpload
+
+      can [:deliver], Newsletter, hidden_at: nil
```
bertocq: Not sure we should specify all the other actions an admin can do over a newsletter 🤔  (like create, edit, update..) Maybe I'm not aware of how default cancancan works

bertocq:
Oh! One last thing 🙇  it would be awesome to increase dev_seeds with an example Newsletter

app/views/admin/newsletters/show.html.erb
MariaCheca: Totally! I try to use it as a _personal convention_, but we can discuss if is worth it to use it as our standard 🙂

MariaCheca:
Closed PR in https://github.com/AyuntamientoMadrid/consul/pull/1274. Those changes will be back to Consul with a backport.

PR #2463, opened 2018-02-14
Make budget investment links always use budget slug

bertocq:
Where
=====
* **Related Issue:** https://github.com/AyuntamientoMadrid/consul/issues/1163* 

What
====
This is a partial backport from Madrid's fork.

We've found out that many budget investment's paths/urls where badly composed (using budget id's instead of budget slugs). That doesn't play quite well for SEO. 

Also there is no need to be as explicit as `budget_investment_url(budget_id: investment.budget_id, id: investment.id)` if `budget_investment_url(investment.budget, investment)` would have the same result... The difference is that the firsts one harcodes the usage of budget ID for urls! The second one leaves that to whatever Budget model says with a `to_param` method.

How
===
- Fixed all explicity budget investment url/paths to just rely on budget & investment objects to know how to parameterize themselves. (This will be hugely needed whenever we backport the SpendingProposal => Budget::Investment migration as you can see at Madrid's issue explanation) https://github.com/consul/consul/commit/6d0d0cd7ad77d8f522ee9ea48e89bcbfe0176f07

- Overwritten default parameterization for Budget class to always use slug instead of id https://github.com/consul/consul/commit/e95c70036f939219c0432738e9496bc6b1b9dc97

- Moving all "load_budget" or "load_investment" methods from controllers to ApplicationController to have only one place to fix this things https://github.com/consul/consul/pull/2463/commits/823c264785a609acea981f58b4f0139ce250fa59 When we do the SpendingProposal => Budget::Investment migration we'll need to modify all `load_investment` calls... having just one would be 🙆‍♂️ for that refactor to be easy to be understood and tested

Screenshots
===========
No need, backend change affecting links only

Test
====
Had to change many tests as well, lets see if everything keeps being okey.

Deployment
==========
As usual

Warnings
========
None

### DIFF
--- a/app/controllers/admin/budget_groups_controller.rb
+++ b/app/controllers/admin/budget_groups_controller.rb
@@ -1,9 +1,9 @@
 class Admin::BudgetGroupsController < Admin::BaseController
   include FeatureFlags
   feature_flag :budgets
+  before_action :load_budget_by_budget_id
 
   def create
-    @budget = Budget.find params[:budget_id]
     @budget.groups.create(budget_group_params)
     @groups = @budget.groups.includes(:headings)
   end
--- a/app/controllers/admin/budget_headings_controller.rb
+++ b/app/controllers/admin/budget_headings_controller.rb
@@ -1,22 +1,20 @@
 class Admin::BudgetHeadingsController < Admin::BaseController
   include FeatureFlags
   feature_flag :budgets
+  before_action :load_budget_by_budget_id
 
   def create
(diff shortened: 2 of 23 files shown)


bertocq:
There is some default behaviours searching budgets by id with devise that needs to be solve/retought, leaving this in stalemate until I've a clear mind to solve the issue (Tried setting Budget.slug as primary key but that's a no-go path)

voodoorai2000:
This issue has become obsolete, reworking in the upcoming spending proposals migration

PR #2464, opened 2018-02-14
[WIP] Url & Paths refactor

bertocq:
Where
=====
* **Related PR's:** https://github.com/consul/consul/pull/2463

What
====
This is a WIP PR that also carries 2 commits from PR https://github.com/consul/consul/pull/2463

This is meant to open a discussion about the best way to fix some internal issues we've while linking to polymorphic relations.

We have many `case`, `if-elsif...` and `is_a?` scenarios that just hold the same logic again and again... "How to link to a resource that has a parent (and even grandparent) without knowing what class it has?"

This refactor is needed to correctly link stuff around, it was started after hotfixing https://github.com/AyuntamientoMadrid/consul/issues/1189 and I suspect it will help fix another production crash https://github.com/AyuntamientoMadrid/consul/issues/1245

How
===
Maybe "Linkable" is not the best name for a concern, but for me "Url's" and "Paths" are just ways to link a resource (absolute or relative links). But the main purpose is to have a way to specify for each "problematic" Model that is involved in polymorphic relationships... how to link/path to it with a clear convention.

As you can see [Linkable](https://github.com/consul/consul/blob/d6385c0c855762db4b5c982a6d8db6696bed04c9/app/models/concerns/linkable.rb) adds two methods `url` and `path`, that will be used everywhere to know if the polymorphic resource knows how to compose its own url/path. Also it adds a class method `nested_linkable_with` that lets those "complex" Models to specify who is the parent/grandparent and also the route (when it doesn't follow conventions).

Also we where mixing path & url concepts by mistake on relationable objects :/ 

Screenshots
===========
No need, backend change only

Test
====
Hopefully passing!

Deployment
==========
As usual

Warnings
========
None


voodoorai2000:
Closed in favor of https://github.com/AyuntamientoMadrid/consul/pull/1617/files#diff-9931b6866b512f5f14df42ec715bc6f8R1

PR #2465, opened 2018-02-15
Design minor fixes

decabeza:
What
====
This PR include the following design minor fixes:

- Fixes back link to on "how to use" help page: changes `<%= back_link_to %>` to `<%= back_link_to help_path %>`

- Fixes heading structure and i18n format on legislation processes: avoid jump from `<h3>` to `<h5>`, changing with a correct `<h4>` tag.

- Refactors custom page view and include specs: if the subtitle field was empty this generates an empty `<h2>` tag, now only shows this tag if have content. Also allows include links to custom page texts.

- Removes unnecessary `<p></p>` tags on investment partial: it was generating a HTML error: with double tags `<p><p><%= investment.description %></p></p>`

- Adds links to milestone description: allows include links on milestone descriptions.

- Removes old unused CSS.

- Fixes image height on items list with tags (see screenshots).

Screenshots
===========

**`Proposals list before`**
proposal_before

**`Proposals list after`**
proposal_after

Test
====

Specs updated with new changes and also added new ones.

### DIFF
--- a/app/views/legislation/processes/_process.html.erb
+++ b/app/views/legislation/processes/_process.html.erb
@@ -7,8 +7,8 @@
     </div>
 
     <div class="small-12 medium-4 column">
-      <%= link_to process, class: "button hollow big expanded", title:  t('.see_latest_comments_title') do %>
-        <span class="icon-comments"></span>&nbsp; <%= t('.see_latest_comments') %>
+      <%= link_to process, class: "button hollow big expanded", title:  t(".see_latest_comments_title") do %>
+        <span class="icon-comments"></span>&nbsp; <%= t(".see_latest_comments") %>
       <% end %>
     </div>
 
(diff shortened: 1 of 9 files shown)


app/views/legislation/processes/_process.html.erb
```diff
@@ -21,43 +21,43 @@
...
+          <h4><%= t("legislation.processes.shared.allegations_dates") %></h4>
+          <p><%= format_date(process.allegations_start_date) %> - <%= format_date(process.allegations_end_date) %></p>
         </div>
       <% end %>
 
       <% if process.result_publication.enabled? %>
         <div class="small-6 medium-<%= column_width %> column">
-            <h5><%= t('legislation.processes.shared.result_publication_date') %></h5>
-            <p><%= format_date(process.result_publication_date) %></p>
+          <h4><%= t("legislation.processes.shared.result_publication_date") %></h4>
```
bertocq: I think in next reunion we should talk about this... "Should we use double or single quotes on I18n.t/l calls". I don't have a strong opinion about it, although I feel the urge to use single quotes but I see that the convention in the codebase is using double quotes because its normal use is in html.erb files that use double quotes for everything.

decabeza: +1 to choose one style. I just have changed it to be consistent with the rest of code 😉

bertocq: Should we create an issue about it or you'll just mention it along your opinion on next meeting?

decabeza: I would create the issue to unify all i18n keys after deciding the format to use 😉

Issue #2498, opened 2018-03-01
Store headings where a user has supported investments


bertocq:
To accomplish https://github.com/consul/consul/issues/2496, https://github.com/consul/consul/issues/2320 and https://github.com/consul/consul/issues/2497 we need to:

Store the headings where the user has supported investments. Both to know in how much (the count) and in which ones in particular.

- Supports will continue to be stored as they are in the Votes table with only user & investment info.

- For each user there will be an internal Headings list for each Budget on which he has manually supported.

- The Headings list will get a new element each time a user supports an investment on a heading for the first time, if he hasn't yet reached the max number of headings in which he can support, defined by the system config on a Budget Group level.

- No Headings will be removed from the list as users cannot remove supports from investments.

- When an investment is reclassified from a Heading A on to another B, the users that have supported that investment will only be able to support investments in the new heading B if they haven't reached the max number of headings in which they can support for the Budget Group config.

For this last scenario there is a corner case that should be covered by a test:
1. For a Budget with a single Group with max_heading_supports = 1, there are two Headings: A & B
2. The user supports a single Investment on Heading A, so his internal heading list only contains "Heading A"
3. The investment gets reclassified on to "Heading B"
4. The user cannot support investments in "Heading B", he can only continue to support investments in "Heading A"

- We'll need a rake task to backfill all internal heading lists for each user, by looking at all votes (supports) casted on investments.

voodoorai2000:
This issue's objective is to calculate correctly if a user can vote in an investment that has been reclassified from one heading to a different one

Instead of changing the standard behaviour we should consider this scenario separately

PR #2503, opened 2018-03-01
Store budget heading voters in database

MariaCheca:
Where
=====
* **Related Issue:** https://github.com/consul/consul/issues/2498

What
====
In order to be able to store in the database in which budget headings a user has supported investments, I've created a new model called `Budget::Heading::Voter` for a table that associates users with budget headings.

When a user supports *for the first time* an investment in a budget heading, a new record is created. So in the following supports the users give in the same heading the relation won't be overwritten (obviously but just in case).

### DIFF
--- a/app/controllers/budgets/investments_controller.rb
+++ b/app/controllers/budgets/investments_controller.rb
@@ -66,6 +66,12 @@ module Budgets
 
     def vote
       @investment.register_selection(current_user)
+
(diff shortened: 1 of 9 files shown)


voodoorai2000:
Rethinking implementation https://github.com/consul/consul/issues/2498#issuecomment-393678672
Closing for now

PR #2517, opened 2018-03-05
[WIP] Base Factories validation spec

bertocq:
References
=====
None

Objectives
====
We [had 39 "is valid" scenarios](https://github.com/consul/consul/commit/6e031a477d191b96ef19dae255a62b412b032984) on model specs that where checking if the model's base factory was valid "as it is"

But apart from that "redundancy" the problem was many model specs didn't had that scenario, as its a bit tedious to remember adding this scenario on every model spec. Also many base factories weren't valid and needed to be fix.

As a **bonus** the Identity model spec was completed (after removing the only scenario it had and realizing either the file should have to be removed or the spec completed https://github.com/consul/consul/commit/3ae7a4105f5fd8bdecc82401f04e3c27206286ba)

Screenshots
===========
No need

Notes
========
There is obviously an alternative... that is checking all model specs for that  basic `is valid` scenario and adding it wherever its missing.

But thinking about we should be ensuring all models have a model spec, and each and everyone of them has an associated factory present at `spec/factories.rb`... maybe "manual checks" are not the way to go. Automating that check might be a good  topic for next PR...

### DIFF
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -650,7 +650,7 @@ FactoryBot.define do
   factory :officing_residence, class: 'Officing::Residence' do
     user
     association :officer, factory: :poll_officer
-    document_number
+    document_number  "12345678Z"
     document_type    "1"
     year_of_birth    "1980"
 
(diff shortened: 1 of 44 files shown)


voodoorai2000:
Trying to maintain abstractions to a minimum, don't see a great benefit in drying this up
Going for clarity over dryness

PR #2543, opened 2018-03-21
Fix for flaky spec in Officing spec test file

raul-fuentes:
References
==========

This is a backport of madrid's https://github.com/AyuntamientoMadrid/consul/pull/1368 PR that tackles issue https://github.com/AyuntamientoMadrid/consul/issues/1200

Objectives
==========
* Fix the flaky, i seems to happen when the container of the selector for dni/passport isn't present. At first it looked like an error due the date used for the 2 poll_shift created (as it was [here](https://github.com/AyuntamientoMadrid/consul/pull/1343) ) but a quick check using different dates provoked a different error, so suspicions recalled about a race condition in the selector container. The solution that I proposes here is just  check the existence of the container using have_selector, which waits for js to be completed

Visual Changes (if any)
=======================
* none

Notes
=====================
* none

### DIFF
--- a/spec/features/officing_spec.rb
+++ b/spec/features/officing_spec.rb
@@ -152,6 +152,8 @@ feature 'Poll Officing' do
       expect(page).to have_content("Here you can validate user documents and store voting results")
 
       visit new_officing_residence_path
+      expect(page).to have_selector('#residence_document_type')
+
       select 'DNI', from: 'residence_document_type'
       fill_in 'residence_document_number', with: "12345678Z"
       fill_in 'residence_year_of_birth', with: '1980'
@@ -169,6 +171,8 @@ feature 'Poll Officing' do
       expect(page).to have_content("Here you can validate user documents and store voting results")
 
       visit new_officing_residence_path
+      expect(page).to have_selector('#residence_document_type')
+
       select 'DNI', from: 'residence_document_type'
       fill_in 'residence_document_number', with: "12345678Y"
       fill_in 'residence_year_of_birth', with: '1980'


spec/features/officing_spec.rb
```diff
@@ -169,8 +171,10 @@
       expect(page).to have_content("Here you can validate user documents and store voting results")
 
       visit new_officing_residence_path
-      select 'DNI', from: 'residence_document_type'
-      fill_in 'residence_document_number', with: "12345678Y"
+      expect(page).to have_selector('#residence_document_type')
+
+      select 'Passport', from: 'residence_document_type'
+      fill_in 'residence_document_number', with: "12345678A"
```
bertocq: Why is necessary to change the residence document number value?

bertocq:
Please @raul-fuentes remember PR titles are included in changelog's on each realease. State in the title what a fork maintainer or just any other dev should know about this PR. Think about something like `Fix flaky spec at Poll Officing dashboard spec`

spec/features/officing_spec.rb
raul-fuentes: That's the result of the cherry-pick. I just changed it back after seeing the results on travis

raul-fuentes:
Looks better now?, Shouldn't I remark that is a backport?

spec/features/officing_spec.rb
bertocq: Reviewing the diff is the second thing (after reading the PR description) that a PR reviewer will do. Authors should always review PR after creating to catch this simple mistakes, reviewers have 100x PR's pending

bertocq:
Yes @raul-fuentes but please don't do it in the title, as its really not important for anyone that reads it... only for the ones that open the PR description to understand:
- Where this changes come
- Are this changes tested in a production environment
- Where this changes already approved somewhere else?

So the `References` section is best place in my opinion to do it like `This is a backport of madrid's https://github.com/AyuntamientoMadrid/consul/pull/1368 PR that tackles issue https://github.com/AyuntamientoMadrid/consul/issues/1200` Instead of just a simple link without context that needs a click to have a quick understanding of what's going on

raul-fuentes:
OK, I have changed the references section following your ideas (I'll keep the description of this PR as scaffold for future backports)

Thanks!

spec/features/officing_spec.rb
```diff
@@ -169,7 +171,9 @@
       expect(page).to have_content("Here you can validate user documents and store voting results")
 
       visit new_officing_residence_path
-      select 'DNI', from: 'residence_document_type'
+      expect(page).to have_selector('#residence_document_type')
+
+      select 'Passport', from: 'residence_document_type'
```
bertocq: Again.. is the change of "DNI" for "Passport" a mistake or is there a reason behind? If so please explain why.

bertocq:
BTW About the references, its not only something we think, but something its already been done in many PR's like https://github.com/consul/consul/pull/2542

spec/features/officing_spec.rb
raul-fuentes: There are none, but come with the cherry pick, I change the commit to the minimum necessary to maintain the functionality in this repository. I had done it thinking in the opposite case, that the repos were similar

bertocq: Madrid is a fork with many particular custom changes, cherry-picking its nice to avoid reproducing changes by hand, but reviewing changes done its always a must for the reviewer, and should be for the author. If human attention wasn't needed we could automate this things entirely and go take some vacations, right?

raul-fuentes: True, I take note

PR #2600, opened 2018-04-17
return 404 instead of 500 when budget_group url is invalid (budget or group not found)

raul-fuentes:
Where
=====
* **Related Issue:** https://github.com/consul/consul/issues/2365

What
====
- Changed the error shown on budgets/groups#show from 500 to 404 when the budget is'n found, also adds the possibility to use slugs instead of ids in the url (mainly for the fork of the city council of Madrid)

How
===
- Removing before_action :load_and_authorize_resource (both budget and group) from budgets/groups_controller and loading manually or raising an ActionController::RoutingError when the budget isn't found

Screenshots
===========
screenshot from 2018-01-25 12 09 32

Test
====
- Created a new  spec file for budgets/group that checks the correct operation of the url with slugs and id's, and the correct type of error when the budget is not found

Deployment
==========
- As usual

Warnings
========
- None

### DIFF
--- a/app/controllers/budgets/groups_controller.rb
+++ b/app/controllers/budgets/groups_controller.rb
@@ -1,12 +1,18 @@
 module Budgets
   class GroupsController < ApplicationController
-    load_and_authorize_resource :budget
-    load_and_authorize_resource :group, class: "Budget::Group"
 
     before_action :set_default_budget_filter, only: :show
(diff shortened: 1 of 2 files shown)


voodoorai2000:
This problem is related to customs routes used in Madrid's fork
Not a problem in Consul right now

PR #2626, opened 2018-05-14
Upstream pull

bertocq:
Objectives
==========
We want to keep consul<=>madrid diff clear as soon as possible, massive diffs tend to give massive conflicts and headaches

Notes
=====================
This should be checked/tested manually to ensure we're not breaking a thing


bertocq:
Wrong flipping repo!

Issue #2389, opened 2018-01-26
Backport Budgets homepage map fixes


MariaCheca:
We need to backport the fixes for the budgets homepage map from https://github.com/AyuntamientoMadrid/consul/pull/1162 back to Consul 😁

vicentemendozam:
I'll take this one!

aitbw:
I'm gonna go with this one :)

voodoorai2000:
Closed via https://github.com/consul/consul/pull/2654

PR #2654, opened 2018-05-28
Budgets homepage map fixes

aitbw:
References
==========
* Related issues: #2380, #2389
* Related PR: AyuntamientoMadrid/consul#1162

Objectives
==========
* Backport the fixes implemented in AyuntamientoMadrid/consul#1162 and correction of its related tests

Visual Changes (if any)
=======================
None

Notes
=====================
It's worth noting that, during the tests, the markers **do appear** in the map but since we can't move the map to where they are located, they show up as "not visible", hence the need to use the `visible: false` syntax

### DIFF
--- a/app/assets/javascripts/map.js.coffee
+++ b/app/assets/javascripts/map.js.coffee
@@ -12,6 +12,7 @@ App.Map =
           App.Map.toogleMap()
 
   initializeMap: (element) ->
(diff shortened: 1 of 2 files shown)


voodoorai2000:
Appreciate it @aitbw 👍👌

PR #2686, opened 2018-06-18
Text customization

ejohnst2:
References
===================
> Add references to related Issues/Pull Requests/Travis Builds/Rollbar errors/etc...

Objectives
===================
> What are the objectives of this changes? (If there is no related Issue with an explanation)

Visual Changes
===================

Notes
===================


decabeza:
Hi @ejohnst2! Maybe wrong repo? 🤔 

I think all these are custom changes to a specific fork, aren't they? 😌

ejohnst2:
@decabeza yes indeed, that was my bad, I opened up the PR against the consul/consul instead of our specific fork. All good. I do have a quick question though, if you're able to answer, in the 'collaborative legislation' portion, we basically want to have each clause/subsection of the legislation be clickable (instead of highlighted ranges) for the interaction with it, and we also want to add in a series of 3 multiple choice questions before they are able to provide a comment/suggestion. After looking through, was wondering if that's possible to do with the built-in text editor that Consul has? Just want to make a judgment call on whether it would be better to do a consul fork and retrofit the collab legislation part OR to build a custom tool. Any thoughts here would help!

Issue #2573, opened 2018-04-04
Flaky spec:  Admin budgets Manage groups and headings Create group


bertocq:
# What
Tests that fail randomly are called "flakies", this one seems to be one:

**Randomized seed:** 34049

**Travis failed build:** https://travis-ci.org/consul/consul/jobs/362251287

**Failure:**

```
    1) Admin budgets Manage groups and headings Create group
     Failure/Error: expect(page).to have_content '1 Group of budget headings'
       expected to find text "1 Group of budget headings" in "Language: English Español Français Nederlands Português CONSUL ADMINISTRATION Admin Notifications My activity My account Sign out Categories Moderated content Polls Participatory budgets Profiles Collaborative Legislation Banners Site customization Manage geozones Signature Sheets Sending of emails Go back Yearly budget 0 Groups of budget headings No groups created yet. Each user will be able to vote in only one heading per group. Add new group Group name"
     # ./spec/features/admin/budgets_spec.rb:237:in `block (3 levels) in <top (required)>'
```

# How
- [ ] Explain why the test is flaky, or under which conditions/scenario it fails randomly
- [ ] Explain why your PR fixes it

### Random values issues
If the problem comes from randomly generated values, running multiple times a single spec could help you reproduce the failure by running at your command line:
```bash
for run in {1..10}
do
  bin/rspec ./spec/features/budgets/investments_spec.rb:256
done
```

You can also try running a single spec in Travis:
Add option `:focus` to the spec and push your branch to Github, for example:
```ruby
scenario 'Show', :focus do
```

But remember to remove that `:focus` changes afterwards when submitting your PR changes!

### Test order issues
Running specs in the order they failed may discover that the problem is that a previous test sets an state in the test environment that makes our flaky fail/pass. Tests should be independent from the rest.

After executing rspec you can see the seed used, add it as an option to rspec, for example:
`bin/rspec --seed 55638` (check **Randomized seed** value at beginning of issue)

### Other things to watch for
- Time related issues (current time, two time or date comparisons with miliseconds/time when its not needed)
- https://semaphoreci.com/community/tutorials/how-to-deal-with-and-eliminate-flaky-tests

### Final thoughts
The true objective of this issue is not "to fix" a flaky spec, but to change a spec that randomly fails into one that we can trust when it passes (or fails). 

That means you've to think "What are we testing here?" and "Can we test the same thing in another way?" or "Are we already testing this somewhere else at least partially?".

[...]

PR #2696, opened 2018-06-23
Fix flaky spec: Admin budgets Manage groups and headings Create group

javierv:
# References

* Issue #2573.

# Objectives

* Fix the flaky spec in `spec/features/admin/budgets_spec.rb:237`: "Admin budgets Manage groups and headings Create group"

To show the form, the JavaScript function `App.Forms.toggleLink()` calls jQuery's `toggle`, which animates the way the form appears on the screen.

Sometimes Capybara clicks the button while it's still moving, resulting in no request being sent to the server.

By disabling CSS and jQuery animations in the test environment, the button shows up immediately at the right position, and Capybara always clicks it correctly.

# Notes

* The implementation uses a partial which is only rendered in the test environment. Using a rack middleware (see references), which is also an option the latest Capybara offers as a beta feature, might be another option.
* These changes also fix a flaky spec in `spec/features/admin/homepage/homepage_spec.rb:43` ("Homepage Feeds Debates"), which fails on my machine sometimes with the message: `Capybara::ModalNotFound:  Unable to find modal dialog (...) Selenium::WebDriver::Error::TimeOutError: timed out after 2 seconds.`

# Related links

* [Rack::NoAnimations](https://github.com/mikz/rack-no_animations)
* Beta option to [disable animations in Capybara](https://github.com/teamcapybara/capybara/commit/e8b49b6)
* [How CSS Animations Can Break Your Tests](http://marcgg.com/blog/2015/01/05/css-animations-failing-capybara-specs/)
* [Turn off animations in Capybara tests](https://stackoverflow.com/questions/19662287/turn-off-animations-in-capybara-tests)


voodoorai2000:
Great solution and nice refactoring 😌

PR #2707, opened 2018-06-28
Change proposals validations

taitus:
References
===================
Add new proposals validations.

Objectives
===================
Add validations and his texts on proposals fields.
1700 characters -> 300 words
6000 characters -> 1200 words


taitus:
Sorry, I was wrong about the origin of the PR.

Issue #2520, opened 2018-03-06
Flaky spec: Officing Results Edit result


aitbw:
# What

Tests that fail randomly are called "flakies", this one seems to be one:

**Randomized seed:** 5891

**Travis failed build:** [Travis CI build 824, job 2](https://travis-ci.org/wairbut-m2c/consul/jobs/350063830)

**Failure:**

```
  1) Officing Results Edit result
     Failure/Error: select booth_name, from: 'officer_assignment_id'
     
     Capybara::ElementNotFound:
       Unable to find visible select box "officer_assignment_id" that is not disabled
     # ./spec/features/officing/results_spec.rb:100:in `block (2 levels) in <top (required)>'
```

# How
- [ ] Explain why the test is flaky, or under which conditions/scenario it fails randomly
- [ ] Explain why your PR fixes it

### Random values issues
If the problem comes from randomly generated values, running multiple times a single spec could help you reproduce the failure by running at your command line:
```bash
for run in {1..10}
do
  bin/rspec ./spec/features/budgets/investments_spec.rb:256
done
```

You can also try running a single spec in Travis:
Add option `:focus` to the spec and push your branch to Github, for example:
```ruby
scenario 'Show', :focus do
```

But remember to remove that `:focus` changes afterwards when submitting your PR changes!

### Test order issues
Running specs in the order they failed may discover that the problem is that a previous test sets an state in the test environment that makes our flaky fail/pass. Tests should be independent from the rest.

After executing rspec you can see the seed used, add it as an option to rspec, for example:
`bin/rspec --seed 55638` (check **Randomized seed** value at beginning of issue)

### Other things to watch for
- Time related issues (current time, two time or date comparisons with miliseconds/time when its not needed)
- [How to Deal With and Eliminate Flaky Tests - Semaphore](https://semaphoreci.com/community/tutorials/how-to-deal-with-and-eliminate-flaky-tests)

### Final thoughts
The true objective of this issue is not "to fix" a flaky spec, but to change a spec that randomly fails into one that we can trust when it passes (or fails). 

That means you've to think "What are we testing here?" and "Can we test the same thing in another way?" or "Are we already testing this somewhere else at least partially?".

Sometimes the fix is to re-write the entire tests with a different approach, or to extend an existing spec. Tests are sometimes written without taking a look at other tests files neither near scenarios.

javierv:
[...]

PR #2712, opened 2018-07-01
Fix flaky specs: Officing Results Add/Edit results

javierv:
# References

* Issue #2520
* Issue #2521
* Pull Request AyuntamientoMadrid#1342
* Pull Request AyuntamientoMadrid#1349
* <del>Pull Request AyuntamientoMadrid#1351</del> [Forked repo build with failed residence specs](https://travis-ci.org/javierv/consul/jobs/400560476)

# Objectives

Fix the flaky specs that appeared in `spec/features/officing/results_spec.rb:48` ("Officing Results Add results"), `spec/features/officing/results_spec.rb:85` ("Officing Results Edit result"), `spec/features/officing/residence_spec.rb:35` ("Residence Assigned officers Verify voter"), `spec/features/officing/residence_spec.rb:83` ("Residence Assigned officers Error on Census (year of birth)"), `spec/features/officing/residence_spec.rb:49` ("Residence Assigned officers Error on verify") and `spec/features/officing/residence_spec.rb:61` ("Residence Assigned officers Error on Census (document number)").

**This explanation is copied from AyuntamientoMadrid#1342 because the problem is exactly the same**. All credit to @iagirre.

The problems was that the test was not reaching to the residence verification page, so it couldn't find the select box for the DNI (and that's the error reported in the issue). To figure out what was going on there, \[she\] studied the flow that the test follows so that \[she\] could find the scenarios where it would crash, and, after doing it, step by step, \[she\] determined that the test will fail when:

* There are more than 1 `officer_assignments` (booths) for this user, so the page will redirect to the select booth page, instead to the residence verification page.
* There aren't any `officer_assignments`, so that the user can't even reach the target page (it redirects to the root).

The code that determines that is (this functions are run in `before_action` blocks):

```ruby
# app/controller/officing/base_controller.rb
# [...]

  def load_officer_assignment
    @officer_assignments ||= current_user.poll_officer.officer_assignments.where(date: Time.current.to_date)
    end

    def verify_officer_assignment
      if @officer_assignments.blank?
        redirect_to officing_root_path, notice: t("officing.residence.flash.not_allowed")
      end
    end

    def verify_booth
      return unless current_booth.blank?
      booths = todays_booths_for_officer(current_user.poll_officer)
      case booths.count
      when 0
        redirect_to officing_root_path
      when 1
        session[:booth_id] = booths.first.id
      else
        redirect_to new_officing_booth_path
      end
    end

    [...]

    def todays_booths_for_officer(officer)
      officer.officer_assignments.by_date(Date.today).map(&:booth).uniq
    end
```

After trying to reproduce it (without success), \[she\] saw the new piece of information about the probable failing hours, and \[she\] realized that the problem could be in the moment of creating the objects. If the `officer_assignments` are created at 23:59:59 and the rest of the test is executed after 00:00:00, the dates for the objects and the Date.current (used to check if there are any shifts today) won't be the same, because the shift will be for, lets say, 07/03/2018 and Date.current will be 08/03/2018, so, there are no shifts and we find the first scenario described above.

To prove that, \[she\] forced in the `factories.rb` the dates of the `officer_assignments` to an earlier date, and it failed in the exact same way as the reported one.

Freezing time ensures the date won't change in the middle of a test, and using dynamic attributes for dates ensures factories don't assign static dates before freezing time. According to the [factory bot documentation](https://www.rubydoc.info/gems/factory_bot/file/GETTING_STARTED.md):

> Most factory attributes can be added using static values that are evaluated when the factory is defined, but some attributes (such as associations and other attributes that must be dynamically generated) will need values assigned each time an instance is generated. These "dynamic" attributes can be added by passing a block instead of a parameter:
> (...)
> `date_of_birth   { 21.years.ago }`

# Notes

* I wasn't sure whether to use the same solution used in AyuntamientoMadrid#1342 or to add [Timecop](https://github.com/travisjeffery/timecop) to the `Gemfile`, but then realized Rails provides helpers to do exactly that and thought it was the easiest way. This decision is of course debatable; comments are welcome!
* Issues AyuntamientoMadrid#1195 and AyuntamientoMadrid#1208 are similar and their fixes (referenced above) haven't been backported yet.
* After upgrading to Rails 5, monkey-patching `ActiveSupport::Testing::TimeHelpers` won't be necessary. The `after { travel_back }` block won't be necessary either.

### DIFF
--- a/spec/features/officing/residence_spec.rb
+++ b/spec/features/officing/residence_spec.rb
@@ -3,6 +3,14 @@ require 'rails_helper'
 feature 'Residence' do
   let(:officer) { create(:poll_officer) }
 
+  background do
+    travel_to Time.now # TODO: use `freeze_time` after migrating to Rails 5.
+  end
+
+  after do
+    travel_back
+  end
+
   feature "Officers without assignments" do
 
     scenario "Can not access residence verification" do
--- a/spec/features/officing/results_spec.rb
+++ b/spec/features/officing/results_spec.rb
@@ -1,10 +1,9 @@
 require 'rails_helper'
-require 'time_helper'
 
 feature 'Officing Results' do
 
   background do
-    freeze_time
+    travel_to Time.now # TODO: use `freeze_time` after migrating to Rails 5.
     @poll_officer = create(:poll_officer)
     @officer_assignment = create(:poll_officer_assignment, :final, officer: @poll_officer)
     @poll = @officer_assignment.booth_assignment.poll
--- a/spec/features/officing/results_spec.rb
+++ b/spec/features/officing/results_spec.rb
@@ -1,8 +1,10 @@
 require 'rails_helper'
+require 'time_helper'
 
 feature 'Officing Results' do
 
   background do
+    freeze_time
     @poll_officer = create(:poll_officer)
     @officer_assignment = create(:poll_officer_assignment, :final, officer: @poll_officer)
     @poll = @officer_assignment.booth_assignment.poll
@@ -15,6 +17,10 @@ feature 'Officing Results' do
     create(:poll_question_answer, title: 'Tomorrow', question: @question_2)
 
     login_as(@poll_officer.user)
+  end
+
+  after do
+    travel_back
   end
 
   scenario 'Only polls where user is officer for results are accessible' do
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -24,7 +24,7 @@ FactoryBot.define do
     end
 
     trait :level_two do
-      residence_verified_at Time.current
+      residence_verified_at { Time.current }
       unconfirmed_phone "611111111"
       confirmed_phone "611111111"
       sms_confirmation_code "1234"
@@ -36,28 +36,28 @@ FactoryBot.define do
     end
 
     trait :level_three do
-      verified_at Time.current
+      verified_at { Time.current }
       document_type "1"
       document_number
     end
 
     trait :hidden do
-      hidden_at Time.current
+      hidden_at { Time.current }
     end
 
     trait :with_confirmed_hide do
-      confirmed_hide_at Time.current
+      confirmed_hide_at { Time.current }
     end
 
     trait :verified do
-      residence_verified_at Time.current
-      verified_at Time.current
+      residence_verified_at { Time.current }
+      verified_at { Time.current }
     end
 
     trait :in_census do
       document_number "12345678Z"
       document_type "1"
-      verified_at Time.current
+      verified_at { Time.current }
     end
   end
 
@@ -112,7 +112,7 @@ FactoryBot.define do
   factory :lock do
     user
     tries 0
-    locked_until Time.current
+    locked_until { Time.current }
   end
 
   factory :verified_user do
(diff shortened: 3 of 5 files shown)


javierv:
Hi @aitbw. Thank you for your comments :thumbsup:! I've removed the monkey patch.

> I was wondering if is it possible to implement a solution similar to those found in the PRs you mentioned.

Would you mind helping me with that? I'm not familiar with the source code and even after checking the other PRs am not sure which factories need to be changed :scream_cat:.

aitbw:
:thinking: You added a new solution on the latest commit. Does this solves the flakies as well? @javierv 

Because if that's the case, I don't think we'd need to implement something similar to the PRs you mentioned

javierv:
@aitbw Yes, the latest commit only refactors the code, doing the same thing without monkey patching. According to what I've tested locally, it solves the flakies. But it would be great to have another pair of eyes have a look :wink:.

javierv:
@aitbw Do you think it makes sense to backport the mentioned pull requests now that we're working on a similar topic?

P.S. (Sorry for the off-topic) I've just sent you an email. Hope it's OK :cold_sweat:! If you haven't received it, it might be inside your spam folder.

javierv:
@aitbw I've done some more testing and I've realized I hadn't fixed anything :sweat_smile:. I had forgotten to make sure factories were assigning dates dynamically.

I've tested it better, both manually and using Timecop, and it looks like this time the flakies are actually fixed. Somehow I hadn't realized until now that by changing my computer's date this can be tested anytime, and had also forgotten about `Timecop.travel` :astonished:. Sorry for that!

javierv:
I've realized AyuntamientoMadrid#1351 only fixes one test in `spec/features/officing/residence_spec.rb`, and that test isn't present on this repo. So I've added another commit fixing the flakies in `spec/features/officing/residence_spec.rb`. 

AyuntamientoMadrid#1343 also fixes tests which aren't present on this repo. So only two pull requests would need to be backported. I've updated the description of this pull request accordingly.

spec/features/officing/results_spec.rb
```diff
@@ -158,4 +159,7 @@
     within('#total_results') { expect(page).to have_content('66') }
   end
 
+  after do
```
aitbw: Can you move this block right after the `background` block? It's easier to understand what's going to happen that way :smile_cat:

spec/features/officing/residence_spec.rb
```diff
@@ -96,4 +100,7 @@
 
   end
 
+  after do
```
aitbw: Can you move this block right after the `background` block? It's easier to understand what's going to happen that way :smile_cat:

Amend if you can, there's no need to create a new commit for this.

aitbw:
Hello once again @javierv, thanks big time for this PR :tada:

As usual, I'll be testing this thoroughly locally and on Travis to see how well it fares but seems like it'll do :smile_cat: 

Re: the tests found on Ayuntamiento Madrid but not here —we have some custom functionality over there we don't plan to backport. Also, there are flakies which you'll only find on Ayuntamiento Madrid so that's why some specs, while they do test the same scenarios, will differ.

spec/features/officing/residence_spec.rb
javierv: Done!

voodoorai2000:
Very glad you figured out how this part of the codebase works, not trivial 👏

Issue #2718, opened 2018-07-02
Flaky spec: Budget Investments Show milestones


javierv:
# What

Tests that fail randomly are called "flakies", this one seems to be one:

**Randomized seed:** 12275

**Travis failed build:** [Travis build 24013, job 2](https://travis-ci.org/consul/consul/jobs/399210734)

**Failure:**

```
1) Budget Investments Show milestones
     Failure/Error: text.index(earlier_content) < text.index(later_content)

     NoMethodError:
       undefined method `<' for nil:NilClass
     # ./spec/support/matchers/appear_before.rb:4:in `block (2 levels) in <top (required)>'
     # ./spec/features/budgets/investments_spec.rb:1013:in `block (3 levels) in <top (required)>'
     # ./spec/features/budgets/investments_spec.rb:1012:in `block (2 levels) in <top (required)>'
```

# How
- [ ] Explain why the test is flaky, or under which conditions/scenario it fails randomly
- [ ] Explain why your PR fixes it

### Random values issues
If the problem comes from randomly generated values, running multiple times a single spec could help you reproduce the failure by running at your command line:
```bash
for run in {1..10}
do
  bin/rspec ./spec/features/budgets/investments_spec.rb:256
done
```

You can also try running a single spec in Travis:
Add option `:focus` to the spec and push your branch to Github, for example:
```ruby
scenario 'Show', :focus do
```

But remember to remove that `:focus` changes afterwards when submitting your PR changes!

### Test order issues
Running specs in the order they failed may discover that the problem is that a previous test sets an state in the test environment that makes our flaky fail/pass. Tests should be independent from the rest.

After executing rspec you can see the seed used, add it as an option to rspec, for example:
`bin/rspec --seed 55638` (check **Randomized seed** value at beginning of issue)

### Other things to watch for
- Time related issues (current time, two time or date comparisons with miliseconds/time when its not needed)
- https://semaphoreci.com/community/tutorials/how-to-deal-with-and-eliminate-flaky-tests

### Final thoughts
The true objective of this issue is not "to fix" a flaky spec, but to change a spec that randomly fails into one that we can trust when it passes (or fails).

That means you've to think "What are we testing here?" and "Can we test the same thing in another way?" or "Are we already testing this somewhere else at least partially?".

Sometimes the fix is to re-write the entire tests with a different approach, or to extend an existing spec. Tests are sometimes written without taking a look at other tests files neither near scenarios.

PR #2719, opened 2018-07-02
Fix flaky spec: Budget Investments Show milestones

javierv:
# References

* Issue #2718

# Objectives

Fix the flaky spec that appeared in `spec/features/budgets/investments_spec.rb:994` ("Budget Investments Show milestones").

The test fails if `Globalize.locale != I18n.locale` when the test starts. AFAIK there's only one test which can modify that locale: `spec/features/translations_spec.rb:76` ("Translations Milestones Globalize javascript interface Highlight current locale"). So, to reproduce it locally, run those tests making sure the translations test runs first. On my machine, this command makes the test fail every time:

`bin/rspec spec/features/budgets/investments_spec.rb:994 spec/features/translations_spec.rb:76 --seed 6700`

Here's what happens during those two tests:

1. The translations test sets `Globalize.locale = :es`.
2. The milestones test `config.before` block sets `I18n.locale = :en`
3. The test creates data like `first_milestone`, which will get assigned `description_es` (as set by Globalize).
4. The `visit budget_investment_path` renders the milestones, which include the line `globalize(neutral_locale(locale))`.
5. Since `locale` is `I18n.locale` (that is, `:en`), the block will call `first_milestone.description_en`, which will return `nil`.
6. The test fails because it expects a description and finds `nil` instead.

By setting `Globalize.locale = I18n.locale` before each test, the changes one test might cause don't affect any other test.

# Notes

* Changing code in `ApplicationController` and `Management::BaseController` isn't necessary to make the test pass. However, since having different values in `Globalize.locale` and `I18n.locale` has proven to be an issue on the test enviroment, I've also changed application code in order to avoid similar situations on production. Different opinions are welcome!

### DIFF
--- a/app/controllers/concerns/translatable.rb
+++ b/app/controllers/concerns/translatable.rb
@@ -2,7 +2,6 @@ module Translatable
   extend ActiveSupport::Concern
 
   included do
-    before_action :set_translation_locale
     before_action :delete_translations, only: [:update]
   end
 
@@ -10,10 +9,6 @@ module Translatable
 
     def translation_params(params)
       resource_model.globalize_attribute_names.select { |k, v| params.include?(k.to_sym) && params[k].present? }
-    end
-
-    def set_translation_locale
-      Globalize.locale = I18n.locale
     end
 
     def delete_translations
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -59,6 +59,7 @@ class ApplicationController < ActionController::Base
       end
 
       I18n.locale = locale
+      Globalize.locale = I18n.locale
     end
 
     def set_layout
--- a/app/controllers/management/base_controller.rb
+++ b/app/controllers/management/base_controller.rb
@@ -39,6 +39,7 @@ class Management::BaseController < ActionController::Base
       session[:locale] ||= I18n.default_locale
 
       I18n.locale = session[:locale]
+      Globalize.locale = I18n.locale
     end
 
     def current_budget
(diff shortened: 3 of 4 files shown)


app/controllers/concerns/translatable.rb
```diff
@@ -2,7 +2,6 @@ module Translatable
   extend ActiveSupport::Concern
 
   included do
-    before_action :set_translation_locale
```
aitbw: What will happen if we don't delete this callback? (as well as its related action)

javierv:
@aitbw Funny thing: I don't think the code can cause problems on production. But I didn't think it could cause problems while testing, and it did, so I don't trust myself that much :laughing:.

Without these changes, on production there are moments where `I18n.locale` and `Globalize.locale` are different. I'm afraid I don't know whether that could cause problems, at least with the current code :thinking:.

However, I see an advantage while developing, particularly for developers who aren't so familiar with the code like myself. Setting `Globalize.locale` after `I18n.locale` makes the code easier to follow and easier to debug.

Needless to say, I'm not familiar enough with CONSUL to be sure one option is better than the other one, so whatever you think is fine for me :+1:!

app/controllers/concerns/translatable.rb
javierv: With these changes, this callback would be redundant, since the same thing is now done in `ApplicationController`.

javierv: Before these changes, `ApplicationController#set_locale` configured `I18n.locale` and `Translatable#set_translation_locale` configured `Globalize.locale`.

After these changes, `ApplicationController#set_locale` configures both `I18n.locale` and `Globalize.locale`. So there's no need for `Translatable#set_translation_locale` anymore, since `Globalize.locale` is already configured.

PR #2711, opened 2018-06-29
Fix flaky specs using CKEditor

javierv:
# References

* AyuntamientoMadrid#1317 ([recent Travis fail](https://travis-ci.org/consul/consul/jobs/396908742))
* AyuntamientoMadrid#1429 ([recent Travis fail](https://travis-ci.org/consul/consul/jobs/395938528))
* AyuntamientoMadrid#1430 ([recent Travis fail](https://travis-ci.org/consul/consul/jobs/397298334))
* AyuntamientoMadrid#1431 ([recent Travis fail](https://travis-ci.org/consul/consul/jobs/397957410))
* [Failed suggesting debates shows up suggestions spec](https://travis-ci.org/consul/consul/jobs/395890858)
* [Failed nested documentable investments spec](https://travis-ci.org/consul/consul/jobs/395890857)
* <del>[Failed nested documentable proposals spec](https://travis-ci.org/consul/consul/jobs/399587929)</del> Failure log not available anymore.
* [Failed nested imageable spec](https://travis-ci.org/AyuntamientoMadrid/consul/jobs/370507529)
* [Failed Turbolinks sanity check from budget's show spec](https://travis-ci.org/AyuntamientoMadrid/consul/jobs/370709360)
* [Failed investment creation with category tags spec](https://travis-ci.org/consul/consul/jobs/389821901)
* Pull Request #1196

# Objectives

[...]

PR #2720, opened 2018-07-03
Fix moderation index views information positioning

aitbw:
References
===================
* This is a backport of AyuntamientoMadrid#1530

Objectives
===================
* Correct entries information and filter positioning for all `moderation#index` views, similar to what was done in AyuntamientoMadrid#1415
* Add missing `thead` and `tbody` tags

Visual Changes
===================
**Note**: For brevity, and since the proposed changes are all the same, only one pair of `before/after` screenshots will be included

## Before

before

## After

after

Notes
===================
* None


javierv:
@aitbw I've just added the failure shown in this pull request to the list of flakies fixed by #2711.

Issue #2713, opened 2018-07-01
Flaky spec: Votes Debates Single debate Update


javierv:
# What
Tests that fail randomly are called "flakies", this one seems to be one:

**Randomized seed:** <del>3315</del> 57656

**Travis failed build:** <del>[Travis build 24008, job 1](https://travis-ci.org/consul/consul/jobs/398882809)</del> [Forked repo Travis build](https://travis-ci.org/javierv/consul/jobs/399449443)

**Failure:**

```
2) Votes Debates Single debate Update
     Failure/Error: expect(page).to have_css("a.no-voted")
       expected to find visible css "a.no-voted" within #<Capybara::Node::Element tag="div" path="/html/body/div[1]/div[1]/div/aside/div[2]/div/div[1]"> but there were no matches
     # ./spec/features/votes_spec.rb:90:in `block (5 levels) in <top (required)>'
     # ./spec/features/votes_spec.rb:88:in `block (4 levels) in <top (required)>'
```

# How
- [ ] Explain why the test is flaky, or under which conditions/scenario it fails randomly
- [ ] Explain why your PR fixes it

### Random values issues
If the problem comes from randomly generated values, running multiple times a single spec could help you reproduce the failure by running at your command line:
```bash
for run in {1..10}
do
  bin/rspec ./spec/features/budgets/investments_spec.rb:256
done
```

You can also try running a single spec in Travis:
Add option `:focus` to the spec and push your branch to Github, for example:
```ruby
scenario 'Show', :focus do
```

But remember to remove that `:focus` changes afterwards when submitting your PR changes!

### Test order issues
Running specs in the order they failed may discover that the problem is that a previous test sets an state in the test environment that makes our flaky fail/pass. Tests should be independent from the rest.

After executing rspec you can see the seed used, add it as an option to rspec, for example:
`bin/rspec --seed 55638` (check **Randomized seed** value at beginning of issue)

### Other things to watch for
- Time related issues (current time, two time or date comparisons with miliseconds/time when its not needed)
- https://semaphoreci.com/community/tutorials/how-to-deal-with-and-eliminate-flaky-tests

### Final thoughts
The true objective of this issue is not "to fix" a flaky spec, but to change a spec that randomly fails into one that we can trust when it passes (or fails). 

That means you've to think "What are we testing here?" and "Can we test the same thing in another way?" or "Are we already testing this somewhere else at least partially?".

[...]

PR #2448, opened 2018-02-06
Fix for `voting comments` flaky specs

aitbw:
Where
=====
* **Related PR:** AyuntamientoMadrid/consul#1212

What
====
* Backport the aforementioned PR into CONSUL to avoid possible flaky specs, see original PR for more details

Test
====
* Fixed coverage, all green! ✅

PR #2455, opened 2018-02-08
Fixed flaky spec: missing comment on legislation annotation

raul-fuentes:
Where
=====
* **Related Issue:** https://github.com/AyuntamientoMadrid/consul/issues/1180
* **Related PR's:** https://github.com/AyuntamientoMadrid/consul/pull/1232

What
====
Fix flaky spec, rspec ./spec/features/comments/legislation_annotations_spec.rb:139 

How
===
Explained in the related PR.

Screenshots
===========
There aren't

Test
====
Explained in the related PR

Deployment
==========
Nothing to apply

Warnings
========
Nothing to apply

Issue #2129, opened 2017-11-14
Hunt Flaky tests 


bertocq:
This one is tricky, but will be so much appreciated by devs/contributors (in particular by first-time-contributors)

We have some usual suspects that normally break once in a while (1 out of 3 travis builds):

- spec/features/tracks_spec.rb (I can't find any examples right now on travis)

- spec/shared/features/nested_imageable.rb
https://travis-ci.org/consul/consul/jobs/301885841

- spec/features/admin/stats_spec.rb
https://travis-ci.org/consul/consul/jobs/285994433
https://travis-ci.org/consul/consul/jobs/278592658
https://travis-ci.org/consul/consul/jobs/285995193

There are many others that fail because of `Unknown error while speaking to annotation store! [object Object]` like https://travis-ci.org/consul/consul/jobs/291110815 or `TypeError: undefined is not an object (evaluating 'this.document.getWindow().$.getSelection')` like https://travis-ci.org/consul/consul/jobs/301885842

We'll try to be conscious about this issue, and instead of rebuilding failed travis builds, copy&paste here tests that seem to be flaky (the rspec failure output) to make it easier to narrow whats going on at each one.

bertocq:
Another set of suspected flaky specs:

## Failures
```ruby
Failures:
  1) Votes Debates Single debate Update
     Failure/Error: expect(page).to have_css("a.no-voted")
       expected to find css "a.no-voted" but there were no matches
     # ./spec/features/votes_spec.rb:90:in `block (5 levels) in <top (required)>'
     # ./spec/features/votes_spec.rb:88:in `block (4 levels) in <top (required)>'
  2) Commenting legislation questions Sanitizes comment body for security
     Failure/Error: expect(page).to have_content "click me http://www.url.com"
       expected to find text "click me http://www.url.com" in "Manuela2653 • Author • 2017-11-15 19:42:03 a comment No votes | I agree 0 I disagree 0 No responses"
     # ./spec/features/comments/legislation_annotations_spec.rb:163:in `block (3 levels) in <top (required)>'
     # ./spec/features/comments/legislation_annotations_spec.rb:162:in `block (2 levels) in <top (required)>'
  3) Commenting legislation questions Index
     Failure/Error: expect(page).to have_content comment.user.name
       expected to find text "Manuela2655" in "Manuela2658 • 2017-11-15 19:42:04 Comment body 207 No votes | I agree 0 I disagree 0 No responses"
[...]

PR #2734, opened 2018-07-05
Fix flaky specs: Votes Debates and Voting comments Update

javierv:
# References

* Issue #2713
* Pull Request #2448
* AyuntamientoMadrid#1173
* AyuntamientoMadrid#1174
* AyuntamientoMadrid#1177
* AyuntamientoMadrid#1192
* AyuntamientoMadrid#1201

# Objectives

Fix the flaky specs that appeared in `spec/features/votes_spec.rb:82` ("Votes Debates Single debate Update"), `spec/features/comments/polls_spec.rb:495` ("Commenting polls Voting comments Update"), `spec/features/comments/topics_spec.rb:525` ("Commenting topics from proposals Voting comments Update"), `spec/features/comments/topics_spec.rb:525` ("Commenting topics from budget investments Voting comments Update"), `spec/features/comments/debates_spec.rb:486` ("Commenting debates Voting comments Update"), `spec/features/comments/budget_investments_spec.rb:481` ("Commenting Budget::Investments Voting comments Update") and `spec/features/comments/legislation_annotations_spec.rb:562` ("Commenting legislation questions Voting comments Update").

Here's the relevant part of the spec:

```ruby
visit debate_path(create(:debate))

find('.in-favor a').click
find('.against a').click
```

After clicking the first link, there's an AJAX request which replaces the existing `.in-favor a` and `.against a` links with new elements. So if Capybara tries to click the existing `.against a` link at the same moment it's being replaced, clicking the link won't generate a new request, similar to what happened in #2696.

Making Capybara check the page for new content before clicking the second link solves the problem.

## Original solution (obsolete)

Another option would be to update the content without replacing the existing links, so it's always available for Capybara to click. That way would also solve a small issue affecting keyboard users (note the way the focus disappears after clicking an icon, and how it goes to the "like" icon when pressing "tab"):

The link loses focus after being clicked and using tab focuses the like icon

Since the links are regenerated after being clicked, the focus goes back to the element being replaced, and that leads to awkward behaviour like clicking "dislike", pressing "tab" and the focus going back to "like".

By changing the content without regenerating the links, the links wouldn't lose focus after being clicked and the "tab" key behaves as expected:

The link is still focused after being clicked and using tab focuses the next link

### DIFF
--- a/spec/features/comments/budget_investments_spec.rb
+++ b/spec/features/comments/budget_investments_spec.rb
@@ -471,6 +471,11 @@ feature 'Commenting Budget::Investments' do
 
       within("#comment_#{@comment.id}_votes") do
         find('.in_favor a').click
+
+        within('.in_favor') do
+          expect(page).to have_content "1"
+        end
+
         find('.against a').click
 
         within('.in_favor') do
--- a/spec/features/comments/debates_spec.rb
+++ b/spec/features/comments/debates_spec.rb
@@ -476,6 +476,11 @@ feature 'Commenting debates' do
 
       within("#comment_#{@comment.id}_votes") do
         find('.in_favor a').click
+
+        within('.in_favor') do
+          expect(page).to have_content "1"
+        end
+
         find('.against a').click
 
         within('.in_favor') do
--- a/spec/features/comments/legislation_annotations_spec.rb
+++ b/spec/features/comments/legislation_annotations_spec.rb
@@ -553,6 +553,11 @@ feature 'Commenting legislation questions' do
 
       within("#comment_#{@comment.id}_votes") do
         find('.in_favor a').click
+
+        within('.in_favor') do
+          expect(page).to have_content "1"
+        end
+
         find('.against a').click
 
         within('.in_favor') do
--- a/spec/features/comments/polls_spec.rb
+++ b/spec/features/comments/polls_spec.rb
@@ -485,6 +485,11 @@ feature 'Commenting polls' do
 
       within("#comment_#{@comment.id}_votes") do
         find('.in_favor a').click
+
+        within('.in_favor') do
+          expect(page).to have_content "1"
+        end
+
         find('.against a').click
 
         within('.in_favor') do
--- a/spec/features/comments/topics_spec.rb
+++ b/spec/features/comments/topics_spec.rb
@@ -515,6 +515,11 @@ feature 'Commenting topics from proposals' do
 
       within("#comment_#{@comment.id}_votes") do
         find('.in_favor a').click
+
+        within('.in_favor') do
+          expect(page).to have_content "1"
+        end
+
         find('.against a').click
 
         within('.in_favor') do
@@ -1065,6 +1070,11 @@ feature 'Commenting topics from budget investments' do
 
       within("#comment_#{@comment.id}_votes") do
         find('.in_favor a').click
+
+        within('.in_favor') do
+          expect(page).to have_content "1"
+        end
+
         find('.against a').click
 
         within('.in_favor') do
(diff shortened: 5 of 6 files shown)


javierv:
This is the last flaky spec I can work on without needing more help than usual.

Here's the status of the rest of the flakies I've seen fail during the last two weeks:

* As mentioned in #2712, issues like AyuntamientoMadrid#1195, AyuntamientoMadrid#1202, AyuntamientoMadrid#1203, AyuntamientoMadrid#1204, AyuntamientoMadrid#1208 and AyuntamientoMadrid#1210 are already fixed in [AyuntamientoMadrid's fork](https://github.com/AyuntamientoMadrid/consul) and need to be backported.
* AyuntamientoMadrid#1194 is failing on my machine sometimes. I've left [a comment in  AyuntamientoMadrid#1256](https://github.com/AyuntamientoMadrid/consul/pull/1256#issuecomment-402863138) with what I think might be a possible fix.
* AyuntamientoMadrid#1180 and [its related flakies](https://github.com/consul/consul/issues/2129#issuecomment-344741720) are also failing on my machine and I think I've seen a recent Travis build where one of them failed. I've also left [a comment in #2455](https://github.com/consul/consul/pull/2455#issuecomment-401971618) saying what I think is causing the tests to fail.

So any help would be appreciated! Maybe after fixing those ones and merging the pull requests related to flakies, Travis' build will be green 100% of the time and there'll be no need to restart builds or wonder if the failures are flakies :smile:. At least for this repo; AyuntamientoMadrid's fork is probably a different beast :thinking:.

Sorry for intruding so much regarding this topic! The first time I configured CONSUL on my machine I saw many tests fail and thought there was something wrong with my configuration. The same happened when I first opened a pull request and Travis reported a failure: I thought I had introduced a bug. So I thought fixing the flakies would help new developers know whether they're doing something wrong or everything is fine. Sorry if I went too far, and sorry sometimes I didn't say I was working on an issue before starting to work on it (mostly because I've done most of them during weekends, but still that isn't an excuse).

voodoorai2000:
Hmmm @javierv we try to use a as little javascript as possible, it's not very pretty and it can have some unexpected side effects. 

We are trying to solve just a flaky spec in this PR. I don't see the need to add javascript to change, the always delicate, voting system.

+1 to using Capybara to check for the correct links being present

javierv:
@voodoorai2000 :+1: I'll make the changes before Monday.

After opening this pull request, I realized #2448 solved a similar issue affecting different tests. However, it wasn't merged because switching to chromedriver apparently made the flakies disappear (although unfortunately it only made them happen less frequently).

@voodoorai2000 @aitbw So I'd suggest I change this pull request so it uses the same solution #2448 does, and then we merge both pull requests. What do you think?

voodoorai2000:
+1 @javierv 👌

aitbw:
I wouldn't merge #2448 at all since it didn't solve anything before the Headless Chrome migration

javierv:
@aitbw My first impression was #2448 would fix the flakies it was supposed to fix :thinking: (as noted in the comments there, those specs still fail on my machine sometimes). Since it uses the same solution I was going to use on this pull request, I'll test it and will let you know before Monday.

voodoorai2000:
Mine too 😌

I don’t quite see how it solves the flackiness, apart from possible session data. In any case I would expect Capybara to make a clean session slate on every test

But if it fails without the fix... and it passes with the fix 👍🏻👍🏻👍🏻

javierv:
Sorry I forgot to clarify #2448 did two things:

1. It added a `logout` after each affected test.
2. It checked the page before clicking `.against a`.

AFAIK the first change didn't affect these specs; however, the second did. There was a small typo in that PR, though, since after voting in favor it was checking for "0" instead of "1" as the content of votes in favor.

So now this PR has two commits, rebased against current master: one keeping the changes in #2448 which solved the problem and fixing the "1" typo (all credit to @aitbw), and another one implemeting the same solution when voting a debate.

On my machine all these specs still fail on master sometimes. After applying these changes, none of them do.

Issue #2557, opened 2018-03-28
Flaky spec: Admin budget investments Edit Adds existing valuation tags


aitbw:
# What

Tests that fail randomly are called "flakies", this one seems to be one:

**Randomized seed:** 52021

**Travis failed build:** [Travis build 20903, job 1](https://travis-ci.org/consul/consul/jobs/359034088)

**Failure:**

```
  1) Admin budget investments Edit Adds existing valuation tags
     Failure/Error: expect(page).to have_content 'Investment project updated succesfully.'
       expected to find text "Investment project updated succesfully." in "Language: English Español Français Nederlands Português CONSUL ADMINISTRATION Admin Notifications My activity My account Sign out Categories Moderated content Polls Participatory budgets Profiles Collaborative Legislation Banners Site customization Manage geozones Signature Sheets Sending of emails Go back dolorem 179 - Group: Group 100 - Investment project 2 Budget Investment 64 title Heading: Heading 96 By: Manuela626 Sent: 2018-03-27 20:55:12 Compatibility: Compatible Selection: Not selected Winner: No Spend money on this Image: Without image Documents: Without documents User tags: Edit Classification Assigned administrator: Undefined Tags: Education Assigned valuators: Undefined Edit classification Dossier Price (€): 10 Cost during the first year (€): Undefined Time scope: Undefined Feasibility: Undecided Edit dossier Valuation comments Leave your comment Comment as admin Milestone Don't have defined milestones Create new milestone"
     # ./spec/features/admin/budget_investments_spec.rb:595:in `block (3 levels) in <top (required)>'
```

# How
- [ ] Explain why the test is flaky, or under which conditions/scenario it fails randomly
- [ ] Explain why your PR fixes it

### Random values issues
If the problem comes from randomly generated values, running multiple times a single spec could help you reproduce the failure by running at your command line:
```bash
for run in {1..10}
do
  bin/rspec ./spec/features/budgets/investments_spec.rb:256
done
```

You can also try running a single spec in Travis:
Add option `:focus` to the spec and push your branch to Github, for example:
```ruby
scenario 'Show', :focus do
```

But remember to remove that `:focus` changes afterwards when submitting your PR changes!

### Test order issues
Running specs in the order they failed may discover that the problem is that a previous test sets an state in the test environment that makes our flaky fail/pass. Tests should be independent from the rest.

After executing rspec you can see the seed used, add it as an option to rspec, for example:
[...]

PR #2534, opened 2018-03-09
Migration from PhantomJS to Headless Chrome

aitbw:
References
==========
* Related issues: #2523, [flaky specs-related issues @ consul/consul](https://github.com/consul/consul/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+flaky) and [flaky specs-related issues @ AyuntamientoMadrid/consul](https://github.com/AyuntamientoMadrid/consul/issues?q=is%3Aissue+is%3Aopen+label%3A%22Flaky+spec%22) —closes #2523 once merged into `master`

* Related PRs: [flaky specs-related PRs @ consul/consul](https://github.com/consul/consul/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Aopen+flaky) and [flaky specs-related PRs @ AyuntamientoMadrid/consul](https://github.com/AyuntamientoMadrid/consul/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Aopen+flaky)

* Related links:
  * [How GitLab switched to Headless Chrome for testing](https://about.gitlab.com/2017/12/19/moving-to-headless-chrome/)
  * [Switching to Headless Chrome for Rails System Tests](http://technopragmatica.blogspot.com/2017/10/switching-to-headless-chrome-for-rails_31.html)
  * [Flaky tests & Capybara best practices](https://www.simplybusiness.co.uk/about-us/tech/2015/02/flaky-tests-and-capybara-best-practices/)
  * [Ruby Capybara with selenium Cheat Sheat](https://blog.morizyun.com/blog/capybara-selenium-webdriver-ruby/)
  * [Getting Started with Headless Chrome ](https://developers.google.com/web/updates/2017/04/headless-chrome)
  * [Running feature specs with Capybara and Chrome headless](https://drivy.engineering/running-capybara-headless-chrome/)
  * [Headless Capybara Feature Specs with Chrome](https://robots.thoughtbot.com/headless-feature-specs-with-chrome)
  * [Moving From PhantomJS to Headless Chrome in Rails](https://medium.com/@weilandia/moving-from-phantomjs-to-headless-chrome-in-rails-fewer-hacks-simpler-debugging-and-more-c2e2558742c5)
  * [Selenium With Headless Chrome On Travis CI ](http://www.amihaiemil.com/2017/07/14/selenium-headless-chrome-travis.html)

Objectives
==========
* Migrate completely from PhantomJS to Headless Chrome, mitigating entirely **all** flaky specs :tada: 

[...]

PR #2754, opened 2018-07-10
Flaky spec: Admin budget investments Index Filtering by admin

decabeza:
References
=====
This is a backport of https://github.com/AyuntamientoMadrid/consul/pull/1234
Related Issue: https://github.com/AyuntamientoMadrid/consul/issues/1217

What
====
- Hunt the flaky that appears in `spec/features/admin/budget_investments_spec.rb:148`
- Specific error in test: 
```  
2) Admin budget investments Index Filtering by admin
     Failure/Error: expect(page).to have_content('There are 2 investments')
       expected to find text "There are 2 investments" in "Language: Decide Madrid ADMINISTRATION Admin Notifications Delegation My activity My account Sign out Categories Moderated content Polls Participatory budgets Profiles Collaborative Legislation Banners Site customization Moderator activity Configuration settings Newsletters Statistics Manage geozones Signature Sheets iste 378 - Investment projects Advanced filters Filter: All Without assigned admin Without assigned valuator Under valuation Valuation finished Winners Download current selectionThere is 1 investment ID Title Supports Administrator Valuator Scope of operation Feasibility Val. Fin. Selected 3 Realocate visitors 0 Admin 1 No valuators assigned Heading 312 Undecided No"
     # ./spec/features/admin/budget_investments_spec.rb:148:in `block (3 levels) in <top (required)>'
```

How
===

**1.  Explanation:**

- After creating a budget investment in the index of the page, it is verified that the previously created is shown.
- Then the filters are used to show investment of specific budgets.

**2.  Flaky Explanation:**

- The test fails because it does not find the budget investments after being filtered several times. It is assumed that the entire page can not be fully rendered due to the existence of several selects

**3.  Fix Explantion:**

I have divided the test into three parts to create a context with the creation of the variables and the visit to the page. Then individually try the selects
```       
     context "Filtering by admin" do
            background do
            end
     
            scenario "Should have budgets links" , :js do
            end
      end
```

Notes
========
- This flaky is related to [https://github.com/AyuntamientoMadrid/consul/issues/1218](url) and has been treated in the same way.

### DIFF
--- a/spec/features/admin/budget_investments_spec.rb
+++ b/spec/features/admin/budget_investments_spec.rb
@@ -134,33 +134,46 @@ feature 'Admin budget investments' do
       expect(page).not_to have_link("Plant trees")
     end
 
-    scenario "Filtering by admin", :js do
-      user = create(:user, username: 'Admin 1')
-      administrator = create(:administrator, user: user)
+    context "Filtering by admin" do
+      background do
+        user = create(:user, username: 'Admin 1')
+        administrator = create(:administrator, user: user)
 
-      create(:budget_investment, title: "Realocate visitors", budget: budget, administrator: administrator)
-      create(:budget_investment, title: "Destroy the city", budget: budget)
+        create(:budget_investment, title: "Realocate visitors", budget: budget, administrator: administrator)
+        create(:budget_investment, title: "Destroy the city", budget: budget)
 
-      visit admin_budget_budget_investments_path(budget_id: budget.id)
-      expect(page).to have_link("Realocate visitors")
-      expect(page).to have_link("Destroy the city")
+        visit admin_budget_budget_investments_path(budget_id: budget.id)
+      end
 
-      select "Admin 1", from: "administrator_id"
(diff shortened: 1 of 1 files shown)


aitbw:
I don't think we need this backport, @decabeza —as per #2534 and AyuntamientoMadrid#1387

javierv:
Pardon me for intruding :pray:. Just in case it helps, I've run the specs in that file hundreds of times while trying to reproduce #2557 and AyuntamientoMadrid#1194, and I've never seen this failure. It doesn't mean it's fixed, but it could be, as @aitbw says.

Also note that after these changes, the test starts selecting "All administrators" instead of doing it after selecting "Admin 1", like it's currently done in the master branch.

IMHO the behaviour in the master branch makes more sense because by default the page is already filtered for "All administrators", so selecting it at the beginning of the test will not trigger any changes and the page will stay the same way.

decabeza:
Thanks, @aitbw and @javierv! 😉

Issue #2361, opened 2018-01-22
Make Community & Communitable truly polymorphic


bertocq:
Work on this issue was done on pull request #2771. Applying the changes on that pull request to our current master branch might be all we need to implement this refactoring.

We have so many `community.from_proposal?` around that its really hard to make a fourth model "communitable".

Most of the methods (if not all) at https://github.com/consul/consul/blob/master/app/helpers/communities_helper.rb should me moved into the concern or just dissapear

The relations as well doesn't quite seem right (Commuties table should have the communitable id & type columns instead of the Proposal / Investment ones..)

This is not an urgent refactor but we're seeing problems in production at Madrid because of non-existent proposals that have their associated community still around (that's another issue to be tackled at: https://github.com/consul/consul/issues/2360)

javierv:
I'll have a look at this one.

voodoorai2000:
Awesome!

javierv:
I've opened #2771 with a very basic "quick and under pressure" :wink: refactoring. I'll continue working on this issue next week.

Here's my first bunch of questions regarding this issue:

* I see `Legislation::Proposal` includes `Communitable`, but I haven't found any references to communities there, nor to legislation proposals in communities. Is there something I'm missing?
* The I18n keys used in this section use the key `investment`. Other sections use `budget_investment`, and Active Record uses `budget/investment`. Although it would be easier to use `budget/investment` here, I've left the original `investment` key because I don't know the impact changing a key could have on the CONSUL community. Is it usually safe to change a translation key, or is it recommended not to do so?

Issue #2360, opened 2018-01-22
Ensure community is destroyed if associated communitable dissapears


bertocq:
Ensure (first with an test, then with code) that after totally destroying a Proposal, Investment or Legislation Proposal... the associated Community is completely destroyed as well.

iagirre:
I'll do this one.

PR #2771, opened 2018-07-17
Make Community & Communitable truly polymorphic 

javierv:
# References

* Issue #2361

# Context

Both proposals and budget investments have communities. However, their relationship in the database isn't properly defined, which makes it difficult to know when a community belongs to a proposal and when a community belongs to a budget investment. That leads to complex code to circumvent this lack of relationship.

# Objectives

Make a polymorphic relationship between community and communitable models in order to simplify the code and make it easier to add new communitable models in the future.

# Pending questions

* `Legislation::Proposal` includes `Communitable`, but I haven't found any references to communities there, nor to legislation proposals in communities. Is there something I'm missing?
* The I18n keys used in this section use the key `investment`. Other sections use `budget_investment`, and Active Record uses `budget/investment`. Although it would be easier to use `budget/investment here`, I've left the original `investment` key because I don't know the impact changing a key could have on the CONSUL community. Is it usually safe to change a translation key, or is it recommended not to do so?
* I've removed some helper methods but then I've realized some forks might be using those methods somewhere else. Is it safe to remove them, or shall we keep them in order to maintain the API?

# Notes

* :warning: After applying these changes, run `rake communities:migrate_communitable`.
* The method `communitable_key` is in the model, although similar funcionality was originally in the helper. IMHO the method doesn't belong in the model nor the helper, but since we aren't using classes like decorators, I've left it there.
* The `community_back_link_path` method is very hard to refactor, since the `budget_investment_path` method requires both the budget and the investment. I've left it (mostly) the same way it was.
* I thought implementing #2360 as well would be trivial, but it isn't :smile:, since usually we don't really destroy communitable objects but only hide them. So I haven't done anything related to that issue.

# Deprecation warnings:

Release v0.17 should include two warnings in the changelog:

* The column `community_id` from `proposals` and `budget_investments` will be removed in v0.18. Run `rake communities:migrate_communitable`.
* The methods `Community#proposal` and `Community#investment` are deprecated and will be removed in v0.18.

### DIFF
--- a/app/controllers/communities_controller.rb
+++ b/app/controllers/communities_controller.rb
@@ -34,6 +34,6 @@ class CommunitiesController < ApplicationController
   end
 
   def communitable_exists?
-    @community.proposal.present? || @community.investment.present?
+    @community.communitable.present?
   end
 end
--- a/app/helpers/communities_helper.rb
+++ b/app/helpers/communities_helper.rb
@@ -1,9 +1,9 @@
 module CommunitiesHelper
   def community_back_link_path(community)
     if community.from_proposal?
-      proposal_path(community.proposal)
+      proposal_path(community.communitable)
     else
-      budget_investment_path(community.investment.budget_id, community.investment)
+      budget_investment_path(community.communitable.budget_id, community.communitable)
     end
   end
 
--- a/app/helpers/communities_helper.rb
+++ b/app/helpers/communities_helper.rb
@@ -1,35 +1,10 @@
 module CommunitiesHelper
-
-  def community_title(community)
-    community.from_proposal? ? community.proposal.title : community.investment.title
-  end
-
-  def community_text(community)
-    community.from_proposal? ? t("community.show.title.proposal") : t("community.show.title.investment")
-  end
-
-  def community_description(community)
-    community.from_proposal? ? t("community.show.description.proposal") : t("community.show.description.investment")
-  end
-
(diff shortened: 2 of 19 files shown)


javierv:
I've just realized I had forgotten to include the migration file in my last commit. It's fixed now.

aitbw:
Hello again @javierv !

My two cents re: this PR:

* #2360 is solved through AyuntamientoMadrid#1260 but it's not an urgent patch so we haven't merged and backported it here yet

* We usually document Rake tasks migration on the `Notes` section of the PR between :warning: emojis

* When we're deprecating tables/attributes from the database, we tend to create 2 pull requests:

  * One for the feature/patch that deprecates the table(s) and/or attribute(s)
  * Another one right after with **just** the ActiveRecord migration that destroys said table(s) and/or attribute(s) —take as example #2319 and #2382 

This way we can include the PR with the deprecation warnings on one CONSUL release and on the following release, we include the migration to destroy the deprecated table(s) and/or attribute(s)

What do you think?

javierm:
(I'm the same developer) @aitbw I've created javierm#1 to remove the obsolete columns. Since this PR isn't merged yet, I've created it on my fork so the changes reference this branch.

javierm:
@aitbw Thanks for the suggestion! Rebased.

javierm:
Closing due to inactivity.

Issue #2781, opened 2018-07-23
Configurable Top Right and Main Navigation Menus


devinbalkind:
### User Story

As an admin user I want to be able to control which menu items appear on the top_right menu and in the main_navigation.

### Why is this Needed?

Currently, the top_right menu automatically generates links to Open Data and Transparency, and it appears only Blog links can be editted via the front end. I want to be able to place whatever items I deem appropriate in that menu.

The main_navigation menu is automatically generated depending on the enabled features. I want to place items before and after those auto-generated feature links. 

### Description

*Type*: Feature

We'll use the existing "content blocks" feature to achieve our objective by doing the following:
* Make the top_links content block control the entire top right menu
* Add two new content block areas: main_navigation_menu_left, which will create items to the left of the auto-generated menu items and main_navigation_menu_right to the same to the right side of the menu.

### Current Behavior

Currently, the top_right menu automatically generates links to Open Data and Transparency, and it appears only Blog links can be editted via the front end. I want to be able to place whatever items I deem appropriate in that menu.

The main_navigation menu is automatically generated depending on the enabled features.

### Expected Behavior
 I want to be able to edit top_right and main_navigation menus by creating and editing use content_blocks.

### Steps to reproduce
NA

### Definition of Done
* Go to /admin/site_customization/content_blocks
* Create a content block for top_links and items there appear as top right menu
* Create a content block for main_navigation_menu_left and items there will appear to the left of the auto-generated menu items 
* Create a content block for main_navigation_menu_left and items there will appear to the left of the auto-generated menu items 

@ziyan-junaideen is working on this and will submit a PR for this feature shortly.

voodoorai2000:
Closed via https://github.com/consul/consul/pull/2785

PR #2785, opened 2018-07-24
Configurable Top Right and Main Navigation Menus

ziyan-junaideen:
# References

- Addresses issue: #2781 - Configurable Top Right and Main Navigation Menus

# Objectives

- Introduce extra content blocks  for main navigation (main_navigation_left, main_navigation_right)
- Auto hide auto generate `top_links` if a content_block for the section exists


voodoorai2000:
Hey @ziyan-junaideen!

This is great!
We are missing some specs though... all changes must be fully covered with specs.
Could you add them? 😌

app/views/shared/_subnavigation.html.erb
```diff
@@ -52,12 +54,17 @@
...
     <% end %>
-    <li>
-      <%= link_to t("layouts.header.help"),
-                    help_path,
-                    accesskey: "6",
-                    class: ("is-active" if current_page?(help_path)),
-                    title: t("shared.go_to_page") + t("layouts.header.help") %>
-    </li>
+
+    <% unless content_block("main_navigation_left", I18n.locale) || content_block("main_navigation_right", I18n.locale) %>
```
decabeza: Hi @ziyan-junaideen,

I don't understand why hide "Help" menu if there are a `content_block main_navigation_left or content_block main_navigation_right` 🤔

app/models/site_customization/content_block.rb
```diff
@@ -1,5 +1,5 @@
 class SiteCustomization::ContentBlock < ActiveRecord::Base
-  VALID_BLOCKS = %w(top_links footer)
+  VALID_BLOCKS = %w(top_links footer main_navigation_left main_navigation_right)
```
decabeza: What about switching to `subnavigation_left` and `subnavigation_right` to add consistency with `shared/_subnavigation.html.erb` name partial 😌

app/views/shared/_subnavigation.html.erb
ziyan-junaideen: I understand. I guess Devin is the best to answer the question. @devinbalkind ^^^

I think he is not part of the team. Will ping him on Slack.

app/models/site_customization/content_block.rb
ziyan-junaideen: I would agree. I noticed the partial names used in the other two. But I think again Devin was looking at the english wording that shows in the menu.

I think we can make the select option text be sent through i18n to get wording based on locale? That way the select list can be worded as organizations prefer.

ziyan-junaideen:
@voodoorai2000 The unit spec for content bock only uses the validation and uniqueness to locale. It doesn't concern individual keys. Thus I don't think there is a need for unit tests?

app/views/shared/_subnavigation.html.erb
devinbalkind: Is there a better way to hide "Help" than doing it through this method?

decabeza: I think is better remove this condition for now. If any fork wants to remove this link can create a custom view on `/views/custom/shared/_subnavigation.html.erb` and remove it. ☺️ 

Anyway, I take note of this to create a setting to "enable/disable" Help page and can use something like `<% if feature?(:help_page) %>` how the rest of the subnavigation links 😌

devinbalkind: Sounds good to me. Thank you.

app/models/site_customization/content_block.rb
```diff
@@ -1,5 +1,5 @@
 class SiteCustomization::ContentBlock < ActiveRecord::Base
-  VALID_BLOCKS = %w(top_links footer)
+  VALID_BLOCKS = %w(top_links footer subnavigation_left subnavigation_right)
```
ziyan-junaideen: @decabeza I made some changes in relation with the keys used for the content block. They are now named as you suggested (subnavigation_left, subnavigation_right).

app/views/admin/site_customization/content_blocks/_form.html.erb
```diff
@@ -17,7 +17,7 @@
   <div class="row">
     <div class="small-12 medium-6 column">
       <%= f.label :name %>
-      <%= f.select :name, SiteCustomization::ContentBlock::VALID_BLOCKS, label: false %>
+      <%= f.select :name, SiteCustomization::ContentBlock::VALID_BLOCKS.map { |key| [t("admin.site_customization.content_blocks.content_block.names.#{key}"), key] }, label: false %>
```
ziyan-junaideen: @decabeza while i named the keys as suggested, the orignial reason to why they were named as such was a language issue. I did some thing like this to pull the wording from the `admin.yml` file.

Will that be OK or shall I just include it in our fork?

app/models/site_customization/content_block.rb
decabeza: 👌

app/views/admin/site_customization/content_blocks/_form.html.erb
decabeza: It's OK @ziyan-junaideen 😉 

Run `i18n-tasks health`, maybe `admin.site_customization.content_blocks.content_block.names.` appears as unused one. Sometimes there is to add dynamic keys to `config/i18n-tasks.yml` under `ignore_missing` or `ignore_unused` 😌

ziyan-junaideen: I am not sure what you are saying. Could you elaborate. The `admin.site_customization.content_blocks.content_block.names` is used to give humanized text to users.

How ever I only updated the `en` file. One reason is I am not sure if it will need changes. The other is English is the only language I am familiar with and will have to use Google to translate the text.

    Forest (en) has 2951 keys in total. On average, values are 25 characters long, keys have 3.9 segments.
    ✓ Perfect! No translations are missing.
    ✓ Well done! Every translation is in use.
    The following data requires normalization:
    config/locales/en/general.yml
    config/locales/en/activerecord.yml
    config/locales/en/activemodel.yml
    config/locales/en/admin.yml
    config/locales/en/moderation.yml
    config/locales/en/valuation.yml
    config/locales/en/management.yml
    config/locales/en/verification.yml
    config/locales/en/mailers.yml
    config/locales/en/pages.yml
    config/locales/en/devise_views.yml
    config/locales/en/responders.yml
    config/locales/en/kaminari.yml
    config/locales/en/officing.yml
    config/locales/en/budgets.yml
    config/locales/en/legislation.yml
    config/locales/en/community.yml
    config/locales/en/documents.yml
    config/locales/en/images.yml
    config/locales/en/guides.yml
    Run `i18n-tasks normalize` to fix

decabeza: Sometimes if you use an i18n key generated dynamically `content_blocks.content_block.names.#{key}` i18n health thinks there unused keys, but it seems is not the case.

Could you please include this keys on `es/admin.yml`? (we always include this two locales, don't worry for use Google translate) 😉 

```yml
names:
  top_links: Enlaces superiores
  footer: Pie de página
  subnavigation_left: Navegación principal izquierda 
  subnavigation_right: Navegación principal derecha
```

Thank you very much again! 🙏

ziyan-junaideen: I thought it was the case when i checked the libs docs. I saw a flag to enable dynamic keys. It will support up to one wild card key.

I will add the keys.

voodoorai2000:
Sure, this looks good @ziyan-junaideen 👍

PR #2797, opened 2018-07-26
Fix order-dependent spec

mlovic:
References
===================
 - Fixes tests from: https://github.com/consul/consul/pull/2723
 - Example of test failure: https://travis-ci.org/wairbut-m2c/consul/jobs/408398729

Notes
=============
> Due to how DB IDs are assigned with AUTO INCREMENT, tests
> cannot assume that any specific ID will be available


mlovic:
Travis failure is due to the spec fixed by https://github.com/consul/consul/pull/2788

Issue #2812, opened 2018-07-28
Admin menu doesn't show up for some admins


javierv:
## Steps to reproduce

1. Create a user with an ID higher than 10.
2. Log in with that user.
3. Go to the "Management" section.

## Expected behaviour

There should be a menu showing links to sections like "My activity" or "My account":

![expected](https://user-images.githubusercontent.com/348557/43359933-e611b012-929b-11e8-8f0b-02c083b7aea4.png)

## Actual behaviour

The menu doesn't show up:

![actual](https://user-images.githubusercontent.com/348557/43359934-ea89cdfa-929b-11e8-924b-338d88802c9f.png)

voodoorai2000:
Closed via https://github.com/consul/consul/pull/2813

PR #2813, opened 2018-07-28
Fix admin menu not showing for high IDs

javierv:
# References

* Issue #2812.
* Issue AyuntamientoMadrid#1591
* [Travis build 24150, job 2](https://travis-ci.org/consul/consul/jobs/408923975)

# Objectives

Fix a bug preventing an admin from seeing the admin menu when their user ID is 10 or higher.

The query searching for the user was:

```ruby
User.find_by(id: session[:manager]["login"].last(1))
```

That line worked when `session[:manager]["login"]` had a value such as `"admin_user_1"`, but failed when it had a value like `"admin_user_76"`, since it only took the last character of the ID.

Taking everything after the last underscore instead of just the last character solves the issue.

# Notes

* Unless I've overlooked something, version 0.16 is **not** affected.
* As seen in the referenced travis build, there already was a failing test covering this issue. Since the test sometimes passed if the user's ID was less than 10, I've added a model test too. Now Travis' build should be green again (except when it's red :sweat_smile:).
* This solution is still fragile. Currently it always works (or so I think), but it assumes the manager login will always end with an underscore followed by the ID. Given the way we handle `session[:manager]`, that might not always be the case in the future. Suggestions are welcome! :wink:.

### DIFF
--- a/app/controllers/management/base_controller.rb
+++ b/app/controllers/management/base_controller.rb
@@ -56,7 +56,7 @@ class Management::BaseController < ActionController::Base
 
     def manager_logged_in
       if current_manager
-        @manager_logged_in = User.find_by(id: session[:manager]["login"].last(1))
+        @manager_logged_in = User.find_by_manager_login(session[:manager]["login"])
       end
     end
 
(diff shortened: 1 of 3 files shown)


voodoorai2000:
Awesome @javierm nice catch!

PR #2565, opened 2018-03-29
Upgrade Ruby to 2.3.7

bertocq:
References
==========
https://www.ruby-lang.org/en/news/2018/03/28/ruby-2-3-7-released/

Objectives
==========
Keep up with latest Ruby version for security & performance reasons

Visual Changes (if any)
=======================
None

Notes
=====================
I would test this in a fork (like Madrid's) before merging the PR, even if Travis gets full ✅

PR #2563, opened 2018-03-28
Update Dockerfile to use Headless Chrome

aitbw:
References
==========
* Related PR: #2534

Objectives
==========
* Update Dockerfile to use Headless Chrome for E2E integration tests

Visual Changes (if any)
=======================
None

Notes
=====================
None

PR #2818, opened 2018-07-30
Rollback chromedriver version in Docker image

mlovic:
References
===================
 - https://github.com/consul/consul/pull/2563

Objectives
===================
Rollback chromedriver version in Docker image to a version compatible
with the version of chromium used.

Commit message:

> Our Docker image uses Chromium v57, which is not supported by
> chromedriver versions after 2.38. See
> https://bugs.chromium.org/p/chromedriver/issues/detail?id=2445
> 
> The reason that the version of chromium is so old is that our base
> image ruby:2.3.6 (the standard image for Ruby 2.3.6 on DockerHub) is
> built on Debian 8 (jessie), which was made obsolete over a year ago.
> 
> The best solution is probably to upgrade the application to Ruby 2.3.7
> (if not later), so we can use the official ruby:2.3.7 image, which is
> kept up-to-date (it has Chromium 68).
> 
> Alternatively, we could try to install Chromium from a more up-to-date
> repository, but this is probably not worth the trouble.

There is already an open PR to upgrade to Ruby 2.3.7: #2565

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,7 +43,7 @@ RUN bundle install --full-index
 
 # Install Chromium and ChromeDriver for E2E integration tests
 RUN apt-get update -qq && apt-get install -y chromium
-RUN wget -N http://chromedriver.storage.googleapis.com/2.40/chromedriver_linux64.zip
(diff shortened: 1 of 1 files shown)


voodoorai2000:
Appreciate it @mlovic 👌

Issue #2772, opened 2018-07-17
Adds "Follow community" on proposals


decabeza:
Background
==
Now all proposals have their community on `/communities/N`. When a user creates a new topic or comments on a topic is added to the participants list.

What we need
==

- Add "Follow community" button on `/communities/N` (see _screenshot_1_)

- The participants should be **only the users who are following the community**.

- If user create a topic or comment on existing one but is no followinf the community ***shouldn't appear** on participants list.

Notes
==
There is a similar feature "Follow citizen proposal" in `proposals/N`.

Screenshots
==

**Screenshot_1**

![follow_community](https://user-images.githubusercontent.com/631897/42809860-db59772e-89b6-11e8-8172-93c21db0b1ee.png)

hpissei:
Hello, @decabeza I am new to ROR and open source I would like to work on this work item.

decabeza:
Hi @hpissei! 

Of course! All yours, thanks! 👌  😌

hpissei:
Thank You Very Much

PR #2825, opened 2018-07-31
follow button added to communities page 

hpissei2:
Objectives
===================
1.Add "Follow community" button on /communities/N (see screenshot_1)
2.The participants should be only the users who are following the community.
3.If user create a topic or comment on existing one but is no followinf the community *shouldn't appear on participants list.

References
===================

* Closes https://github.com/consul/consul/issues/2772

Visual Changes
===================
42809860-db59772e-89b6-11e8-8172-93c21db0b1ee

### DIFF
--- a/app/models/community.rb
+++ b/app/models/community.rb
@@ -4,9 +4,7 @@ class Community < ActiveRecord::Base
   has_many :topics
 
   def participants
-    users_participants = users_who_commented +
-                         users_who_topics_author +
-                         author_from_community
+    users_participants = users_who_follow_community
     users_participants.uniq
   end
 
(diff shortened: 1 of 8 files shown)


javierm:
Hi, @hpissei2 :smile:. First of all, thank you for this pull request :pray:, and sorry it took us more than three years to reply.

I've just updated the original issue saying that the requirements need to be changed in order to make them compatible with recent changes we've done to the way we display what users are following. The mentioned changes have been done due to privacy issues.

So I guess this pull request doesn't much sense anymore. Sorry for that!

Since you took time to develop it, I've taken some time to review it in order to return the favor :thinking:. Please note that the following comments are just in case you were interested in a review (you probably aren't, but I can't think of anything else I can do for you); please don't work on fixing them because, as mentioned, the pull request doesn't make much sense anymore.

1. Adding tests is a great way to make sure the code works as expected :smile:
2. The method `users_who_follow_community` seems to return users who follow anything, and not necessarily the community :thinking:; I guess we could add `include Followable` to the model and then return `followers` instead?
3. The methods `users_who_commented`, `users_who_topics_author`, `author_from_community` and `by_authors` weren't used anymore :broom:

As I said earlier, feel free to ignore this message completely, and once again sorry for the really late review :pray:.

Issue #2735, opened 2018-07-06
Make polls translatable


decabeza:
Background
==

Now on admin is possible to create content in some languages. You can try this feature on `/admin/budgets/N/budget_investments/N/budget_investment_milestones/new`.

What we need
==

Make translatable the **polls** section (including questions and answers) created from admin panel on:

- `/admin/polls/new` 
- `/admin/questions/new` 
- `/admin/questions/N/answers/new` 

Screenshots
==

<img width="1241" alt="translate_poll" src="https://user-images.githubusercontent.com/631897/42386224-2980318e-813f-11e8-81b6-62ba77dd3274.png">
<img width="1245" alt="translate_question" src="https://user-images.githubusercontent.com/631897/42386225-299e98ae-813f-11e8-93ec-1a0b7b2068d4.png">
<img width="1251" alt="translate_answer" src="https://user-images.githubusercontent.com/631897/42386226-29b7b3ac-813f-11e8-9e0e-9dcae92f82dc.png">

Notes
==

You can use as reference the PR [Allow admin generated content to be translatable](https://github.com/consul/consul/pull/2619).

microweb10:
I can work on this one

microweb10:
@voodoorai2000 I was going to use [this PR](https://github.com/consul/consul/pull/2619) as a reference, as the description says. But I will also have a look at the one you mentioned. What's the difference? As I understood the PR in this repo is a backport from AyuntamientoMadrid repo.

voodoorai2000:
Yes 😌

The originally referenced [PR](https://github.com/consul/consul/pull/2619) is the one that introduced into CONSUL Globalization: gems, javascript, views, etc

This other [PR](https://github.com/AyuntamientoMadrid/consul/pull/1585) uses the Globalization implementation and just applies it to a new section 👌

microweb10:
Oh, yeah! No worries 😉
I realized about that when I had a look at the PR. The rest was very helpful to "understand" what it does BTW 😊

voodoorai2000:
Glad to hear 😊👍

microweb10:
@voodoorai2000 I have a couple of questions:

- I see that Spanish translation for polls is currently "Votaciones". Shouldn't it be "Encuestas"?

[...]

PR #2833, opened 2018-08-02
Make polls translatable

microweb10:
References
===================
Issue https://github.com/consul/consul/issues/2735

Visual Changes
===================
captura de pantalla 2018-08-04 a las 12 56 25
captura de pantalla 2018-08-04 a las 12 56 51
captura de pantalla 2018-08-04 a las 12 58 09
captura de pantalla 2018-08-04 a las 12 58 41
captura de pantalla 2018-08-04 a las 12 59 09
captura de pantalla 2018-08-04 a las 13 00 37

Notes
===================
closes https://github.com/consul/consul/issues/2735


aitbw:
Hi @microweb10 ! Thanks for this PR :tada:

We'll review this pull request as soon as possible and if all goes well, it'll be merged into `master` :smile_cat:

Nevertheless, I'd like you to address two things first:

1) Could you include screenshots/GIFs of how this looks/behaves? This is extremely useful for those folks who read the whole changelog and like to have a quick glimpse of the latest features

2) Can you rebase this with `master`? This way we can avoid that unnecessary extra `merge` commit :relieved:

Let me know what you think!

mlovic:
Hi @microweb10, thanks for the contribution! Nothing is being merged in August, but I think by early September this should be reviewed and merged.

I just want to let you know that we've made some changes to the code shared by all admin-translatable resources (https://github.com/AyuntamientoMadrid/consul/pull/1613). Some of these changes will conflict with yours, but they shouldn't be too hard to resolve. See https://gist.github.com/mlovic/c7c4575c38a6c814032f05bb1d016669 for an example of how to adapt to the new code.

app/views/admin/poll/questions/_form.html.erb
```diff
@@ -1,18 +1,36 @@
...
+    <%= hidden_field_tag "delete_translations[#{locale}]", 0 %>
+	<% end %>
+
   <%= f.hidden_field :proposal_id %>
 
   <div class="small-12">
     <div class="small-12 medium-6 large-4">
+      <% select_options = Poll.joins(:translations)
+                              .where('poll_translations.locale = ?', Globalize.locale)
+                              .pluck("poll_translations.name", :id) %>
```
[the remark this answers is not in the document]
microweb10: OK, I will use `Poll.all.map...`
I didn't know that was the desired behavior, for me it didn't make sense to create translated questions/answers for a not translated poll 😅

db/dev_seeds/polls.rb
```diff
@@ -28,19 +28,49 @@
               ends_at:   1.month.ago,
               results_enabled: true,
               stats_enabled: true)
+
+  Poll.all.each do |poll, i|
```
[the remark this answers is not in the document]
microweb10: You are right, I think I wanted to use it at first but at the end, I didn't and I forgot to remove it. Thanks!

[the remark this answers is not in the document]
aitbw: @microweb10 I'd also like to add that you should use `Poll.find_each` instead of `Poll.all.each` as the latter can exceed the amount of memory available if said model is associated with a table with a considerable amount of records, as mentioned [here](https://guides.rubyonrails.org/v4.2/active_record_querying.html#retrieving-multiple-objects-in-batches)

app/views/admin/poll/questions/_form.html.erb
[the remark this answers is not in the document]
mlovic: That's a good point, but the current locale of the application while the admin is using it does not necessarily correspond to the languages that the polls are translated to (even though in practice this is probably unlikely). For example, an admin might be trying to add French translations to a Spanish-by-default application, but have his own user session set to English. In this case, the select options will be rendered in English, even though neither the polls nor its questions are translated to English.

The spec examples I wrote cover this edge case. Could you include them (or something similar) in the PR?

I could still be missing something, so do let me know if you think anything doesn't make sense. Thanks!

app/controllers/admin/poll/polls_controller.rb
```diff
@@ -74,4 +75,13 @@ def load_search
       @search = search_params[:search]
     end
 
+    def resource
+      @poll = Poll.find(params[:id]) unless @poll
+      @poll
```
aitbw: Is this line necessary? As far as I know, Ruby always returns the value of the last statement but if you can provide some insight @microweb10, that'd be great :smile_cat:

app/controllers/admin/poll/questions/answers_controller.rb
```diff
@@ -49,11 +51,20 @@ def order_answers
...
+      params.require(:poll_question_answer).permit(*attributes, *translation_params(params[:poll_question_answer]))
     end
 
     def load_answer
       @answer = ::Poll::Question::Answer.find(params[:id] || params[:answer_id])
     end
 
+    def resource
+      load_answer unless @answer
+      @answer
```
aitbw: Is this line necessary? As far as I know, Ruby always returns the value of the last statement but if you can provide some insight @microweb10, that'd be great :smile_cat:

app/controllers/admin/poll/questions_controller.rb
```diff
@@ -55,11 +56,21 @@ def destroy
...
+      params.require(:poll_question).permit(*attributes, *translation_params(params[:poll_question]))
     end
 
     def search_params
       params.permit(:poll_id, :search)
     end
 
+    def resource
+      @poll_question = Poll::Question.find(params[:id]) unless @poll_question
+      @poll_question
```
aitbw: Is this line necessary? As far as I know, Ruby always returns the value of the last statement but if you can provide some insight @microweb10, that'd be great :smile_cat:

app/views/admin/poll/questions/_form.html.erb
```diff
@@ -1,18 +1,34 @@
...
 
+	<% @question.globalize_locales.each do |locale| %>
+    <%= hidden_field_tag "delete_translations[#{locale}]", 0 %>
+	<% end %>
+
   <%= f.hidden_field :proposal_id %>
 
   <div class="small-12">
     <div class="small-12 medium-6 large-4">
+      <% select_options = Poll.all.map { |p| [p.name, p.id] } %>
```
microweb10: We will use `Poll.all.map...` instead of `Poll.pluck` cause we want to make use of [the fallbacks set for i18n](https://github.com/consul/consul/blob/master/config/application.rb#L24) for missing translations

app/views/admin/poll/questions/_form.html.erb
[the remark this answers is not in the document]
microweb10: Yeah, I will include them

db/dev_seeds/polls.rb
```diff
@@ -28,19 +28,49 @@
               ends_at:   1.month.ago,
               results_enabled: true,
               stats_enabled: true)
+
+  Poll.all.each do |poll|
```
aitbw: Avoid using `Poll.all.each` as it can exceed the amount of memory available if said model has an associated table with a considerable amount of records, as shown [here](https://guides.rubyonrails.org/v4.2/active_record_querying.html#retrieving-multiple-objects-in-batches)

app/views/admin/poll/questions/_form.html.erb
aitbw: :thinking: I think [this comment](https://github.com/consul/consul/pull/2833#discussion_r215298180) can also be applied in here. What do you think, @microweb10 ?

app/controllers/admin/poll/polls_controller.rb
microweb10: Well, it will not be necessary if you don't use `unless @poll`. But, what will happen if `@poll` was already defined? `@poll = Poll.find(params[:id])` will never be executed and the method `resource` can return something different to `@poll`. Am I wrong?

app/controllers/admin/poll/questions/answers_controller.rb
mlovic: Sorry, @microweb10 may have copied that pattern from my banners PR.

The reason I explicitly set the return value is to avoid depending on the return value of `load_answer`, whose main purpose is to change state. I think it would be easy for someone to add more state-changing code and inadvertently change the return value, e.g.:

```ruby
def load_answer
  @answer = ::Poll::Question::Answer.find(params[:id] || params[:answer_id])
  Rails.logger.debug "Loaded answer #{@id.answer}"
end
```

app/controllers/admin/poll/polls_controller.rb
aitbw: Oh yes, you're absolutely right! Sorry about the inconvenience! :bowing_man:

app/controllers/admin/poll/questions_controller.rb
aitbw: Resolved as per [this comment](https://github.com/consul/consul/pull/2833#discussion_r215309891)

app/controllers/admin/poll/questions/answers_controller.rb
aitbw: That is correct, @mlovic. Sorry about the inconvenience. As always, thanks for the detailed explanation! :bowing_man:

db/dev_seeds/polls.rb
microweb10: In this case is not a problem. This is the seeds file and we don't have thousands of records, but anyway I will update it. It's never a bad idea to use the "Best Practices" 😉

microweb10: Resolved as per commit https://github.com/consul/consul/pull/2833/commits/723eacff0f412bc1d7641c3db1106dedccda6dde

app/views/admin/poll/questions/_form.html.erb
microweb10: Not the same at all. 
Here we don't iterate over (`each`). 
We actually need to `map`, which involves storing in memory (`array`) the values 😊

app/controllers/admin/poll/polls_controller.rb
javierm: What about `@poll ||= Poll.find(params[:id])`?

aitbw: I thought about it @javierm but wouldn't that approach avoid fetching a poll with a different ID? :thinking:

javierm: Mmm... I thought `@poll ||= Poll.find(params[:id])` was (almost) exactly equivalent to `Poll.find(params[:id]) unless @poll`, the only difference being it returned `@poll` if already defined. Is there something I'm missing :thinking:?

microweb10: Yeah, exactly the same. I thought about that, but I just wanted to use the same patterns I've seen in other controllers. I thought it was maybe more readable to other users. But, personally, I like it more. Changed!

app/controllers/admin/poll/questions_controller.rb
aitbw: @microweb10 As per commit 27b90cb, I think you could apply the same patch in here :ok_hand:

microweb10: 👍 https://github.com/consul/consul/pull/2833/commits/7373a41a2868c982c3caea924378091d4052b890

db/dev_seeds/polls.rb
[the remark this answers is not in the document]
microweb10: resolved https://github.com/consul/consul/pull/2833/commits/dc01291ad26ae5189f421a4a915ab9a6353f611a

app/views/admin/poll/questions/_form.html.erb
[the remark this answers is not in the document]
microweb10: resolved https://github.com/consul/consul/pull/2833/commits/900ca790cd522cac714c6e8440277c9b152c0dfd

javierm:
Hi, @microweb10 Thank you so much for working on this issue :heart:! Please excuse us for being a bit late implementing related changes :pray:, mainly due to summer holidays :beach_umbrella:.

We've been working on refactoring our translatable sections in order to make the process easier for new sections. Would it be possible for you to rebase this branch against current master and check implementations like commit AyuntamientoMadrid/consul@efc4e37 or commit AyuntamientoMadrid/consul@188c2bc as examples of the current (hopefully simplified) code we change in order to make a section translatable?

Thank you very much!

microweb10:
@javierm I've made the changes. I hope it's OK. Let me know if you need anything else.

javierm:
Thanks again @microweb10!

It looks like there's a small problem with the rebasing and this pull request now shows 199 commits, most of them unrelated to making polls translatable. I'm not sure why; could it be that you're trying to rebase against an old version of the `master` branch?

If there's something we can help with, ask without hesitation :wink:.

microweb10:
@javierm I have no idea why neither 😔. I don't think that's the reason cause I fetched upstream first and merge into my master. If you have a look at [my master branch](https://github.com/microweb10/consul) you will see that it has the same commits than [the master branch in consul](https://github.com/consul/consul). Then I did the rebase against that master branch, which should be up to date. Although I'm not an expert with rebase command. I also don't know why the migration error in travis. It's not related to my code neither.

microweb10:
I don't know if it's a good idea, but I could create a new PR and we could discard this one

javierm:
@microweb10 I can't figure out what's going so, so I think a new pull request would be great! Sorry we've forced you to do so many adjustments on this one! And thank you so much for your undestanding :relieved:!

mlovic:
That _is_ strange. The branch looks fine to me too. And if you use the GitHub ["compare" page](https://github.com/consul/consul/compare/master...microweb10:issue-2735) everything looks as expected. It even tells you that this PR exists with the same "diff".

PR #2840, opened 2018-08-07
Fix management helper when user id has multiple digits

mlovic:
References
===================
 -  Bug introduced in #2137

Objectives
===================
Fix helper in `Management::BaseController`:

> The old code was only reading the last digit of the id.
> This caused failures when the id was >9. Should also
> fix `spec/features/management_spec.rb:15`, which would fail
> occasionally because of this.

Should also prevent all of the recent failures from `spec/features/management_spec.rb:15` (e.g. https://travis-ci.org/wairbut-m2c-aytomadrid/consul/jobs/412676445)

Notes
====================
This change was also pulled into AyuntamientoMadrid/consul, so it should be ported there as well.

### DIFF
--- a/app/controllers/management/base_controller.rb
+++ b/app/controllers/management/base_controller.rb
@@ -56,7 +56,8 @@ class Management::BaseController < ActionController::Base
 
     def manager_logged_in
       if current_manager
-        @manager_logged_in = User.find_by(id: session[:manager]["login"].last(1))
+        user_id = session[:manager]["login"].split('_').last.to_i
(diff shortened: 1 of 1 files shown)


aitbw:
I think this is a duplicate for #2813 :thinking: can you confirm that, @mlovic?

mlovic:
Oh yeah, you're right... I tried searching for open PRs/issues before looking into the test failure but I didn't read that one.

Oh well. Closing this then.

PR #2844, opened 2018-08-13
Fix a misleading comment

PierreMesure:
References
===================
No reference

Objectives
===================
Fixes a comment that had become wrong due to changes in the code.

Visual Changes
===================
No change whatsoever

Notes
===================
Don't know if that's big enough to have its own PR but I didn't want to include it in a non-related one.


voodoorai2000:
Appreciate it @PierreMesure 👌

Issue #2755, opened 2018-07-11
Improve admin banners section


decabeza:
⚠️ Wait https://github.com/consul/consul/pull/2319/ is merged before to start this issue.

Background
==

Now the admin users can create and manage banners on `/admin/banners`.

What we need
==

Improve this section with following options _[Screenshot_1]_:

- Adds the checkbox to show banner on **Polls**.
- Adds the checkbox to show banner on **Legislation processes**.
- Adds the checkbox to show banner on **each heading** of  participatory budgeting.
- Adds a select for banner position (top or bottom).

Notes
==
If the banner position is **top** show before filters _[Screenshot_2]_
If the banner position is **botom** show before pagination _[Screenshot_3]_

Screenshots
==
**[Screenshot_1]**
![admin_banners](https://user-images.githubusercontent.com/631897/42566114-c70d4e54-8505-11e8-8608-278386003c80.png)

**[Screenshot_2]**
![top](https://user-images.githubusercontent.com/631897/42566208-277fb3e4-8506-11e8-9a9a-af6db402dae8.png)

**[Screenshot_3]**
![bottom](https://user-images.githubusercontent.com/631897/42566211-2a0b61f8-8506-11e8-9821-8ac8d6ebd9bc.png)

papayalabs:
Working on this issue

PR #2847, opened 2018-08-14
2755 improve admin banners section

papayalabs:
References
===================
Issue [#2755](https://github.com/consul/consul/issues/2755)

Objectives
===================
Improve this section with following options:

    Adds the checkbox to show banner on Polls.
    Adds the checkbox to show banner on Legislation processes.
    Adds the checkbox to show banner on each heading of participatory budgeting.
    Adds a select for banner position (top or bottom).

Visual Changes
===================
screen shot 2018-08-14 at 1 17 34 pm
screen shot 2018-08-14 at 1 06 03 pm
screen shot 2018-08-14 at 1 06 16 pm
screen shot 2018-08-14 at 1 06 25 pm
screen shot 2018-08-14 at 1 06 32 pm
screen shot 2018-08-14 at 1 06 39 pm
screen shot 2018-08-14 at 1 06 45 pm
screen shot 2018-08-14 at 1 06 50 pm
screen shot 2018-08-14 at 1 07 06 pm
screen shot 2018-08-14 at 1 07 52 pm

Notes
===================
The basic changes of this functionality are in banners_helper.rb where we add methods for banner_position and for display banners in headings. Headings values are not static, they change when users add new headings or change existings ones. We save this values in an array in banner model ( using serialize ) and display checking the values of the array.

Here is some of the code:

In banner model:

    BANNER_POSITIONS = %w(Top Bottom).freeze
    serialize :headings

In banners_helper.rb

    def is_banner_on_top?
      @banners.sample.banner_position == "Top"
    end
  
    def is_banner_on_bottom?
      @banners.sample.banner_position == "Bottom"
    end
  
    def is_banner_heading_on_top?(heading_id)
      if (@banners.sample.headings != nil) && (@banners.sample.headings.include? heading_id.to_s)
        @banners.sample.banner_position == "Top"
      else
        false
      end  
    end
  
    def is_banner_heading_on_bottom?(heading_id)
      if (@banners.sample.headings != nil) && (@banners.sample.headings.include? heading_id.to_s)
        @banners.sample.banner_position == "Bottom"
      else
        false
      end  
    end

In banner_form:

    <%= f.label :sections, t("admin.banners.banner.sections_label") %>
      <%= f.collection_check_boxes(:web_section_ids, @banner_sections, :id, :name) do |b| %>
        <%= b.check_box %> <%= t("admin.banners.banner.sections.#{b.text}") %><br> 
		<% if b.text == 'budgets'%>
		  <% Budget::Group.all.each do |group| %>
		    <% if group.budget.phase != 'finished' %>
		      <span style="display:inline-block; width: 20px;"></span><%= group.name %><br>
			  <% group.headings.each do |heading| %>
		        <span style="display:inline-block; width: 40px;"></span>
	  		    <% if (@banner.headings!=nil) && (@banner.headings.include? heading.id.to_s) %>	  
	              <%= check_box_tag "banner_headings[]",heading.id, checked = true, id: "banner_headings_#{heading.id}" %> <%= heading.name %><br>
	  		    <% else %>
	              <%= check_box_tag "banner_headings[]",heading.id, checked = false, id: "banner_headings_#{heading.id}" %> <%= heading.name %><br>		  
	  		    <% end %>
			  <% end %>	
			<% end %>  
		  <% end %>	
		<% end %> 
          <% end %>

### DIFF
--- a/app/helpers/banners_helper.rb
+++ b/app/helpers/banners_helper.rb
@@ -5,7 +5,7 @@ module BannersHelper
   end
   
   def is_banner_on_top?
-    @banners.sample.banner_position == "Top" || @banners.sample.banner_position.nil?
+    @banners.sample.banner_position == "Top"
   end
   
   def is_banner_on_bottom?
@@ -14,7 +14,7 @@ module BannersHelper
   
   def is_banner_heading_on_top?(heading_id)
     if (@banners.sample.headings != nil) && (@banners.sample.headings.include? heading_id.to_s)
-      @banners.sample.banner_position == "Top" || @banners.sample.banner_position.nil?
+      @banners.sample.banner_position == "Top"
     else
       false
     end  
--- a/app/helpers/banners_helper.rb
+++ b/app/helpers/banners_helper.rb
@@ -5,7 +5,7 @@ module BannersHelper
   end
   
   def is_banner_on_top?
-    @banners.sample.banner_position == "Top"
+    @banners.sample.banner_position == "Top" || @banners.sample.banner_position.nil?
   end
   
   def is_banner_on_bottom?
--- a/app/helpers/banners_helper.rb
+++ b/app/helpers/banners_helper.rb
@@ -10,6 +10,22 @@ module BannersHelper
   
   def is_banner_on_bottom?
     @banners.sample.banner_position == "Bottom"
-  end    
-
+  end
+  
+  def is_banner_heading_on_top?(heading_id)
+    if (@banners.sample.headings != nil) && (@banners.sample.headings.include? heading_id.to_s)
+      @banners.sample.banner_position == "Top" || @banners.sample.banner_position.nil?
+    else
+      false
+    end  
+  end
+  
+  def is_banner_heading_on_bottom?(heading_id)
+    if (@banners.sample.headings != nil) && (@banners.sample.headings.include? heading_id.to_s)
+      @banners.sample.banner_position == "Bottom"
+    else
+      false
+    end  
+  end
+       
 end
--- a/app/helpers/banners_helper.rb
+++ b/app/helpers/banners_helper.rb
@@ -3,5 +3,13 @@ module BannersHelper
   def has_banners?
     @banners.present? && @banners.count > 0
   end
+  
+  def is_banner_on_top?
+    @banners.sample.banner_position == "Top"
+  end
+  
+  def is_banner_on_bottom?
+    @banners.sample.banner_position == "Bottom"
+  end    
 
 end
--- a/app/views/admin/banners/_form.html.erb
+++ b/app/views/admin/banners/_form.html.erb
@@ -54,19 +54,22 @@
     <div class="small-12 column">
       <%= f.label :sections, t("admin.banners.banner.sections_label") %>
       <%= f.collection_check_boxes(:web_section_ids, @banner_sections, :id, :name) do |b| %>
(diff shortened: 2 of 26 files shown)


app/views/admin/banners/_form.html.erb
```diff
@@ -54,7 +54,22 @@
...
-        <%= b.check_box %> <%= t("admin.banners.banner.sections.#{b.text}") %><br>
+        <%= b.check_box %> <%= t("admin.banners.banner.sections.#{b.text}") %><br> 
+		<% if b.text == 'budgets'%>
+		  <% Budget::Group.all.each do |group| %>
+		    <% if group.budget.phase != 'finished' %>
+		      <span style="display:inline-block; width: 20px;"></span><%= group.name %><br>
+			  <% group.headings.each do |heading| %>
+		        <span style="display:inline-block; width: 40px;"></span>
+	  		    <% if (@banner.headings!=nil) && (@banner.headings.include? heading.id.to_s) %>	  
+	              <%= check_box_tag "banner_headings[]",heading.id, checked = true, id: "banner_headings_#{heading.id}" %> <%= heading.name %><br>
```
javierm: Hi @papayalabs. Using `serialize :headings` is a very interesting idea. However, IMHO its current implementation is making the code a bit too complicated. Ideally the view would be simple and the `banner_headings` parameters would be handled automatically with something like:

```erb
<%= f.collection_check_boxes(:budget_headings) # Probably more parameters are needed in this call %>
```
If something like this is possible using `serialize`, go for it. If not, it'd probably be better to have a `has_many :budget_headings, through: :budget_sections` relationship in the `Banner` model.

app/helpers/banners_helper.rb
```diff
@@ -3,5 +3,29 @@ module BannersHelper
...
+  def is_banner_on_top?
+    @banners.sample.banner_position == "Top"
+  end
+  
+  def is_banner_on_bottom?
+    @banners.sample.banner_position == "Bottom"
+  end
+  
+  def is_banner_heading_on_top?(heading_id)
+    if (@banners.sample.headings != nil) && (@banners.sample.headings.include? heading_id.to_s)
```
javierm: Why are we using `sample`? Are we assuming all banners have the same headings and the same position?

app/views/admin/banners/_form.html.erb
```diff
@@ -54,7 +54,22 @@
...
+	  		    <% else %>
+	              <%= check_box_tag "banner_headings[]",heading.id, checked = false, id: "banner_headings_#{heading.id}" %> <%= heading.name %><br>		  
+	  		    <% end %>
+			  <% end %>	
+			<% end %>  
+		  <% end %>	
+		<% end %>
```
javierm: I think we always use two spaces for indentation; never tabs, and never four spaces. It would be great if you could review this pull request and adjust the indentations as well as trailing whitespaces and newline characters at the end of a file :+1:.

app/views/budgets/index.html.erb
```diff
@@ -176,6 +176,10 @@
   </div>
 <% end %>
 
+<% if has_banners? && is_banner_on_bottom? %>
+  <%= render "shared/banner" %>
+<% end %>
```
javierm: This block is repeated in many places. Would it be possible to reduce duplication by using maybe a partial or a helper, or by replacing the current logic with `content_for`? Probably a partial is the easiest way:
```erb
<%= render "shared/bottom_banners" %>
```
Then we could do something similar for the top banners.

Issue #2850, opened 2018-08-15
Error 500 when creating a budget investment with S3


PierreMesure:
### Description

[comment]: # (Feature or Bug? eg. Type: Bug)
*Type*: Bug

When using AWS S3 to store external files, an error 500 appears when creating a budget investment with an image. This is due to the variable `investment` not being defined inside the following method:

```
def image_absolute_url(image, version)
  return "" unless image
  if Paperclip::Attachment.default_options[:storage] == :filesystem
    URI(request.url) + image.attachment.url(version)
  else
    investment.image_url(version)
  end
end
```

### Current Behavior
[comment]: # (Describe what actually happened.)

The budget investment is created but an error message appears full page.

### Expected Behavior
[comment]: # (Describe what you expected to happen.)

No error message

### Steps to reproduce
[comment]: # (Describe how we can replicate the bug step by step.)

- configure Consul with AWS S3 as an external storage system for Paperclip (should work with any other)
- create a budget investment with an image

### Definition of Done
[comment]: # (Any other information that would be useful, bullets are helpful.)

The fix is really easy. However, it seems that no one uses Consul with an external storage system right now, which explains why this bug can slip in. Adding specs to the uploader feature to test it in these conditions seems necessary.

PR #2851, opened 2018-08-15
Fix for error 500 when using an external storage system

PierreMesure:
References
===================
#2850

Objectives
===================
- Fixes a bug when creating a budget investment with an image stored on an external storage system

Visual Changes
===================
No visual changes


aitbw:
Hey @PierreMesure, care to edit the PR's title with something more descriptive like `Fix for error 500 when using an external storage system`? Same with the commit message

Whilst it might seem something minor, it actually is very useful for CONSUL's changelogs or if we need to backtrace something through Git history :smile_cat:

Let me know what you think!

PierreMesure:
I haven't had time to write specs for this fix but I think it'd be good to merge it anyway as it fixes a bug that I'm probably not the only one to have.

javierm:
@PierreMesure Sorry it took 3 years to finally have a look at this :pray: :sweat_smile:.

We're now migrating to Active Storage and so are having a look at everything related to storing documents and images. Since we've refactored some of the code, the fix is now easier, so we're closing this pull request in favor of #4701 (we're keeping you as co-author of the changes). Thank you very much for bringing up this issue and a solution! :pray:

Issue #2495, opened 2018-02-28
Improves Admin pages editor


decabeza:
Background
==

Now in the Admin Panel, there is a section to create pages `/admin/site_customization/pages` and we're having problems with the content that it generates.

![ck_editor](https://user-images.githubusercontent.com/631897/36801742-2ddd81d2-1cb3-11e8-96f3-f31542a313e5.png)

In `/views/pages/custom_page.html.erb` the content is rendered with 
`<%= text_with_links @custom_page.content %>` but doesn't allow the user to insert headings or links.

What we need
==

Replace **CK Editor** with another tool or update it to [CK Editor 5](https://github.com/ckeditor/ckeditor5) to change the configuration and allow:
- Insert headings
- Insert links
- Optional but wonderful: Insert images

The resulting content should have a correct HTML structure and format.

Screenshot
==
### **`CK Editor 5`**
![ckeditor5](https://user-images.githubusercontent.com/631897/36802157-739d0822-1cb4-11e8-8b24-15f474b2a4d0.png)

Other option to explore
==

[Trix editor](https://trix-editor.org/) used on basecamp. [Trix repo](https://github.com/basecamp/trix)

raul-fuentes:
I'll do this one

raul-fuentes:
Hi @decabeza, this is the actual look of the field with ckeditor5 (with a very basic setup)
![screenshot from 22](https://user-images.githubusercontent.com/33748390/40615529-6d027d76-6287-11e8-8c97-4f334ed0377b.png)


Before starting with the images I must warn that it will be a bit long, you have to add an image uploader to each model where the editor is used.
I imagine that the imageable concern can handle it but I would like to have the opinion of  @voodoorai2000 

Apart from that, should I include more things in the editor's toolbar?

voodoorai2000:
Sure @raul-fuentes the current imageable concern should work fine integrating it with ckeditor 👍 

Thinking about the upcoming [migration from paperclip to ActiveStorage](https://github.com/consul/consul/issues/2629) it's probably helpful to send the image upload for ckeditor in a separate PR 👌

decabeza:
@raul-fuentes I think this basic setup is enough! 👏  And how @voodoorai2000 says, better send the image upload option in a separate PR 😉

javierm:
Closed via #2876.

PR #2854, opened 2018-08-16
Ckeditor4 update

raul-fuentes:
References
==========
https://github.com/consul/consul/issues/2495

Objectives
==========
Update ckeditor4 configuration adding headings, links and images

Visual Changes (if any)
=======================
- Interface
captura de pantalla de 2018-08-14 14-03-55

- Changing heading
peek 14-08-2018 14-05

- Adding a link
peek 14-08-2018 14-05-8

- Adding an image
peek 14-08-2018 14-07-9

Notes
=====================
*****

### DIFF
--- a/app/controllers/ckeditor/pictures_controller.rb
+++ b/app/controllers/ckeditor/pictures_controller.rb
@@ -1,0 +1,36 @@
+class Ckeditor::PicturesController < Ckeditor::ApplicationController
+  load_and_authorize_resource
+  def index
+    @pictures = Ckeditor.picture_adapter.find_all(ckeditor_pictures_scope)
+    @pictures = Ckeditor::Paginatable.new(@pictures).page(params[:page])
+
+    respond_to do |format|
+      format.html { render :layout => @pictures.first_page? }
+    end
+  end
+
+  def create
+    @picture = Ckeditor.picture_model.new
+    respond_with_asset(@picture)
+  end
+
+  def destroy
+    @picture.destroy
+
+    respond_to do |format|
+      format.html { redirect_to pictures_path }
+      format.json { render :nothing => true, :status => 204 }
+    end
+  end
+
+  protected
+
+    def find_asset
+      @picture = Ckeditor.picture_adapter.get!(params[:id])
+    end
+
+    def authorize_resource
+      model = (@picture || Ckeditor.picture_model)
+      @authorization_adapter.try(:authorize, params[:action], model)
+    end
+end
--- a/app/models/abilities/everyone.rb
+++ b/app/models/abilities/everyone.rb
@@ -3,6 +3,10 @@ module Abilities
     include CanCan::Ability
 
     def initialize(user)
+      can :access, :ckeditor   # needed to access Ckeditor filebrowser
+      can [:access, :read, :create, :destroy], Ckeditor::Picture
+      can [:access, :read, :create, :destroy], Ckeditor::AttachmentFile
+
       can [:read, :map], Debate
       can [:read, :map, :summary, :share], Proposal
       can :read, Comment
--- a/config/initializers/ckeditor.rb
+++ b/config/initializers/ckeditor.rb
@@ -1,4 +1,58 @@
+# Use this hook to configure ckeditor
 Ckeditor.setup do |config|
-  config.assets_languages = Rails.application.config.i18n.available_locales.map{|l| l.to_s.downcase}
-  config.assets_plugins = %w[copyformatting tableselection scayt wsc]
+  # ==> ORM configuration
+  # Load and configure the ORM. Supports :active_record (default), :mongo_mapper and
+  # :mongoid (bson_ext recommended) by default. Other ORMs may be
+  # available as additional gems.
+  require 'ckeditor/orm/active_record'
+
+  # Allowed image file types for upload.
+  # Set to nil or [] (empty array) for all file types
+  # By default: %w(jpg jpeg png gif tiff)
+  # config.image_file_types = %w(jpg jpeg png gif tiff)
+
+  # Allowed flash file types for upload.
+  # Set to nil or [] (empty array) for all file types
+  # By default: %w(jpg jpeg png gif tiff)
+  # config.flash_file_types = %w(swf)
+
+  # Allowed attachment file types for upload.
+  # Set to nil or [] (empty array) for all file types
+  # By default: %w(doc docx xls odt ods pdf rar zip tar tar.gz swf)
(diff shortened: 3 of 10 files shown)


PierreMesure:
Hey @raul-fuentes, have you tried the image upload with a remote storage? I can do it if you haven't

raul-fuentes:
Hi @PierreMesure !

Thanks for the offer, but as far as I know, the storage must be on the servers themselves

PierreMesure:
It seems that you're just letting Paperclip handle it. In that case, if Paperclip [has been configured to use another storage](https://github.com/consul/docs/blob/d439e66702e424ccf75cfb1d8301a358b8dbd7e1/en/getting_started/using-aws-s3-as-storage.md), it should work I guess? The [documentation](https://github.com/galetahub/ckeditor) doesn't give much more detail so I guess it must be pretty straight-forward.

Anyway, I guess that means that this hasn't been tested.

raul-fuentes:
as you can see in app/models/ckeditor/attachment_file.rb
the route is set to  ':rails_root/public/ckeditor_assets/attachments/:id/:filename'

PierreMesure:
I have really little knowledge on that but I would guess that Paperclip uses the url to decide where to save (and retrieve) the files?
The other models that have attached files don't mention a path, by the way, for example [models/image.rb](https://github.com/consul/consul/blob/master/app/models/image.rb). The images are saved using the url.

app/models/abilities/everyone.rb
```diff
@@ -3,6 +3,10 @@ class Everyone
     include CanCan::Ability
 
     def initialize(user)
+      can :access, :ckeditor   # needed to access Ckeditor filebrowser
+      can [:access, :read, :create, :destroy], Ckeditor::Picture
```
javierm: Note that with this configuration any user can delete any image, even the ones uploaded by different users.

config/initializers/ckeditor.rb
```diff
@@ -1,4 +1,58 @@
+# Use this hook to configure ckeditor
 Ckeditor.setup do |config|
-  config.assets_languages = Rails.application.config.i18n.available_locales.map{|l| l.to_s.downcase}
-  config.assets_plugins = %w[copyformatting tableselection scayt wsc]
```
javierm: It looks like these lines were added in #1196 in order to reduce the time the asset compilation takes. Is it safe to remove them now?

app/models/abilities/everyone.rb
aitbw: To consider: this model allows actions for **literally everyone**, even non-logged/registered users

The proper model to use is `common.rb`

config/initializers/ckeditor.rb
aitbw: The `config.assets_plugins` line was added via #2711 in order to fix CKEditor-dependant flaky specs, is there a good reason to delete said line?

javierm: @aitbw To be a bit more precise, it was changed to fix the flakies. Removing it completely (like this PR does) would have also fixed the flakies, but then compilation time would have been long, as mentioned in #1196.

aitbw: :thinking: how so, @javierm ? I mean, how removing that line could have fixed the flakies?

app/controllers/ckeditor/pictures_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  protected
+
+    def find_asset
+      @picture = Ckeditor.picture_adapter.get!(params[:id])
+    end
+
+    def authorize_resource
+      model = (@picture || Ckeditor.picture_model)
```
aitbw: Are the parenthesis around `picture || Ckeditor.picture_model` necessary?

app/controllers/ckeditor/pictures_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+      @authorization_adapter.try(:authorize, params[:action], model)
+    end
+end
```
aitbw: Don't forget a trailing newline at the end of this file :pray:

db/migrate/20180813141443_create_ckeditor_assets.rb
```diff
@@ -0,0 +1,23 @@
+class CreateCkeditorAssets < ActiveRecord::Migration
+  def self.up
+    create_table :ckeditor_assets do |t|
+      t.string  :data_file_name, null: false
+      t.string  :data_content_type
+      t.integer :data_file_size
+      t.string  :data_fingerprint
+      t.string  :type, limit: 30
+
+      # Uncomment	it to save images dimensions, if your need it
```
aitbw: Typo: you*

But the bigger questions here are the following:
* What should we uncomment? Are you referring to the 2 next lines for `width` and `integer`?
* Wouldn't it be better to allow those fields to be empty and then handle through the model if we want to save those values?

db/migrate/20180813141443_create_ckeditor_assets.rb
```diff
@@ -0,0 +1,23 @@
...
+      t.integer :width
+      t.integer :height
+
+      t.timestamps null: false
```
aitbw: I don't think the `null: false` constraint is necessary here

config/initializers/ckeditor.rb
aitbw: Ohhhhhhh, that makes sense. Thanks for the concise explanation :bowing_man:

app/controllers/ckeditor/pictures_controller.rb
javierm: @aitbw This file is almost completely copied from the `ckeditor` gem. That's why its code conventions have nothing to do with the rest of the project. IMHO, if it's possible, it'd be great if we didn't have to copy the file and could just add one or two changes to this class using something like the `CKEditor` initializer.

config/initializers/ckeditor.rb
```diff
@@ -1,4 +1,60 @@
...
+  # By default: nil (CDN disabled)
+  # config.cdn_url = '//cdn.ckeditor.com/4.7.1/standard/ckeditor.js'
+
+  # JS config url
+  # Used when CKEditor CDN enabled
+  # By default: "ckeditor/config.js"
+  # config.js_config_url = 'ckeditor/config.js'
 end
+
+Ckeditor::PicturesController.send(:load_and_authorize_resource)
```
aitbw: Don't forget to add a trailing new line in here :pray:

Issue #2855, opened 2018-08-16
500 error when we created a budget investment from management panel


sturpin:
### Description
When we try to create a budget investment from the management panel with a user, who has either registered in the portal or we manually, we receive a 500 error.

*Type*: Bug

### Steps to reproduce
1. We select user from the management panel
![step_1](https://user-images.githubusercontent.com/16426243/44199392-49096280-a144-11e8-89c3-7969bd19d23b.png)

2. The user is verified
![step_2](https://user-images.githubusercontent.com/16426243/44199423-62121380-a144-11e8-8311-df73fa2cc844.png)

3. We create a budget investment from management panel with user identification
![step_3](https://user-images.githubusercontent.com/16426243/44204358-84ab2900-a152-11e8-89a6-46668f7cc1b5.png)

### Definition of Done
We always receive the 500 error, fill in or not the form and even if you do not accept the privacy policy.

### Traceback Rollbar
NoMethodError: undefined method `each' for nil:NilClass
``File "/consul/app/views/budgets/investments/_form.html.erb" line 62 in block in _app_views_budgets_investments__form_html_erb___2841398156537415368_33401000``

``File "/consul/app/views/budgets/investments/_form.html.erb" line 1 in _app_views_budgets_investments__form_html_erb___2841398156537415368_33401000``

``File "/consul/app/views/management/budgets/investments/new.html.erb" line 10 in _app_views_management_budgets_investments_new_html_erb__2031065292089547621_33642880``

``File "/consul/app/controllers/management/budgets/investments_controller.rb" line 26 in create``

16yuki0702:
I can fix it.

sturpin:
Fantastic @16yuki0702 !!
That error has been solved, but even if you put a location on the map, it gives you an error and does not allow you to save the investment.

### Images 
1. Map location can't be blank ??
![pic_01](https://user-images.githubusercontent.com/16426243/44271731-30c83f00-a23b-11e8-877c-394b1b6770a9.png)

2. I put the location on the map
![pic_02](https://user-images.githubusercontent.com/16426243/44271818-72f18080-a23b-11e8-99de-062348ce5185.png)

If you delete the geolocation, it keeps giving you the same error.

16yuki0702:
thank you for your checking !!
oh, that's strange.
I'll check it.

16yuki0702:
I fixed so that you can store map location.

sturpin:
[...]

PR #2856, opened 2018-08-17
Call load_categories when failed to save investment

16yuki0702:
References
===================
- Addresses issue: https://github.com/consul/consul/issues/2855 - 500 error when we created a budget investment from management panel

Objectives
===================
- Fixed NoMethodError.
It occurs When it fails to @investment.save in create.
because it attempt to render new page without categories variable.
So I added a load_caregories method when it failed.

- I Fixed storing map location error.
Error occurs when storing map location, or skipping map.
I fixed it so that you can store map location, or skip map location.

### DIFF
--- a/app/controllers/management/budgets/investments_controller.rb
+++ b/app/controllers/management/budgets/investments_controller.rb
@@ -23,6 +23,7 @@ class Management::Budgets::InvestmentsController < Management::BaseController
       notice = t('flash.actions.create.notice', resource_name: Budget::Investment.model_name.human, count: 1)
       redirect_to management_budget_investment_path(@budget, @investment), notice: notice
     else
+      load_categories
       render :new
     end
   end
--- a/app/controllers/management/budgets/investments_controller.rb
+++ b/app/controllers/management/budgets/investments_controller.rb
@@ -70,9 +70,15 @@ class Management::Budgets::InvestmentsController < Management::BaseController
     end
 
     def set_map_location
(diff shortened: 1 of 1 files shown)


sturpin:
Good job @16yuki0702 !! :+1:

sturpin:
Hi @16yuki0702 , I found a small detail... I tell you in case you think it's not very complicated to solved.
I've verified that if you uncheck the privacy policy, allows you to create the investment equally.
For me there is no problem as it's an internal panel....

16yuki0702:
@sturpin Thanks, I'll check it.

16yuki0702:
I investigated why it passed validation when if you uncheck the privacy policy,
I found that terms_of_service is assigned with "1" in controller.

It is not only
```
app/controllers/management/budgets/investments_controller.rb
```
but also
```
app/controllers/management/users_controller.rb
```

It seems to attempt to pass the validation whether terms_of_service is checked or not.
I don't know why it does, so would anyone teach me about this ??

If the policy of modification it is settled,
I want to fix it in another PR, OK??
(Because I think it is out of range in this PR)

sturpin:
Yes @16yuki0702 I agree with you, keep me informed :)

javierm:
@16yuki0702 Thank you very much for your contribution :heart:!

Since you're working on the issue, would it be possible to add a test so we catch this bug immediately if we reintroduce it in the future?

Thanks a lot!

PR #2887, opened 2018-09-10
[Backport] Replace `.all.each` with `.find_each` to reduce memory usage

aitbw:
References
===================
* **Related PR**: AyuntamientoMadrid#1566

Objectives
===================
* As per [this section](http://guides.rubyonrails.org/v4.2/active_record_querying.html#retrieving-multiple-objects-in-batches) in the Ruby on Rails documentation, the usage of `.all.each` over an ActiveRecord model might exceed the amount of memory available if said model is associated with a table with a considerable amount of records. Thus, it has been replaced where possible with `.find_each`, which has considerable performance gains

Visual Changes
===================
* None

Notes
===================
* None


voodoorai2000:
Yes, great! 👌

Issue #2741, opened 2018-07-09
Updates translatable custom pages


decabeza:
Background
==

Now on admin section is possible to create custom pages in `/admin/site_customization/pages/new`.

What we need
==

Update translatable feature for custom pages created from admin panel. 

You can use as reference the PR [Allow admin generated content to be translatable](https://github.com/consul/consul/pull/2619). And try this feature on `/admin/budgets/N/budget_investments/N/budget_investment_milestones/new`.

Screenshots
==

**Actual**
<img width="1218" alt="actual" src="https://user-images.githubusercontent.com/631897/42449329-867de44c-8380-11e8-841e-669948361e41.png">

**After changes**
<img width="1220" alt="after" src="https://user-images.githubusercontent.com/631897/42449332-8b980732-8380-11e8-924d-6f5cfcb7e085.png">

papayalabs:
Working on this issue

javierm:
Closed via #2913.

PR #2913, opened 2018-09-20
Updates translatable custom pages

papayalabs:
References
===================
Issue [#2741](https://github.com/consul/consul/issues/2741)

Visual Changes
===================
> In custom page index we replace 'title' with 'slug' because title could be in many languages. We drop 'locale' from table because does not make sense now.
screen shot 2018-08-06 at 3 51 03 pm
Editing with many languages
screen shot 2018-08-06 at 3 51 16 pm

Notes
===================
In custom page index we replace 'title' with 'slug' because title could be in many languages. We drop 'locale' from table because does not make sense now.

In the migration of translation tables of custom pages ( 20180802221830_add_translate_pages.rb ) we add a change_column to allow null in title. Title is not used anymore as index in table and now is used in many languages title_en,title_es,title_fr, etc. Give NullConstraint from database if do not make this change.

    class AddTranslatePages < ActiveRecord::Migration
      def self.up
        SiteCustomization::Page.create_translation_table!({
          title: :string,
          subtitle: :string,
          content: :text
        })
        change_column :site_customization_pages, :title, :string, :null => true
      end

      def self.down
        SiteCustomization::Page.drop_translation_table!
        change_column :site_customization_pages, :title, :string, :null => false
      end
    end


papayalabs:
Hi @mlovic, all the functionality works fine, the only detail is when you click other translation tag in the Custom Page form the CKEDITOR goes down one line and keeps like that if you continue navigating. I try to fix this from the beginning but I could not find how.

spec/features/site_customization/custom_pages_spec.rb
```diff
@@ -5,10 +5,9 @@
     scenario "See default content when custom page is not published" do
       custom_page = create(:site_customization_page,
         slug: "conditions",
-        title: "Custom conditions",
-        content: "New text for conditions page",
-        print_content_flag: true,
-        locale: "en"
+        title_en: "Custom conditions",
+        content_en: "New text for conditions page",
```
javierm: Thanks to the beauty of globalize we can still write:

```ruby
      custom_page = create(:site_customization_page,
        slug: "conditions",
        title: "Custom conditions",
        content: "New text for conditions page",
        print_content_flag: true
      )
```
No need to add the `_en` prefix when creating records.

javierm: The same idea applies to a few more specs changed in the pull request.

spec/features/site_customization/custom_pages_spec.rb
```diff
@@ -142,25 +134,125 @@
...
+      click_link "Remove language"
+
+      expect(page).not_to have_link "Español"
+
+      click_button "Update Custom page"
+      visit @edit_page_url
+      expect(page).not_to have_link "Español"
+    end
+
+    context "Globalize javascript interface" do
```
javierm: It looks like some of these tests could be replaced by `it_behaves_like "translatable"`. I'm not sure, though. @mlovic @papayalabs What do you think?

db/schema.rb
```diff
@@ -116,6 +116,8 @@
     t.datetime "updated_at",                  null: false
     t.text     "background_color"
     t.text     "font_color"
+    t.string   "banner_position"
+    t.text     "headings"
```
javierm: I think these lines aren't related to this pull request.

db/schema.rb
```diff
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20180813141443) do
+ActiveRecord::Schema.define(version: 20180831000000) do
```
javierm: This schema version doesn't seem to be related to any migration files. I wonder how it got there :confused:.

db/schema.rb
```diff
@@ -1074,17 +1076,29 @@
...
-    t.string   "title",                                null: false
+    t.string   "title"
     t.string   "subtitle"
     t.text     "content"
     t.boolean  "more_info_flag"
     t.boolean  "print_content_flag"
     t.string   "status",             default: "draft"
     t.datetime "created_at",                           null: false
     t.datetime "updated_at",                           null: false
-    t.string   "locale"
```
javierm: I didn't find a migration file where this column was dropped. Did I miss it?

spec/features/site_customization/custom_pages_spec.rb
papayalabs: I think is good to maintain uniformity with the prefixs ( much more easy for anyone that see the code for the first time). There are places in the this same spec where is needed to define this fields in differents languages and you need to use prefixs. But this my opinion, if you guys like without prefixs where is no needed I will changed. ;)

spec/features/site_customization/custom_pages_spec.rb
papayalabs: Using it_behaves_like "translatable" will use test "Add a translation scenario" with "Français" and I think is important to test with "Português" because for this language has some particularities that may trigger some errors that will not trigger with other languages. But again is my perspective, if you guys prefer with "translatable" code I will changed ;)

papayalabs:
Hi @javierm! Yes you are right that I need to setup everything to be more aware of indentation and other rubocop rules. I fix all you tell me about this. I also change schema.rb ( hope I do it right). I only comment the changes in "spec/features/site_customization/custom_pages_spec.rb" with my perspective but is not fixed. Please tell me if you need more changes.

papayalabs:
@javierm almost all the changes you request were made but I unified everything in one commit ( first one ) so it looks that were not made

app/views/admin/site_customization/pages/_form.html.erb
```diff
@@ -30,28 +31,30 @@
...
+          <% @page.globalize_locales.each do |locale| %>
+            <% globalize(locale) do %>
+              <%= content_tag :span , class: "js-globalize-attribute",
+                                data: { locale: locale },
+                                style: display_translation?(locale) do %>
+              <%= f.cktext_area "content_#{locale.to_s.downcase.underscore.to_sym}",
+                                class: "js-globalize-attribute",
+                                ckeditor: { language: locale },
+                                data: { locale: locale },
+                                label: false %>
```
javierm: I think we need (somehow) a label here. Maybe you could use the code @microweb10 hast just implemented (still not merged, though): https://github.com/consul/consul/pull/2914/files#diff-4d169e27f1a8740fdcd328f352bf82d2R38

app/views/admin/site_customization/pages/edit.html.erb
```diff
@@ -1,11 +1,11 @@
-<% provide :title do %>
-  <%= t("admin.header.title") %> - <%= t("admin.menu.site_customization.pages") %> - <%= @page.title %>
+<% provide :slug do %>
```
javierm: Actually, we need to use `provide :title` :wink:, since the parameter here is used to generate the HTML `<title>` tag (see `app/views/layouts/_common_head.html.erb`, the line saying `<title><%= content_for?(:title) ? yield(:title) : default_title  %></title>`).

app/views/admin/site_customization/pages/edit.html.erb
```diff
@@ -1,11 +1,11 @@
...
+  <%= t("admin.header.title") %> - <%= t("admin.menu.site_customization.pages") %> - <%= @page.slug %>
```
javierm: I don't have a strong preference about which title the browser should show, which is what we're changing here. Which one do you think it's better?

app/views/admin/site_customization/pages/edit.html.erb
```diff
@@ -1,11 +1,11 @@
...
 <% end %>
 
 <%= back_link_to admin_site_customization_pages_path %>
 
 <div class="small-12 column">
-  <h2 class="inline-block"><%= t("admin.site_customization.pages.edit.title", page_title: @page.title) %></h2>
+  <h2 class="inline-block"><%= t("admin.site_customization.pages.edit.title", page_title: @page.slug) %></h2>
```
javierm: Similar comment as the previous one. Should admins see "Editing My Page" or "Editing my-page"? Maybe @decabeza can tell better.

app/views/admin/site_customization/pages/index.html.erb
```diff
@@ -11,24 +11,22 @@
...
-        <th><%= t("admin.site_customization.pages.page.locale") %></th>
         <th class="small-4"><%= t("admin.actions.actions") %></th>
       </tr>
     </thead>
     <tbody>
     <% @pages.each do |page| %>
       <tr id="<%= dom_id(page) %>">
         <td>
-          <%= link_to page.title, edit_admin_site_customization_page_path(page) %>
+          <%= link_to page.slug, edit_admin_site_customization_page_path(page) %>
```
javierm: In all other translatable sections we show the title. @decabeza @papayalabs  Which one do you guys think is more convenient here?

spec/features/site_customization/custom_pages_spec.rb
javierm: I agree it's important to test Brazilian Portuguese :+1:! In the `translatable` shared examples there's a scenario for that case, called "Add a translation for a locale with non-underscored name". Would that be enough, or do we need something specific to custom pages? If we need something specific, we could add `it_behaves_like "translatable"` and then add tests for specific cases, like in [collaborative legislation specs](https://github.com/AyuntamientoMadrid/consul/pull/1628/files#diff-8417d724f680660eb405aab8cda022adR106). What do you think?

app/views/pages/help/_other.html.erb
```diff
@@ -3,7 +3,7 @@
 <ul class="features">
   <li><%= link_to t("pages.help.other.how_to_use", org_name: setting['org_name']), how_to_use_path %></li>
 
-  <% SiteCustomization::Page.with_more_info_flag.with_same_locale.each do |custom_page| %>
+  <% SiteCustomization::Page.with_more_info_flag.each do |custom_page| %>
```
javierm: This one is tricky. IMHO we should only show pages which have a translation for the current locale. How about keeping this line the way it was and changing the scope to:
```ruby
   scope :with_same_locale, -> { joins(:translations).where("site_customization_page_translations.locale": I18n.locale) }
```

Feel free to simplify the condition! This is only an example.

spec/features/site_customization/custom_pages_spec.rb
papayalabs: I had not seen that scenario in the translatable shared. I think that would be enough, Thanks for the clarification @javierm! I am going to updated with translatable shared

app/views/admin/site_customization/pages/_form.html.erb
papayalabs: Yes, our functionlity with ckeditor is similar. Will work just fine anything that @microweb10 implement. 👍

app/views/admin/site_customization/pages/edit.html.erb
papayalabs: Ok I will check this.  I took the daring to replace `slug`with `title`but I am not sure is the best way to go. 😉 I take this way because slug like status and created,updated at are not translatable in many languages but I could change everything to use `title`again. Just confirm me!

app/views/admin/site_customization/pages/edit.html.erb
papayalabs: I don´t have preference either 😉. Just confirm me what to use. :)

app/views/pages/help/_other.html.erb
papayalabs: I am agree. Will add again with_same_local scope and check to simplify the condition.

app/views/admin/site_customization/pages/edit.html.erb
decabeza: Hi, @papayalabs and @javierm! I think it's better always to use the **page title**. Usually, the slugs will have a dash or underscore and it'll a little more uncomprehensive. 😌 

I think is easier to read 
`Administration - Editing My custom page` 
instead of 
`Administration - Editing my-custom-page`

spec/features/site_customization/custom_pages_spec.rb
papayalabs: Hi @javierm! in the ´translatable shared class´is there a way to pass some field that are not translatable? For example in cutom pages the default for ´status´is ´draft´and we need to pass ´status´´published´for tests.

db/migrate/20180924071722_add_translate_pages.rb
```diff
@@ -0,0 +1,15 @@
...
+      content: :text
+    })
+    change_column :site_customization_pages, :title, :string, :null => true
+  end
+
+  def self.down
+    SiteCustomization::Page.drop_translation_table!
+    change_column :site_customization_pages, :title, :string, :null => false
+  end
+end
```
javierm: I think this missing newline character is the only spacing issue left.

papayalabs: Yes, sorry. Updated.

app/views/admin/site_customization/pages/_form.html.erb
```diff
@@ -1,5 +1,5 @@
-<%= form_for [:admin, @page], html: {class: "edit_page", data: {watch_changes: true}} do |f| %>
-
+<%= translatable_form_for [:admin, @page], html: {class: "edit_page", data: {watch_changes: true}} do |f| %>
+<%= render "admin/shared/globalize_locales", resource: @page %>
```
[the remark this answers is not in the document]
papayalabs: Updated! 👍

spec/features/site_customization/custom_pages_spec.rb
javierm: How about adding a new factory? In `spec/factories/administration.rb`:

```ruby
  factory :site_customization_page, class: 'SiteCustomization::Page' do
    slug "example-page"
   # (...)

    trait :published do
      status "published"
    end
    
    # (...)

    factory :published_site_customization_page, traits: [:published] # <---- Add this line
  end
```
Maybe that way we could pass `published_site_customization_page` as the factory name to `translatable`.

You might run into issues using the translatable behaviour because (so far) some of the specs there don't work well with CKEditor. If that happens, it's OK to leave the tests where they are for now.

papayalabs: Thanks so much for the tip!!

app/helpers/translatable_form_helper.rb
```diff
@@ -47,7 +51,14 @@ def translatable_field(field_type, method, options = {})
...
 
-              @template.concat send(field_type, localized_attr_name, final_options)
+              if field_type == :cktext_area
+                @template.concat content_tag :div, send(field_type, localized_attr_name, final_options),
+                                             class: "js-globalize-attribute",
+                                             style: @template.display_translation?(locale),
+                                             data: { locale: locale }
+              else
+                @template.concat send(field_type, localized_attr_name, final_options)
+              end
```
javierm: We've already merged #2914 which incorporated these changes, in case you'd like to rebase against master.

spec/features/site_customization/custom_pages_spec.rb
papayalabs: Was not necessary to create published_site_customization_page, the thing was that in translatable class it uses a ´update_button_text´ and for default uses the name ´Save changes´but the button in Custom Pages is ´Update Custom page´and I thought that the page was not published. I already add ´Update Custom page´ name in translatable class. But you are right does not recognize CKEditor fields nevertheless it works fine with the others fields and the tests now does not check CKEditor fields. So how I proceed? Leave tests as they are or use translatable with (title,subtitle) only? Thanks!!

javierm: Leaving tests as they are is fine :ok_hand:. We'll try to better integrate CKEditor with Capybara in the future.

app/views/admin/site_customization/pages/_form.html.erb
```diff
@@ -30,33 +31,20 @@
...
-        <%= f.label :content %>
-        <%= f.cktext_area :content, label: false, cols: 80, rows: 10,
-                          ckeditor: { language: I18n.locale, toolbar: "admin" } %>
-      </div>
-
-      <div class="small-12 medium-6 large-3">
-        <%= f.submit class: "button success expanded" %>
-      </div>
+  <div class="small-12 column">
+	<hr>
```
javierm: Could you remove the trailing whitespaces in this line? By the way, IMHO this line should be either indented or before the previous line (I'm not sure which one is right; you probably know it better).

app/views/admin/site_customization/pages/_form.html.erb
```diff
@@ -30,33 +31,20 @@
...
+    <%= f.translatable_text_field :title %>
+    <%= f.translatable_text_field :subtitle %>
+    <div class="ckeditor">
+      <%= f.translatable_cktext_area :content %>
     </div>
-  </div>
+    <div class="small-12 medium-6 large-3">
+      <%= f.submit class: "button success expanded" %>
+    </div>
+   </div>
```
javierm: Could you move this line one space to the left :wink:? Sorry I'm being a bit annoying about whitespace issues. Hope you understand we've got many contributors and sometimes we need to try to keep the codebase as consistent as possible.

spec/features/site_customization/custom_pages_spec.rb
papayalabs: Ok @javierm! So I thing all the requested changes are ready. I unified everything in one commit. Thanks!

app/views/admin/site_customization/pages/_form.html.erb
```diff
@@ -30,33 +31,20 @@
...
-        <%= f.label :content %>
-        <%= f.cktext_area :content, label: false, cols: 80, rows: 10,
-                          ckeditor: { language: I18n.locale, toolbar: "admin" } %>
-      </div>
-
-      <div class="small-12 medium-6 large-3">
-        <%= f.submit class: "button success expanded" %>
-      </div>
+  <div class="small-12 column">
+    <hr>
```
javierm: Thanks for the indentation fix! Would you mind removing the trailing whitespace here :pray:? Thank you so much!

papayalabs: Sure! No problem. I revised with an option of the editor the trailing whitespaces but I think I am doing something wrong. I am going to checked again.

papayalabs: Ready! Thanks @javierm for your patience 🙏

spec/features/admin/site_customization/pages_spec.rb
```diff
@@ -11,7 +11,7 @@
     custom_page = create(:site_customization_page)
     visit admin_site_customization_pages_path
 
-    expect(page).to have_content(custom_page.title)
```
decabeza: As we decided to keep the title and the slug on `/admin/site_customization/pages/index.html.erb` could you keep both also in the spec? 🙏 😌 

Thanks!

spec/features/admin/site_customization/pages_spec.rb
```diff
@@ -22,19 +22,18 @@
         click_link "Custom pages"
       end
 
-      expect(page).not_to have_content "An example custom page"
```
decabeza: Same the previous 😉

spec/features/admin/site_customization/pages_spec.rb
```diff
@@ -22,19 +22,18 @@
...
+      fill_in "site_customization_page_title_en", with: "An example custom page"
+      fill_in "site_customization_page_subtitle_en", with: "Page subtitle"
       fill_in "site_customization_page_slug", with: "example-page"
-      fill_in "site_customization_page_content", with: "This page is about..."
-      select 'English', from:"site_customization_page_locale"
+      fill_in "site_customization_page_content_en", with: "This page is about..."
 
       click_button "Create Custom page"
 
-      expect(page).to have_content "An example custom page"
```
decabeza: Same the previous 😉

spec/features/admin/site_customization/pages_spec.rb
papayalabs: Of course @decabeza! I already added in all scenarios in spec. I forgot to update the spec. 🐰 Thanks!!

spec/features/admin/site_customization/pages_spec.rb
papayalabs: Ready ✌️

spec/features/admin/site_customization/pages_spec.rb
papayalabs: Ready ✌️

papayalabs:
My pleasure! Thank you guys to make so nice to contribute in this project!

Issue #2846, opened 2018-08-13
Require the answer of all questions in a poll


tiagozini:
### User Story
As a user
I want to have a way to allow a poll to have all questions as mandatory, showing the related error to the user that did not select an answer in a question. If an error occurs the answer shouldn't be sent. It should be possible to configured this in the poll admin settings.
So that partial participations don't occur when it is not desired.

### Why is this Needed?

The problem is that now any question answered is registered, even when a user abandons the poll after the first questions. Answering to the questions, maybe, without much attention. Also, answering the poll partially may result in an unrealistic idea of what people really think about the subject.

### Description
*Feature*: Allow a poll to have all question as mandatory, displaying errors to users when they don't select an answer for a question and don't update answers if these errors occur. It should be possible to configured this in the poll admin settings.

PR #2917, opened 2018-09-23
GH-2846 - Add setting to require the answer choose for all questions …

tiagozini:


voodoorai2000:
Hey @tiagozini!

Awesome yes, this is a configuration option that sounds perfectly reasonable 😌, and very well explained in the issue 🙌 

Regarding the implementation, we try to stay away as much as possible from javascript, as it's harder to implement and test. And we also try to use sparingly and only when really necessary, session variables.

To solve this issue, I suggest using more in depth the validation capabilities of Rails. 
Maybe `validates_each`[1]

You could check using Rails' validations, before saving the answers, at the model level, and if not all of them have been answered, we could display another standard rails validation error in the form, and life will be beautiful 😊

Testing this validation error, in model specs, should also be simpler 😌 

What do you think?

[1] https://guides.rubyonrails.org/active_record_validations.html#validates-each

tiagozini:
> Hey @tiagozini!
> 
> Awesome yes, this is a configuration option that sounds perfectly reasonable 😌, and very well explained in the issue 🙌
> 
> Regarding the implementation, we try to stay away as much as possible from javascript, as it's harder to implement and test. And we also try to use sparingly and only when really necessary, session variables.
> 
> To solve this issue, I suggest using more in depth the validation capabilities of Rails.
> Maybe `validates_each`[1]
> 
> You could check using Rails' validations, before saving the answers, at the model level, and if not all of them have been answered, we could display another standard rails validation error in the form, and life will be beautiful 😊
> 
> Testing this validation error, in model specs, should also be simpler 😌
> 
> What do you think?
> 
> [1] https://guides.rubyonrails.org/active_record_validations.html#validates-each

About Javascript use I can´t undestand how avoid its use in this feature. About sessions, I changed the code to don´t use them here. About use of standart validation I can´t see a way to use on the view to show error or not for each question even using validates_each or validates_associated . 
I am open to sugestions.

voodoorai2000:
Continuing this conversation in a chat room, if someone is interested in participating, just ping 😌
If you want this feature in your fork put a 👍 in the [issue](https://github.com/consul/consul/issues/2846)

javierm:
Hi, @tiagozini :smile:. Are you still working on this pull request? I wasn't part of the chat room mentioned by Rai; where there any conclusions? I see the last commits include "WIP", so I guess this wasn't finished :thinking:.

Issue #2954, opened 2018-10-06
Ability to catch and forward emails


ziyan-junaideen:
### Background

We are required to migrate the production database to a another instance of the app for testing purposes . This raises the possibility of blasting our users with unintended emails. It would be good to be able to on config prevent emails been sent to users.

### Description

*Type*: Feature

To be able to configure the application to catch and forward emails to a specified email address.

### Current Behavior

There exists no configurable option to prevent / forward emails unless by updating the code base.

### Expected Behavior

- Add key `email_interceptor_recipients: user@example.com`
- Reboot the app
- Generated emails delivered to the forwarded to the specified emails

### Notes

- This is work in progress
- Uses https://github.com/croaky/recipient_interceptor
- Following discussion with @voodoorai2000 in Slack

microweb10:
Hi @ziyan-junaideen 

We have created a new [PR](https://github.com/consul/consul/pull/3251) using your code, added some tests, and it will be merged shortly. ☺️ 

Thank you very much for your contribution! 🎉

PR #2955, opened 2018-10-06
Configurable e-mail intercept and forward

ziyan-junaideen:
## References
- Issue #2954 - Ability to catch and forward emails

## Visual Changes
There exists no visual changes

## Concerns
- Not sure how to cover the updates from tests, and tested it manually. Any ideas to on how it can be tested?
- Does it make sense to have it across environments? Do we need to limit it to a specific env like staging? We would be using it in a 'staging' environment but I guess that would depend on the situation.
- Where would we document the changes? I added it to the general section of the secrets yml file.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -54,6 +54,7 @@ gem 'unicorn', '~> 5.4.1'
 gem 'whenever', '~> 0.10.0', require: false
 gem 'globalize', '~> 5.0.0'
 gem 'globalize-accessors', '~> 0.2.1'
+gem 'recipient_interceptor'
 
 source 'https://rails-assets.org' do
   gem 'rails-assets-leaflet'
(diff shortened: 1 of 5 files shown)


voodoorai2000:
Hey @ziyan-junaideen!

Very nice 👌

Sure, to test this maybe by setting the `Mail.register_interceptor(interceptor)` in the test could work

Yes, +1 to using this configuration only for `Staging` and `Preproduction` environments

To document this feature, we would need to add the `email_interceptor_recipients` to `secrets.example.yml` and please leave a comment in the PR Notes using a ⚠️symbol, so that we remember to add to the Release Notes that a new `secret` has to be added to production environments in use 👍

microweb10:
Hi @ziyan-junaideen 

We have created a new [PR](https://github.com/consul/consul/pull/3251) using your code, added some tests, and it will be merged shortly. ☺️ 

Thank you very much for your contribution! 🎉

PR #2969, opened 2018-10-12
Add missing language names

PierreMesure:
References
===================
Recently, the Consul team changed the locale which defines the name of a language. Although some languages got the information transferred, some others didn't get that chance.
Even though I hope that people will quickly update the information on Crowdin, new translations are added very seldom to the master branch so I figured it would be a good thing to fix this with a quick PR rather than waiting for it or doing the work manually on each of our forks.

Objectives
===================
- Adds language name for a dozen of languages. The names were taken from [this page](https://meta.wikimedia.org/wiki/Template:List_of_language_names_ordered_by_code).

Visual Changes
===================
capture d ecran 2018-10-12 a 09 17 38

capture d ecran 2018-10-12 a 09 19 07


javierm:
Hi @PierreMesure. Thank you for having a look at this issue! Could you confirm whether these changes are already part of #2964?

PierreMesure:
Hi @javierm, yes this other PR seems even better than mine! I feel stupid, I should have looked at the open PRs before submitting mine. On the other hand, I feel like it should have been merged at the same time as the one that change the location of language names in the locales.

In any case, thank you for taking the time to add so many languages that you don't use in the base repo of Consul!

javierm:
Don't feel stupid, no worries, better to have two pull requests than no pull request at all :+1:. We also wish we had merged everything working perfectly from the start, but unfortunately translations are very complex and there are details we don't catch immediately :sweat_smile:.

Hope everything will be in a usable state real soon! :smile:. In the meantime, thanks for being alert, and if there's something you think we might have overlooked (particularly after merging the open pull requests), please let us know!

PR #3029, opened 2018-11-10
Update debates_spec.rb

clairezed:
Objectives
===================
> Fix an incorrect feature setting title in `debates_spec.rb` (I noticed it because it raised an error in my cloned project)

Notes
===================
If I trust https://github.com/consul/consul/blob/master/db/seeds.rb#L85, debates recommendations Setting title is `recommendations_on_debates`, and not `recommendations_for_debates`.


voodoorai2000:
Nice catch @clairezed 👏

PR #3030, opened 2018-11-11
Fix poll results accuracy

### LINKED ISSUE #3004: Poll results accuracy
### Description

*Type*: Bug

This [template](https://github.com/consul/consul/blob/master/app/views/polls/results.html.erb#L39) expects a float number but in this model [file](https://github.com/consul/consul/blob/master/app/models/poll/question/answer.rb#L49) is calculated using integer numbers so ruby returns an integer value. 

### Definition of Done
A posible solution is replace this [line](https://github.com/consul/consul/blob/master/app/models/poll/question/answer.rb#L49) with this
`question.answers_total_votes.zero? ? 0 : (total_votes * 100).fdiv(question.answers_total_votes)`

xuanxu:
Poll's questions results were returning integer instead of floats. This PR increase accuracy using a float value (as expected in the view).
This PR closes #3004

### DIFF
--- a/app/models/poll/question/answer.rb
+++ b/app/models/poll/question/answer.rb
@@ -46,7 +46,7 @@ class Poll::Question::Answer < ActiveRecord::Base
   end
 
   def total_votes_percentage
-    question.answers_total_votes.zero? ? 0 : (total_votes * 100) / question.answers_total_votes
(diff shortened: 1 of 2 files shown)


voodoorai2000:
Thank you @xuanxu 😃

Issue #3013, opened 2018-11-06
Bug on budget investments pagination


decabeza:
## Description

*Type*: Bug

### Current Behavior
When use pagination on `admin/budgets/N/budget_investments` after "Select/unselect" a budget investment appear an error.

### Expected Behavior
Use pagination on `admin/budgets/N/budget_investments` after "Select/unselect" without generate an error. 🤓 

### Steps to reproduce
1. Visit `admin/budgets/N/budget_investments` 
2.  "Select/unselect" a budget investment.
3. Use pagination.

### Definition of Done
[comment]: # (Any other information that would be useful, bullets are helpful.)

behraaang:
Hello dear @decabeza 
I had prepared a solution can I make a pull request?

behraaang:
I have made a [pull request](https://github.com/consul/consul/pull/3034) to this problem, my fork had 3 failing tests which are not relate and are not passing the test suite as result, is that okay?

Moisi:
This issue considered closed?

PR #3034, opened 2018-11-12
Fix pagination after selecting/unselecting budget investment

behraaang:
References
===================
Issue https://github.com/consul/consul/issues/3013

Objectives
===================
Fix pagination link by rendering the specific selected investment on the list

Visual Changes
===================
There won't be any reload to the page when selecting/unselecting the investment 

Notes
===================
There are 3 fail tests on the test suite which are not related to the fix and have been on the fork when I got it


decabeza:
Hi @behraaang, thanks for the PR! 😌 

There is a failing spec `features/admin/budget_investments_spec.rb:1069` related with this changes. Now the page doesn't refresh and the spec fails, but you can add another `click_button('Filter')` on **L1082** to fix it. 😉 

Also, could you create a new spec to test pagination not fails anymore. 🙏

.gitignore
```diff
@@ -34,3 +34,4 @@
 public/sitemap.xml
 public/system/
 /public/ckeditor_assets/
+.idea/
```
decabeza: I think this no necessary add this change on the PR 😉

behraaang: Hi @decabeza 
I have done the things you have asked, I hope next fix would be with no problem ;).
Thank you

voodoorai2000:
This is great, thank you @behraaang! 🎉

Issue #3035, opened 2018-11-13
Flaky spec: Budget Investments - Balloting Phase - Confirm


microweb10:
## Failure

**Travis failed build**: [Travis build 29543, job 3](https://travis-ci.org/consul/consul/jobs/536654919)

```
1) Budget Investments Balloting Phase Confirm
     Failure/Error: expect(page).to have_content "Remove"
       expected to find text "Remove" in "Budget Investment 338 title 2019-05-24 • Manuela1296 • Global Heading Spend money on this €1 Vote"
     # ./spec/support/common_actions/budgets.rb:10:in `block in add_to_ballot'
     # ./spec/support/common_actions/budgets.rb:8:in `add_to_ballot'
     # ./spec/features/budgets/investments_spec.rb:1559:in `block (3 levels) in <top (required)>'
```

## Cause

We don't know the cause yet. It might be related to issues with turbolinks when an element with an AJAX callback has been removed from the DOM, but that's only a wild guess.

javierm:
It hasn't failed for months, so it might be fixed by #3113 or by another change.

PR #3036, opened 2018-11-13
[WIP] Fix flaky spec: Budget Investments - Balloting Phase - Confirm

microweb10:
References
===================
Issue https://github.com/consul/consul/issues/3035

Explain why the test is flaky
===================
Apparently Capybara can execute a piece of JS before other execution of JS has completely finished.

Explain why your PR fixes it
===================
Using `execute_script` method will add `click()` function to the bottom of JS call stack and the click will be executed after other JS functions are finished except for the ones using async callbacks.

Notes
===================
I've executed the test 150 times after the change and passed 100% of the times
```bash
for run in {1..150}
do
  bin/rspec ./spec/features/budgets/investments_spec.rb:1450
done
```

### DIFF
--- a/app/views/budgets/ballot/lines/create.js.erb
+++ b/app/views/budgets/ballot/lines/create.js.erb
@@ -1,9 +1,5 @@
 $("#progress_bar").html('<%= j render("/budgets/ballot/progress_bar", ballot: @ballot) %>');
 $("#sidebar").html('<%= j render("/budgets/investments/sidebar") %>');
-$("#<%= dom_id(@investment) %>_ballot").html('<%= j render("/budgets/investments/ballot",
-                                                           investment: @investment,
-                                                           investment_ids: @investment_ids,
-                                                           ballot: @ballot) %>');
 
 <%= render 'refresh_ballots',
             investment: @investment,
--- a/app/views/budgets/ballot/lines/create.js.erb
+++ b/app/views/budgets/ballot/lines/create.js.erb
@@ -1,7 +1,8 @@
 $("#progress_bar").html('<%= j render("/budgets/ballot/progress_bar", ballot: @ballot) %>');
-$("#sidebar").html('<%= j render("/budgets/investments/sidebar") %>');
 
 <%= render 'refresh_ballots',
             investment: @investment,
             investment_ids: @investment_ids,
             ballot: @ballot %>
+
+$("#sidebar").html('<%= j render("/budgets/investments/sidebar") %>');
--- a/app/views/budgets/ballot/lines/destroy.js.erb
+++ b/app/views/budgets/ballot/lines/destroy.js.erb
@@ -1,8 +1,9 @@
 $("#progress_bar").html('<%= j render("budgets/ballot/progress_bar", ballot: @ballot) %>');
 $("#ballot").html('<%= j render("budgets/ballot/ballot") %>')
-$("#sidebar").html('<%= j render("budgets/investments/sidebar") %>');
 
 <%= render 'refresh_ballots',
             investment: @investment,
             investment_ids: @investment_ids,
             ballot: @ballot %>
+
+$("#sidebar").html('<%= j render("budgets/investments/sidebar") %>');
--- a/app/views/budgets/ballot/lines/destroy.js.erb
+++ b/app/views/budgets/ballot/lines/destroy.js.erb
@@ -1,11 +1,7 @@
 $("#progress_bar").html('<%= j render("budgets/ballot/progress_bar", ballot: @ballot) %>');
-$("#sidebar").html('<%= j render("budgets/investments/sidebar") %>');
 $("#ballot").html('<%= j render("budgets/ballot/ballot") %>')
+$("#sidebar").html('<%= j render("budgets/investments/sidebar") %>');
 
-$("#<%= dom_id(@investment) %>_ballot").html('<%= j render("/budgets/investments/ballot",
-                                                            investment: @investment,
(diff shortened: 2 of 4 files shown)


javierm:
Hi @microweb10. Thank you for this pull request (and bug report)!

Unfortunately I couldn't reproduce the issue locally after running the test 60 times on the `master` branch, so most of what I'm saying are guesses based on similar flaky tests we've had in the past.

It looks like the view in `app/views/budgets/ballot/lines/create.js.erb` re-renders the ballot the user has selected and then re-renders each ballot.

So then, if capybara clicks the "add" link in the second ballot at the same time it's being re-rendered, the click is ignored.

Since we try to describe feature specs as close to what users do as possible, we try not to execute JavaScript manually during a test. So, if possible, try to keep the original line `find('.add a').click` (or maybe change it to `click_link "Vote"`, which does exactly the same but it's closer to what users see).

My suggestion would be either to add an extra line after `expect(page).to have_content "Remove"` checking all ballots have been re-rendered or, if that's not possible, to remove the `render 'refresh_ballots'` part in `app/views/budgets/ballot/lines/create.js.erb`. I'm not familiar with this code so I don't know whether it makes sense.

What do you think?

microweb10:
@javierm it's curious that you cannot reproduce the failure. For me, it was not the first time I saw the same spec failing.

Anyway, I agree and will keep investigating. Hopefully, I will find a good solution 😅 

Thanks for the feedback!

javierm:
Yes, sometimes these flaky tests are quite a pain to debug. I remember there was one which used to fail on my machine about once every five times (I think it was the one fixed in #2995), but after we added some unrelated code which made requests a bit slower, I couldn't reproduce it on my machine unless I deleted the mentioned code :smile:, even though it still failed on Travis builds once in a while.

So it looks like depending on our machines we will be able to reproduce different flaky specs :roll_eyes:.

microweb10:
@javierm I have not very good news 😕

I've tried:

- To check all the ballots are re-rendered after clicking the link. 
- To remove the part that refreshes all ballots.
- To remove the part that re-renders the ballot the user has selected and leave the part that refreshes all ballots.

Actually I think the last option is the one we should use. I don't see the point of rendering twice the ballot the user has selected. `render 'refresh_ballots'` will render all of them 😄.

But I'm still able to reproduce the flaky spec 😓.
I'm running out of ideas 😭.

javierm:
Interesting! How did you check all ballots had been re-rendered?

A tip on debugging; you add the following code to `spec/spec_helper.rb`:

```
  config.after(:each, type: :feature) do |example|
    if example.exception
      save_and_open_page
      save_and_open_screenshot
    end
  end
```

If the screenshots are too small, you can change the size by changing a line in `spec/rails_helper.rb`:
```ruby
    chromeOptions: { args: %w(headless no-sandbox window-size=1200,1600) }
```
Feel free to share the screenshots with the GitHub community :wink:.

microweb10:
To check if all ballots are re-rendered I added the following code at the end of the method `add_to_ballot(budget_investment)` 
```ruby
    Budget::Investment.selected.by_heading(heading.id).each do |budget_investment|
      within("#budget_investment_#{budget_investment.id}") { find('.ballot a') }
    end
```

Apparently that code works fine cause the method `find` doesn't raise any exception, but the flaky spec fails at the same point, on the line `expect(page).to have_content "Remove"`

Yeah, I knew about the `save_and_open_page` and `save_and_open_screenshot` methods (thanks anyway ☺️), but actually it doesn't give me any useful information, the ballot appears like the `"Vote"` button hasn't been clicked (while apparently has been found and clicked correclty as no other errors are raised 😐)

I attached an example just tested now:

- This is the error

```
Failures:

  1) Budget Investments Balloting Phase Confirm
     Failure/Error: expect(page).to have_content "Remove"
       expected to find text "Remove" in "Budget Investment 2 title 2018-11-15 • Manuela3 • Global Heading Spend money on this €10 Vote"
```

- This is the screenshot

capybara-201811151336501433444517

javierm:
Correct me if I'm wrong. I think the code:

```ruby
    Budget::Investment.selected.by_heading(heading.id).each do |budget_investment|
      within("#budget_investment_#{budget_investment.id}") { find('.ballot a') }
    end
```
Doesn't really check all ballots have been re-rendered, because the `.ballot a` is also there before re-rendering.

Thank you for sharing the screenshot! After seeing it, I'm more confident the problem is related to the `find('.add a').click` code being executed at the same time ballots are being re-rendered. I wonder if there's a way to make sure ballots have been re-rendered before clicking the `.add a` link :thinking:.

microweb10:
I'm afraid you are more than right, it doesn't really check it 😅

But, there should be a way, I will keep thinking about it.

The funny thing is that I also tried to remove the part that re-render all ballots and only re-render the one selected, but I was still getting failures.

Thanks for your help! I will keep investigating 😊

javierm:
> The funny thing is that I also tried to remove the part that re-render all ballots and only re-render the one selected, but I was still getting failures.

That's what puzzles me the most about this issue :thinking:.

javierm:
> The part that is not really needed is the one that re-renders the ballot that has been selected, as all ballots will be re-rendered after. So I removed that part.

That sounds good :+1:.

> And the method create in app/controllers/budgets/ballot/lines_controller.rb apperars to not be triggered at all 

That's what I expected, since that's what always happens in similar flaky specs where links are replaced at the same time capybara tries to click them: no request is sent to the server.

I'll let you know if I've got more ideas about how to fix the spec.

javierm:
Closing in favor of #3113.

PR #3063, opened 2018-11-25
Check for missing feature specs for Poll::Question::Answer

microweb10:
Objectives
===================
Make sure all scenarios for Poll::Question::Answer and Poll::Answer were tested


spec/features/polls/answers_spec.rb
```diff
@@ -61,7 +61,25 @@
     end
   end
 
-  pending "Update"
+  scenario "Update" do
+    answer = create(:poll_question_answer, question: question,
```
javierm: It looks like this pull request was open before we had the useless assignment rule active :smile:.

microweb10:
> Why is this pull request WIP? I think you told me months ago, but I've forgotten 😅.

Since we have `Poll::Answer` and `Poll::Question::Answer` this PR was wrongly created. We have to check what tests do we currently have and which ones do we need to add in order to fully cover the 2 different kinds of answers.

It's been in WIP because I haven't had time to look at it since then 😅

microweb10:
After reviewing this PR I have concluded in:

- The test `behave like nested imageable` should be in the admin part (first commit)
- The scenario about adding a video to an answer was already tested in the admin part (second commit)
- The scenarios about listing and creating answers were already tested in the admin part (third commit)
- No extra tests are needed since non-admin users cannot CRUD the answers, they can only vote and voting tests are already covered ☺️ 

So I have removed the WIP, edited the description of the PR and I think it can be merged

PR #3084, opened 2018-12-05
[WIP] Add translations to Proposals, Debates, BudgetInvestments and Comments

Senen:
References
===================
> Related issues: https://github.com/rockandror/consul/issues/15

Objectives
===================
1. Add database translations to Proposals, Debates, BudgetInvestments and Comments.
2. Add translations interface to frontend forms and create new feature setting to allow to enable/disable this interface.
3. Admin panel views keep showing translation interface when is disabled

Visual Changes
===================
**New setting** 
captura de pantalla 2018-12-05 a las 11 43 10

**Proposal form when translation interface is enabled (Desktop version)**
captura de pantalla 2018-12-05 a las 11 46 09

**Proposal form when translation interface is disabled (Desktop version)**
captura de pantalla 2018-12-05 a las 12 15 07

**Proposal form when translation interface is enabled (Mobile version)**
screen shot 2018-12-05 at 11 52 54

**Proposal form when translation interface is enabled (Mobile version)**
screen shot 2018-12-05 at 11 50 40

**Managment proposal form (Desktop version)**
captura de pantalla 2018-12-05 a las 12 00 04

Notes
===================
This PR is in WIP status, not ready for merge yet!


Senen:
Hello @consul/core, here we have a WIP PR with database translations for proposals and discussions. We were able to use the existing translation interface and shared feature specs with some modifications to check the behavior at FrontEnd. We will add BudgetInvestments and comments once we have reviewed this solution and its problems with all of you.

Some things about the behavior of our current implementation to discuss/improve:

1. As you can see in the PR section "Visual changes", the mobile version is a disaster when the translation interface is enabled and some languages are active at the same time. Perhaps here we should convert the language tabs to a drop-down system for mobile phones only 🤔. What do you think of this @decabeza?

2. When the translation interface is disabled, users can only enter a proposal in the interface language. For example: If a verified user of decide.madrid.es is browsing in Italian language and creates a proposal, it will be stored as an Italian translation, so other users will not be able to read this user proposal in Spanish language. Another solution could be to force the default consul language in the forms when the translation interface is disabled so that users can only enter contents in the default language of the application.

3. We found that translatable_form_helper is rendering all the translatable fields nested in the selected interface language, this could be a bit confusing for the user because the only visual difference between them are the input values when switching between the language tabs. Perhaps rendering translatable fields in each group language would help users better understand the interface and really know what language they are working in.

Your opinions will be very much appreciated 🤗

Senen:
I'm going to split into simpler PR's!

PR #3087, opened 2018-12-06
Marker clustering

PierreMesure:
Objectives
===================
This PR adds support for marker clustering on the PB investments map.
It uses the standard [leaflet component](https://leafletjs.com/2012/08/20/guest-post-markerclusterer-0-1-released.html) for clustering.

Visual Changes
===================
ezgif-1-5021e65a0f61

Notes
===================
- I updated leaflet to the latest version because the current one is too old to work with the latest version of the clustering component ~~(also done by PR #2959)~~.
- ~~I'm not sure if my code is efficient, considering I am adding the markers to the map every time the createMarker method is called. It seems to work with the demo seeds but I hope you can test it with bigger numbers.~~


PierreMesure:
I haven't managed to fix the tests yet. Not sure if it's due to the library update or to the clustering but the markers or the clusters aren't loaded when Selenium is rendering the page.

PierreMesure:
The tests are fixed!

A few notes:
- the reason why the tests were failing is because the clusters were not appearing on the map. While Leaflet seems to load single pins even when they are outside of the map, it doesn't do so for clusters.
- to solve this, I had to make sure that the created pins are located inside the boundaries of the map
- to see what the problem was, I used Capybara running with Chrome instead of Headless Chrome (I had no idea you could do that, super powerful!). I feel stupid for learning about it only now but I'm hoping someone else reading this benefits from this note. Just set `Capybara.javascript_driver = :chrome` and enjoy!

javierm:
Closing in favor of #5116.

Issue #2791, opened 2018-07-25
Add homepage for legislation processes


decabeza:
⚠️ **Wait to https://github.com/consul/consul/issues/2495 is merged before to start this issue.**

Background
==
Admins can create new processes on `/admin/legislation/processes`. Sometimes the processes are complex and need to give more context to users with texts, images...

What we need
==
- Add a new "Homepage" tab on admin legislation processes `/admin/legislation/processes/N/edit`.
- Include a checkbox to enable/disable this homepage.
- A CKEditor textarea (wait to https://github.com/consul/consul/issues/2495 for use last version).
- Now the process index URL redirects to the active phase, but if **homepage is enabled** redirect always to this page (independent of the current phase).

Screenshots
==

**Admin form**
![processes_homepage_admin](https://user-images.githubusercontent.com/631897/43212172-8e17eda0-9033-11e8-8c30-5b09dce2ee89.jpg)

**Front view of process index**
![processes_homepage](https://user-images.githubusercontent.com/631897/43212180-90826426-9033-11e8-9407-b5d5588b98d0.jpg)

dpgigorro:
maybe this resolves it: https://github.com/dpgigorro/consul/tree/2791-homepage-for-legislation-process

javierm:
@dpgigorro Thank you very much! Would you mind opening a pull request with the relevant changes against our `master` branch so we can easily check those changes and make comments about them? Thanks again :wink:!

javierm:
Closed via #3091.

PR #3091, opened 2018-12-07
add homepage for legislation processes

dpgigorro:
References
===================
> [#2791](https://github.com/consul/consul/issues/2791)

Objectives
===================
> Added new tab in admin section for legislative processes, to enable a default homepage for each process

Visual Changes
===================
> Admin:

admin

> Front:
front

Notes
===================
> No need to execute any rake.
> [#2791](https://github.com/consul/consul/issues/2791)

### DIFF
--- a/app/views/admin/legislation/homepages/_form.html.erb
+++ b/app/views/admin/legislation/homepages/_form.html.erb
@@ -1,0 +1,47 @@
+<%= render "admin/shared/globalize_locales", resource: @process  %>
+
+<%= translatable_form_for [:admin, @process], url: url, html: {data: {watch_changes: true}} do |f| %>
+
+  <% if @process.errors.any? %>
+
+  <div id="error_explanation" data-alert class="callout alert" data-closable>
+    <button class="close-button" aria-label="<%= t("application.close") %>" type="button" data-close>
+      <span aria-hidden="true">&times;</span>
+    </button>
+
+    <strong>
+      <%= @process.errors.count %>
+      <%= t("admin.legislation.processes.errors.form.error", count: @process.errors.count) %>
+    </strong>
+  </div>
+
+  <% end %>
+
+  <div class="small-12 column margin-top">
+    <%= f.check_box :homepage_enabled, checked: @process.homepage_enabled?, label: t("admin.legislation.processes.form.homepage_enabled") %>
+  </div>
+
+  <div class="ckeditor small-12 column">
+    <label><%= t("admin.legislation.processes.form.homepage") %></label>
+    <p class="help-text"><%= t("admin.legislation.processes.form.homepage_description") %></p>
+    <%= f.cktext_area :homepage,
+                      label: false,
+                      ckeditor: { language: I18n.locale, height: 500, toolbar: 'admin' } %>
+  </div>
+
+  <div class="small-12 column">
+    <hr>
+  </div>
+
+  <%= f.translatable_fields do |translations_form| %>
+    <div class="small-12 column end">
+      <%= translations_form.cktext_area :homepage,
+                                        ckeditor: { height: 500, toolbar: 'admin' },
+                                        hint: t("admin.legislation.processes.form.use_markdown") %>
+    </div>
+  <% end %>
+
+  <div class="small-12 medium-3 column clear end">
+    <%= f.submit(class: "button success expanded", value: t("admin.legislation.processes.#{admin_submit_action(@process)}.submit_button")) %>
+  </div>
+<% end %>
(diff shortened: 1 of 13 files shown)


app/views/admin/legislation/homepages/_form.html.erb
```diff
@@ -0,0 +1,47 @@
+<%= render "admin/shared/globalize_locales", resource: @process  %>
```
javierm: We've recently added another partial which doesn't display the options to add and remove languages (that should be done from the "Information" section; otherwise adding new languages will result in validation errors). If you rebase this branch agasint latest master, we could replace this code with:

```erb
<%= render "admin/shared/globalize_tabs",
           resource: @process,
           display_style: lambda { |locale| enable_translation_style(@process, locale) } %>
```
Not very pretty, I know :sweat_smile: ; we've just added the new partial, so it can certainly be improved.

If you aren't familiar with `git rebase`, feel free to leave this line as it is now, and we'll apply the suggested change later :wink:.

db/migrate/20181206153510_add_home_page_to_legislation_processes.rb
```diff
@@ -0,0 +1,16 @@
+class AddHomePageToLegislationProcesses < ActiveRecord::Migration
+  def change
+    add_column :legislation_processes, :homepage, :text
```
javierm: Adding the translation column is enough :wink:, so in my humble opinion this line (and the one in the `schema.rb` file) should be removed. The reason we still keep both original and translation columns in other tables is we've only added translations for other models recently, but we're planning to remove the original columns in a future release.

app/views/legislation/processes/show.html.erb
```diff
@@ -0,0 +1,13 @@
...
+
+<%= render 'legislation/processes/header', process: @process, header: :full %>
+
+<%= render 'documents/additional_documents', documents: @process.documents %>
+
+<%= render 'key_dates', process: @process, phase: :debate_phase %>
+
+<div class="row">
+  <div class="small-12 medium-9 column">
+    <%= @process.homepage.html_safe %>
```
javierm: In order to remove tags which aren't allowed, how about:
```suggestion
    <%= AdminWYSIWYGSanitizer.new.sanitize(@process.homepage)  %>
```

app/views/admin/legislation/homepages/_form.html.erb
```diff
@@ -0,0 +1,47 @@
...
+      <span aria-hidden="true">&times;</span>
+    </button>
+
+    <strong>
+      <%= @process.errors.count %>
+      <%= t("admin.legislation.processes.errors.form.error", count: @process.errors.count) %>
+    </strong>
+  </div>
+
+  <% end %>
```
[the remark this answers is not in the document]
dpgigorro: Ok, thanks

app/views/admin/legislation/homepages/_form.html.erb
```diff
@@ -0,0 +1,47 @@
...
+
+  <div class="small-12 column margin-top">
+    <%= f.check_box :homepage_enabled, checked: @process.homepage_enabled?, label: t("admin.legislation.processes.form.homepage_enabled") %>
```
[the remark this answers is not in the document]
dpgigorro: yes, you are right

app/views/admin/legislation/homepages/_form.html.erb
```diff
@@ -0,0 +1,47 @@
...
+  </div>
+
+  <div class="ckeditor small-12 column">
+    <label><%= t("admin.legislation.processes.form.homepage") %></label>
+    <p class="help-text"><%= t("admin.legislation.processes.form.homepage_description") %></p>
+    <%= f.cktext_area :homepage,
```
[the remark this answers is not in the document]
dpgigorro: Done

app/views/admin/legislation/homepages/_form.html.erb
dpgigorro: ok, thanks

db/migrate/20181206153510_add_home_page_to_legislation_processes.rb
dpgigorro: ok

app/views/legislation/processes/show.html.erb
dpgigorro: If the homepage is not going to contain tables or so, it's ok

voodoorai2000:
This is great @dpgigorro! 🙌 Thank you so much 🎉

PR #3120, opened 2018-12-19
[Backport] Enable useless assignment rubocop rule

javierm:
## References

* Backports AyuntamientoMadrid#1765

## Objectives

Help developers who don't run the ruby syntax check locally realize they might be involuntarily assigning a value to a local variable which is never used, and so they might want to check whether they need to refactor the code a bit.


javierm:
Travis failure isn't related to these changes.

Issue #2824, opened 2018-07-31
Add options to display results for online polls


voodoorai2000:
### Context

When a Poll is finished, an admin can decide to [show the results publicly](https://github.com/consul/consul/pull/2095)

However this option is only allowed if there are counted Booth Votes(`PartialResults`) for a Poll 

### What is needed?

Make available the options to display results, when there are only Web Votes(`Question::Answers`) for a Poll

### How
Probably the simplest implementation is to check if there are `Voters` for a Poll, as Voters are stored for both Booth Votes and Web Votes, in that case display the options 👍

Also take into account that in the admin section where these options are displayed(`admin/polls/1/results`) other data is also included, like Booth Votes per Question.
This data should still be displayed and the options to display results should be independent and available for both Booth Votes and Web Votes

### Screenshots
<img width="1049" alt="screen shot 2017-10-20 at 18 44 19" src="https://user-images.githubusercontent.com/2141690/31832434-4960ce8a-b5c7-11e7-96df-12365e5aeead.png">

aitbw:
I'm on it

voodoorai2000:
@aitbw are you still up for working on this one? 😃

jaflutz:
Can I try this one (if @aitbw  isn't working on it anymore, of course)?

microweb10:
Sure! @jaflutz go for it! 👍

microweb10:
Done in PR https://github.com/consul/consul/pull/3155

PR #3155, opened 2019-01-05
Enable options to show stats and results with any type of voter

jaflutz:
## References

#2824 - Add options to display results for online polls

## Objectives

The options to display stats and results are now showing even when there are only web voters in a poll. They'll will be shown independent from the partial results in the admin results page. 

## Visual Changes

None. Just display logic.


microweb10:
@jaflutz thank you very much for this PR! 🎉 ☺️ ❤️

Issue #3153, opened 2019-01-03
Refactor image attributes


javierm:
## Background

We've got several forms accepting image attachments. In all their controllers, there's a line like this one:

```ruby
image_attributes: [:id, :title, :attachment, :cached_attachment, :user_id, :_destroy]
```

If we ever change those image attributes (and it might happen when we switch to `ActiveSupport`, we'll need to change them in many places.

## Objectives

Define those attributes in only one place, and reuse them everywhere they're used.

rogelio-o:
I can give a try if no one is working on it!

microweb10:
Sure @rogelio-o! All yours ☺️.

javierm:
Closed via #3170.

PR #3170, opened 2019-01-12
Solves #3153. It refactors images attributes.

rogelio-o:
## References

Refactor image attributes #3153

## Objectives

Define those attributes in only one place, and reuse them everywhere they're used.


javierm:
@rogelio-o Thank you very much for this pull request! :heart_eyes:

There's only one thing I can think about which would make this pull request even better. Including `ImagesHelper` would include all its methods (like `image_absolute_url`) in these controllers, and it would also make the `image_attributes` available in the view layer.

What do you think about moving the `image_attributes` method to a controller concern (maybe in a file like `app/controllers/concerns/image_attributes.rb`) so there are no side effects?

rogelio-o:
Sure! You're right, it's better to move it!

rogelio-o:
Great! Thanks!

javierm:
Merged! :tada: Thank you very much @rogelio-o! :smile:

PR #3156, opened 2019-01-07
[WIP] Frontend translations

Senen:
## References

> Related Issues: #3130 

## Objectives

This PR solves first five bullets os the issue description.

1. ~Add globalizable concern and migrations to add database translations to Proposals,Debates, BudgetInsvestments and Comments~
2. ~Add globalizable behavior to frontend forms~
3. ~Add feature setting to allow administrators to enable/disable translation interface at frontend~
4. ~Create new feature setting to enable/disable translation interface at frontend~
5. ~Add translation interface to frontend translatable forms~
6. Enhance translation interface for frontend views. There are so many language tabs active at the same time

We will make another PR with new translations interface to solve 6th bullet.

## Visual Changes

**New setting** 
captura de pantalla 2018-12-05 a las 11 43 10

**Proposal form when translation interface is enabled (Desktop version)**
captura de pantalla 2018-12-05 a las 11 46 09

**Proposal form when translation interface is disabled (Desktop version)**
captura de pantalla 2018-12-05 a las 12 15 07

**Proposal form when translation interface is enabled (Mobile version)**
screen shot 2018-12-05 at 11 52 54

**Proposal form when translation interface is enabled (Mobile version)**
screen shot 2018-12-05 at 11 50 40

**Managment proposal form (Desktop version)**
captura de pantalla 2018-12-05 a las 12 00 04

## Notes
⚠️ For release notes:

Execute this command to enable help page:
```bin/rake settings:enable_translation_interface RAILS_ENV=production```

Issue #3130, opened 2018-12-24
Translations for proposals, debates, budget investments and comments


Senen:
### User Story
As a citizen, i want to read proposals, debates, budget investments and their comments in my preferred language.

### Why is this Needed?
Consul interface is available in 17 languages actually and growing but proposals, debates, budget::investments and comments are only available in default application language. 

### Description
*Type*: Feature

The purpose of this issue is to make these models translatable. So we will need to provide users an optional interface to add language translations if they like.

1. Add globalizable concern and migrations to add database translations to Proposals,Debates, BudgetInsvestments and Comments 
2. Add globalizable behavior to frontend forms
3. Add feature setting to allow administrators to enable/disable translation interface at frontend
4. Create new feature setting to enable/disable translation interface at frontend
5. Add translation interface to frontend translatable forms
6. Enhance translation interface for frontend views. There are so many language tabs active at the same time

### Current Behavior
Proposals, Debates, BudgetInvestments and Comments are not translatable and only readable in default application language.

### Expected Behavior
Proposals, Debates, BudgetInvestments and Comments are translatable and readable in all application available locales.

### Definition of Done
* Users are able to read Proposals, Debates, BudgetInvestments and Comments in their preferred language when desired translations exists
* Users can introduce Proposals, Debates, BudgetInvestments and Comments(🤔) in all languages if they like and only when feature setting is enabled.

javierm:
Closed via #3359.

PR #3189, opened 2019-01-15
[WIP] Update globalize gem

Senen:
## References

> Related Issues: #3130
> Related Pull Requests: #3156 

## Objectives
Update globalize gem to the latest. There are two reasons to update:
1. Current version of globalize is 5.0.0 which is not ThreadSafe. This bug is fixed at v5.1.0 release.
2. Latest version is v5.2.0 (this is the reason why i need to update), it works pretty good with paranoia gem and related translations simply adding hidden_at column to translations table. Without this update all translations will be destroyed (really) when calling destroy method of parent model, this will break restore feature because translation cannot be recovered.

Here is the globalize [CHANGELOG](https://github.com/globalize/globalize/blob/master/CHANGELOG.md). Here you can check ThreadSafe bug fix.

As example:
```
class Banner < ActiveRecord::Base

  acts_as_paranoid column: :hidden_at
  include ActsAsParanoidAliases

  translates :title,       touch: true
  translates :description, touch: true
  include Globalizable

  ....
end
```
With current globalize version (v5.0.0) we are not able to recover soft_deleted Banner records because translations were really destroyed.

To make globalize models to work good with "paranoia" gem we will only need to add **hidden_at** column to Banner::Translation class as shown below (This feature es available only at v5.2.0 release):

```
class Banner::Translation < Globalize::ActiveRecord::Translation
  acts_as_paranoid column: :hidden_at
end
```

## Visual Changes
Without visual changes

## Notes
None

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -52,7 +52,7 @@ gem 'turnout', '~> 2.4.0'
 gem 'uglifier', '~> 4.1.19'
 gem 'unicorn', '~> 5.4.1'
 gem 'whenever', '~> 0.10.0', require: false
-gem 'globalize', '~> 5.0.0'
+gem 'globalize', '~> 5.2.0'
 gem 'globalize-accessors', '~> 0.2.1'
 
 source 'https://rails-assets.org' do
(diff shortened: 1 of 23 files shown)


Senen:
Closed in favor of #3194

PR #3201, opened 2019-01-22
Pananoid translations (Translations soft deletion)

Senen:
## References

Related Issues: #3156 
Related PR's: This PR contains this another one #3194 commits 

## Objectives
Add soft deletion behavior to translations records.

## Why?
All translations were being lost after soft deletion of parent models. We want to keep these translations in database until they are really destroyed through parent model record destroy method.

## Visual Changes
None

## Notes
None


Senen:
@voodoorai2000 i forgot to add database indexes for new `:hidden_at` database columns. I'll add them the asap.

PR #3239, opened 2019-01-28
Globalize and I18n fallbacks

Senen:
## References

* Related Issues: #3130
* Related Pull Requests: #3194

## Objectives
After globalize update to v5.2.0 we have fixed ThreadSafe bug. This version of globalize uses RequestStore gem to store I18n.locale and Globalize.fallbacks in a per request basis to avoid collissions between different requests. This change broke Globalize.fallback results because it tries to fetch from RequestStore, where there is no locale fallbacks definition.

With this solution we rely on I18n fallbacks definition which is not using request store and keep inmutable during aplicacion run. New globalize code will try to fetch Globalize fallbacks from RequestStore but it will always be empty so Globalize will ask to I18n for its
locale fallbacks.

This could be a good solution if we want Globalize to have the same I18n fallbacks definition. Other possible solution could be to invoke Globalize.fallbaks calculation method on each request but we prefer no to do this processing for every single request.

## Visual Changes
None

## Notes
None


Senen:
Hi @voodoorai2000, i'm updating PR description because its from older version! I changed the way to solve this recently! 🙏

Senen:
@voodoorai2000 Now it's updated with the latest code version! 😄

PR #3286, opened 2019-02-10
Replace sass-rails gem by sassc-rails

### LINKED ISSUE #3284: Ruby Sass is Deprecated
http://sass.logdown.com/posts/7081811

PierreMesure:
## References

- Closes #3284 

## Objectives

- Replacing sass-rails by its C version, the former relying on a deprecated gem.

## Notes

- `DEPRECATION WARNING: Passing a string to call() is deprecated and will be illegal
in Sass 4.0. Use call(get-function("shake")) instead.`


PierreMesure:
One failed spec so far:
```
1) Legislation Proposals Random pagination Random order maintained with pagination
     Failure/Error: expect(first_page_proposals_order & legislation_proposals_order).to eq([])
     
       expected: []
            got: ["legislation_proposal_24", "legislation_proposal_24_votes", "legislation_proposal_26", "legislation_proposal_26_votes"]
     
       (compared using ==)
     # ./spec/features/legislation/proposals_spec.rb:68:in `block (3 levels) in <top (required)>'
```

PierreMesure:
OK, seems like it's just a flaky.

voodoorai2000:
Thanks @milovanderlinden, @PierreMesure!

We are still analysing the system dependencies that this gem has and the consequences for forks. We'll come back to it 👍

javierm:
Just for the record, the reason we haven't approved this change yet is it broke the installation in the system we tried it (Madrid's preproduction server). It looks like the required version of GCC was not installed on that server, as mentioned in AyuntamientoMadrid#70.

So we need to consider some CONSUL installations will crash if we use `sassc` :thinking:.

javierm:
After investigating this topic, it looks like everything will be fine with systems with GCC >= 4.7 (released in 2012). Madrid is probably the only place where this will be an issue, since they've got GCC 4.4.

@PierreMesure Could you rebase this pull request and use double quotes instead of single quotes? :pray: After that, I think it'll be safe to merge it.

PierreMesure:
Done.

PR #3336, opened 2019-03-04
Add CoffeeLint rules

javierm:
## Objectives

* Enable CoffeeLint in Hound
* Replace existing CoffeeLint rules with the ones we use in the code
* Update code not following the defined rules

## Notes

As we do in the rest of the code, we're adding a rule setting the max line length to 100 characters, but we're not updating existing code in order to follow that rule.


javierm:
Wrong repository.

Issue #3346, opened 2019-03-08
Disable social media logins on register page when credentials are not set


milovanderlinden:
Currently, the register page will always show all social media buttons, even when the credentials are not set, causing a 500 page when somebody clicks the log in.

We want to achieve a situation where only configured social media logins are shown.

decabeza:
Good point! 👌 

For now you can disable these features on `admin/settings#tab-feature-flags` but yes, It would be nice only show to users if the credentials are set. 😌 

Also we should include a little description on feature, something like: 

"Even if this feature is enabled you have to set the X credentials on `config/secrets.yml` in order to show to users."

![screenshot 2019-03-08 at 14 42 19](https://user-images.githubusercontent.com/631897/54031921-666d4f00-41b0-11e9-8696-293003ed4d1d.png)

milovanderlinden:
Thank you @decabeza. I have created an isolated branch for this. I have just simply disabled them, but by what you told, I now see that I also have to re-enable the check for `feature?`

milovanderlinden:
Mmm, I think I cannot use `<% if feature? :google_login %>` anymore as the loop is on the omniauth providers. Any advice? Can I compare f.i. the name of the feature with the name of the omniauth_provider?

milovanderlinden:
If a omniauth service is present but not configured, it will not show up. The screenshot is an example where all login features are enabled, but only one is configured:
![image](https://user-images.githubusercontent.com/106038/54075356-c6daba00-429e-11e9-9880-58bc2a06bb70.png)

This screenshot shows the admin interface. Omniauth providers that have empty strings in `secrets.yml` will show up as disabled without the possibility to configure them.
![image](https://user-images.githubusercontent.com/106038/54075397-28028d80-429f-11e9-985f-f78d282ffbed.png)

javierm:
Duplicate of #2744.

PR #3347, opened 2019-03-09
Issue 3346

milovanderlinden:
## References

#3346 

## Objectives

To hide omniauth logins that are not configured.


PierreMesure:
Hey @milovanderlinden, it seems like there are more changes on this branch than just the ones you are describing. Could you clean it so it can be merged?

Issue #3375, opened 2019-03-18
Add alias attribute for administrator users


abelardogilm:
### User Story

As an admin, I want to add an alias for administrators users so that make easy to identify them

### Why is this Needed?

As we have on Valuators, adding an alias for administrators make easy and comfortable identify administrators 

### Description

*Type*: Feature

### Current Behavior

N/A

### Expected Behavior

Add new edit administrator page where administrator users have the ability of update administrator alias and update it.
![image](https://user-images.githubusercontent.com/1029265/54522212-f3609700-496c-11e9-8400-2591da599f67.png)

Display new administrator alias on administrators index page
![image](https://user-images.githubusercontent.com/1029265/54522242-05423a00-496d-11e9-9fab-9574a4e5d874.png)

Modify admin selectors to filter by admin to display alias if exists(name as fallback)
![image](https://user-images.githubusercontent.com/1029265/54549774-07c38480-49ab-11e9-9ab0-3f4bbf58395b.png)

Modify investments table to display admin alias on Administrator column
![image](https://user-images.githubusercontent.com/1029265/54592408-9cba9200-4a2c-11e9-8052-964eaa8ce227.png)

### Steps to reproduce

N/A

### Definition of Done

An administrator user can edit other administrators alias
Administrators alias will be stored as administrator attribute
Only users with admin role can edit administrators

PR #3389, opened 2019-03-20
Add description field to administrator users like evaluators description

abelardogilm:
## References

https://github.com/consul/consul/issues/3375

## Objectives

Add model, controller and views changes to let admin users manage administrator description
Display this new description as administrator name if exists

## Visual Changes

**Edit administrator view**
image

**Administrators index**
image

**Budget investments filter_by administrator options and table information**
image

## Notes

- Needs a migration to add new administrator field
- Administrator name will be used as fallback in case of alias was not set


abelardogilm:
I don't know why spec/features/notifications_spec.rb:83 test is failing, maybe flaky test fail? I run it in local and it pass

Issue #2688, opened 2018-06-19
Proposal's dashboard


decabeza:
Proposal's dashboard
==

We are working in a new section for improving proposals management by authors. 

A proposal's dashboard that contains:

- [ ] **Chart** with the ideal number of supports, and the actual reached.
- [ ] List of **goals** (these goals unlock different resources).
- [ ] List of **recommended actions** to get more support for your proposal.
- [ ] List of  **available resources**

Notes
==

This functionality is being developed on [this repo](https://github.com/medialab-prado/consul) and you can see the [roadmap](https://github.com/medialab-prado/consul/projects/2).

Screenshots
==

A first mockup version (more in progress)

![proposal_dashboard](https://user-images.githubusercontent.com/631897/41592297-681229f6-73bc-11e8-92e3-1eed9ea598e8.jpg)

javierm:
Closed via #3412.

PR #3398, opened 2019-03-21
Platoniq dashboard/polls form bugs

Senen:
## References

Related issues: #2688 

## Objectives

Fix some bugs:
* **Database data corruption**: Removed Poll::Question::Answer callback model was not working with nested anwers when users submits more that one new answers at the same time, new nested answers where saved with the same give_order and that was colliding with uniqueness validation.

* **Invalid HTML**: We cannot use id html attribute for nested answers cocoon wrapper because there could be many answers sets at the same time, one per  each question. This was causing wrong behaviour of interface. To solve this we have used `classes` instead of `ids`.

* **Given order managment**: We have changed the way to manage given order:
1º Remove model callback who was causing data corruption 
2º Add given_order to answers_fields at dashboard polls and to admin polls
3ª Add JS scritps for questions and answers to keep answers given_order updated

## Visual Changes
None

## Notes
None

### DIFF
--- a/app/assets/javascripts/answers.js.coffee
+++ b/app/assets/javascripts/answers.js.coffee
@@ -1,0 +1,9 @@
+App.Answers =
+
+  nestedAnswers: ->
+    $('.nested-answers').on 'cocoon:after-insert', (e, insertedItem) ->
+      nestedAnswersCount = $("input[type='hidden'][name$='[given_order]']").size()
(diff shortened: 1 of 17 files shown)


taitus:
Related Issue: [Dashboard: Fix proposal dashboard polls](https://github.com/Platoniq/consul/issues/2)

Issue #2831, opened 2018-07-31
Add a rake task to upgrade CONSUL from a previous version


voodoorai2000:
Context
===
Most [Releases](https://github.com/consul/consul/releases) require a custom rake task to be executed in order to upgrade to the next version

What
===
Add a rake task `consul:upgrade` that understands the [current version](https://github.com/consul/consul/blob/master/app/controllers/installation_controller.rb#L15) and executes the necessary rake tasks for the next version

How
===
Suggested implementation:
- Creating a database table to store the the last version upgraded
- Executing any previous rake tasks in `consul:upgrade`
- Updating the last version executed in the database table

gabteles:
I can do this one @voodoorai2000

javierm:
The process to upgrade was simplified in #3590. After a few years using it, I'd say it's good enough :thinking:.

PR #3401, opened 2019-03-22
Remove obsolete rake tasks

javierm:
## References

* Backports AyuntamientoMadrid#1942

## Objectives

* Remove obsolete code
* Remove useless tests which make Travis builds take longer

## Notes

These tasks dealt with data migrations or stats generations which were done only once, so we don't need them anymore. New CONSUL installations don't need these tasks, and existing CONSUL installations will execute them when upgrading one release at a time.


PierreMesure:
Not super fun upgrading one version at a time 😬. I'm not sure how many will face this problem but it probably would have been better to leave these files for at least a few more updates.

javierm:
Thank you for the feedback, Pierre :smile:. For now we don't plan to support skipping versions when upgrading because the amount of obsolete code/data we'd need to maintain would be just too much. However, we know the upgrading process should be easier :relieved:. We've got some ideas to simplify the upgrade, like automatically adding new settings (javierm#6) and adding the `consul:upgrade` task (issue #2831).

Issue #3428, opened 2019-04-08
Create a new subset of valuators that can only comment but not evaluate


abelardogilm:
### User Story

As a budget investment reviewer, I want to only have the ability of create evaluation comments so that make dossier data more secure

### Why is this Needed?

We need a new kind of user profile at the level of a valuator that can create evaluation comments but can not edit dossier information.  

### Description

*Type*: Feature

### Current Behavior

We only have valuator user role which allow user to edit dossier information and create evaluation comments

### Expected Behavior

We will have the possibility of create valuator users by choosing the actions that the user can do.
Valuator user actions:
- Show budget investment
- Edit dossier
- Show evaluation comments
- Create evaluation comments

### Definition of Done

Administrator could manage valuator user to enable or disable to
- edit budget investment dossier information
- create evaluation comments

javierm:
Closed via #3437.

PR #3437, opened 2019-04-11
Add abilities to valuator model for comment and edit dossiers

abelardogilm:
## References

#3428 

## Objectives

Adding that new abilities, admin users can manage valuators to allow them to comment or edit dossier information, enabling and updating the resource. With that, we can have a new type of valuator that can only make evaluation comments, or can only edit dossiers, or can do both. And it will allow to create new abilities easily in future feautures

### Admin
- valuators
image

- valuator edit
image

- valuator show (with checks enabled)
image
- valuator show (without checks enabled)
image

### Valuator
- Can comment && !can edit dossier
image

- Can edit dossier && !can comment
image

- Can comment && can edit dossier
image

- !Can comment && !can edit dossier
image
 
## Notes

By default, but properties will be set as enabled to allow current users have current functionality
Admin users will be allowed to comment and edit dossiers.


abelardogilm:
Failing SPEC: rspec ./spec/features/comments/budget_investments_spec.rb:404
I run it on local and pass, run  describe container and pass, but runnign file suite fail randomly

As this spec is not in that PR, I think we need to investigate it to tag it as flaky test,

javierm:
@abelardogilm Fixed by #3595 :wink:

Issue #3451, opened 2019-04-23
Flaky spec: Ballots Groups Change my heading


javierm:
## Failure

**Travis failed builds:** [Travis build 28983, job 2](https://travis-ci.org/consul/consul/jobs/523559558), [Madrid's build 11441, job 9](https://travis-ci.org/AyuntamientoMadrid/consul/jobs/523571391)
and [forked repo build 20, job 1](https://travis-ci.org/javierm/consul/jobs/417194624)
**Messages:**
```
Failures:
1) Ballots Groups Change my heading
     Failure/Error: investments.where(group: group).first.heading
     NoMethodError:
       undefined method `heading' for nil:NilClass
     # ./app/models/budget/ballot.rb:71:in `heading_for_group'
     # ./app/controllers/budgets/investments_controller.rb:158:in `load_assigned_heading'
     # ./app/controllers/budgets/investments_controller.rb:147:in `load_heading'
     # ------------------
     # --- Caused by: ---
     # Capybara::ElementNotFound:
     #   Unable to find visible css "#budget_investment_10"
     #   ./spec/support/common_actions/budgets.rb:8:in `add_to_ballot'
```
Another message:
```
 1) Ballots Groups Change my heading
     Failure/Error: expect(page).to have_content "Remove"
     Selenium::WebDriver::Error::StaleElementReferenceError:
       stale element reference: element is not attached to the page document
         (Session info: headless chrome=71.0.3578.80)
         (Driver info: chromedriver=2.36.540471 (9c759b81a907e70363c6312294d30b6ccccc2752),platform=Linux 4.4.0-101-generic x86_64)
     # ./spec/support/common_actions/budgets.rb:10:in `block in add_to_ballot'
     # ./spec/support/common_actions/budgets.rb:8:in `add_to_ballot'
     # ./spec/features/budgets/ballots_spec.rb:282:in `block (3 levels) in <top (required)>'
```

## Cause

The code:
```
      within("#budget_investment_#{investment1.id}") do
        find(".remove a").click
      end

      visit budget_investments_path(budget, heading_id: new_york.id)
```

Generates a "normal" request after an AJAX request. A possible reason why the tests fails is we don't check the AJAX request has finished before visiting the new page, and so both requests enter a race condition.

Note this test only [fails after upgrading to Rails 5](https://github.com/consul/consul/pull/2631#issuecomment-414689146), probably due to a different behaviour in AJAX calls during tests.

PR #3454, opened 2019-04-25
Fix Invalid Authenticity Token with Rails 5

javierm:
## References

* Backport commit AyuntamientoMadrid@01309528 from AyuntamientoMadrid#1947

## Objectives

Fix a bug which caused `ActionController::InvalidAuthenticityToken` after upgrading to Rails 5.

## Notes

* The exception was confirmed in development; I haven't checked it on production.
* It looks like we missed this commit when porting  AyuntamientoMadrid#1947 to CONSUL

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -4,6 +4,8 @@ class ApplicationController < ActionController::Base
   include HasFilters
   include HasOrders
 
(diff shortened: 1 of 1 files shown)


javierm:
Travis build failed due to issue #3451, which isn't related to this pull request.

Issue #3374, opened 2019-03-17
City map bug


ppeczek:
Updating map at URL: `admin/site_customization/images` doesn't update it at: `proposals/map`

PierreMesure:
Hey Piotr, I don't see any setting to update the map at the address you mention. You can change the map by adding the file named `map.jpg` at `app/assets/images/custom`.

This is what is recommended in the [docs](https://github.com/consul/docs/blob/master/en/customization/images.md).

If you want to use a PNG (which makes more sense sometimes), you'll need to modify the reference to the map directly on the page templates.

Dr4u95:
Hi, I did the same step  ppeczek and have the same issue ( The map appears in /proposals but in /proposals/map still is the default). 
  - I added an image in the file app/assets/images/custom/map.jpg , no works.
  - I modify the file map.jpg in app/assets/images/ and nothing too.

PierreMesure:
Not sure what to tell you besides making sure that you deployed the code and restarted the server. It might be that the assets haven't been precompiled correctly.

See our [demo instance](https://consul.demokratiskastader.se/proposals/map) and our [source code](https://github.com/digidemlab/consul/tree/master/app/assets/images/custom) for comparison.

Dr4u95:
First of all, i only install the Production Server. I did the step in the other post and compile the assets using :

   1. RAILS_ENV=production bundle exec rake assets:clobber 
   2. RAILS_ENV=production bundle exec rake assets:precompile

All the files that I touch, they modified, while the map (districts) disappears and does not show any.

decabeza:
Hi @ppeczek @PierreMesure @Dr4u95!

Admin users can change **map image** on `/admin/site_customization/images`:
![site_customization_images](https://user-images.githubusercontent.com/631897/57291838-86da5c80-70c0-11e9-87a3-3405fee61769.png)

But this custom image is not rendering in all pages with the map 😅. 

I create a PR https://github.com/consul/consul/pull/3472 to fix it! 🙏

Dr4u95:
@decabeza Ty, that works 💯  ... only forget close the ("map.jpg") in the second file 👍 .

@PierreMesure Ty for let me see your demo, I check a fault that had with the coordinates, should insert only the coordinates and not the entire label 👍 .

PR #3472, opened 2019-05-07
Always use map image from admin site customization images

decabeza:
## References

This closes https://github.com/consul/consul/issues/3374.

## Objectives

Admin users can change map image on `/admin/site_customization/images`.

This custom image is not rendering in all pages with the map. With this PR we always render custom map image from admin site customization images if it's present.

### DIFF
--- a/app/views/legislation/proposals/_geozones.html.erb
+++ b/app/views/legislation/proposals/_geozones.html.erb
@@ -2,5 +2,5 @@
 <h2 class="sidebar-title"><%= t("shared.tags_cloud.districts") %></h2>
 <br>
 <%= link_to map_proposals_path, id: "map", title: t("shared.tags_cloud.districts_list") do %>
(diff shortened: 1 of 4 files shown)


app/views/legislation/proposals/_geozones.html.erb
```diff
@@ -2,5 +2,5 @@
 <h2 class="sidebar-title"><%= t("shared.tags_cloud.districts") %></h2>
 <br>
 <%= link_to map_proposals_path, id: "map", title: t("shared.tags_cloud.districts_list") do %>
-  <%= image_tag("map.jpg", alt: t("shared.tags_cloud.districts_list")) %>
+  <%= image_tag(image_path_for("map.jpg", alt: t("shared.tags_cloud.districts_list")) %>
```
milovanderlinden: I believe there is one bracket missing from this statement.

PR #3477, opened 2019-05-08
Fix a bug where a category can't be created if it already exists as a tag

### LINKED ISSUE #2975: Can't create proposal topics (tags) if they have already been added by users
### Description

*Type*: Bug

capture d ecran 2018-10-17 a 14 44 21

I created a proposal as a normal user and wrote a couple of tags. Later on, when I tried to create a few tags on the admin backoffice, I realised that it doesn't work for the tags that have already been created by users.

It's impossible to know though because there is no error message and the user tags don't show up in the list. It seems that the list only shows the tags with `kind=category` and user tags don't have any `kind`.

### Current Behavior

See description.

### Expected Behavior

I understand that having official tags is useful. When an admin "creates" a tag in the admin backoffice, if it already exists as a user tag, its `kind` should be set to `category` to "promote" it.

### Steps to reproduce

1. Create a proposal with a new non-official tag.
2. In the admin page > Settings > Proposal topics, try to create this exact same tag.

### Definition of Done

- What do you think? Is my suggestion satisfying?

- As a bonus, I think destroying a tag in the admin section should not remove it altogether from the proposals which are tagged with it. This is the case today but my approach would be: 
```
if taggings_count = 0
  delete
else
  kind = nil
end
```
You want to be able to remove it from the suggestions without losing that information on the proposals.

capture d ecran 2018-10-17 a 14 50 06

PierreMesure:
## References

* Closes #2975

## Objectives

Fixes #2975 and adds a new spec to ensure that it works in this case

### DIFF
--- a/app/controllers/admin/tags_controller.rb
+++ b/app/controllers/admin/tags_controller.rb
@@ -9,13 +9,8 @@ class Admin::TagsController < Admin::BaseController
   end
 
   def create
-    @existing_tag = Tag.where(name: tag_params.first).first
-    if @existing_tag
-      @existing_tag.kind = "category"
-      @existing_tag.save
-    else
-      Tag.category.create(tag_params)
-    end
+    Tag.find_or_create_by!(name: tag_params["name"]).update!(kind: "category")
+
     redirect_to admin_tags_path
   end
 
--- a/app/controllers/admin/tags_controller.rb
+++ b/app/controllers/admin/tags_controller.rb
@@ -9,7 +9,13 @@ class Admin::TagsController < Admin::BaseController
   end
 
   def create
-    Tag.category.create!(tag_params)
+    @existing_tag = Tag.where(name: tag_params.first).first
+    if @existing_tag
+      @existing_tag.kind = "category"
+      @existing_tag.save
+    else
+      Tag.category.create(tag_params)
+    end
     redirect_to admin_tags_path
   end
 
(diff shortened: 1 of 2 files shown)


PierreMesure:
When you are done reviewing this PR, it would be interesting to get your opinion on the bonus improvement I talk about in #2975.

PierreMesure:
Rebased.

app/controllers/admin/tags_controller.rb
```diff
@@ -9,7 +9,14 @@ def index
   end
 
   def create
-    Tag.category.create!(tag_params)
+    @existing_tag = Tag.where(name: tag_params.first).first
```
javierm: The method `tag_params.first` is raising a warning:

```
DEPRECATION WARNING: Method first is deprecated and will be removed in Rails 5.1, as `ActionController::Parameters` no longer inherits from hash. Using this deprecated behavior exposes potential security problems. If you continue to use this method you may be creating a security vulnerability in your app that can be exploited. Instead, consider using one of these documented methods which are not deprecated: http://api.rubyonrails.org/v5.0.7.2/classes/ActionController/Parameters.html (called from create at /home/cjt/dev/consul/app/controllers/admin/tags_controller.rb:12)
```

How about `tag_params["name"]`? :thinking:

Maybe we could simplity the code and write:

```ruby
    Tag.find_or_create_by!(name: tag_params["name"]).update!(kind: "category")
```

What do you think?

PierreMesure:
Great review, I fixed everything you mentioned!

PR #3501, opened 2019-05-16
Validate poll token

javierm:
## References

* Backports commit AyuntamientoMadrid@a1f09188
* Backports commit AyuntamientoMadrid@1803f479
* Backports AyuntamientoMadrid#957

## Objectives

* Add message in case vote token is not properly received
* Bring code related to stats which will be used when we add our new stats layout
* Reduce differences between Madrid's fork and consul source codes


javierm:
Since this feature isn't used in Madrid anymore because it's only half-done, we've decided not to add this code to CONSUL.

PR #3518, opened 2019-05-21
Link DelayedJob's configuration file to Capistrano's shared folder

voodoorai2000:
## Context
The `delayed_job_config.rb` file is used to configure the background queue to send emails asyncronously with [DelayedJobs](https://github.com/collectiveidea/delayed_job).

## Objectives
Add `delayed_job_config.rb` to the capistrano shared folder so that the [Installer's specific DelayedJob configuration](https://github.com/consul/installer/blob/master/roles/queue/templates/delayed_job_config.rb) does not get overwritten after a deploy by the [default DelayedJob configuration](https://github.com/consul/consul/blob/master/config/initializers/delayed_job_config.rb).

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -21,7 +21,7 @@ set :log_level, :info
 set :pty, true
 set :use_sudo, false
 
-set :linked_files, %w{config/database.yml config/secrets.yml config/unicorn.rb config/environments/production.rb}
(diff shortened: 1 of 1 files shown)


javierm:
Closing this one since we're going to try a different approach where the installer doesn't add any extra files to the `shared` folder.

Maybe we could apply something like: https://github.com/voodoorai2000/installer/pull/8/files so delayed jobs are always enabled :thinking:. Or maybe we could move the configuration to `secrets.yml`, similar to SMTP settings in #3695.

PR #3533, opened 2019-05-24
Fix column order in schema file

javierm:
## Objectives

Use a `schema.rb` file consistent with the one generated during a fresh installation.

## Notes

When introducing Rails 5 and the dashboard, we generated the schema file without dropping the database first. When doing so locally, migrations aren't always executed in order, changing the order of the columns.


javierm:
Travis build failed because of issue #3035, not related to this pull request.

PR #3537, opened 2019-05-24
Set locales in test environment to avoid failed specs

### LINKED ISSUE #3526: Translatable spec should pass even if some languages are disabled
### Description

When disabling some of the languages used for the [translatable specs](https://github.com/consul/consul/blob/translations/spec/shared/features/edit_translatable.rb), for example Spanish, English, French or Portuguese, the corresponding tests fail. This should not happen.

I don't know which solution would be best:
- that the test suite overrides the available languages in `config/application.rb` and uses whichever languages it wants. With the risk that tests passing with the default languages fail with the ones actually used in production
- that the test suite dynamically picks the available languages and uses only these to test. To put it in other way, no harcoding of the languages used for test, just variables. And if there is only one language available, these translatable specs should not run as they test unused functions.

*Type*: Bug

Related issues: #3130
Related PRs: #3359 #3455 

### Steps to reproduce

- Remove `pt-BR` from the list of available languages
- Get the following message for 18 failing specs:
```
Failure/Error: select "Português brasileiro", from: "translation_locale"
     
     Capybara::ElementNotFound:
       Unable to find visible option "Português brasileiro" within #<Capybara::Node::Element tag="select" path="/html/body/div/div/div[3]/div/div[2]/div/div/div[3]/div[6]/select">
```

### Definition of Done
Specs should pass with any combination of available languages, even just one.

PierreMesure:
## References

- Closes #3526

## Objectives

A few different specs fail when you inactivate certain languages.

This is a problem because you want people to keep a clean test suite for their forks but they might want to disable languages.

We solve this by overriding the available languages for the test environment.

## Notes

- Right now, I am setting only the necessary languages (en, es, fr, nl, pt-BR). This is open to discussion.

### DIFF
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -4,14 +4,7 @@ Rails.application.configure do
   # Some tests require the following languages (en, es, fr, nl, pt-BR)
   # so we override the available languages for the test environment.
   config.i18n.default_locale = :en
-  available_locales = [
-    "en",
-    "es",
-    "fr",
-    "nl",
-    "pt-BR"
-  ]
-  config.i18n.available_locales = available_locales
+  config.i18n.available_locales = %w[de en es fr nl pt-BR zh-CN]
 
   # The test environment is used exclusively to run your application's
   # test suite. You never need to work with it otherwise. Remember that
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -1,6 +1,18 @@
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
 
+  # Some tests require the following languages (en, es, fr, nl, pt-BR)
+  # so we override the available languages for the test environment.
+  config.i18n.default_locale = :en
+  available_locales = [
+    "en",
+    "es",
+    "fr",
+    "nl",
+    "pt-BR"
+  ]
+  config.i18n.available_locales = available_locales
+
   # The test environment is used exclusively to run your application's
   # test suite. You never need to work with it otherwise. Remember that
   # your test database is "scratch space" for the test suite and is wiped
--- a/spec/features/remote_translations_spec.rb
+++ b/spec/features/remote_translations_spec.rb
@@ -5,8 +5,7 @@ describe "Remote Translations" do
   before do
     Setting["feature.remote_translations"] = true
     create(:proposal)
-    available_locales_response = ["ar", "de", "en", "es", "fa", "fr", "he", "it", "nl", "pl",
-                                  "pt", "sv", "zh-Hans", "zh-Hant"]
+    available_locales_response = %w[de en es fr pt zh-Hans]
     expect(RemoteTranslations::Microsoft::AvailableLocales).to receive(:available_locales).
                                                             and_return(available_locales_response)
   end
@@ -27,18 +26,14 @@ describe "Remote Translations" do
       end
 
       scenario "should display text in English" do
-        available_locales_with_fallback_en = [:ar, :de, :fa, :he, :nl, :pl, :sv]
-
-        visit root_path(locale: available_locales_with_fallback_en.sample)
+        visit root_path(locale: :de)
 
         expect(page).to have_css ".remote-translations-button"
         expect(page).to have_content "The content of this page is not available in your language"
       end
 
       scenario "should display text in English after parse key" do
-        available_locales_with_fallback_en = [:"zh-CN", :"zh-TW"]
-
-        visit root_path(locale: available_locales_with_fallback_en.sample)
+        visit root_path(locale: :"zh-CN")
 
         expect(page).to have_css ".remote-translations-button"
         expect(page).to have_content "The content of this page is not available in your language"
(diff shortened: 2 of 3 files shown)


javierm:
I really like the idea of setting only the necessary languages :+1:. We definitely need `pt-BR` (or a similar language) available in the test environment so we make sure we handle languages with a `-` properly. Maybe in the future we find a way for test to dynamically choose which languages to use among the ones enabled, but for now I think this idea is simple and does the job really well.

config/environments/test.rb
```diff
@@ -1,6 +1,18 @@
...
+  # so we override the available languages for the test environment.
+  config.i18n.default_locale = :en
+  available_locales = [
+    "en",
+    "es",
+    "fr",
+    "nl",
+    "pt-BR"
+  ]
+  config.i18n.available_locales = available_locales
```
javierm: Maybe we can omit the local variable and assign the value to `config.i18n.available_locales` directly :thinking:. Since there are only a few languages, we could also have everything on one line:
```
  config.i18n.available_locales = %w[en es fr nl pt-BR]
```
What do you think?

config/environments/test.rb
```diff
@@ -1,6 +1,18 @@
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
 
+  # Some tests require the following languages (en, es, fr, nl, pt-BR)
+  # so we override the available languages for the test environment.
+  config.i18n.default_locale = :en
```
javierm: I think we can omit this line, since we set this value in `spec/rails_helper`: https://github.com/consul/consul/blob/master/spec/rails_helper.rb#L16

javierm:
There's only a small problem after applying these changes to current `master`: the tests for remote translations assume a few more languages are available. @taitus What do you think about changing the remote translations spec to use less languages? For example:
```diff
diff --git a/spec/features/remote_translations_spec.rb b/spec/features/remote_translations_spec.rb
index 1ff98d9..d46fdfe 100644
--- a/spec/features/remote_translations_spec.rb
+++ b/spec/features/remote_translations_spec.rb
@@ -5,8 +5,7 @@ describe "Remote Translations" do
   before do
     Setting["feature.remote_translations"] = true
     create(:proposal)
-    available_locales_response = ["ar", "de", "en", "es", "fa", "fr", "he", "it", "nl", "pl",
-                                  "pt", "sv", "zh-Hans", "zh-Hant"]
+    available_locales_response = ["de", "en", "es", "fr", "pt", "zh-Hans"]
     expect(RemoteTranslations::Microsoft::AvailableLocales).to receive(:available_locales).
                                                             and_return(available_locales_response)
   end
@@ -27,18 +26,14 @@ describe "Remote Translations" do
       end
 
       scenario "should display text in English" do
-        available_locales_with_fallback_en = [:ar, :de, :fa, :he, :nl, :pl, :sv]
-
-        visit root_path(locale: available_locales_with_fallback_en.sample)
+        visit root_path(locale: :de)
 
         expect(page).to have_css ".remote-translations-button"
         expect(page).to have_content "The content of this page is not available in your language"
       end
 
       scenario "should display text in English after parse key" do
-        available_locales_with_fallback_en = [:"zh-CN", :"zh-TW"]
-
-        visit root_path(locale: available_locales_with_fallback_en.sample)
+        visit root_path(locale: :"zh-CN")
 
         expect(page).to have_css ".remote-translations-button"
         expect(page).to have_content "The content of this page is not available in your language"
@@ -53,7 +48,7 @@ describe "Remote Translations" do
       end
 
       scenario "with locale that has :es fallback" do
-        available_locales_with_fallback_es = [:es, :fr, :it]
+        available_locales_with_fallback_es = [:es, :fr]
 
         visit root_path(locale: available_locales_with_fallback_es.sample)
 
@@ -72,9 +67,7 @@ describe "Remote Translations" do
   end
 
   scenario "Not display remote translation button when locale is not included in microsoft translate client" do
-    not_available_locales = [:val, :gl, :sq]
-
-    visit root_path(locale: not_available_locales.sample)
+    visit root_path(locale: :nl)
 
     expect(page).not_to have_css ".remote-translations-button"
   end
```

config/environments/test.rb
javierm: Actually, we need this line because the test environment will fail to load if the default locale defined in `config/application.rb` isn't present in the available locales defined in the test environment.

So what do you think about removing the `I18n.default_locale = :en` line in `spec/rails_helper.rb`?

PierreMesure:
Hey Javier, great that you're looking at this! I can apply the first two changes you're requesting.

I'm not sure enabling more languages is a problem performance-wise so why don't we enable them all? If that's a problem, I suggest we rationalise the number of languages used in the specs in a future PR so we can already close this one.

javierm:
Hi, Pierre. Here's a patch which will (hopefully) make the remote translation tests compatible with this pull request. You can apply it using `git am 0001-Reduce-locales-used-in-remote-translations-tests.txt`.

[0001-Reduce-locales-used-in-remote-translations-tests.txt](https://github.com/consul/consul/files/3590428/0001-Reduce-locales-used-in-remote-translations-tests.txt)

Once you apply the patch, push the changes and then I think we can safely merge this pull request :tada:.

javierm:
@PierreMesure Awesome!  Thank you very much! :tada:

PR #3474, opened 2019-05-08
Remove deprecated translations code

javierm:
## References

* Adapts pull request #3327 to the current master branch
* Backports AyuntamientoMadrid#1985 and AyuntamientoMadrid#1996

## Objectives

* Remove obsolete columns which might still accidentally be used by active record scopes
* Fix incompatibilities between Globalize and Rails 5

PR #3578, opened 2019-06-03
Rename deprecated translation columns

javierm:
## References

* Adapts pull request #3327 to the current master branch
* Backports and adapts AyuntamientoMadrid#1985 and AyuntamientoMadrid#1996

## Objectives

* Rename obsolete columns which might still accidentally be used by active record scopes
* Fix incompatibilities between Globalize and Rails 5


javierm:
We've decide we'll either remove the columns (#3474) or keep them with the same name.

PR #3592, opened 2019-06-06
Don't show links to disabled budget results

javierm:
## References

* Backports AyuntamientoMadrid#2037

## Background

We were blocking the access to the budget results path when results were not enabled. However, links to the page were still available even if clicking them resulted in a "you are not authorized to access this page" message.

## Objectives

* Don't show links to disabled budget results


javierm:

PR #3359, opened 2019-03-13
Translate user generated content

voodoorai2000:
## References
- Update globalize gem #3194
- Initialize graphql after application initialization #3159 
- Pananoid translations (Translations soft deletion) #3201
- Globalize and I18n fallbacks #3239
- Adapt translatable shared specs #3240
- Adapt suggest script for translations #3241
- Translations sanitization #3242
- Microsoft translate client #3235 
- Avoid exception on paranoid translations migrations #3311
- Remove deprecated translations code #3327
- Budget translations update #3415
- Comments translations #3246
- Proposal translations #3243
- Responsive translation interface https://github.com/consul/consul/pull/3579
- Block translatable resource updates with no translations https://github.com/consul/consul/pull/3599
- Add Translations Interface docs https://github.com/consul/docs/pull/67
- Add Remote Translations docs https://github.com/consul/docs/pull/66

## Objectives
Make all user generated content translatable.

PR #3604, opened 2019-06-11
Test translations against latest master

javierm:
## Objectives

Check what the translations branch would look like when rebased against our current master branch.


javierm:
The test was successful. I've updated #3359 with the contents of this pull request.

PR #3606, opened 2019-06-11
Add Font Awesome icons

decabeza:
## Objectives

Add Font Awesome icons using [font-awesome-sass gem](https://github.com/FortAwesome/font-awesome-sass) to replace the customised icons font. 

Now you can use any icon from Free version of [Font Awesome](https://fontawesome.com/icons?m=free).

## Visual Changes

First row shows "old" icons and second row the Font Awesome version.

icons

## Notes

You can still use existing icons with `<span class="icon-X"></span>` markup but is highly recommended use the [font-awesome-sass gem rails helper](https://github.com/FortAwesome/font-awesome-sass#rails-helper-usage) for example:

`<%= icon("far", "address-book", class: "my-class") %>`


Gemfile.lock
```diff
@@ -185,6 +185,8 @@ GEM
     faraday (0.12.2)
       multipart-post (>= 1.2, < 3)
     ffi (1.10.0)
+    font-awesome-sass (5.8.1)
+      sassc (>= 1.11)
```
javierm: Re-reviewing this pull request I've noticed we include the `sassc` gem. We had some problems when adding the `sassc` to Madrid's preproduction server and had to undo those changes :thinking:. What do you think about waiting for pull request #3286 to be tested and merged before adding this one?

javierm: I've rebased this pull request now that we're using `sassc` :smile:

PR #3608, opened 2019-06-11
Remove system recounts in the admin section

javierm:
## References

* Backports AyuntamientoMadrid#2036

## Background

In the past, we had two numbers to track the total numbers of votes in a booth; what we called "final" recounts and "system" recounts. However, sytem recounts are not relevant anymore.

## Objectives

* Remove system recounts in the admin section, so administrators are not confused when they see two numbers for the same result.

## Visual Changes

Before these changes:

Poll recounts show final and system recounts
Booth recounts show final and system recounts

After these changes:

Poll recounts show only final recounts
Booth recounts show only final recounts

### DIFF
--- a/app/models/poll.rb
+++ b/app/models/poll.rb
@@ -81,6 +81,10 @@ class Poll < ApplicationRecord
     ends_at < timestamp
   end
 
+  def recounts_confirmed?
+    ends_at < 1.month.ago
+  end
+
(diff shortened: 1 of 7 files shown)


javierm:
Travis failure was caused by network issues in Travis and is not related to this pull request.

Issue #3610, opened 2019-06-11
Upgrading to v1.0.0 crashes the settings page


PierreMesure:
### Description

[comment]: # (Feature or Bug? eg. Type: Bug)
*Type*: Bug

### Current Behavior
[comment]: # (Describe what actually happened.)

When trying to access the settings page, an error 500 is thrown because some settings are missing in the table, resulting in `@uploads_settings` being nil.

### Expected Behavior
[comment]: # (Describe what you expected to happen.)

- The new upgrade task should make sure that all the new settings are created.

- Moreover, the whole settings management should be more robust. It should take the hardcoded list present in [models/setting.rb](https://github.com/consul/consul/blob/22787151f55529cefa0ea23418ffc356b5d49c21/app/models/setting.rb#L79) as a primary source of information so that any missing or corrupted key in the table doesn't lead to an error 500 or to a missing key in the interface. If a key or its value is missing, it could be recreated from the default value on the fly.

### Steps to reproduce
[comment]: # (Describe how we can replicate the bug step by step.)

- Install v1.0.0.beta or v0.19 with dev seeds
- Upgrade to v1.0.0
- Go to `admin/settings` and enjoy an error 500.
- To fix it, just run the `settings:add_new_settings` task.

### Definition of Done

I'm shipping a PR to fix the first point. The second needs to be done too to close this task but more discussion is probably needed with the developers of the new system.

voodoorai2000:
Hi @PierreMesure!,

Thank you so much for letting us know about this. We’re looking into it. :pray:

Just to make sure, you did use Capistrano to deploy version 1.0.0 from 1.0.0-beta right?. That should have executed the first task `add_new_settings` 
https://github.com/consul/consul/blob/master/config/deploy.rb#L48

And you upgraded from 0.19 to 1.0.0-beta and then to 1.0.0 right? When upgrading from 0.19 to 1.0.0-beta the `rename_setting_keys` task should  have already been executed.
https://github.com/consul/consul/releases/tag/1.0.0-beta

PierreMesure:
Hey @voodoorai2000,

Yes, the upgrade went smoothly when using Capistrano. But if you are just upgrading a local test machine or Heroku instances, `add_new_settings` will never be run. It's great that it's in the Capistrano  deployment flow but in my opinion, it should also be present in the upgrade script. It doesn't hurt to run it several times and it covers all cases.

voodoorai2000:
Hey @PierreMesure,

[...]

PR #3611, opened 2019-06-11
Adding missing subtasks to upgrade task and adding info log

PierreMesure:
## References

- #3610

## Objectives

Fixes a bug when upgrading to v1.0.0 because the taks `add_new_settings` is not run and upload settings are missing, crashing the whole `admin/settings` page

## Visual Changes

- None

## Notes

- Please patch quickly to avoid any problem for other people who will try to update to v1.0.0 in the coming days
- The settings management needs to be made more robust, see more explanation in #3610.


javierm:
@PierreMesure Merged :tada: Thank you very much!

PR #3615, opened 2019-06-12
Test votation types

javierm:
## Objectives

Check a rebase


javierm:
Checked and applied to #3586.

PR #3618, opened 2019-06-17
Use Date.current and Time.current

javierm:
## Background

Using `Date.today` and `Time.now` might lead to inconsistencies if the time zone the application uses is not the same as the system time zone.

## Objectives

Fix all possible date errors we might have when the computer the application is running uses a different time zone than the one we define in `config.time_zone`.


javierm:
Travis failure is not related to this pull request, but to issue #3642.

javierm:

Issue #3559, opened 2019-05-28
Hide related content proposals when these proposals are retired by their authors


decabeza:
## Background
The debates and proposals have a "related content" section where users can add related proposals and other content. Proposals can be retired by the proposal author through the "Edit my proposal" menu in the author's proposal dashboard.

## What we need
If a proposal is retired by his/her author, it should disappear from all "related content" sections where it appears.

nileshkumar:
@decabeza : I will start working on it..

jaredholdcroft:
Hi. I'd like to take this if no-one else is working on it?

javierm:
@jaredholdcroft Thanks! Go ahead :smile:.

karunya-tota:
@javierm @jaredholdcroft Is this issue still up for grabs? First-time contributor here!

jaredholdcroft:
@karunya-tota Yeah, I didn't get chance to do anything here :(

javierm:
@karunya-tota Sorry for the late reply!

If you'd like to take this issue, you're more than welcome! Thanks a lot! :pray:

karunya-tota:
Thank you! I hope to start taking a look today!

matisnape:
@javierm Hi, is this is still up for grabs? 

And are you joining Hacktoberfest with the new rules? It's now opt-in for open source projects, you just need to add hacktoberfest topic https://hacktoberfest.digitalocean.com/hacktoberfest-update

javierm:
@karunya-tota Are you still working on this issue? If not, @matisnape, sure, go ahead :wink:.

matisnape:
@javierm done, although dunno if that's the good way to do it :)

javierm:
> On another note - when I run Rubocop locally, there's much more warnings than from Hound here - how come? It's mostly too long lines in specs.

Hound only barks when there are offenses in the changes you make; it doesn't say anything about code which this pull request doesn't touch. Otherwise all contributors would get hundreds of comments about long lines whenever they open a pull request, which would be really annoying.

Also note the long lines Rubocop rule uses the "refactor" severity, meaning we aren't very strict about that rule, although it's usually better to follow it.

PR #3619, opened 2019-06-18
show only not-retired proposal in Related content section

nileshkumar:
## References

Issue #3559

## Objectives

The debates and proposals have a "related content" section where users can add related proposals and other content. Proposals can be retired by the proposal author through the "Edit my proposal" menu in the author's proposal dashboard.

## Visual Changes

If a proposal is retired by his/her author, it should disappear from all "related content" sections where it appears.

### DIFF
--- a/app/controllers/proposals_controller.rb
+++ b/app/controllers/proposals_controller.rb
@@ -25,6 +25,7 @@ class ProposalsController < ApplicationController
 
   def show
     super
+    @proposal = @proposal.not_retired  #showing only proposal which is not Retired
     @notifications = @proposal.notifications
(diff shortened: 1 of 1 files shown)


javierm:
Hi, @nileshkumar. Thank you very much for this pull request.

Unfortunately, it looks like many tests are failing because the method `@proposal.not_retired` doesn't exist, so I'm afraid we cannot approve it :thinking:.

Also note the pull request should have tests for the scenarios the code is trying to fix. In this case I'd recommend you start writing the tests which check debates and proposals don't have retired proposals in their "related content" section.

PR #3620, opened 2019-06-19
Use Rails 5.1 conventions in migrations and specs

javierm:
## Background

We forgot to add these changes to pull requests which were in development before we upgraded to Rails 5.

## Objectives

* Make migrations, models and specs compatible with Rails 5.1


javierm:

PR #3623, opened 2019-06-20
Remove commented code in Capistrano tasks

javierm:
## Objectives

* Make the code easier to read by removing unused and commented code
* Remove an unused reference to Ruby version 2.3.2, which would make it a bit harder to upgrade Ruby


javierm:
Closed in favor of #3627.

PR #3624, opened 2019-06-21
Add support for Errbit self-hosted exception management

javierm:
## References

* [Errbit - the open source error catcher that's Airbrake API compliant](https://github.com/errbit/errbit/)

## Background

We're currently using the `rollbar` gem, so anyone can handle their exceptions with Rollbar.

However, managing exceptions with Rollbar means sending data about your application to a third party, which is something some users might be concerned about.

## Objectives

* Add support for self-hosted exception mangement with Errbit

## Notes

* To install and configure Errbit, check their [installation](https://github.com/errbit/errbit/#installation) and [notes on deployment](https://github.com/errbit/errbit/blob/master/docs/deployment.md) documentation
* The `airbrake` gem also works with Airbrake's hosting service, the same way the `rollbar` gem works with Rollbar's services


javierm:
Travis failure is not related to this pull request and will be addressed shortly.

PR #3627, opened 2019-06-22
Upgrade ruby to 2.4.6

javierm:
## Background

Many gems have dropped support for Ruby 2.3, including Rails 6, which is going to be released soon.

## Objectives

* Upgrade Ruby to a more recent and well supported version


javierm:

PR #3628, opened 2019-06-22
Simplify calls to render partial

javierm:
## Objectives

* Simplify the code rendering partials
* Make the code more consistent, since most of our calls to partial rendering omit the `partial` and `locals` keys


javierm:

PR #3630, opened 2019-06-23
Fix typos and remove unused code

javierm:
## Objectives

* Use translations instead of plain English texts
* Remove unused code
* Fix label in form to add own answer
* Fix typos


javierm:
Travis failures are not related to this pull request, but due to network timeouts probably when connecting to map and translations services.

PR #3634, opened 2019-06-24
Add title to differentiate signature sheets

cassianoblonski:
## References

Related to #3557 

## Objectives

Add a title to differentiate between signature sheets on `admin/signature_sheets`

## Visual Changes
image

PR #3647, opened 2019-06-28
Add option to soft delete signature sheets

### LINKED ISSUE #3543: Option to delete signature sheets
## Background
Proposals can be supported by signatures, and those supports can be included in the system through the "Signature sheets" menu `/admin/signature_sheets` In case a sheet is wrongly included (for example associated to the wrong proposal) it can't be deleted from the interface.

## What we need
Include a button "Delete sheet and supports" next to each included sheet of supports. When clicking it will delete the supports from the proposal and will mark the sheet as deleted. But will keep showing the information that was shown before deleting, as a record for accountability of the process.

cassianoblonski:
## References

* Resolves #3543
* Pull request #3634 should probably be merged first

## Objectives

Add an option to soft delete signature sheets on admin/signature_sheets.

## Visual Changes
image

### DIFF
--- a/app/controllers/admin/signature_sheets_controller.rb
+++ b/app/controllers/admin/signature_sheets_controller.rb
@@ -1,4 +1,5 @@
 class Admin::SignatureSheetsController < Admin::BaseController
+  before_action :set_signature_sheet, only: [:show, :destroy]
 
   def index
     @signature_sheets = SignatureSheet.all.order(created_at: :desc)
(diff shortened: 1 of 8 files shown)


javierm:
@cassianoblonski Thanks for this pull request! It looks like it would better to merge #3634 first :thinking:, so if you don't mind, I'll have a look at this pull request after we merge #3634.

javierm:
Closing since the original author doesn't seem to be available anymore.

Issue #3560, opened 2019-05-28
Allow proposals and participatory budgeting projects to be search by id in the search forms


decabeza:
## What we need
The admin interface and the user's frontend search forms for  proposals and participatory budgeting projects should allow to search content by introducing **the id of the content**.

eloisemaun:
I'll take this one, if it's not already been claimed. Long time developer, first time open source github contribution. :)

voodoorai2000:
Great! 🎉
All yours @eloisemaun! 👏

eloisemaun:
Some clarification questions:

1. Which of the following functionality pieces were you hoping for:
    1. the basic search form to also look at the ID when it is searching, and/or 
    1. a "Search by ID" field added to the advanced search
1. Should the search functionality only look at the actual database ID (17) or the parameterized ID that is shown in the URL (17-sint-ipsa-non-in-saepe) ? Or were you referring to the code (MAD-2019-06-17)?

eloisemaun:
I implemented searching by the ID and by the parameterized ID+title (for proposals only) for the search field in the nav bar. If you would like a "Search by ID" field implemented for the advanced search, too, just say the word. Otherwise, there's a pull request ready for this.

PR #3648, opened 2019-06-30
ID searching for Proposals, Investments

eloisemaun:
Investment searching is now responsive to investment.id. Proposals are now responsive to proposal.id and to the string returned by proposal.to_param().

## References

Closes: #3560

## Objectives

To allow users and admins to input ID's or the parameterized ID+title (for proposals) in the search bar for Participatory Budget Investments and Proposals.

## Visual Changes

No visual changes. Functional only.

## Notes

For this functionality to work, contributors with existing local copies of the project will need to reset the database in order to add these additional columns to the tsvector column.

```
bin/rake db.reset
bin/rake db.migrate
bin/rake db.dev_seed
```

### DIFF
--- a/app/models/budget/investment.rb
+++ b/app/models/budget/investment.rb
@@ -193,6 +193,7 @@ class Budget
 
     def searchable_values
       { title              => "A",
+        id.to_s            => "A",
         author.username    => "B",
         heading.try(:name) => "B",
         tag_list.join(" ") => "B",
--- a/app/models/proposal.rb
+++ b/app/models/proposal.rb
@@ -122,8 +122,10 @@ class Proposal < ApplicationRecord
 
   def searchable_values
     { title              => "A",
+      id.to_s            => "A",
       author.username    => "B",
       tag_list.join(" ") => "B",
+      to_param()         => "B",
       geozone.try(:name) => "B",
       summary            => "C",
       description        => "D"
@@ -131,6 +133,7 @@ class Proposal < ApplicationRecord
   end
 
   def self.search(terms)
+    terms = terms.to_s if terms.is_a? Numeric
     by_code = search_by_code(terms.strip)
     by_code.present? ? by_code : pg_search(terms)
   end
--- a/spec/models/budget/investment_spec.rb
+++ b/spec/models/budget/investment_spec.rb
@@ -647,6 +647,12 @@ describe Budget::Investment do
         expect(results).to eq([budget_investment])
       end
 
+      it "searches by id" do
+        budget_investment = create(:budget_investment, title: "free willy")
+        results = described_class.search(budget_investment.id)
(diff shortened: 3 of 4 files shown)


PierreMesure:
Hey @eloisemaun, great addition!

Do you know if there would be any way to apply this change to existing installations without needing to reset the database?

eloisemaun:
@PierreMesure - I am not a Postgres expert, so it is possible that there is a way to apply it without resetting with additional SQL statements. But as this is in development, it is cleaner to reset the database so Postgres can fill in those TSvector columns fresh from the start. If this is a problem and it would be preferable to have a set of SQL to run to fill in those columns, just let me know.

PierreMesure:
Hi @eloisemaun, not a Postgres expert either.

From what I understand after digging in the world of Ruby/Postgres full text search, the tsvector column needs to be updated to take into account the changes. But it seems like we should also be done whenever the content changes (for example editing a proposal title) and I cannot find where we do that in the source code.

@voodoorai2000, you're the one who implemented most of the logic here. Any reason why there is no trigger for tsvector updates?

Bonus question: why not using [ransack](https://github.com/activerecord-hackery/ransack) instead of a complex implementation of [pg_search](https://github.com/Casecommons/pg_search)? Or even [searchkick](https://github.com/ankane/searchkick) (which seems a bit overkill but a switch happens...)?

javierm:
> why not using ransack instead of a complex implementation of pg_search? Or even searchkick (which seems a bit overkill but a switch happens...)?

Hi, Pierre.

I haven't checked Ransack for a while (back then it was MetaSearch, actually). We're using `pg_search` because it makes it easy to give more relevance to results containing the search query in the title.

Regarding searchkick, the main reason we're not using it is it would require installing and running ElasticSearch, which IMHO would make CONSUL a bit harder to install and (more importantly) maintain :thinking:.

voodoorai2000:
Hey,

Thank you so much @eloisemaun this looks great! 🎉

Sure @PierreMesure, we've implemented [search_cache](https://github.com/consul/consul/blob/master/app/models/concerns/search_cache.rb#L5) concern which is included in every [searchable](https://github.com/consul/consul/blob/master/app/models/concerns/searchable.rb#L6) model. That deals with updating the tsv vector after saving a record. 👍 

The caching system is sort of implied in the [specs](https://github.com/consul/consul/blob/836f9ba7b35971365e2ce9f16be05e8d0da1d06c/spec/models/proposal_spec.rb#L460), however we probably should add some explicit model specs for the `search_cache` concern. 👌 

@eloisemaun would you mind trying out a rake task where you `save` every record that should update its tsv vector? 🙏That would make it production ready 👌We will add it it to the next release's [upgrade rake task](https://github.com/consul/consul/pull/3590) 😊

mrdocR (approved):

PR #3656, opened 2019-07-08
Fix text confirming investment heading support

javierm:
## Background

We accidentally removed the `count` option in the confirm group text in commit 55fb14ac, which made the translation return a hash.

## Visual Changes

Before this change:

Confirmation text shows Object object

After this change:

Confirmation text shows a useful message


javierm:
Travis build failed because of the issue fixed by #3641, so the failure is not related to this pull request.

Issue #3645, opened 2019-06-25
Flaky spec: Admin Active polls Add


javierm:
## Failure

**Travis failed builds**:

* [Build 29469, job 2](https://travis-ci.org/consul/consul/jobs/535358057)
* [Build 30734, job 1](https://travis-ci.org/consul/consul/jobs/539814361)
* [Madrid's build 11452, job 1](https://travis-ci.org/AyuntamientoMadrid/consul/jobs/524024410)

**Message**:

```
  1) Admin Active polls Add
     Failure/Error: locator = find("label", text: locator)[:for] if page.has_css?("label", text: locator)
     Selenium::WebDriver::Error::StaleElementReferenceError:
       stale element reference: element is not attached to the page document
         (Session info: headless chrome=74.0.3729.169)
         (Driver info: chromedriver=2.36.540471 (9c759b81a907e70363c6312294d30b6ccccc2752),platform=Linux 4.4.0-101-generic x86_64)
     # ./spec/support/common_actions/verifications.rb:56:in `fill_in_ckeditor'
     # ./spec/features/admin/poll/active_polls_spec.rb:22:in `block (2 levels) in <top (required)>'
```

## Cause

We're not sure 100% sure about the cause, since the error message seems to imply the label element was in the document but then was removed from it.

However, this is one of the few tests (or the only one) where we try to fill in CKEditor immediately after clicking a link leading to a form. And we've also had issues with this test in the branch upgrading to Rails 5.1 because CKEditor takes too long to load and the textarea is filled before a CKEditor replaces it. CKEditor takes more time to load since we upgraded to Rails 5 (at least in the development environment), so it might be related.

One possible fix would probably be checking CKEditor is loaded properly before trying to fill it. However, using CKEditor inline like suggested in pull request #3659 seems to solve the problem because it makes CKEditor load faster. This test fails about once every 500 attempts (check [javierm's build 457](https://travis-ci.org/javierm/consul/builds/556501452), [javierm's build 463](https://travis-ci.org/javierm/consul/builds/556558520) or [javierm's build 465](https://travis-ci.org/javierm/consul/builds/556567839) for some examples), but after applying the changes in #3659 I've executed the test more than 10,000 times and it hasn't failed so far. So for now we'll assume it was caused by CKEditor taking too long to load and it's solved by making CKEditor faster.

PR #3657, opened 2019-07-08
Precompile CKEditor dialog plugins

javierm:
## References

* Based on commit 54c82a53 from pull request #2711
* Might be related to issue #3645, although it's unlikely since tests don't try to load any CKEditor dialogs

## Objectives

Make CKEditor dialogs for images, links and tables work on all production environments.

## Notes

This bug only affects production environments. These dialogs load (although they might take time to compile) correctly on the development and test environments, so this pull request doesn't add any tests.


javierm:

PR #3663, opened 2019-07-15
Load custom locales after everything is loaded

javierm:
## Background

The line:

```
config.i18n.load_path += Dir[Rails.root.join("config", "locales", "**", "*.{rb,yml}")]
```

Was adding every locale under the `config/locales` folder, including the ones inside `config/locales/custom/`. The files were loaded in the same order as listed using `ls -f`.

So if the custom locales were loaded before the folder `config/locales/#{I18n.locale}`, the default locales would override the custom ones, and we want the custom locales to override the default
ones so CONSUL is easier to customize.

## Objectives

Make sure custom locales are loaded after every other locale is loaded.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -54,7 +54,8 @@ module Consul
       "it"    => "es",
       "pt-BR" => "es"
     }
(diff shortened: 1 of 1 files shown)


javierm:
Travis failure is not related to this pull request and will be fixed by #3661.

PR #3680, opened 2019-08-23
Add missing link and button on management interface

PierreMesure:
## Objectives

- Adds the notification bell and the link to go back to the homepage in the header of the management interface.

## Visual Changes

Before:

Capture d’écran 2019-08-23 à 23 41 46

After:

Capture d’écran 2019-08-23 à 23 42 08

### DIFF
--- a/app/views/layouts/management.html.erb
+++ b/app/views/layouts/management.html.erb
@@ -12,6 +12,10 @@
       <div class="top-links">
         <div class="expanded row">
           <%= render "shared/locale_switcher" %>
+
+          <%= link_to root_path, class: "float-right" do %>
+            <%= t("admin.dashboard.index.back", org: setting["org_name"]) %>
+          <% end %>
         </div>
       </div>
 
@@ -35,6 +39,7 @@
               <div class="top-bar-right">
                 <ul class="menu" data-responsive-menu="medium-dropdown">
                   <%= render "shared/admin_login_items", current_user: manager_logged_in %>
+                  <%= render "layouts/notification_item", current_user: manager_logged_in %>
                   <%= render "devise/menu/login_items", current_user: manager_logged_in %>
                 </ul>
               </div>


microweb10:
Hi Pierre! Thank you very much for spotting this! Of course, it would be great to have that content in the management section too ☺️ 

I think I would go for reusing the `_admin_header` partial we already have to avoid duplication of the code 🤔 
Please have a look at this commit https://github.com/consul/consul/commit/474c9c5f6e9aaa4bf046db6e46dda5f623af8e7e

Although some modifications need to be done, such as creating some variables and adding some translations...

I'm curious about what Javi thinks about this, what approach he likes more, or if he maybe prefers a totally different approach 😅

PierreMesure:
I like your approach! It's cleaner!

I think we could probably also move the translations manually from the place where they already exist.

javierm:
From a technical point of view, a partial sounds good :smile:.

What I'm not sure about is whether the link is currently missing because we forgot to add it, or it was left this way on purpose because managers play a specific role and they wouldn't want to visit the public area :thinking:. @decabeza Do you know what's the reason?

decabeza:
@PierreMesure @javierm indeed, these links are not visible on purpose. The "management" section is intended to be used by public administration officials to create/manage content on behalf of other users, so it's not necessary for them to visit the front page or use the notifications.

In any case, I don't have a problem with them being added. ☺️

PierreMesure:
In our cities, the people using this interface are often admins (unlike a big city like Madrid, we only have a few civil servants doing most of the admin and management work) and some of them were a big confused because they are used to see these buttons and suddenly they are missing even though it feels like they are on the admin side.

So I'd say we should add them for consistency's skull.

PierreMesure:
I can update the PR with a partial in the coming weeks.

javierm:
@PierreMesure That's a fair point. So I'm fine with adding the link :+1:. I'll review the pull request once it's updated (or a few weeks later; I'm afraid I can't promise anything :sweat_smile:).

javierm:
@PierreMesure Once again thanks for bringing this up :smile:, and @microweb10 great suggestion about the partial!

I've applied that suggestion and added a couple of test expectations in pull request #4734; closing this one.

PR #3694, opened 2019-09-11
Use puma instead of unicorn

javierm:
## References

* Pull request #3517 provides a configuration in case we decide to keep both unicorn and puma

## Objectives

* Replace unicorn with puma, a server which is more advanced and we already use in the development environment.
* Simplify configuration files thanks to `capistrano3-puma`

## Release notes

:warning: We've replaced unicorn with puma. There's no need to change your nginx configuration if you were using a standard unicorn configuration. However, on your nginx configuration you might want to change the line saying:

```
server unix:/home/elretirao/consul/shared/sockets/unicorn.sock;
```

To:
```
server unix:/home/elretirao/consul/shared/tmp/sockets/puma.sock fail_timeout=0;
```

So when reading that file it's obvious you're using puma and not unicorn.


config/deploy.rb
```diff
@@ -94,3 +95,16 @@ def deploysecret(key)
...
 end
+
+desc "Convert files which were unicorn sockets into symbolic links to the puma socket, \
+      so legacy nginx configurations pointing to the unicorn socket keep working"
+task :replace_unicorn_with_puma do
+  on roles(:app) do
+    with rails_env: fetch(:rails_env) do
+      execute "mkdir -p #{shared_path}/tmp/sockets; true"
+      execute "mkdir -p #{shared_path}/tmp/pids; true"
+      execute "ln -sf #{shared_path}/tmp/sockets/puma.sock #{shared_path}/sockets/unicorn.sock; true"
```
javierm: This will give a warning while deploying if the folder `#{shared_path}/sockets` does not exist :thinking:.

PR #3716, opened 2019-09-22
Edit Budget Investment only in accepting phase

### LINKED ISSUE #3691: Let the users edit his budget investment proposals only during the acceptance phase
### User Story
Hello, im Daniel from Mexico and I worked in the project of implementing consul in the city of San Pedro in Monterrey
https://decide.sanpedro.gob.mx

### Why is this Needed?
During the process of generating proposals we get complaints from many people who said they could not edit their proposals because they had a spelling error or forgot to upload an image.
and they didn't like the idea of erasing everything and starting from scratch

### Description

The idea is that the user can only edit their proposals during the acceptance phase, during the following phases, access to this will be limited

### Current Behavior

We already have a fix for this, so I would like to do a PR for this based on what we implement in San Pedro

denialtorres:
## References

resolves #3691 

> let the users modify his budget investments only in the accepting phase 

## Visual Changes

The `Edit` option will appear under the `Delete` button if they user goes to `My Content >> Investments` only during the `Accepting Phase`, after that the button will disappear 

image

if they try to go to the edit url during another phase or without being the author of the proposal then they will be rejected

Screen Shot 2019-09-22 at 3 50 44 PM

## Notes

this is only working to regular users, if you're an admin you still will be able to edit your proposals in the following phases


app/views/budgets/investments/edit.html.erb
```diff
@@ -0,0 +1,6 @@
+<div class="budget-investment-new row">
+  <div class="small-12 medium-9 column">
+    <h1><%= t("management.budget_investments.edit") %></h1>
+    <%= render '/budgets/investments/form_edit', form_url: budget_investments_path(@budget) %>
```
javierm: Is there a way to reuse the `/budgets/investments/form` partial here? Then we could remove the `form_edit` file and so remove the duplication between both files :thinking:.

config/routes/budget.rb
```diff
@@ -1,6 +1,6 @@
 resources :budgets, only: [:show, :index] do
   resources :groups, controller: "budgets/groups", only: [:show]
-  resources :investments, controller: "budgets/investments", only: [:index, :new, :create, :show, :destroy] do
+  resources :investments, controller: "budgets/investments", only: [:index, :new, :create, :show, :destroy, :edit, :update] do
```
javierm: Since now we're using all seven default controller actions, we could probably remove the `only` key and write: `resources :investments, controller: "budgets/investments" do`

spec/models/abilities/common_spec.rb
```diff
@@ -259,6 +259,11 @@
       it { should_not be_able_to(:destroy, own_investment_in_selecting_budget) }
       it { should_not be_able_to(:destroy, own_investment_in_balloting_budget) }
 
+      it { should be_able_to(:edit, own_investment_in_accepting_budget) }
+      it { should_not be_able_to(:edit, own_investment_in_reviewing_budget) }
+      it { should_not be_able_to(:edit, own_investment_in_selecting_budget) }
+      it { should_not be_able_to(:edit, own_investment_in_balloting_budget) }
+
```
javierm: Would it be possible to add a feature test as well, checking users can correctly click on the link to edit a form, fill in the form, and save the changes? Thanks a lot! :pray:

spec/models/abilities/common_spec.rb
```diff
@@ -259,6 +259,11 @@
       it { should_not be_able_to(:destroy, own_investment_in_selecting_budget) }
       it { should_not be_able_to(:destroy, own_investment_in_balloting_budget) }
 
+      it { should be_able_to(:edit, own_investment_in_accepting_budget) }
+      it { should_not be_able_to(:edit, own_investment_in_reviewing_budget) }
+      it { should_not be_able_to(:edit, own_investment_in_selecting_budget) }
+      it { should_not be_able_to(:edit, own_investment_in_balloting_budget) }
```
javierm: Another test checking a user is not able to edit another author's proposal would be awesome! :smile:

javierm: Sorry, I meant to **update** another author's proposal :smile:.

app/controllers/budgets/investments_controller.rb
```diff
@@ -77,6 +77,21 @@ def create
       end
     end
 
+    def edit
+      if @investment.author == current_user
```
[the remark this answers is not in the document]
denialtorres: hey, yes, you're right on this 

but when i add `author_id: user.id` inside `abilities/common.rb`
It only works for normal users, the administrator is able to enter to the edit view and modify other users' proposals, is this the desired behavior?

app/views/budgets/investments/edit.html.erb
denialtorres: hey yes, I was having problems at first with the `translatable_fields` so I decided to use a separate form, but it seems to be working now 🤷

spec/models/abilities/common_spec.rb
denialtorres: yes, I think I can do it, are you using capybara for the features?

javierm: Yes. I guess the right places to add the test would be `spec/features/budgets/investments_spec.rb`.

app/controllers/budgets/investments_controller.rb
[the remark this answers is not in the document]
javierm: Good catch! :+1: No, administrators should only be able to edit investments in the admin section.

Maybe we could have two different permissions, like `update` and `admin_update` :thinking:, so users could update their own investments and aministrators could "admin update" everyone's investments :thinking:. Any suggestions?

denialtorres:
hi @javierm, it seems to be ready now 

do you know why i keep getting errors in the `coveralls` part ?? regards!

javierm:
Great! I'll have a look at the changes during the week.

Lately coveralls integration with our Travis builds isn't working so well :thinking:. I've just disabled it, since as it is right now it's very confusing.

denialtorres:
hey, hi @javierm the changes are ready, can you check them when you have a chance?, you were right in the part of the form, it is much better if I put `budget_investment_path (@budget, @investment` in the edit view.

also i add a new scenario to activate the validation error and make sure that part is working

PR #3726, opened 2019-09-25
Add more tests to calculate winners

javierm:
## Objectives

* Use test descriptions which show the specific case they're testing
* Add test cases we weren't testing


javierm:

PR #3767, opened 2019-10-11
Mark fields as required automatically

javierm:
## References

* Depends on pull request #4087

## Objectives

* Improve user experience when filling in forms
* Simplify labels stating whether a field is optional

## Visual Changes

Now labels have an asterisk indicating a field is required:

proposal title and summary are marked as required, while proposal text is not

## Notes

* There should be a text saying "fields mark with * are required" at the beginning of each form with required fields
* Fields with conditions or custom validation rules are not affected; for example, the map location of a proposal is required unless the field "this proposal doesn't have a concrete location" is checked


javierm:
These changes aren't compatible with the form changes we've done in pull requests #4580 and #4225. Closing for now. We might have to re-visit this issue in the future, since the forms in the admin section still don't show which fields are mandatory.

PR #3781, opened 2019-10-20
Don't send newsletters to unconfirmed accounts

### LINKED ISSUE #3761: Consul sends newsletters to unconfirmed accounts
### Description

*Type*: Bug

I checked the logs of my e-mail server and one of the bounces is due to an incorrect e-mail address (it ends with "@vmail.com" instead of "@gmail.com". What happened is the person created a first account with a typo in the e-mail, never got the confirmation mail and then created a new one with the correct e-mail.

### Current Behavior

To my surprise, sending a newsletter to all the users of the platform also sends it to this incorrect address (hence the bounce).

### Expected Behavior

Mails should never be sent to users that haven't even confirmed their e-mail address. It's creating bounces, ineffective and even if the address was correct, someone who didn't bother to finish creating their account probably doesn't want to be bothered by e-mails later on.

### Steps to reproduce

- Sign up but don't confirm your account.
- Send a newsletter to all users as an admin.
- You should receive an e-mail to your unconfirmed address.

### Definition of Done

- Unconfirmed accounts don't receive anything
- Check for other places where unconfirmed users might be included when they shouldn't.

mohsinkhansymc:
## References

* Closes #3761 

## Objectives

Prior to this PR, an unconfirmed user was considered as active. After this change a user who has confirmed his account is an active user.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -67,7 +67,7 @@ class User < ApplicationRecord
     where(document_type: document_type, document_number: document_number)
   end
   scope :email_digest,   -> { where(email_digest: true) }
-  scope :active,         -> { where(erased_at: nil) }
+  scope :active,         -> { where(erased_at: nil).where.not(confirmed_at: nil) }
   scope :erased,         -> { where.not(erased_at: nil) }
   scope :public_for_api, -> { all }
   scope :by_comments,    ->(query, topics_ids) { joins(:comments).where(query, topics_ids).distinct }
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -67,7 +67,7 @@ class User < ApplicationRecord
     where(document_type: document_type, document_number: document_number)
   end
   scope :email_digest,   -> { where(email_digest: true) }
-  scope :active,         -> { where(erased_at: nil).where.not(confirmed_at: nil) }
+  scope :active,         -> { where(erased_at: nil) }
   scope :erased,         -> { where.not(erased_at: nil) }
   scope :public_for_api, -> { all }
   scope :by_comments,    ->(query, topics_ids) { joins(:comments).where(query, topics_ids).distinct }
--- a/lib/user_segments.rb
+++ b/lib/user_segments.rb
@@ -10,7 +10,7 @@ class UserSegments
                 not_supported_on_current_budget]
 
   def self.all_users
-    User.active
+    User.where(erased_at: nil).where.not(confirmed_at: nil)
   end
 
   def self.administrators
--- a/lib/user_segments.rb
+++ b/lib/user_segments.rb
@@ -10,7 +10,7 @@ class UserSegments
                 not_supported_on_current_budget]
 
   def self.all_users
-    User.where(erased_at: nil).where.not(confirmed_at: nil)
+    User.active.where.not(confirmed_at: nil)
   end
 
   def self.administrators
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -399,6 +399,15 @@ describe User do
         expect(User.active).not_to include(user3)
       end
 
+      it "returns users who have confirmed their account" do
+        user1 = create(:user, erased_at: nil, confirmed_at: Time.current)
+        user2 = create(:user, erased_at: nil, confirmed_at: Time.current)
+        user3 = create(:user, erased_at: nil, confirmed_at: nil)
+
+        expect(User.active).to match_array [user1, user2]
+        expect(User.active).not_to include(user3)
+      end
+
     end
 
     describe "erased" do
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -399,15 +399,6 @@ describe User do
         expect(User.active).not_to include(user3)
       end
 
-      it "returns users who have confirmed their account" do
-        user1 = create(:user, erased_at: nil, confirmed_at: Time.current)
-        user2 = create(:user, erased_at: nil, confirmed_at: Time.current)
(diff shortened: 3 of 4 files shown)


mohsinkhansymc:
@PierreMesure Please review, thanks

javierm:
@mohsinkhansymc Thanks for this pull request! :smile: I'll have a look at it, hopefully next week.

In the meantime, could you add a test for the scenario we're trying to fix (sending newsletters to unconfirmed account)? :pray: Thanks!

PierreMesure:
Thanks a lot!

Did you do any manual test? Because by looking at the other scopes like:
```
scope :newsletter,     -> { where(newsletter: true) }
scope :email_digest,   -> { where(email_digest: true) }
```
It doesn't seem like they filter by active user. Shouldn't we add that criterium to all the scopes (except `erased`)?

mohsinkhansymc:
@PierreMesure The active check is enough as first all the users are fetched and then 

> Thanks a lot!
> 
> Did you do any manual test? Because by looking at the other scopes like:
> 
> ```
> scope :newsletter,     -> { where(newsletter: true) }
> scope :email_digest,   -> { where(email_digest: true) }
> ```
> 
> It doesn't seem like they filter by active user, but booleans like `newsletter` or `email_digest` are already set to true before the account is confirmed. Shouldn't we add that criterium to all the scopes (except `erased`)?
> 
> I will have time to try it myself on Friday.

This code should work if the said scope sends to all users, but sure you can have a look on Friday

mohsinkhansymc:
@javierm Please review the test case, thanks

PierreMesure:
Ah, thanks! That answers my question!

@mohsinkhansymc your test looks great. I'd say you don't need to test that `user3` is not in the array since you already tested that the array is exactly `[user1, user2]`.

app/models/user.rb
```diff
@@ -67,7 +67,7 @@ class User < ApplicationRecord
     where(document_type: document_type, document_number: document_number)
   end
   scope :email_digest,   -> { where(email_digest: true) }
-  scope :active,         -> { where(erased_at: nil) }
+  scope :active,         -> { where(erased_at: nil).where.not(confirmed_at: nil) }
```
javierm: Changing the `active` scope might have some side effects :thinking:. For instance, this scope is used while verifying users. What do you think about changing the `UserSegments.all_users` method instead, which will only affect newsletters?

spec/models/user_spec.rb
```diff
@@ -399,6 +399,15 @@
         expect(User.active).not_to include(user3)
       end
 
+      it "returns users who have confirmed their account" do
```
javierm: Since we're trying to test that newsletters aren't sent to unconfirmed accounts, how about adding a test for `Newsletter#list_of_recipient_emails` instead? https://github.com/consul/consul/blob/master/spec/models/newsletter_spec.rb#L50

app/models/user.rb
mohsinkhansymc: @javierm You are right about this, we don't know what might break. I will revert and do the change to all_users method

mohsinkhansymc:
@javierm I have done the changes you proposed. Thanks

lib/user_segments.rb
```diff
@@ -10,7 +10,7 @@ class UserSegments
                 not_supported_on_current_budget]
 
   def self.all_users
-    User.active
+    User.where(erased_at: nil).where.not(confirmed_at: nil)
```
javierm: Maybe we can keep using the `active` scope :thinking:.
```suggestion
    User.active.where.not(confirmed_at: nil)
```

mohsinkhansymc: @javierm makes sense as if the condition for active users changes in future, this will work flawlessly :+1:

PierreMesure:
Before we close this, I'd like to point out that `User.active` is also used to calculate the number of users shown in the Statistics page on the admin panel. Is this what we want? Or do we want to reflect the number of users who actually confirmed their e-mail address?

Capture d’écran 2019-10-23 à 19 36 56

javierm:
> In general, shouldn't we replace all the occurrences of `User.active` by a new scope that dismisses the users that didn't confirm their e-mail addresses?

The `active` scope is used in several places related to verification, which is a part of the application most forks change, so I'd rather be conservative here and keep the existing behavior :thinking:.

Regarding the admin statistics, I don't have a strong opinion; I guess both approaches are valid :thinking:.

PierreMesure:
Yes, I agree with you that touching `active` is not a good idea. But creating a `confirmed_users` scope which we can then use in [all the places](https://github.com/consul/consul/search?p=2&q=%22User.active%22&unscoped_q=%22User.active%22) where `active` is used today?

mohsinkhansymc:
@PierreMesure we can have a separate label on the stats page that shows confirmed accounts or we can modify the total users label to `confirmedcount/totalusers`. Also instead of showing unverified users we can show active users that will return the confirmed accounts count. What do you guys think? @javierm

javierm:
This is an interesting topic :thinking:. Since we're trying to get everything ready for version 1.1, I'd rather be conservative and only change the code affecting newsletters. Feel free to open a new issue and continue this conversation once we release version 1.1, and maybe then we can study the implications of chaging the `active` scope.

@mohsinkhansymc For now, it would be awesome if we kept this pull request as it is, except for the [suggested change in the all_users method](https://github.com/consul/consul/pull/3781/files#r337603211), and then we can merge it and it will be included in version 1.1 :tada:.

javierm:
@mohsinkhansymc Thanks a lot! :smile:

PR #3784, opened 2019-10-21
Reduce security risk on remote_census_api

taitus:
## References
Related PR: #3498 
Related Documentation PR: [Update remote census doc #86](https://github.com/consul/docs/pull/86) 

## Objectives
The use of eval is a serious security risk, so we change by JSON.parse method.
Describe expected format for Setting["remote_census.request.structure"]

## Visual Changes
Captura de pantalla 2019-10-21 a las 17 39 38

## Notes
⚠️ If your installation is using the "Remote Census configuration" feature and you have already configured the "Request Structure" field, please be sure to update the format of this field.
You can find this section on:  `Settings` > `Global Settings` > `Remote Census configuration`

Example for old expected format value:
```
{
  request: {
    codigo_institucion: 12,
    codigo_portal: 5,
    documento:   ,
    tipo_documento:  ,
  }
}
```

Example for new expected format value:
```
{
  "request": {
    "codigo_institucion": 12,
    "codigo_portal":  5,
    "documento":  null,
    "tipo_documento":  null,
  }
}
```

### DIFF
--- a/config/locales/en/settings.yml
+++ b/config/locales/en/settings.yml
@@ -132,7 +132,7 @@ en:
         method_name: "Request method name"
         method_name_description: "Request method name accepted by the City Census WebService."
         structure: "Request Structure"
-        structure_description: 'Request Structure that receives the Census WebService of the City council. The "static" values of this request should be filled. Values related to Document Type, Document Number, Date of Birth and Postal Code should be blank.'
+        structure_description: 'Request Structure that receives the Census WebService of the City council. The "static" values of this request should be filled. Values related to Document Type, Document Number, Date of Birth and Postal Code should be null. Example of a valid format: { "request": { "codigo_institucion": 1, "codigo_portal": 1, "codigo_usuario": 1, "documento": null, "tipo_documento": null, "codigo_idioma": 102, "nivel": 3 } }'
         document_type: "Path for Document Type"
         document_type_description: "Path in the request structure that sends the Document Type. DO NOT FILL IN if the WebService does not require the Document Type to verify a user."
         document_number: "Path for Document Number"
--- a/lib/remote_census_api.rb
+++ b/lib/remote_census_api.rb
@@ -84,7 +84,7 @@ class RemoteCensusApi
     end
 
     def request(document_type, document_number, date_of_birth, postal_code)
-      structure = eval(Setting["remote_census.request.structure"])
+      structure = JSON.parse(Setting["remote_census.request.structure"])
 
       fill_in(structure, Setting["remote_census.request.document_type"], document_type)
       fill_in(structure, Setting["remote_census.request.document_number"], document_number)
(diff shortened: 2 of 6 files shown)


lib/remote_census_api.rb
```diff
@@ -100,12 +100,11 @@ def request(document_type, document_number, date_of_birth, postal_code)
 
     def fill_in(structure, path_value, value)
       path = parse_path(path_value)
-
       update_value(structure, path, value) if path.present?
     end
 
     def parse_path(path_value)
-      path_value.split(".").map { |section| section.to_sym } if path_value.present?
+      path_value.split(".") if path_value.present?
```
javierm: This change is necessary because we now use strings instead of symbols, right?

By the way, I've just noticed the method `parse_path` is defined twice on this file. I wonder why our `Lint/DuplicateMethods` rule didn't catch it :thinking:.

config/locales/en/settings.yml
```diff
@@ -132,7 +132,7 @@ en:
         method_name: "Request method name"
         method_name_description: "Request method name accepted by the City Census WebService."
         structure: "Request Structure"
-        structure_description: 'Request Structure that receives the Census WebService of the City council. The "static" values of this request should be filled. Values related to Document Type, Document Number, Date of Birth and Postal Code should be blank.'
+        structure_description: 'Request Structure that receives the Census WebService of the City council. The "static" values of this request should be filled. Values related to Document Type, Document Number, Date of Birth and Postal Code should be null. Example of a valid format: { "request": { "codigo_institucion": 1, "codigo_portal": 1, "codigo_usuario": 1, "documento": null, "tipo_documento": null, "codigo_idioma": 102, "nivel": 3 } }'
```
javierm: Does it make sense to use English names in the parameters here? I've no idea; just asking :smile:.

lib/remote_census_api.rb
taitus: Yes @javierm , now we use strings instead of symbols.
I think the `Lint/DuplicateMethods` rule doesn't detect it because although it's in the same file it's not in the same class? 🤔 
Anyway I will rename the methods to make it clearer (`parse_response_path`, `parse_request_path`)

config/locales/en/settings.yml
taitus: Nice catch @javierm 
But I did it consciously. The reason is because I have copied the request of CensusAPI so that it can be more easily understood what is intended to be done with these fields.
But I can change it to English if you don't think that's motive enough.

lib/remote_census_api.rb
javierm: > although it's in the same file it's not in the same class?

Of course :smile:. Didn't realize the `Response` class finished before the `private` keyword :sweat_smile:.

config/locales/en/settings.yml
javierm: Leaving it like this is fine with me :+1:.

PR #3646, opened 2019-06-27
Make user verification configurable from the admin interface

voodoorai2000:
## References

- Add CRUD to manage local census records https://github.com/consul/consul/pull/3479
- Add the ability to import local census records through a CSV file https://github.com/consul/consul/pull/3500
- New setting tab remote Census https://github.com/consul/consul/pull/3434
- New remote Census API https://github.com/consul/consul/pull/3498
- Use RemoteCensusAPI in Officing residence verification https://github.com/consul/consul/pull/3521
- Use RemoteCensusAPI for Signatures https://github.com/consul/consul/pull/3522
- Use RemoteCensusAPI for Management verifications https://github.com/consul/consul/pull/3523
- New remote Census API https://github.com/consul/consul/pull/3498/
- Add local census management documentation https://github.com/consul/docs/pull/80
- Add census conexion configuration documentation https://github.com/consul/docs/pull/79

## Objectives

Make user verification configurable from the admin interface.

PR #3829, opened 2019-11-08
Remove duplicated local census records on deployment

Senen:
## References

Related Pull Requests: #3646  

PR #3646 includes a migration that adds an unique database constraint to LocalCensusRecord model so we need to provide a way to remove duplicated records before to apply the new index to database.

## Objectives

Remove duplicated local census records automatically during capistrano deployment and before applying new aforementioned migration.

Also provide a rake task to remove duplicated records manually for those Consul installations that are not using caspistrano.

## Release notes

With this PR people using capistrano may not worry about deleting local census records dupicates manually from the database, but people using other deployment strategies that capistrano, for example Heroku, should manually run the rake task `local_census_records:remove_duplicates` before running `db:migrate`.

## Production server test results
**Deployment test steps**:
1. Prepare a production server with the Ansible installer and deploy the Consul 1.0.0 version.
2. Create dups in the LocalCensusRecord database table
2. Deploy with capistrano the code of this PR
3. Check during deployment and before the capistrano step deploy:migrate  those duplicated records were removed successfully

Below the deployment log that demonstrates this PR is working successfully.
```
01:02 remove_local_census_records_duplicates
      01 /home/deploy/consul/rvm1scripts/rvm-auto.sh . bundle exec rake local_census_records:remove_duplicates
      01 INFO Removing local census records duplicates
      01
      01 INFO Removed 6 records.
      01
    ✔ 01 deploy@ec2-63-32-46-29.eu-west-1.compute.amazonaws.com 4.907s
01:07 deploy:migrate
      [deploy:migrate] Run `rake db:migrate`
01:07 deploy:migrating
```

**This PR wil be in WIP status until all review issues are solved**

### DIFF
--- a/lib/tasks/local_census_records.rake
+++ b/lib/tasks/local_census_records.rake
@@ -1,0 +1,13 @@
+namespace :local_census_records do
+  desc "Remove duplicated records from database"
+  task remove_duplicates: :environment do
+    ids = LocalCensusRecord.group(:document_type, :document_number).pluck("MIN(id) as id")
+    duplicates = LocalCensusRecord.count - ids.size
+
+    if duplicates > 0
+      ApplicationLogger.new.info "Removing local census records duplicates"
+      LocalCensusRecord.where("id NOT IN (?)", ids).destroy_all
+      ApplicationLogger.new.info "Removed #{duplicates} records."
+    end
+  end
+end
--- a/spec/lib/tasks/local_census_records_spec.rb
+++ b/spec/lib/tasks/local_census_records_spec.rb
@@ -1,0 +1,36 @@
+require "rails_helper"
+require Rails.root.join("db", "migrate", "20190530082138_add_unique_index_to_local_census_records")
+
+describe "LocalCensusRecord tasks" do
+  let(:run_rake_task) do
+    Rake::Task["local_census_records:remove_duplicates"].reenable
+    Rake.application.invoke_task("local_census_records:remove_duplicates")
+  end
+
+  describe "#remove_duplicates" do
+    around do |example|
+      ActiveRecord::Migration.suppress_messages do
+        example.run
+      end
+    end
+
+    before { AddUniqueIndexToLocalCensusRecords.new.down }
+    after { AddUniqueIndexToLocalCensusRecords.new.up }
+
+    it "Remove all duplicates keeping older records" do
+      record1 = create(:local_census_record, document_type: "1", document_number: "#DOCUMENT_NUMBER")
+      record2 = create(:local_census_record, document_type: "2", document_number: "#DOCUMENT_NUMBER")
+      dup_record1 = build(:local_census_record, document_type: "1", document_number: "#DOCUMENT_NUMBER")
+      dup_record1.save!(validate: false)
+      dup_record2 = build(:local_census_record, document_type: "2", document_number: "#DOCUMENT_NUMBER")
+      dup_record2.save!(validate: false)
+      record3 = create(:local_census_record, document_type: "3", document_number: "#DOCUMENT_NUMBER")
+
+      expect(LocalCensusRecord.count).to eq(5)
+
+      run_rake_task
+
+      expect(LocalCensusRecord.all).to match_array([record1, record2, record3])
+    end
+  end
+end
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -43,6 +43,7 @@ namespace :deploy do
   before :starting, "rvm1:install:rvm"
   before :starting, "rvm1:install:ruby"
   before :starting, "install_bundler_gem"
+  before "deploy:migrate", "remove_local_census_records_duplicates"
 
   after "deploy:migrate", "add_new_settings"
   after :publishing, "deploy:restart"
@@ -64,6 +65,16 @@ task :install_bundler_gem do
   on roles(:app) do
     within release_path do
       execute :rvm, fetch(:rvm1_ruby_version), "do", "gem install bundler"
+    end
+  end
+end
+
+task :remove_local_census_records_duplicates do
+  on roles(:db) do
+    within release_path do
+      with rails_env: fetch(:rails_env) do
+        execute :rake, "local_census_records:remove_duplicates"
+      end
     end
   end
 end


lib/tasks/local_census_records.rake
```diff
@@ -0,0 +1,11 @@
+namespace :local_census_records do
+  desc "Remove duplicated records from database"
+  task remove_duplicates: :environment do
+    ApplicationLogger.new.info "Removing local census records duplicates"
+    ids = LocalCensusRecord.group(:document_type, :document_number).pluck("MIN(id) as id")
+
+    duplicates = LocalCensusRecord.count - ids.size
+    LocalCensusRecord.where("id NOT IN (?)", ids).destroy_all
+    ApplicationLogger.new.info "Removed #{duplicates} records."
```
javierm: I'm thinking the log messages might generate unneeded noise during deployments :thinking:. I mean, now this task will be executed every time you deploy between the moment you upgrade to version 1.1 and the moment you upgrade to version 1.2, but duplicates will only be deleted once.

So what about something like:

```ruby
    ids = LocalCensusRecord.group(:document_type, :document_number).pluck("MIN(id) as id")
    duplicates = LocalCensusRecord.count - ids.size

    if duplicates > 0
      ApplicationLogger.new.info "Removing local census records duplicates"
      LocalCensusRecord.where("id NOT IN (?)", ids).destroy_all
      ApplicationLogger.new.info "Removed #{duplicates} records."
    end
```

Another option would be to check whether the migration to add the unique index has already been run :thinking:.

spec/lib/tasks/local_census_records_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+      duplicated_record2 = build(:local_census_record, document_type: "2",
+                                                      document_number: "#DOCUMENT_NUMBER")
+      duplicated_record2.save!(validate: false)
+
+      record3 = create(:local_census_record, document_type: "3", document_number: "#DOCUMENT_NUMBER")
+      expect(LocalCensusRecord.count).to eq(5)
+
+      run_rake_task
+      expect(LocalCensusRecord.all).to eq([record1, record2, record3])
```
javierm: The `.all` method doesn't guarantee the order of the records. So if we don't want this spec to be flaky, we have to use `match_array` instead of `eq`. Unfortunately we've got a lot of experience with similar situations :sweat_smile:.

spec/lib/tasks/local_census_records_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+      duplicated_record1 = build(:local_census_record, document_type: "1",
+                                                      document_number: "#DOCUMENT_NUMBER")
+      duplicated_record1.save!(validate: false)
+
+      duplicated_record2 = build(:local_census_record, document_type: "2",
+                                                      document_number: "#DOCUMENT_NUMBER")
+      duplicated_record2.save!(validate: false)
+
+      record3 = create(:local_census_record, document_type: "3", document_number: "#DOCUMENT_NUMBER")
+      expect(LocalCensusRecord.count).to eq(5)
```
javierm: We usually add a blank line between setup and expectations.

```ruby
      record3 = create(:local_census_record, document_type: "3", document_number: "#DOCUMENT_NUMBER")

      expect(LocalCensusRecord.count).to eq(5)
```

spec/lib/tasks/local_census_records_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+      run_rake_task
```
javierm: We usually add a blank line between actions and expectations:

```ruby
run_rake_task

expect(LocalCensusRecord.all) # ...
```

spec/lib/tasks/local_census_records_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    end
+
+    before { AddUniqueIndexToLocalCensusRecords.new.down }
+    after { AddUniqueIndexToLocalCensusRecords.new.up }
+
+    it "Remove all duplicates keeping older records" do
+      record1 = create(:local_census_record, document_type: "1", document_number: "#DOCUMENT_NUMBER")
+      record2 = create(:local_census_record, document_type: "2", document_number: "#DOCUMENT_NUMBER")
+      duplicated_record1 = build(:local_census_record, document_type: "1",
+                                                      document_number: "#DOCUMENT_NUMBER")
```
javierm: Is the intention to align the `d` from `document_number` with the `d` from `document_type`? :thinking: Just curious; we don't have consistent criteria about multi-line alignments.

Senen: Yes it was, but refactoring i miss one whitespace 😅 ! Personally prefer
```
duplicated_record1 = build(:local_census_record, document_type: "1", 
                                                 document_number: "#DOCUMENT_NUMBER")
``` 

over

```
duplicated_record1 = build(:local_census_record, document_type: "1", 
  document_number: "#DOCUMENT_NUMBER")
```

lib/tasks/local_census_records.rake
Senen: Of course, better to have the minimum noise possible!

spec/lib/tasks/local_census_records_spec.rb
Senen: Done! 👍

spec/lib/tasks/local_census_records_spec.rb
Senen: Done! 👍

spec/lib/tasks/local_census_records_spec.rb
Senen: Here finally replaced `duplicated_record1` with `dup_record1` so there is no need to use multiline and at the same time accomplish Rubocop max line length rule.

spec/lib/tasks/local_census_records_spec.rb
Senen: Done! 👍 🙏

lib/tasks/local_census_records.rake
Senen: Done!

PR #3830, opened 2019-11-08
Fix pagination problem on mobile

PierreMesure:
## Objectives

- Fixes a bug in which the pagination buttons wouldn't be displayed correctly if there are more than two pages.
- Backports the fix from [AyuntamientoMadrid/consul](https://github.com/AyuntamientoMadrid/consul/pull/248)

### Before:
Capture d’écran 2019-11-08 à 16 34 15
### After:
Capture d’écran 2019-11-08 à 16 34 56

## Notes

- Fixed thanks to @javierm

### DIFF
--- a/app/assets/stylesheets/layout.scss
+++ b/app/assets/stylesheets/layout.scss
@@ -409,6 +409,10 @@ a {
   background: $brand;
 }
 
+.pagination li {
+  display: inline-block;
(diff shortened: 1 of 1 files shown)


javierm:
@PierreMesure Thanks a lot! :smile:.

PR #3836, opened 2019-11-08
Make the dev_seeds based on available locales

### LINKED ISSUE #3835: Dev seeds fail if Spanish is not enabled
### Description

*Type*: Bug

Related to #3526. The dev seeds seem to require English and Spanish to pass. That's a problem because someone might want to disable any of them but still get some example content.

### Steps to reproduce

- Remove `:en` or `:es` from the list of available locales in `application.rb`.

### Definition of Done

- Maybe we could use `available_locales.first` and `available_locales.second` as the locales for the seeds? That wouldn't solve the problem for monolingual instances.
- To solve the latter, an `if available_locales.size > 1` would be required.

PierreMesure:
## References

- Closes #3835

## Objectives

- Fixes the dev_seeds if `:en` or `:es` are disabled as locales. Only works if at least two locales are enabled.

## Visual Changes

- None

## Notes

- The seeds fail if only one locale is enabled.

### DIFF
--- a/db/dev_seeds/budgets.rb
+++ b/db/dev_seeds/budgets.rb
@@ -25,16 +25,19 @@ def add_image_to(imageable)
 end
 
 section "Creating Budgets" do
+  locale_1 = I18n.available_locales.first
+  locale_2 = I18n.available_locales.second
+
   Budget.create!(
-    name_en: "#{I18n.t("seeds.budgets.budget", locale: :en)} #{Date.current.year - 1}",
-    name_es: "#{I18n.t("seeds.budgets.budget", locale: :es)} #{Date.current.year - 1}",
+    "name_#{locale_1}": "#{I18n.t("seeds.budgets.budget", locale: locale_1)} #{Date.current.year - 1}",
+    "name_#{locale_2}": "#{I18n.t("seeds.budgets.budget", locale: locale_2)} #{Date.current.year - 1}",
     currency_symbol: I18n.t("seeds.budgets.currency"),
     phase: "finished"
   )
 
   Budget.create!(
-    name_en: "#{I18n.t("seeds.budgets.budget", locale: :en)} #{Date.current.year}",
-    name_es: "#{I18n.t("seeds.budgets.budget", locale: :es)} #{Date.current.year}",
+    "name_#{locale_1}": "#{I18n.t("seeds.budgets.budget", locale: locale_1)} #{Date.current.year}",
+    "name_#{locale_2}": "#{I18n.t("seeds.budgets.budget", locale: locale_2)} #{Date.current.year}",
     currency_symbol: I18n.t("seeds.budgets.currency"),
     phase: "accepting"
   )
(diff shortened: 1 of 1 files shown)


microweb10:
Hi @PierreMesure! Thank you very much for your contribution ❤️ 

This is a good point for sure! All forks don't have English and/or Spanish as available locales, and some of them may have only one language available. I think we should also cover the scenario of having only one language in `available_locales`. 

So we could either:
- Create the content only for the currently available locale and forget about creating content in different locales. Is the content in different languages really needed for development? 🤔 In my opinion, it's not.
- Somehow make English and Spanish available for running the task `db:dev_seed` and then restore to the previous value of `available_locales`.

What are your thoughts on this? 🤔

PierreMesure:
I’d go for the first one

microweb10:
Hi @PierreMesure 

Maybe even a better solution would be to use the `random_locale` helper that does not depend on any specific language and will work with just one or more than one `available_locales`. Something similar than what's done in this line https://github.com/consul/consul/blob/master/db/dev_seeds/budgets.rb#L44
```
random_locales.map do |locale|
  Globalize.with_locale(locale) do
    #Code creating the budget/group/heading...
  end
end
```
What do you think? 😄

PierreMesure:
That sounds great! But shouldn't we try to generate seeds in all the locales that are enabled?

I'm thinking of a case where the developer is only proficient in some of the languages enabled (which can happen in case a city wants a lot of enabled languages, like Decide Madrid). The chance that the random locale is one they can understand is small.

It really doesn't matter that much though, we're talking about dev seeds so I don't think we have to worry too much about understanding the seed content, what matters is that running dev_seeds.rb works.

PierreMesure:
That sounds great! But shouldn't we try to generate seeds in all the locales that are enabled?

I'm thinking of a case where the developer is only proficient in some of the languages enabled (which can happen in case a city wants a lot of enabled languages, like Decide Madrid). The chance that the random locale is one they can understand is small.

It really doesn't matter that much though, we're talking about dev seeds so I don't think we have to worry too much about understanding the seed content, what matters is that running dev_seeds.rb works.

microweb10:
Hi @PierreMesure 

Not really 😅 
We already introduced a change to actually not do that, see this commit https://github.com/consul/consul/commit/731723838213f0d85d66d5b7b75dec27e859ff42

Using the `random_locales` should be fine, and we will keep consistency with the rest of the `dev_seeds`

Also, please note that the content is not being generated in different languages, it's all generated in English or in Latin using "lorem ipsum". 
It's just content for development purposes, it doesn't really need to be in any specific language. 👍

PierreMesure:
Yes, agreed. Should I update the PR based on your suggestion? I might have some time to do it tomorrow. Otherwise next week.

microweb10:
Yes ☺️ please 🙏

microweb10:
Please note that more files need to be changed, not only the one for budgets. If I'm not wrong you will have to adjust these:
```
db/dev_seeds/admin_notifications.rb
db/dev_seeds/budgets.rb
db/dev_seeds/legislation_processes.rb
db/dev_seeds/polls.rb
db/dev_seeds/widgets.rb
```

I advise you to set your `available_locales` to a language different from English and Spanish, so you can properly test if the rake `bin/rake db:dev_seed` works properly for only one language that is not English or Spanish.

Thanks! 🙏

javierm:
For these cases were two languages are generated (which I find useful in development, particularly if you're developing multi-language features), maybe we could use the default locale as the first language and another random one as the second language :thinking:. What do you think?

By the way, there are a few more places where we use texts in English/Spanish instead of using `I18n.t("seeds.XXX")`, like the seeds for admin notifications and widgets. Would it be possible to change those as well?

PierreMesure:
Hey, I'm really overworked right now so I won't have time to work on it in the coming two weeks at least. We'll see later in April. I guess that's OK unless you are in a hurry?

javierm:
We don't expect a new release anytime soon, so that's totally fine :smile:.

javierm:
@PierreMesure Thank you for this pull request and for bringing up the issue :pray:, and @microweb10 awesome review :smile:.

I've applied the suggestions you've mentioned about using random locales, prioritizing English and Spanish when available in order to keep compatibility with the current version.

Closing in favor of #4733.

PR #3838, opened 2019-11-09
Replace old Spanish text with org name

PierreMesure:
## Objectives

It's a tiny PR, it just removes an old Spanish text that would show as title of the page displayed after creating an account: Gobierno abierto.

Right now, I'm using the org name as title, that's open for improvements. It could also say something like "Account created" or "Check your e-mails".

### Before
Capture d’écran 2019-11-09 à 13 40 41

### After
Capture d’écran 2019-11-09 à 13 43 37

### DIFF
--- a/app/views/layouts/devise.html.erb
+++ b/app/views/layouts/devise.html.erb
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html lang="<%= I18n.locale %>">
   <head>
-    <%= render "layouts/common_head", default_title: "Gobierno abierto" %>
+    <%= render "layouts/common_head", default_title: setting["org_name"] %>
     <%= render "layouts/meta_tags" %>
     <%= raw setting["html.per_page_code_head"] %>
   </head>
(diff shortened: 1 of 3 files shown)


app/views/layouts/devise.html.erb
```diff
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html lang="<%= I18n.locale %>">
   <head>
-    <%= render "layouts/common_head", default_title: "Gobierno abierto" %>
+    <%= render "layouts/common_head", default_title: setting["org_name"] %>
```
javierm: It would be great to have a test for this change so we don't accidentally remove it, for example when solving git conflicts :pray:.

How about adding:

```ruby
expect(page).to have_title "CONSUL"
```

To: https://github.com/consul/consul/blob/master/spec/features/registration_form_spec.rb#L38

If we want a better title, we could add:

```erb
<% provide(:title) { t("devise_views.users.registrations.success.title") } %>
```

To: https://github.com/consul/consul/blob/master/app/views/users/registrations/success.html.erb#L1

And then change the mentioned test to:

```ruby
expect(page).to have_title "Confirm your email address"
```

What do you think?

PierreMesure: Great suggestion

Issue #3837, opened 2019-11-09
Add e-mail verification when creating an account


PierreMesure:
### User Story

- As an administrator, I want to make sure that as many e-mail addresses as possible are correct in my database.
- As a user, I want to get an error message if there is a typo in my e-mail address.

### Why is this Needed?

[comment]: # (Describe the problem and why this task is needed. Provide a description of the current state, what you would like to happen, and what actually happened)

I just discovered that Consul doesn't check the e-mail address at all when creating an account!

There are [some verifications](https://github.com/consul/consul/blob/00145d54ca35ddbbff3c0b71f1440239f9ee5fe3/app/models/verification/management/email.rb#L9) when creating an account with the management interface but not through the usual method.

It is a big problem as many people do mistype their e-mail address. On our Consul installations, you can see a lot of `user@gmail` or `user@gmail.vom`.

### Description

[comment]: # (Feature or Bug? eg. Type: Bug)
*Type*: Feature

I suggest that we add at least a regex validation on the form and the model.

It seems that [there is no perfect solution](https://stackoverflow.com/questions/22993545/ruby-email-validation-with-regex) with regex so we could also add a more advanced verification that MX records exist with the gem [valid_email2](https://github.com/micke/valid_email2).

### Current Behavior
[comment]: # (Describe what actually happened.)

You can enter `test@test` as an e-mail address when creating an account.

### Expected Behavior
[comment]: # (Describe what you expected to happen.)

You shouldn't be able to...

### Steps to reproduce
[comment]: # (Describe how we can replicate the bug step by step.)

- Create an account with `test@test` as the e-mail address.

### Definition of Done
[comment]: # (Any other information that would be useful, bullets are helpful.)
- Add Regex validation on the form and the model.
- (optional) Add validation with the `valid_email2` gem.

javierm:
Indeed there seems to be no perfect solution. Here's a summary of the current situation:

When registering users, we aren't checking anything on the server level but letting the browser validate the email. We are using a server-side validation for the email format when managers verify emails.

The regular expressions used both in the server side and the client side mark emails like `test@test` as valid. It looks like the RFC is quite lax and allows email addresses that are most likely invalid.

AFAIK Ruby doesn't come with stricter regular expressions, although there are gems providing them. Adding a stricter regular expression in CONSUL is possible by changing `config.email_regexp` in `config/initializers/devise.rb`.

[...]

PR #3841, opened 2019-11-09
Add e-mail validation when creating user

PierreMesure:
## References

- #3837

## Objectives

This PR adds email validation both at the model level (when saving to the DB) and in real-time on the form.

## Notes

- I had to skip validation when creating users with invalid email on some old specs. I didn't remove them as I assumed that existing installations will have invalid e-mails in their DB and need a filter when sending newsletters and notifications. But ideally, we could remove them as the filter is now applied when creating the user.
- I added a validation of MX records for the domain. It works surprisingly fast and catches a few more invalid emails (like `test@gnail.com`).
- I added an exception for `@consul.dev` e-mail addresses as that's what we're using for the seeds.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -78,6 +78,8 @@ class User < ApplicationRecord
 
   validates :username, presence: true, if: :username_required?
   validates :username, uniqueness: { scope: :registering_with_oauth }, if: :username_required?
+  validates :email, presence: true, 'valid_email_2/email': true, allow_blank: true
+  validates :email, 'valid_email_2/email': { mx: true }
   validates :document_number, uniqueness: { scope: :document_type }, allow_nil: true
 
   validate :validate_username_length
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -78,8 +78,8 @@ class User < ApplicationRecord
 
   validates :username, presence: true, if: :username_required?
   validates :username, uniqueness: { scope: :registering_with_oauth }, if: :username_required?
-  validates :email, presence: true, 'valid_email_2/email': true, allow_blank: true
-  validates :email, 'valid_email_2/email': { mx: true }
+  validates :email, 'valid_email_2/email': true, allow_blank: true, if: :not_email_consul_dev?
+  validates :email, 'valid_email_2/email': { mx: true }, if: :not_email_consul_dev?
   validates :document_number, uniqueness: { scope: :document_type }, allow_nil: true
 
   validate :validate_username_length
@@ -415,5 +415,9 @@ class User < ApplicationRecord
         attributes: :username,
         maximum: User.username_max_length)
       validator.validate(self)
+    end
+
+    def not_email_consul_dev?
+      ->(user) { user.email.end_with?("@consul.dev") }
     end
 end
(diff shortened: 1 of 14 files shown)


app/models/user.rb
```diff
@@ -78,6 +78,8 @@ class User < ApplicationRecord
 
   validates :username, presence: true, if: :username_required?
   validates :username, uniqueness: { scope: :registering_with_oauth }, if: :username_required?
+  validates :email, presence: true, 'valid_email_2/email': true, allow_blank: true, unless: :email_consul_dev?
```
javierm: Hi, Pierre!

This issue is very tricky. I thought about adding the `presence` rule a few months ago, but then realized there were dozens (maybe hundreds? I don't remember :thinking:) of users in production databases with no emails. I don't remember the circumstances allowing blank emails: maybe it was related to users belonging to organizations, or maybe it was related to managed users.

Since we need to study this issue carefully and make sure there are no side effects in existing installations, I've added the `post-1.1` label, so this pull request will not be included in release 1.1.

PierreMesure: There are definitely blank e-mails for managed users! I think `allow_blank: true` should work for these? But I guess `presence: true` shouldn't be here.

app/models/user.rb
```diff
@@ -414,4 +416,8 @@ def validate_username_length
         maximum: User.username_max_length)
       validator.validate(self)
     end
+
+    def not_email_consul_dev?
```
fig: This method doesn't appear to behave as its name suggests. Can we make a positive assertion and change the guard clause to an `unless`?
```suggestion
    def email_consul_dev?
```

app/models/user.rb
```diff
@@ -78,6 +78,8 @@ class User < ApplicationRecord
 
   validates :username, presence: true, if: :username_required?
   validates :username, uniqueness: { scope: :registering_with_oauth }, if: :username_required?
+  validates :email, 'valid_email_2/email': true, allow_blank: true, if: :not_email_consul_dev?
```
fig:
```suggestion
  validates :email, 'valid_email_2/email': true, allow_blank: true, unless: :email_consul_dev?
```

app/models/user.rb
```diff
@@ -78,6 +78,8 @@ class User < ApplicationRecord
...
+  validates :email, 'valid_email_2/email': { mx: true }, if: :not_email_consul_dev?
```
fig:
```suggestion
  validates :email, 'valid_email_2/email': { mx: true }, unless: :email_consul_dev?
```

javierm:
@PierreMesure Sorry for taking so long to reply! :pray: The whole world of email validations is a mess and it's hard to take a decision :relieved:.

For now, we aren't adding the remote verification of MX records in the registrations form. I'll be honest: I'm not sure whether it's got more advantages or disadvantages, and when in doubt, we're keeping it the way it was.

I'm leaving issue #3837 open, though, and adding a comment there, since this is a problem we'll probably revisit in the future (can't guarantee when, though :sweat_smile:).

Thank you very much for this pull request and for giving us ideas to think about!

PR #3854, opened 2019-11-13
Fix a tiny CSS leak

PierreMesure:
## References

- #2694

## Objectives

I just fixed a tiny CSS leak. The shadow effect of the cards was also applied to a link that didn't need it on hovering. The result was rather unpleasant aesthetically.

### Before
Capture d’écran 2019-11-13 à 20 25 24

### After

Capture d’écran 2019-11-13 à 20 26 00

## Notes

- @decabeza would be the best reviewer if he has time. He was the one to introduce the CSS effect for the cards.

### DIFF
--- a/app/assets/stylesheets/layout.scss
+++ b/app/assets/stylesheets/layout.scss
@@ -2770,8 +2770,11 @@ table {
     }
 
     a:hover {
-      box-shadow: 0 0 12px 0 rgba(0, 0, 0, 0.2);
       text-decoration: none;
+
+      &:not(.see-all) {
+        box-shadow: 0 0 12px 0 rgba(0, 0, 0, 0.2);
+      }
 
       img {
         transform: scale(1.1);


app/assets/stylesheets/layout.scss
```diff
@@ -2770,7 +2770,11 @@ table {
     }
 
     a:hover {
-      box-shadow: 0 0 12px 0 rgba(0, 0, 0, 0.2);
+
```
decabeza: Hi @PierreMesure! Works perfectly! 👌

Only by convention, can you please change the order?

```
a:hover {
  text-decoration: none;

  &:not(.see-all) {
    box-shadow: 0 0 12px 0 rgba(0, 0, 0, 0.2);
  }
```

So `text-decoration: none;` doesn't get left alone 😜 

Many thanks! 🙏

PR #3866, opened 2019-11-22
Make it easier to release a new version of CONSUL

microweb10:
## Objectives

With this change we no longer need the branch `changelog` and it will be easier to release new versions of CONSUL.

## Note
The [wiki page about releasing a new version](https://github.com/consul/consul/wiki/Release-steps) needs to be adjusted


Gemfile
```diff
@@ -93,6 +93,7 @@ group :development do
   gem "capistrano3-delayed-job", "~> 1.7.3"
   gem "capistrano3-puma", "~> 4.0.0"
   gem "erb_lint", require: false
+  gem "github_changelog_generator"
```
javierm: For consistency, how about adding `", ~> 1.15.0"`?

Rakefile
```diff
@@ -5,3 +5,17 @@ require File.expand_path("../config/application", __FILE__)
...
+  require "github_changelog_generator/task"
+
+  GitHubChangelogGenerator::RakeTask.new :changelog do |config|
+    config.user = "consul"
+    config.project = "consul"
+    config.since_tag = "1.0.0"
+    config.future_release = "1.1.0"
+    config.base = "#{Rails.root}/CHANGELOG.md"
+    config.token = Rails.application.secrets.github_changelog_token
+    config.max_issues = 1
```
javierm: In the past, we had `max_issues = 1` in the Rakefile and `max_issues = 0` in the `.github_changelog_generator` file. I don't know which way has precedence :thinking:. Is it this one?

PR #3868, opened 2019-11-24
Categories are still shown when properties of the legislation process are changed

### LINKED ISSUE #3558: Keep categories being shown when change proposal phase date on colaborative legislation processes
## Background
When a proposal phase of a collaborative legislation process is active, in the sidebar are shown the categories of the process. 

When the admin change the date of the proposals phase the categories disappear from the sidebar.

## What we need
Keep showing the categories in the sidebar although the proposal phase date is changed.

## Screenshots

When the admin change the date of the proposals phase on `/admin/legislation/processes/N/edit`

Screenshot 2019-11-04 at 13 05 56

The categories disappear of `/admin/legislation/processes/N/proposals` this causes also are disappearing on the front view.

Screenshot 2019-11-03 at 19 00 36

Screenshot 2019-11-03 at 18 59 10

andrewcameronsims:
## References

Closes #3558 : "Keep categories being shown when change proposal phase date on colaborative legislation processes"

## Notes

In the issue, it is mentioned that date changes to the legislation process cause the categories of the process to be deleted. I discovered that this bug occurs when any change whatsoever is made to the process from the administrator view, not just date. That was because in the update action it calls set_tag_list even when there are no tags in the parameters, and this causes the tag list to be set to empty. The fix for this is to make the set_tag_list call to be conditional on the presence of tags in the parameters. I have also added to one of the existing specs to have confidence to know if this bug reoccurs. This spec fails now without my change, and passes with it. Hope it's ok!

### DIFF
--- a/app/controllers/admin/legislation/processes_controller.rb
+++ b/app/controllers/admin/legislation/processes_controller.rb
@@ -24,8 +24,6 @@ class Admin::Legislation::ProcessesController < Admin::Legislation::BaseControll
 
   def update
     if @process.update(process_params)
-      set_tag_list
-
       link = legislation_process_path(@process)
       redirect_back(fallback_location: (request.referer || root_path),
                     notice: t("admin.legislation.processes.update.notice", link: link))
@@ -75,11 +73,6 @@ class Admin::Legislation::ProcessesController < Admin::Legislation::BaseControll
         documents_attributes: [:id, :title, :attachment, :cached_attachment, :user_id, :_destroy],
         image_attributes: image_attributes
       ]
-    end
-
-    def set_tag_list
-      @process.set_tag_list_on(:customs, process_params[:custom_list])
-      @process.save!
     end
 
     def resource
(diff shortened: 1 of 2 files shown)


app/controllers/admin/legislation/processes_controller.rb
```diff
@@ -24,7 +24,7 @@ def create
 
   def update
     if @process.update(process_params)
-      set_tag_list
+      set_tag_list if process_params[:custom_list]
```
javierm: @andrewcameronsims Awesome! :tada: Nice debugging there to find the problem! :+1: And great test! :clap:

I've just realized we don't even need the `set_tag_list` method at all; the `acts_as_taggable_on :customs` line in `Legislation::Process` generates the `custom_list=` method, and the attribute is correctly updated when we call `@process.update(process_params)`.

So would you mind removing the `set_tag_list` method completely? :pray:

By the way, are you familiar with `git rebase`? If so, could you squash all commits into one? For such a small pull request, having four commits seems too much :thinking:. If not, it's alright :wink:.

andrewcameronsims: No worries. Is this OK? Thanks for the feedback.

Issue #3815, opened 2019-11-04
Add "Other" as a third option for gender


PierreMesure:
### User Story

- As a user, I want to have my gender set to "Other", either by the census or manually.
- As an admin, I want to be able to see which proportion of my users are gender non-conforming.

### Why is this Needed?

Although it might not be needed yet in Madrid, more and more countries and cities recognise the possibility of citizens to define themselves as something else than male or female.

### Description

[comment]: # (Feature or Bug? eg. Type: Bug)
*Type*: Feature

I suggest we start by simply adding the category "other" in the database and the user model. A more advanced implementation could follow if the need is expressed.

### Definition of Done
[comment]: # (Any other information that would be useful, bullets are helpful.)

- Add `other` to the model for a user
- Add it in all the statistics page

gabrielbaldao:
Can i contribute with this feature?

javierm:
@gabrielbaldao Sure, go ahead! :smile:

Regarding statistics, one possible implementation would be to show statistics for the genders defined in the database :thinking:. This way they would work on every CONSUL installation, no matter which genders are defined.

gabrielbaldao:
Thanks @javierm, looking here I was wondering about change this 
```scope :male,           -> { where(gender: "male") }``` to [enumerize gem](https://github.com/brainspec/enumerize), what do you think?

javierm:
The enumerize gem looks interesting :smile:. We might add it in the future if it simplifies the code a lot, but for now I'd try to implement this feature without adding a new dependency. We try not to add too many extra dependencies so we don't end up with hundreds of gems in our Gemfile, making it even harder to keep them up to date.

Regarding the places to display the information, maybe we could start by adding an "other" gender to budgets/polls stats when there are users with other genders in the database:

![Stats by gender, showing male and female participation](https://user-images.githubusercontent.com/35156/70658370-287cea00-1c5e-11ea-87a6-6fa7b531e15a.png)

gabrielbaldao:
Right, I agree with this, for this I sayed about do in another pr, but I was having problems to find this page.
Do you suggest some image to show this another gender?

javierm:
@decabeza is the one in charge of icons :smile:. @decabeza any suggestions about an intersex/other icon?

javierm:
@gabrielbaldao @PierreMesure I'd like to ask your opinion about a couple of things I'm considering.

[...]

PR #3882, opened 2019-12-10
[WIP] add other as gender to users

gabrielbaldao:
## References
Resolver #3815 

## Objectives

## Visual Changes

## Notes


app/views/admin/stats/gender_stats.html.erb
```diff
@@ -0,0 +1,28 @@
...
+  <div class="row stats-numbers">
+    <div class="small-12 medium-3 column">
+      <p class="featured">
+        <%= t("admin.stats.genders.female") %><br>
+        <span id="female_count" class="number"><%= @users_with_female_gender %></span>
+      </p>
+    </div>
+
+    <div class="small-12 medium-3 column">
+    <p class="featured">
```
m-pereira: I think you could fix the identation here.

gabrielbaldao: Thanks for help me

app/assets/stylesheets/stats.scss
```diff
@@ -53,6 +53,10 @@
       content: image-url("stats_woman.png");
     }
 
+    &.other::before {
+      content: image-url("stats_other.png");
```
javierm: @decabeza Do you know of an image we could use here, which is consistent with the images we use for men and women?

lib/census_api.rb
```diff
@@ -40,6 +40,8 @@ def gender
         "male"
       when "Mujer"
         "female"
+      when "Otro"
```
javierm: This part is actually a hack which only works for certain Spanish census APIs :sweat_smile:, so I'm not sure which word we should use here :thinking: (if any). @taitus Any ideas?

app/controllers/admin/stats_controller.rb
```diff
@@ -90,6 +90,12 @@ def polls
     @participants = ::Poll::Voter.where(poll: @polls)
   end
 
+  def gender_stats
```
javierm: For now maybe we could remove this action (alongside its view and the entries in `admin.yml`), since admin stats are a bit different than public stats :thinking:.

javierm:
I forgot to apologize for the delay reviewing this pull request :pray:. Testing and releasing version 1.1 took us quite some time, so we stopped adding features for a few months.

spec/models/budget/stats_spec.rb
```diff
@@ -152,6 +159,12 @@
         expect(stats.female_percentage).to be 40.0
       end
     end
+
+    describe "#other_percentage" do
+      it "returns the percentage of other participants" do
```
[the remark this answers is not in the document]
javierm: Hound's comment seems to be outdated.

decabeza:
I attach `stats_other.png` file as a little mixin between current female and male icons.

stats_other

I'm not sure if it's correct to use gender icons, but also we can use something like:

gender_icons

Anyway, I think we should use [Font Awesome gender/people icons](https://fontawesome.com/icons?d=gallery&c=gender,users-people&m=free) instead of png files, for example:

gender_icons_2

lib/census_api.rb
taitus: It is not clear to me that we need to add this option until some institution using the Census connection includes this new type of "gender description" in their API response. 🤔 
However, if we want consistent with the other changes of this PR, the word "other" seems right to me.

PR #3892, opened 2019-12-20
Remove custom spec for residence verification

microweb10:
## Objectives

Looks like this custom spec was specific for Decide Madrid and should not be included in CONSUL 🤔


javierm:
It's true these tests don't make much sense, but the real issue IMHO is these tests pass because we've got a custom residence model which checks the residence is in Madrid.

So I'd either keep both the (wrong) tests and the (wrong) code or remove both of them :thinking:.

microweb10:
I don't know if it's actually a good idea to remove that custom verification code. 🤔 
I think it could be a good example for the forks that want to have a custom verification, at least until CONSUL will have a proper customizable verification system.

PR #3910, opened 2020-01-28
Force dateFormat on datepicker

taitus:
## Objectives
After this change: 1f059d44cfd960b58cca9fc376b238be8378b394
When locale is :en, after select a date in datepicker the format rendered is "mm/dd/yyyy". The problem is that format is Date invalid. To fix this behaviour, we force dateFormat to the same format that is used to filter on advanced search in the frontend and that format is a valid format "dd/mm/yyyy"

## Visual Changes
- Before PR:
Captura de pantalla 2020-01-28 a las 17 32 57

- After PR:
Captura de pantalla 2020-01-29 a las 12 34 17

## Notes
Pending improve manage Dates on admin section. Unify the format from datepicker render with the form date fields format and with advanced search.
This is a temporary solution that we should analyze in depth to allow different date formats depending on the language in the filters.

### DIFF
--- a/app/assets/javascripts/advanced_search.js
+++ b/app/assets/javascripts/advanced_search.js
@@ -25,6 +25,7 @@
       });
       $(".js-calendar-full").datepicker();
       $.datepicker.setDefaults($.datepicker.regional[locale]);
+      $.datepicker.setDefaults({ dateFormat: "dd/mm/yy" });
     },
     initialize: function() {
       App.AdvancedSearch.init_calendar();
--- a/spec/features/admin/banners_spec.rb
+++ b/spec/features/admin/banners_spec.rb
@@ -179,6 +179,17 @@ describe "Admin banners magement" do
     expect(page).not_to have_content "Wrong text"
   end
 
+  scenario "when change date field on edit banner page display expected format", :js do
+    banner = create(:banner)
+    visit edit_admin_banner_path(banner)
+
+    fill_in "Post started at", with: "20/02/2002"
+    find_field("Post started at").click
+    within(".ui-datepicker") { click_link "22" }
+
+    expect(page).to have_field "Post started at", with: "22/02/2002"
+  end
+
   scenario "Delete a banner" do
     create(:banner, title: "Ugly banner",
                     description: "Bad text",


app/assets/javascripts/advanced_search.js
```diff
@@ -25,6 +25,7 @@
       });
       $(".js-calendar-full").datepicker();
       $.datepicker.setDefaults($.datepicker.regional[locale]);
+      $.datepicker.setDefaults({ dateFormat: "dd/mm/yy" });
```
javierm: How about a test to make sure we don't run into this annoying bug again?

This is what I've come up with. Feel free to improve it! :wink:

```ruby
banner = create(:banner)
login_as(create(:administrator).user)

visit edit_admin_banner_path(banner)
fill_in "Post started at", with: "20/02/2002"
find_field("Post started at").click
within(".ui-datepicker") { click_link "22" }

expect(page).to have_field "Post started at", with: "22/02/2002"

```

taitus: I added the same spec you mentioned...I haven't found a simpler way to test it.

Related with this spec:

When I created the PR I already commented in the notes that an _improve manage dates on admin section_ is still pending. But now doing the spec I don't know if we should apply any improvement in terms of data consistency.

Look at the lines: https://github.com/consul/consul/blob/master/app/views/admin/banners/_form.html.erb#L8
https://github.com/consul/consul/blob/master/app/views/admin/banners/_form.html.erb#L15

These lines apply a format depending on the locale of the administrator. What happens if we edit only one of the two dates of the banner? The date just selected will be seen with the format "dd/mm/yyyy" and the date that has not been modified will be seen with the format "yyyy-mm-dd".
It's true that once you save the changes, they will both be rendered again with the same format, but until you save it, it can be confusing.

Captura de pantalla 2020-02-04 a las 18 52 58

javierm: > What happens if we edit only one of the two dates of the banner? The date just selected will be seen with the format "dd/mm/yyyy" and the date that has not been modified will be seen with the format "yyyy-mm-dd".

Yeah, I think we mentioned it when discussing this issue last week. That's pretty horrible, actually :sweat_smile:. We should definitely use the same format in both cases, although in my opinion fixing it isn't that urgent and we can leave it as it is now until we release version 1.1 :thinking:.

Just one more thing about this pull request: would you mind squashing both commits into one? :pray: Thanks a lot!

PR #3911, opened 2020-01-29
Improve Dashboard emails

### LINKED ISSUE #3852: Dashboard emailing feature not working
Type: Bug

On the proposal dashboard the emailing feature under Email section seems not working.

Current behaviour
User clicks on "Send to <email>" button, message arrives but no email has been received.

Steps to reproduce
I see the same behavior on the demo.cosnulproject.org as well. You can create a new proposal, then visit E-mail under resources of the dashboard section. Please click on "Send to <email>"

Expected Behaviour
Would like to receive emails when author clicks on the button

Definition of Done
If we were able to receive emails when author clicks on the button.

taitus:
## References

* Closes #3852 

## Objectives
- Rename Setting["dashboard.emails"] to Setting["feature.dashboard.notification_emails"]. It currently appeared as Configuration Setting, without enable/disable button. Need that this setting appear as Feature Setting.
- To fix related issue #3852 we allow send the forward email always. 

## Notes
- In this commit [Disable dashboard email deliveries](https://github.com/consul/consul/commit/e1390570), we disabled all emails from Dashboard with a Setting. I think that the reason for this commit was to allow the enable/disable of the dashboard emails related to the notifications to the authors of the proposals, that is, the emails that are automatically sent through a Cron task with the resources/actions available to the dashboard users.
I have decided that it is best to continue allowing these emails to be enable/disable:
- new_actions_notification_rake_published
- new_actions_notification_rake_created
- new_actions_notification_on_create
- new_actions_notification_on_published
 
Except for the "forward" email, which I think should always be able to be sent, since it is a simple email that is sent to the author himself when he requests it from the dashboard. This "forward" email is only a description and a poster of your proposal, so that it can be shared and your proposal supported.

### DIFF
--- a/app/models/setting.rb
+++ b/app/models/setting.rb
@@ -170,7 +170,7 @@ class Setting < ApplicationRecord
         "hot_score_period_in_days": 31,
         "related_content_score_threshold": -0.3,
         "featured_proposals_number": 3,
-        "dashboard.emails": nil,
+        "feature.dashboard.notification_emails": nil,
         "remote_census.general.endpoint": "",
         "remote_census.request.method_name": "",
         "remote_census.request.structure": "",
--- a/config/locales/en/settings.yml
+++ b/config/locales/en/settings.yml
@@ -125,6 +125,9 @@ en:
       valuation_comment_notification: "Valuation comment notification"
       valuation_comment_notification_description: "Send an email to all associated users except valuation commenter to budget investment when a new valuation comment is created"
       graphql_api: "GraphQL API"
+      dashboard:
+        notification_emails: "Dashboard notification emails"
+        notification_emails_description: "Enable sending dashboard notification emails to proposal's authors"
     remote_census:
       general:
         endpoint: "Endpoint"
--- a/config/locales/es/settings.yml
+++ b/config/locales/es/settings.yml
@@ -125,6 +125,9 @@ es:
       valuation_comment_notification: "Notificar comentarios de evaluación"
       valuation_comment_notification_description: "Envía un email a todos los usuarios menos al que haya comentado asociados a un presupuesto participativo cuando se cree un nuevo comentario de evaluación"
       graphql_api: "API GraphQL"
+      dashboard:
+        notification_emails: "Emails del panel de progreso"
+        notification_emails_description: "Activar el envío de emails de notificaciones a los autores de las propuestas en la sección de panel de progreso"
     remote_census:
       general:
         endpoint: "Endpoint"
--- a/lib/tasks/settings.rake
+++ b/lib/tasks/settings.rake
@@ -4,4 +4,10 @@ namespace :settings do
     ApplicationLogger.new.info "Adding new settings"
     Setting.add_new_settings
   end
+
+  desc "Rename existing settings"
+  task rename_setting_keys: :environment do
+    ApplicationLogger.new.info "Renaming existing settings"
+    Setting.rename_key from: "dashboard.emails", to: "feature.dashboard.notification_emails"
+  end
 end
--- a/app/mailers/dashboard/mailer.rb
+++ b/app/mailers/dashboard/mailer.rb
@@ -1,6 +1,6 @@
 class Dashboard::Mailer < ApplicationMailer
   layout "mailer"
-  after_action :check_deliverability
+  after_action :check_deliverability, except: [:forward]
 
   def forward(proposal)
     @proposal = proposal
--- a/app/mailers/dashboard/mailer.rb
+++ b/app/mailers/dashboard/mailer.rb
@@ -41,6 +41,6 @@ class Dashboard::Mailer < ApplicationMailer
     end
 
     def check_deliverability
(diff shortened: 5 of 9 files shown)


config/locales/en/settings.yml
```diff
@@ -125,6 +125,9 @@ en:
       valuation_comment_notification: "Valuation comment notification"
       valuation_comment_notification_description: "Send an email to all associated users except valuation commenter to budget investment when a new valuation comment is created"
       graphql_api: "GraphQL API"
+      dashboard:
+        notification_emails: "Dashboard notification emails"
+        notification_emails_description: "Enable the sending dashboard notification emails to proposal’s authors"
```
javierm: ’"Enable the sending" sounds strange to me; "Enable sending" sounds better, although I might be wrong since I'm not a native speaker.

By the way, I think we use `'` instead of `’` in English (so it would be "proposal's" instead of "proposal’s") :thinking:.

config/locales/es/settings.yml
```diff
@@ -125,6 +125,9 @@ es:
       valuation_comment_notification: "Notificar comentarios de evaluación"
       valuation_comment_notification_description: "Envía un email a todos los usuarios menos al que haya comentado asociados a un presupuesto participativo cuando se cree un nuevo comentario de evaluación"
       graphql_api: "API GraphQL"
+      dashboard:
+        notification_emails: "Emails del Panel de Progreso"
```
javierm: Althoug we don't seem to have a standard name in Spanish, it looks like in other places we write "panel de progreso", using lowercase :thinking:.

config/locales/es/settings.yml
```diff
@@ -125,6 +125,9 @@ es:
...
+        notification_emails_description: "Activar el envio de emails de notificaciones a los autores de las propuestas en la sección de Panel de Progreso"
```
javierm: It looks like an accent is missing in "envío".

Althoug we don't seem to have a standard name in Spanish, it looks like in other places we write "panel de progreso", using lowercase :thinking:.

app/models/setting.rb
```diff
@@ -170,7 +170,7 @@ def defaults
         "hot_score_period_in_days": 31,
         "related_content_score_threshold": -0.3,
         "featured_proposals_number": 3,
-        "dashboard.emails": nil,
+        "feature.dashboard.notification_emails": nil,
```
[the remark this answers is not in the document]
taitus: I think it's a good idea @javierm, but I have a question... 
Currently the key that disappears ("dashboard.emails") was being displayed in Settings > Configuration... that is, it was not an Enable/Disable key, it was a value filled with a input field.
For this reason, I would discard to use: "settings:rename_setting_keys" because even if it works, the new key would save the text introduced by the administrator with the old version, instead of True or False...
And if we only use "settings:remove_deprecated_settings", in case someone already has updated "dashboards.emails" he would lose this feature. 🤔 
How about the next solution:
We create a new task to add in the consul.rake that only removes the old key and in case it has a value we updates the new key with true? in case it has no value we leave it in false/nil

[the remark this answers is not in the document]
javierm: As far as I know, keys in the "Featues" section are not stored as `true` or `false`, but either as a string or `nil`. That's because the type of the "value" column in the "settings" table is a string.

This is something I've just tried in the console:

```ruby
Setting["feature.featured_proposals"] = "aaa" # Makes the setting appear as "enabled"
Setting["feature.featured_proposals"] = "" # Makes the setting appear as "disabled"
```

So I guess renaming the key would maintain the current behaviour :thinking:.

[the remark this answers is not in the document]
taitus: Ok @javierm !!!
I agree with you as "almost" always 😄 
I was a little worried about the bbdd value:
```
> Setting["feature.featured_proposals”] = true
> Setting["feature.featured_proposals”]
=> "t" 
```

```
Setting["feature.featured_proposals”] = false
> Setting["feature.featured_proposals”]
=> nil 
```
But as you rightly say with the rename_key task we will maintain the expected behavior.

lib/tasks/settings.rake
```diff
@@ -5,3 +5,9 @@ namespace :settings do
     Setting.add_new_settings
   end
 end
+
+desc "Rename existing settings"
+task rename_setting_keys: :environment do
+  ApplicationLogger.new.info "Renaming existing settings"
+  Setting.rename_key from: "dashboard.emails",  to: "feature.dashboard.notification_emails"
+end
```
javierm: I think this task needs to be inside the `namespace :settings do` block :thinking:. Right now I'm getting an error message:
```
$ rake settings:rename_setting_keys                                                                 ◼
rake aborted!
Don't know how to build task 'settings:rename_setting_keys' (See the list of available tasks with `rake --tasks`)
Did you mean?  settings:add_new_settings
```

taitus: Nice catch @javierm 
I don't know how I ended up committing it out of the namespace. My fault 🙏 
Now it's solved. 
Thank you for review!

javierm:
Awesome! :tada:

PR #3919, opened 2020-02-13
Align select admins and valuators with the rest of the form

microweb10:
## Objectives

Align the selection of administrators and valuators with the rest of the form when creating or editing a Participatory Budget.

### BEFORE
before

### AFTER
after


javierm:
Interesting :thinking:. I get the results shown in the "after" screenshot even without these changes, in both firefox and chromium. Which browser are you using?

microweb10:
the issue happens In both Chrome and Safari 😅

PR #3920, opened 2020-02-17
Fix style of banner

### LINKED ISSUE #3639: Broken banner styles for some pages
### Description / Current Behavior

In the current master branch I've created a banner from the admin, but the appearance is different depending on the page where it is displayed.

For example, in the debates page it looks nice:

Captura de pantalla 2019-06-24 a las 17 54 38

In the participatory budgets it partially overlaps with the content after it:

Captura de pantalla 2019-06-24 a las 17 54 54

In the help page something similar happens:

Captura de pantalla 2019-06-24 a las 17 55 07

And in the homepage it spreads the full width, which I think looks a bit weird (but this is subjective):

Captura de pantalla 2019-06-24 a las 17 59 48

### Description

*Type*: bug

¿Is this something you can fix in your side for all installations? That'd be appreciated.

edumoreira1506:
## References

* Closes #3639 

## Objectives

This PR fixes banner style in all pages. You can see all broken styles in #3639 

## Visual Changes

Before of the changes, in home page, the banner was outside of site container and now:
Screenshot from 2020-02-17 15-09-22

Before of the changes, in help page, the banner wasn't bottom margin, and this caused an impression that banner was together with default banner of help page, and now:
Screenshot from 2020-02-12 10-44-11

Before of the changes, in budgets page, banner was broken, and now:
Screenshot from 2020-02-17 15-09-41

### DIFF
--- a/app/assets/stylesheets/banner.scss
+++ b/app/assets/stylesheets/banner.scss
@@ -1,0 +1,11 @@
+.banner {
+  margin-bottom: 2rem;
+
+  &__container {
+    max-width: 75rem;
+    height: auto;
+    width: 100%;
+    display: block;
+    margin: 0 auto;
+  }
+}
--- a/app/views/budgets/index.html.erb
+++ b/app/views/budgets/index.html.erb
@@ -9,7 +9,7 @@
 <% end %>
 
 <% if current_budget.present? %>
-  <div id="budget_heading" class="expanded budget no-margin-top">
+  <div id="budget_heading" class="expanded budget">
     <div class="row" data-equalizer data-equalizer-on="medium">
       <div class="small-12 medium-9 column padding" data-equalizer-watch>
 
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -23,3 +23,4 @@
 @import "autocomplete_overrides";
 @import "jquery-ui/sortable";
 @import "leaflet";
+@import "banner";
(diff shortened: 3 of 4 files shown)


app/assets/stylesheets/banner.scss
```diff
@@ -0,0 +1,11 @@
+.banner {
+  margin-bottom: 2rem;
+
+  &__container {
```
javierm, line 4: Instead of adding an extra container, do we get the same results if we apply these styles to the `a` tag inside the `.banner`? If so, I'd rather go for it; we already have too many `div` tags in our views :sweat_smile:.

app/views/budgets/index.html.erb
```diff
@@ -9,7 +9,7 @@
 <% end %>
 
 <% if current_budget.present? %>
-  <div id="budget_heading" class="expanded budget no-margin-top">
+  <div id="budget_heading" class="expanded budget">
```
javierm, line 12: Note removing this class will add an extra margin when there's no banner.

Also, since now we're adding a bottom margin to the banner, IMHO this change isn't necessary :thinking:. What do you think?

app/assets/stylesheets/banner.scss
```diff
@@ -0,0 +1,11 @@
+.banner {
+  margin-bottom: 2rem;
```
javierm, line 2: This one is tricky because the margin is needed in some sections but not in other sections :thinking:. I guess it's better to have too much margin in places like the proposals section rather than too little in places like the help page, so I'm fine with this change :+1:.

javierm:
Hi, @edumoreira1506 :smile:.

Are you still working on this pull request? :thinking:

javierm:
Due to inactivity, closing in favor of #4080.

PR #3926, opened 2020-02-20
Fix hidden active elements in admin menu

javierm:
## Background

In the admin menu, some links take you to a section, and some links open a submenu with more links.

When we disable the "multi-open" property of the menu and the active element is a link which takes you to a section, Foundation will hide it whenever we click a link which opens a submenu.

## Objectives

Make sure active menu items aren't hidden when navigating through the menus.

## Visual Changes

Before these changes:

The current element is hidden when opening a submenu  

After these changes:

The current element is present when opening a submenu


PierreMesure:
Well done! Pretty sure I opened an issue about it two years ago but can’t find it.

PR #3940, opened 2020-03-10
Add title to differentiate signature sheets

### LINKED ISSUE #3557: Add a name field for the signature sheets
## Background
Proposals can be supported by signatures, and those supports can be included in the system through the "Signature sheets" menu in `/admin/signature_sheets`.

## What we need
Include an optional new field "name" in `/admin/signature_sheets/new` that will be shown in the list of sheets `/admin/signature_sheets` to facilitate differentiating between them.

andrewcameronsims:
## References

* Closes #3557

## Objectives

Add a title to differentiate between signature sheets

### DIFF
--- a/app/controllers/admin/signature_sheets_controller.rb
+++ b/app/controllers/admin/signature_sheets_controller.rb
@@ -26,6 +26,6 @@ class Admin::SignatureSheetsController < Admin::BaseController
   private
 
     def signature_sheet_params
-      params.require(:signature_sheet).permit(:signable_type, :signable_id, :required_fields_to_verify)
+      params.require(:signature_sheet).permit(:signable_type, :signable_id, :document_numbers, :title, :required_fields_to_verify)
     end
 end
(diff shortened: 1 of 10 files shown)


app/controllers/admin/signature_sheets_controller.rb
```diff
@@ -26,6 +26,12 @@ def show
   private
 
     def signature_sheet_params
-      params.require(:signature_sheet).permit(:signable_type, :signable_id, :required_fields_to_verify)
+      params.require(:signature_sheet).permit(
+        :signable_type,
+        :signable_id,
+        :document_numbers,
```
javierm, line 32: I guess this line was added due fo a confict on this file :thinking:. The field `document_numbers` no longer exists, so we should remove it.

config/locales/es/activerecord.yml
```diff
@@ -302,6 +302,7 @@ es:
         body: "Mensaje"
         title: "Título"
       signature_sheet:
+        title: "Titulo"
```
javierm, line 305: There's a small typo here; correct Spanish is "Título" (you can copy the text two lines above this one if your keyboard doesn't generate Spanish characters :wink:).

andrewcameronsims:
@javierm Made the requested changes, build is green 😄

PR #3956, opened 2020-03-30
Fix minor design details in admin front

decabeza:
## Objectives

This PR:

- Fix bottom blank space in the admin section sidebar.
- Make a little a refactor in scss to avoid using a mixin for side menu and content styles.
- Improve styles for the admin sidebar aligning the icons correctly and adding a little CSS animation on the arrow icon when the menu opens.
- Remove admin header box-shadow.
- Fix icon sortable styles in the admin budget investment list.

## Visual Changes

admin

animation

icons

### DIFF
--- a/app/assets/stylesheets/admin.scss
+++ b/app/assets/stylesheets/admin.scss
@@ -435,11 +435,11 @@ code {
     [class^="icon-"] {
       color: #fff;
       display: inline-block;
-      font-size: rem-calc(24);
+      font-size: rem-calc(20);
       line-height: $line-height;
-      padding: $line-height / 2 $line-height / 4;
-      padding-left: 0;
+      text-align: center;
       vertical-align: middle;
+      width: $line-height * 1.5;
     }
   }
 
@@ -458,13 +458,20 @@ code {
       border-left: 2px solid $sidebar-active;
       font-weight: bold;
     }
+
+    &[aria-expanded="true"] {
+
+      > a::after {
+        transform: rotate(180deg);
+      }
+    }
   }
 
   li a {
     color: #fff;
     display: block;
-    line-height: rem-calc(48);
-    padding-left: rem-calc(12);
+    line-height: $line-height * 2;
+    padding-left: $line-height / 4;
     vertical-align: top;
 
     &:hover {
@@ -478,12 +485,13 @@ code {
 
     > a::after {
       border: 0;
-      content: "\61" !important;
-      font-family: "icons" !important;
+      content: "\f078";
+      font-family: "Font Awesome 5 Free";
+      font-weight: bold;
       height: auto;
-      position: absolute !important;
-      right: 30px;
-      top: 6px !important;
+      position: absolute;
+      right: 24px;
+      transition: 0.25s;
     }
   }
 
--- a/app/assets/stylesheets/admin.scss
+++ b/app/assets/stylesheets/admin.scss
@@ -63,10 +63,6 @@ $sidebar-active: #f4fcd0;
     color: #000;
     height: auto;
 
-    @include breakpoint(medium) {
-      box-shadow: 0 2px 2px #eee;
-    }
-
     @include breakpoint(small only) {
 
       .top-bar-left ul {
--- a/app/assets/stylesheets/admin.scss
+++ b/app/assets/stylesheets/admin.scss
@@ -364,26 +364,25 @@ $sidebar-active: #f4fcd0;
 }
 
 .icon-sortable {
-  font-family: "icons";
+  font-family: "Font Awesome 5 Free";
   font-size: $small-font-size;
-  padding-right: $line-height / 2;
   position: relative;
 
   &::before,
   &::after {
     left: 6px;
-    opacity: 0.25;
+    opacity: 0.5;
     position: absolute;
   }
 
   &::before {
-    content: "\57";
-    top: -2px;
+    content: "\f0d8";
+    bottom: 0;
   }
 
   &::after {
-    content: "\52";
-    bottom: -10px;
+    content: "\f0d7";
+    top: 0;
   }
 
   &.asc {
--- a/app/assets/stylesheets/layout.scss
+++ b/app/assets/stylesheets/layout.scss
@@ -446,18 +446,18 @@ a {
 }
 
 .menu-and-content {
-  $side-menu-min-width: 250px;
+  $side-menu-min-width: rem-calc(240);
 
   @include breakpoint(medium) {
     display: flex;
 
     > nav {
-      flex: 25%;
+      flex: 20%;
       min-width: $side-menu-min-width;
 
       + * {
-        flex: 75%;
-        padding: $line-height !important;
+        flex: 80%;
+        padding: $line-height;
       }
     }
 
@@ -478,7 +478,7 @@ a {
       z-index: 12;
 
       + * {
-        padding: $line-height !important;
+        padding: $line-height;
       }
     }
 


javierm:
Thanks @decabeza :smile:. A couple of questions before reviewing.

> Fix bottom blank space in the admin section sidebar.

Could you give me a hint about how to reproduce the issue? :pray: It works fine for me using both Firefox and Chrome :thinking:. Problems like this one were supposed to be fixed in pull request #3778.

> Make a little a refactor in scss to avoid using a mixin for side menu and content styles

Would it be possible to elaborate about why this change makes the code better? :smile:

decabeza:
> Could you give me a hint about how to reproduce the issue? 🙏 It works fine for me using both Firefox and Chrome 🤔. Problems like this one were supposed to be fixed in pull request #3778.

On `master` branch I keep seeing the blank space in all browsers 😅 
## Firefox
firefox
## Safari
safari
## Chrome
chrome

decabeza:
> Would it be possible to elaborate about why this change makes the code better? 😄

Personally I don't like mixins, they make me have to look another file to know what that code does. 

Also, I think that they don't add much when only the @include is used but they don't add anything else. 

For example instead of:
```
@mixin border-red {
  border: 1px solid #f00;
}

.border-red-one {
  @include border-red;
}

.border-red-two {
  @include border-red;
}
```
I think it's better:

```
.border-red {
  border: 1px solid #f00;
}
```

Why have two classes that only include the mixin when you can have a single class that does the same thing? 🤔 

This avoids having two classes that do basically the same thing and are referenced in another file. As I say it's a personal preference, but I think in this case it doesn't add anything. Of course, we could create mixins for other styles with a great CSS code refactor although I'm more in favor of eliminating mixins completely.

For now, all the @mixins created are used only once, so they don't make much sense. 😅

javierm:
> On master branch I keep seeing the blank space in all browsers sweat_smile

Now I get it :smile:. I don't have the issue because my resolution is low enough so the menu doesn't fit in the screen, but I can reproduce it by removing some elements in the menu :smile:.

javierm:
OK, not a problem removing the mixin; just thought it was better to elaborate on the reasons, so when we look back at the code we know why it was changed :smile:.

Having said that, in general I prefer mixins :rofl:. Changing your example a little bit:

```scss
@mixin border-red {
  border: 1px solid #f00;
}

.proposal {
  @include border-red;
}

.poll-answer {
  @include border-red;
}
```

Looks better to me than:

```scss
.border-red {
  border: 1px solid green; // Changed in a fork because CSS is easier to change than views
}
```

In the particular case of this pull request, using a mixin doesn't make a big difference to me, so I'm OK with the change :relieved:.

app/assets/stylesheets/layout.scss
```diff
@@ -449,6 +449,40 @@ a {
   margin-top: 0;
 }
 
+.admin {
+
+  .side-menu-and-content {
+    min-height: calc(100vh - 110px);
```
javierm, line 455: Just curious: where are these numbers (110px here and 181px for the dashboard) coming from? :thinking:

On my machine this makes a scrollbar appear in the browser's window. Maybe it's just a me, but I've always found it a usability issue: the scrollbar appears and I feel it's asking me to scroll and check the content below... but then, there's nothing below :sweat_smile:.

Scrollbar appearing when there is no content below

Do you think it's possible to use a more precise unit, or will it always depend on the browser and screen resolution?

By the way, would it be easier to accomplish this goal if we had less `<div>` tags in that part of the code? I made an experiment a few months ago, (see javierm#12), but as we mentioned back then, I wanted to know your opinion before continuing working on it.

app/assets/stylesheets/admin.scss
```diff
@@ -461,13 +452,20 @@ code {
       border-left: 2px solid $sidebar-active;
       font-weight: bold;
     }
+
+    &[aria-expanded="true"] {
+
+      > a::after {
+        transform: rotate(180deg);
```
javierm, line 459: I love this one :smile:.

app/assets/stylesheets/admin.scss
```diff
@@ -62,10 +62,6 @@ $sidebar-active: #f4fcd0;
     color: #000;
     height: auto;
 
-    @include breakpoint(medium) {
-      box-shadow: 0 2px 2px #eee;
-    }
-
```
javierm, line 68: I'm fine with this change :ok_hand:. Would it be possible to know the reason (it would be awesome if it were explained in the commit message!)? I'm asking because I wouldn't know what to say if somebody else opens a pull request and adds these lines again :roll_eyes:.

decabeza: I don't see the need, but I've added it to the commit description. 😅 The reason is that it is only decorative and has been deleted because it was also deleted in the front views.

app/assets/stylesheets/layout.scss
decabeza: The numbers come from headers height: **110px** from admin header and **181px** from the front header.

Those pixels are absolute, I mean, they couldn't be more accurate. 😛 The problem of showing or not the scroll bar I think is complicated because it depends a lot on the screen resolution, I have tried in my laptop and in two external screens and it doesn't appear to me. Anyway, I personally prefer a scroll bar to "nothing", which appears on many websites and I can understand that it is due to the size of my screen instead of blank space that gives the impression that the layout is broken.

I have looked at https://github.com/javierm/consul/pull/12 the PR but I do not see the point. If I understand correctly, to simplify the HTML, JS custom is added. This seems to me more complicated to maintain since CONSUL uses a framework, so... I don't see sense to create "custom" files to add a function that the Foundation framework already does by default. The HTML may seem more complicated but you can always consult the official documentation, if you add JS custom it seems more complicated to me for someone who wants to make a modification.

I don't see a simple solution since it will always depend on the browser and screen resolution. Also, as these changes are related to fixing minor design details administration panel I don't see the point of creating a separate PR. 🙏

app/assets/stylesheets/mixins.scss
```diff
@@ -22,29 +21,7 @@
...
-
-  > :last-child {
-    flex: 75%;
-    overflow-x: auto;
-    padding: $line-height !important;
-  }
-}
-
-// 03. Orbit bullet
+// 02. Orbit bullet
```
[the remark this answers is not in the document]
decabeza: Done! 😉

app/assets/stylesheets/layout.scss
javierm: > Those pixels are absolute, I mean, they couldn't be more accurate. :stuck_out_tongue:

Maybe we're talking about different things :sweat_smile:. I mean, the headers do not always have that height. Here are screenshots of firefox and chrome on my machine:

Header in Firefox: 124.65px

Header in Chrome: 117px

Not to mention they also change depending on the user's font size :relieved:.

It seems to me using relative units wouldn't solve the issue either :thinking:.

Do you think it's feasible to use `flex`, which would solve these problems on every browser?

> if you add JS custom it seems more complicated to me for someone who wants to make a modification.

Fair point. It would be great to know what's harder to maintain: a custom view which will have to be changed every time the original view changes or a custom JavaScript file changing the behavior defined in the original JavaScript file :thinking:.

> I personally prefer a scroll bar to "nothing"

We agree to disagree :smile:. Unfortunately we can't do usability tests to check which option is worse :sweat_smile:.

javierm: I've updated javierm#12 so it doesn't use JavaScript anymore.

The behaviour is not the same as it is now, though :relieved:. Let me know what you think :wink:.

javierm: EDIT: the behaviour is now similar to the `master` branch :smile:.

javierm: I've updated the pull request once again. Now it doesn't use JavaScript and the back button isn't broken either. It uses a (hidden) checkbox to toggle the menu, which IMHO semantically makes more sense than a link.

javierm: One more update. I've added a commit which solves the blank space on the bottom and doesn't add a scrollbar. It seems to be really simple after taking all the `off-canvas` div tags.

Note so far I've only done changes in the admin section; the dashboard layout is probably broken right now.

javierm: @decabeza I've improved the code a bit and tested the cases I could think of. Feel free to check pull requests #4005 and #4006 :wink: (only if you'd like to do so, of course!)

Regarding this pull request, I was thinking maybe we could remove the commits removing the mixin (already included in #4005) and fixing the blank space (solved by #4006) :thinking:. What do you think?

javierm: Proceeding as mentioned, as #4005 has already been merged.

javierm (approved):

Issue #3676, opened 2019-08-12
Translation override - variable not passed


milovanderlinden:
We discovered that with translation overrides from the database table `i18n_content_translations` variables are not interpreted.

This means that when we set a translation to `%{foo} is awesome`, this gets displayed as is. The solution is rather simple:

In the `i18n_translation.rb` helper, we need to check for options and when present, pass them as follows:
```ruby
if translation.present?
  # Allow for options to be translated
  if options.present?
    translation % options
  else
    translation
  end
else
  translate(key, options)
end
```

PR #3959, opened 2020-03-31
Fix custom translations with options

decabeza:
## References

This PR resolves https://github.com/consul/consul/issues/3676.

## Objectives

There was a **bug** when an admin user tries to change a text on the "Custom information text" section `/admin/site_customization/information_texts` with the keys that include some `%{variable}` in the text.

#### BEFORE
interpolation

#### AFTER
after

### DIFF
--- a/config/initializers/i18n_translation.rb
+++ b/config/initializers/i18n_translation.rb
@@ -12,7 +12,11 @@ module ActionView
         i18_content = I18nContent.find_by(key: key)
         translation = I18nContentTranslation.find_by(i18n_content_id: i18_content&.id,
                                                      locale: current_locale)&.value
-        translation.presence || translate(key, options)
+        if translation.present?
+          translation % options
+        else
+          translate(key, options)
+        end
       end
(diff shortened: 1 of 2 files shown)


config/initializers/i18n_translation.rb
```diff
@@ -12,7 +12,15 @@ def t(key, options = {})
         i18_content = I18nContent.find_by(key: key)
         translation = I18nContentTranslation.find_by(i18n_content_id: i18_content&.id,
                                                      locale: current_locale)&.value
-        translation.presence || translate(key, options)
+        if translation.present?
+          if options.present?
```
javierm, line 16: I'm not sure this condition is necessary :thinking:. What do you think about:

```ruby
        if translation.present?
          translation % options
        else
          translate(key, options)
        end
```

decabeza: I used the first option that seems simpler. The second one is also very hard to understand for me. 😂

Issue #2605, opened 2018-04-18
Replaces icons of expand/collapse comments


decabeza:
Background
==

Now on all item with comments we use an arrow to expand/collapse comments.

Background
==

Replaces arrow icon with plus/minus icons to expand/collapse comments.

Screenshots
==

**Actual icons**
![screen shot 2018-04-18 at 13 29 15](https://user-images.githubusercontent.com/631897/38929486-c080c66e-430c-11e8-9daa-4cb2aaf42e11.png)

![screen shot 2018-04-18 at 13 29 18](https://user-images.githubusercontent.com/631897/38929487-c09f2b2c-430c-11e8-9305-cc7387df7bb5.png)

**Replace to this ones**
![screen shot 2018-04-18 at 13 28 47](https://user-images.githubusercontent.com/631897/38929506-ce2650cc-430c-11e8-84fd-58d028fdf2e0.png)
![screen shot 2018-04-18 at 13 28 59](https://user-images.githubusercontent.com/631897/38929508-ce4294f8-430c-11e8-9623-f615df36b664.png)

jesushp:
It would be useful to reflect all that kind of decisions on a common place, as to speak, a Design System.

PR #3972, opened 2020-04-14
Replaces icons of expand/collapse comments

decabeza:
## References

This PR close https://github.com/consul/consul/issues/2605.

## Objectives

Replaces icons of expand/collapse comments.

### Before
before

### After
after

### DIFF
--- a/app/assets/javascripts/comments.js
+++ b/app/assets/javascripts/comments.js
@@ -40,13 +40,7 @@
       $("#js-comment-form-" + id).toggle();
     },
     toggle_arrow: function(id) {
-      var arrow;
-      arrow = "span#" + id + "_arrow";
-      if ($(arrow).hasClass("icon-arrow-right")) {
-        $(arrow).removeClass("icon-arrow-right").addClass("icon-arrow-down");
-      } else {
-        $(arrow).removeClass("icon-arrow-down").addClass("icon-arrow-right");
-      }
+      $("span#" + id + "_arrow").toggleClass("fa-minus-square").toggleClass("fa-plus-square");
     },
     initialize: function() {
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/layout.scss
```diff
@@ -2116,17 +2116,16 @@ table {
...
-    [class^="icon-arrow"] {
-      padding-left: $line-height / 2;
+    .relative {
+      padding-left: rem-calc(18);
     }
 
-    [class^="icon-arrow"] {
-      font-size: $base-font-size;
-      left: -20px;
+    [class^="far"] {
```
javierm, line 2123: I guess here we mean `.far` (correct me if I'm wrong). Otherwise we might accidentally target classes starting with `far` (which isn't a common way to start a word, but you never know, particulary with all these modern class names :smile:).

By the way, ideally we wouldn't have to change the view in order to change an icon and changing CSS would be enough, but that's another story :wink:.

app/assets/javascripts/comments.js
```diff
@@ -42,10 +42,10 @@
     toggle_arrow: function(id) {
       var arrow;
       arrow = "span#" + id + "_arrow";
-      if ($(arrow).hasClass("icon-arrow-right")) {
-        $(arrow).removeClass("icon-arrow-right").addClass("icon-arrow-down");
+      if ($(arrow).hasClass("far fa-minus-square")) {
```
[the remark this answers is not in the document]
decabeza: The `far` class is not necessary for the JS only in the views, good refactor! 😉

app/assets/stylesheets/layout.scss
decabeza: Good one! 😌

Issue #3092, opened 2018-12-08
Foundation equalizer not working correctly on Safari


PierreMesure:
### Description

[comment]: # (Feature or Bug? eg. Type: Bug)
*Type*: Bug

Config: Safari 12.0.1 on macOS 10.13.6

This seems to be a [known](https://foundation.zurb.com/forum/posts/35313-data-equalizer-not-working-in-safari---inconsistant) problem associated to Foundation equalizer.

The solution should be to reload the sizes on page load but I got inconsistent results with it:

```
$(window).load(function(){
    $(document).foundation('equalizer', 'reflow');
});
```

### Current Behavior
[comment]: # (Describe what actually happened.)

![capture d ecran 2018-12-08 a 10 54 24](https://user-images.githubusercontent.com/7223028/49684748-1eb72880-fad8-11e8-801c-55a14d61e140.png)

### Expected Behavior
[comment]: # (Describe what you expected to happen.)

![capture d ecran 2018-12-08 a 10 56 36](https://user-images.githubusercontent.com/7223028/49684749-21198280-fad8-11e8-959a-8e0ee822a553.png)

### Steps to reproduce
[comment]: # (Describe how we can replicate the bug step by step.)

Just visit any consul homepage with cards.

### Related issues

- #3044

PR #3973, opened 2020-04-14
Replace equalizer to display flex on cards

decabeza:
## References

This PR close https://github.com/consul/consul/issues/3092.

## Objectives

Replace Foundation `data-equalizer` to CSS `display: flex` on cards for homepage and custom pages avoiding use JS and prevent the layout break when the user uses the back button in their browser.

Also, move cards views to shared partials to `shared/_cards.html.erb` and `shared/_card.html.erb`.

## Visual Changes

All the cards maintain the same height per row independently of the content.

homepage_cards


app/views/pages/_card.html.erb
```diff
@@ -1,5 +1,5 @@
 <div id="<%= dom_id(card) %>"
-     class="card small-12 medium-<%= card.columns %> column margin-bottom end" data-equalizer-watch>
+     class="card small-12 medium-<%= card.columns %> column margin-bottom end">
```
javierm, line 2: I see this partial and the one in `welcome/_card` are identical :thinking:. Would it make sense to use one file instead, under either `cards/_card` or `shared/_card`? That way we wouldn't have to change the code in two places :smile:.

app/views/welcome/_cards.html.erb
```diff
@@ -1,7 +1,9 @@
...
-  <% @cards.each do |card| %>
-    <%= render "card", card: card %>
-  <% end %>
+<div class="row">
+  <div class="cards-container">
+    <% @cards.each do |card| %>
+      <%= render "card", card: card %>
+    <% end %>
+  </div>
 </div>
```
javierm, line 9: Maybe we could also extract this block into a partial in either `cards/_cards` or `shared/_cards` :thinking:. What do you think?

app/assets/stylesheets/layout.scss
```diff
@@ -2788,11 +2788,18 @@ table {
...
   }
 
+  .cards-container {
+    display: flex;
+    flex-wrap: wrap;
+  }
+
   .card {
     border: 0;
+    flex-grow: 0;
```
javierm, line 2798: AFAIK `flex-grow` is `0` by default :thinking:. Is there a reason to explicitely add this line?

app/views/pages/_card.html.erb
decabeza: Yes, one file it's better! I've created two new partials `shared/_cards` and `shared/_card`. 😌

PR #3976, opened 2020-04-17
Deactivate ckeditor file attachments feature

Senen:
## References
By working on #3659 i realized about this error.

## Objectives
This feature was not working because it was not fully configured and tested so its better to disable it completely.

## Visual Changes
**Before**
Captura de pantalla 2020-04-17 a las 20 17 45

**After**
Now ckeditor buttons related to file uploads are not shown anymore.
Captura de pantalla 2020-04-17 a las 19 55 48

## Notes
-

### DIFF
--- a/spec/features/ckeditor_spec.rb
+++ b/spec/features/ckeditor_spec.rb
@@ -30,4 +30,15 @@ describe "CKEditor" do
 
     expect(page).to have_css "img[src$='clippy.jpg']"
   end
+
+  scenario "cannot upload attachments through link tab", :js do
+    login_as(create(:administrator).user)
+    visit new_admin_site_customization_page_path
+
+    find(".cke_button__link").click
+
+    expect(page).to have_css(".cke_dialog")
+    expect(page).not_to have_link "Upload"
+    expect(page).not_to have_link "Browse Server"
+  end
 end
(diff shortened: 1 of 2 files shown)


spec/features/ckeditor_spec.rb
```diff
@@ -30,4 +30,14 @@
...
     expect(page).to have_css "img[src$='clippy.jpg']"
   end
+
+  scenario "cannot upload attachments through link tab", :js do
+    login_as(create(:administrator).user)
+    visit new_admin_site_customization_page_path
+
+    find(".cke_button__link").click
+
+    expect(page).not_to have_link "Upload"
```
javierm, line 40: This test passes even without the changes in this pull request :thinking:.

I think we need to add an extra expectation before this line. Otherwise this line might be executed before CKEditor's dialog appears, and pass because at that time thi "Upload" link is still not present.

How about:

```ruby
    expect(page).to have_css(".cke_dialog")
    expect(page).not_to have_link "Upload"
    expect(page).not_to have_link "Browse Server"
```

Senen: Hi @javierm you are right, i swear yesterday i saw this test failing in my machine! Maybe i forgot to save a file before running the spec. Now i added the first expectation you suggested! Thanks a lot

PR #3996, opened 2020-04-30
Optmize docker

luchiago:
## References

## Objectives

Make some optimizations on docker. The first one is to remove the server pid before each initialization of the server. The second one I observed that the docker-compose were building the image again because it wasn't using the built and tagged image.

## Visual Changes

## Notes

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -13,9 +13,8 @@ services:
 
   # service configuration for our dockerized Rails app
   app:
-
     # use the Dockerfile next to this file
-    build: .
+    image: consul
     entrypoint: /usr/local/bin/entrypoint.sh
     command: bundle exec rails s -p 3000 -b "0.0.0.0"
     working_dir: /var/www/consul
--- a/scripts/entrypoint.sh
+++ b/scripts/entrypoint.sh
@@ -13,5 +13,5 @@ usermod -g "$USER_GID" consul 2> /dev/null
 chown -R -h "$USER_UID" "$BUNDLE_PATH"
 chgrp -R -h "$USER_GID" "$BUNDLE_PATH"
 
+/usr/bin/sudo rm /var/www/consul/tmp/pids/server.pid
 /usr/bin/sudo -EH -u consul "$@"
-


luchiago:
Hi @javierm, how are you man? Should we close this PR or still needed?

javierm:
@raoulkramer What do you think of the changes in this pull request?

raoulkramer:
@javierm looks interesting, I will test the suggested changes on monday! Especially I like d0914d9 because it will take away, manual/ server level intervention when you use a docker/ container managing interface. Additionally I will compare the current dockerfile and docker-compose.yml with the one we have modified, and see if there are more additions that could improve the file.

luchiago:
@javierm @raoulkramer Thanks for reply my and give me attention, even this PR being too stale 😅. At the time, I had a problem building the docker image, and to remove the PID during docker compose-up, you probably will test my changes looking forward to these problems. Whether my PR is accepted or not, I'm happy with my contribution to you guys, a great project.

raoulkramer:
@luchiago @javierm I have been looking into d0914d9 . The implementation suggested in the commit doesn't work with my setup. I looked further into removing the server.pid with a docker restart. This could be possible to handle the pid remove command as a separate script / .sh file, I am working on that. But a side note, it is not inline with the docker approach where the concept is to replace your container with a fresh one, then server.pid is automatically removed.... But we have some user cases where it is convenient to keep the container.

javierm:
A variation of the first change has already been included  in pull request #4684, so I guess the question now is, does this change make sense? @raoulkramer @abihas Any thoughts? :thinking:

```diff
  # service configuration for our dockerized Rails app
  app:
    # use the Dockerfile next to this file
-    build: .
+    image: consul
    entrypoint: /usr/local/bin/entrypoint.sh
    command: bundle exec rails s -p 3000 -b "0.0.0.0"
    working_dir: /var/www/consul
```

raoulkramer:
@javierm :wave:  I think with #4684 we tackled @luchiago First Objective. The Second Objective needs some research and testing.

luchiago:
@raoulkramer As far as I remember, every time we up the containers the images were rebuild

raoulkramer:
@luchiago I never experienced that images were rebuild with a `docker compose up -d`  :thinking:

javierm:
I can reproduce the issue @luchiago mentioned by following the steps in the [CONSUL documentation for Docker](https://docs.consulproject.org/docs/english-documentation/introduction/servers/docker):

```bash
docker build -t consul .
(...)
POSTGRES_PASSWORD=password docker-compose up
```

Doing so will generate the image twice: first with `docker build` and then with `docker-compose up`.

However, I think we can't really rely on adding `image: consul` because, if you haven't already run `docker build -t consul`, it will download [Hashicorp's Consul Docker image](https://hub.docker.com/_/consul) :sweat_smile:.

So maybe the documentation isn't right and this is a documentation issue? :thinking:

raoulkramer:
@javierm what result do you get when you use `docker compose build` (thats what I use)?   [Difference between `docker build` and `docker compose build`](https://stackoverflow.com/questions/50230399/what-is-the-difference-between-docker-compose-build-and-docker-build) and  than It uses the docker-compose.yml file in order to retrieve the (build) parameters. 
----
ps the `docker-compose` cmd changed in `docker compose` replacing the dash (-) with a space, since Compose V2 https://docs.docker.com/compose/cli-command/ . This should potentially by changed in the docs.. :thinking:

javierm:
@raoulkramer Looks good :+1:. For now, we're keeping `docker-compose` because it looks like `docker compose` has only been introduced lately and so it wouldn't work for people using older versions, but it's great to know we'll have to update it in the future :smile:. Correct me if I'm wrong! :pray:

@luchiago I've just opened consul/docs#111 with updated instructions in order to avoid building the images twice, so I'm closing this pull request. Please reopen if I didn't understand the issue and/or it isn't fixed with the changes in the documentation :pray:. And thanks a lot for reporting the problem and for your patience waiting for our replies :smile:.

luchiago:
No worries @javierm. Happy to open this discussion with you guys.

PR #4000, opened 2020-05-06
Fix ckeditor compatibility with Turbolinks 5.x 

Senen:
## References

Related pull requests: #3699, #3998

This is and update of #3998 to make it work with Turbolinks 5.

## Background
Now we are using Turbolinks 5.X we are initializing application JS on turbolinks:load event which is fired up on first page visits and on subsequent Turbolinks page loads, so with this version of Turbolinks we only need to destroy existing ckeditor instances before rendering the new page to avoid JS errors when using browser history back. This solution was extracted from here [1].

[1] galetahub/ckeditor#575 (comment)

## Objectives
Fix javascript errors when using browser history back to restore a page with ckeditor instances.

## Visual Changes
None

## Notes
**Testing pending on production environment.**

This PR includes commits of #3699 temporarily, and most sure we will need to revisit the commit to solve conflicts if #3998 finally gets merged.


javierm:
I've added the relevant parts to pull request #3699. Will rebase it when we merge #3998.

PR #4002, opened 2020-05-08
Update comment responses count when adding replies

Senen:
## References

This PR needs #3997 commits to work. Will rebase when #3997 is merged.

## Objectives
Update the parent comment "responses count" when a new reply is added.

Replace parent comment of a reply when a reply for a comment is added, this will replace parent comment which will include the reply also.

Changes introduced at 7105f21 (the comments part) are needed to initialize automatically comments added or replaced through ajax calls.

## Steps to reproduce
1. Go to any debate (or any other commentable resource)
2. Add a reply to a comment

At this point you can see your reply but parent comment responses count is not updated!

## Visual Changes

**Before**
ResponsesCountUpdateError

**After**
ResponsesCountUpdateFix

## Notes
None

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -173,5 +173,4 @@ $(function() {
 
   $(document).ready(initialize_modules);
   $(document).on("page:load", initialize_modules);
-  $(document).on("ajax:complete", initialize_modules);
 });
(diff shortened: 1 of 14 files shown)


javierm:
Closing in favor of #4003.

PR #4004, opened 2020-05-08
Move conditional into shared banner partial

decabeza:
## Objectives

A little refactor to avoid always writing `if has_banners?` every time the partial `shared/banner` is used.

Before you have to write:
```erb
<% if has_banners? %>
  <%= render "shared/banner" %>
<% end %>
```
Now it's enough with:   
```erb
<%= render "shared/banner" %>
```

😌


javierm:
It looks like I forgot to add this pull request to our project board :thinking:. I really hope GitHub implements the option to automatically add new pull requests :relieved:.

PR #4022, opened 2020-05-25
Improve watch form changes feature

Senen:
## References
Related PR: #3699  (See Notes section)

## Objectives
1. Show "Unsaved changes" alert dialog when a user leaves a page with a "form to track changes" with unsaved data even when page is restored from browser history.

2. Do not show alert if users restores the original value (by hand) after modifying it.

## Solution
Data attributes dynamically added to the form to store form input values were lost when using browser history so when user navigates back to a page with a form with unsaved changes the form serialization is done again but this time with modified but unpersisted values so application thinks form wasn't changed, by using a hidden input to store form serialized data we can keep warning the user about unsaved changes from original version (persisted).

## Notes
None

### DIFF
--- a/app/assets/javascripts/watch_form_changes.js
+++ b/app/assets/javascripts/watch_form_changes.js
@@ -85,5 +85,10 @@
     }
   };
 
-  $(document).on("turbolinks:before-visit", App.WatchFormChanges.checkChanges);
+  $(document).on("turbolinks:before-visit", function() {
+    // Ckeditor updates textarea input only on form submit. This patch will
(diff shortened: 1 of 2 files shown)


Senen:
Closing because is a too complex solution that will be difficult to maintain. Application will be only watching for changes at legislation draft versions form markdown editor where to warn the user about unsaved changes is useful still.

javierm:
Thanks for going though the nightmare of investigating these issues! :pray: I've applied the (slightly modified) tests to pull request #3699.

PR #4041, opened 2020-06-15
Refactor code to render admin table actions

javierm:
## Objectives

* Remove code duplication
* Make it easier to change the styles of links to edit, destroy, restore, ... a record
* Make the table actions work better at medium resolutions
* Make the user interface more consistent; for example, always ask for confirmation when deleting a record and alwas display the link to the "delete" action after the other links

## Notes

* This pull request introduces components; we need to debate whether we adopt this pattern in future developments
* The "Global settings" and "Custom images" sections have not adopted the new layout because they also use a form; we have to decide whether these actions should look like forms do or like table actions do

## Visual changes

All these changes are relevant using medium resolutions

### Before these changes

The text exceeds its container's boundaries

The text exceeds its container's boundaries

Links are expanded over two lines, breaking the layout

### After these changes

The text fits inside its boundaries

The text fits inside its boundaries

Links stay on one line

## Release notes

We've started to use components, which are a way to group helpers and views together. Existing helpers and views will still work properly. If you'd like to customize a component, check the documentation. TODO: link to the documentation to customize components.


taitus (approved):

Senen (approved):
I think we are going to ❤️ this pattern. 👏

PR #4046, opened 2020-06-18
Remove unused document section on polls

decabeza:
## Objectives

This section is not used because it's only possible to add documents to the poll's answers, not to the poll itself.

## Notes

The `<% if false %>` was [added 3 years ago](https://github.com/consul/consul/commit/70a35fbe58221bf6bd7a04683f9a5a432b9b39dd#diff-9d7534803a942c4e1527d539936492ecR46) so effectively this section is not being used.

The key `t("polls.show.documents")` is still used in another part of the view so it is not necessary to remove it.


javierm:
Cool! :smile: Thanks! :tada:

PR #4048, opened 2020-06-18
Remove WatchFormChanges feature

Senen:
## References
Related Pull Requests: #4022

## Objectives
As the solution created at #4022 is too complex to maintain and now that users can leave a page with unsaved changes and restore it without loosing unsaved changes (#3998) this feature seems to be not so important now.

Check #4022 commits messages for a complete explanation of errors we found reviewing `WatchFormChanges` feature in  deep.

## Visual Changes
Users will no see the alert popup when leaving a page with a form with unsaved chages.

## Notes
None


Senen:
There is still one place where warning the user about unsaved changes could be useful, the legislation draft version markdown editor. This finally will be addressed at #3699.

javierm:
Most of these changes have been applied to pull request #3699. Thanks a lot!

Issue #4011, opened 2020-05-15
Error on processes with the comments phase's blank date


decabeza:
### Description

Type: Bug

### Current Behavior
The page generates an error when visiting a process with the comments phase's blank date.

![error](https://user-images.githubusercontent.com/631897/82036029-abd4b880-96a0-11ea-91c6-8eb9f7c8fa5c.png)

### Steps to reproduce
1. Create a new process in `/admin/legislation/processes/new` page.
2. Mark **Comments phase** as "Enabled" but leaving date fields blank.
3. Visit process page `/legislation/processes/N/`.

### Definition of Done
I think there should be a validation so phases with blank dates cannot be enabled.

![form](https://user-images.githubusercontent.com/631897/82036704-81cfc600-96a1-11ea-9092-3360c5f07783.png)

Right now, the rest of the phases: Draft phase, Debate phase, and Proposals phase can be enabled without a date. This doesn't have much sense for me, anyway I think the key is the coherence, I think all phases should be enabled only with a valid date.

PR #4052, opened 2020-06-24
Fix phase started bug

edumoreira1506:
## References

> Related issue #4011 

## Objectives

> Fix ArgumentError in Legislation::ProcessesController#allegations

## Visual Changes

Peek 2020-06-24 18-18

### DIFF
--- a/app/models/legislation/process/phase.rb
+++ b/app/models/legislation/process/phase.rb
@@ -10,7 +10,7 @@ class Legislation::Process::Phase
   end
 
   def started?
-    enabled? && Date.current >= @start_date
+    @start_date && enabled? && Date.current >= @start_date ? true : false
   end
 
   def open?
--- a/spec/models/legislation/process/phase_spec.rb
+++ b/spec/models/legislation/process/phase_spec.rb
@@ -49,6 +49,11 @@ RSpec.describe Legislation::Process::Phase, type: :model do
       process.update!(debate_start_date: Date.current - 2.days,
                       debate_end_date: Date.current - 1.day)
       expect(process.debate_phase.started?).to be true
+
+      # nil
+      process.update!(debate_start_date: nil,
+                      debate_end_date: nil)
+      expect(process.debate_phase.started?).to be false
     end
 
     it "checks draft phase" do


app/models/legislation/process/phase.rb
```diff
@@ -10,7 +10,7 @@ def enabled?
   end
 
   def started?
-    enabled? && Date.current >= @start_date
+    return enabled? && Date.current >= @start_date unless @start_date.nil?
```
lairtonlelis, line 13: `return` is unnecessary in this case, ruby has an implicit return. And regarding the use of `unless` for the nil check, a simple `if` would make the code more readable, like this:
```suggestion
    enabled? && Date.current >= @start_date if @start_date
```

spec/models/legislation/process/phase_spec.rb
```diff
@@ -49,6 +49,11 @@
       process.update!(debate_start_date: Date.current - 2.days,
                       debate_end_date: Date.current - 1.day)
       expect(process.debate_phase.started?).to be true
+
+      # nil
+      process.update(!debate_start_date: nil,
```
lairtonlelis, line 54: There is a syntax error here. Perhaps you meant to use `update!`?
```suggestion
      process.update!(debate_start_date: nil,
```

edumoreira1506: Yep, my bad, thank you.

app/models/legislation/process/phase.rb
edumoreira1506: Fixed.

edumoreira1506:
Could you review @decabeza ?

javierm:
@edumoreira1506 Thanks for your contribution! :pray:

Before reviewing, could you comment what your thoughts about adding validation rules (as @decabeza suggested in issue #4011) are? Personally I like the idea a lot, but I haven't investigated the issue so I'm not sure whether we could implement it by changing these lines in the `Legislation::Process` or this approach would make the applcation work unexpectedly in some places :thinking::

```ruby
  validates :debate_start_date, presence: true, if: :debate_end_date?
  validates :debate_end_date, presence: true, if: :debate_start_date?
  validates :draft_start_date, presence: true, if: :draft_end_date?
  validates :draft_end_date, presence: true, if: :draft_start_date?
  validates :allegations_start_date, presence: true, if: :allegations_end_date?
  validates :allegations_end_date, presence: true, if: :allegations_start_date?
  validates :proposals_phase_end_date, presence: true, if: :proposals_phase_start_date?
```

javierm:
@edumoreira1506 Any comments about the validation rules I mentioned?

javierm:
Closing due to lack of response.

PR #2151, opened 2017-11-23
Feature flag api

xuanxu:
## References

* Closes #1592 

## Objectives

Add a feature flag for the GraphQL API

PR #4057, opened 2020-07-09
Deactivate GraphQL API feature by default

javierm:
## References

* The option to enable/disable our GraphQL API was introduced in pull request #2151
* In pull request #4028 we stopped loading GraphQL definitions when the application starts

## Objectives

Disable the GraphQL API by default, since it's a feature which most Consul installations don't use.

## Notes

* TODO: update the documentation, and disable access to /graphiql when this feature is disabled

Existing installations will not be affected by this change.


javierm:
The API seems to be used by more Consul installations than we thought, so I'm closing this pull request for now.

PR #4066, opened 2020-07-25
Show Wordpress login button if it's the only one enabled

microweb10:
## References

Wordpress login was introduced in PR https://github.com/consul/consul/pull/3902 but a BUG was also introduced, the button will not appear if Wordpress login is the only one external authentication feature enabled.

## Objectives

- Fix the BUG and show the Wordpress login button if it's the only one enabled.


javierm:
Cool! Thanks! :smile:

PR #4111, opened 2020-08-27
Use native HTML5 date fields in the admin section

javierm:
## References

* Since August the 27th, a test is failing due to the jQuery UI datepicker; for example, in [Travis build 34601, job 4](https://travis-ci.org/github/consul/consul/jobs/721625373)

## Background

We've had to add a couple of hacks in order to make jQuery UI datepicker work with Turbolinks, and one of our tests is failing because the datepicker changes its height when changing from a month with 5 weeks to a month with 6 weeks and makes it impossible to click a chekbox.

Furthermore we were using a date format of date/month/year, which is confusing for users living in countries where month/date/year is used.

The datepicker changes height when selecting August 2020 and the checkbox to enable a process cannot be selected anymore

## Objectives

* Use a date format familiar to users
* Fix an issue with the datepicker covering other fields
* Improve usability for mobile users when selecting dates
* Improve keyboard accessibility when selecting dates
* Simplify the code dealing with date formats

## Notes

Since date fields are not supported in Safari and Internet Explorer, we're still using the jQuery UI datepicker on those browsers (and on any other browser not supporting date fields).

### DIFF
--- a/app/assets/javascripts/advanced_search.js
+++ b/app/assets/javascripts/advanced_search.js
@@ -17,30 +17,7 @@
         $(".js-calendar").datepicker("option", "disabled", true);
       }
     },
-    init_calendar: function() {
-      var locale;
-      locale = $("#js-locale").data("current-locale");
-      $(".js-calendar").datepicker({
-        maxDate: "+0d"
-      });
-      $(".js-calendar-full").datepicker();
-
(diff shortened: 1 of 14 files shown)


Senen (approved):
👏

PR #4114, opened 2020-08-31
Get map center longitude from correct place

Senen:
## References

Related Pull Request: #3699

This error was introduced at commit b9ce68bc821bc2dba1cb8270d875f2e14a21d81b.

## Objectives
We were loading map center longitude from wrong map data attribute.

### DIFF
--- a/spec/shared/system/mappable.rb
+++ b/spec/shared/system/mappable.rb
@@ -283,12 +283,14 @@ shared_examples "mappable" do |mappable_factory_name, mappable_association_name,
       do_login_for(user) if management
     end
 
-    scenario "Should display map on #{mappable_factory_name} show page", :js do
+    scenario "Should display map and marker on #{mappable_factory_name} show page", :js do
       arguments[:id] = mappable.id
 
       visit send(mappable_show_path, arguments)
 
-      expect(page).to have_css(".map_location")
+      within ".map_location" do
+        expect(page).to have_css(".map-icon")
+      end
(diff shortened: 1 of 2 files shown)


spec/shared/system/mappable.rb
```diff
@@ -291,6 +291,16 @@
       expect(page).to have_css(".map_location")
     end
 
+    scenario "Should display marker on #{mappable_factory_name} show page", :js do
+      arguments[:id] = mappable.id
+
+      visit send(mappable_show_path, arguments)
+
```
javierm, line 298: Since the setup for this test is exactly the same as the previous test, and system tests are quite slow, what do you think about merging this test with the previous one? Something like:

```ruby
scenario "Displays map and marker on #{mappable_factory_name} show page", :js do
  arguments[:id] = mappable.id

  visit send(mappable_show_path, arguments)

  expect(page).to have_css(".map_location")

  within ".map_location" do
    expect(page).to have_css(".map-icon")
  end
end
```

javierm (commented):
Nice catch :smile:. I've left a comment about the test. Let me know what you think :wink:.

spec/shared/system/mappable.rb
Senen: Considering that `mappable` show pages only show the map when a marker is defined it has perfect sense to me. Thanks

javierm (approved):

PR #4117, opened 2020-08-31
Apply Legislation Process default colors to dev seeds

taitus:
## References
Related commit: https://github.com/consul/consul/pull/4080/commits/0b83be6

## Objectives
- Apply Legislation Process default colors to dev seeds.
- Simplify code.

## Notes
Legislation Processes created through the admin form were getting the default color.
However, Legislation processes created by other means (like the `db:dev_seed` rake task) were not getting these default values.

This feature was originally implemented when we were using Rails 4. With Rails 5, we can provide default values to all new Legislation processes and simplify the code at the same time thanks to its `attribute` method.

### DIFF
--- a/app/helpers/legislation_helper.rb
+++ b/app/helpers/legislation_helper.rb
@@ -22,22 +22,6 @@ module LegislationHelper
     @process.background_color.present? && @process.font_color.present?
   end
 
-  def default_bg_color
-    "#e7f2fc"
-  end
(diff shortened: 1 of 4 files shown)


javierm (commented):
Cool! There's a small typo in the test. Other than that, it's great :smile:.

javierm (approved):

Issue #4119, opened 2020-09-02
Confusing translation


Senen:
### User Story

A translator warned us through Crowdin application that we have a confusing source string ([link](https://crowdin.com/translate/consul/419/en-sr?filter=advanced&value=12&approvals=translated_not_approved&qa_issues=33&sort_ascending=1#58784)) and this is causing problems to translators to define a correct translation for this string into any other language.

### Description
Fix the [source string](https://github.com/consul/consul/blob/master/config/locales/en/general.yml#L575) so translators can understand the meaning.

The source string is:
```
<strong>I am participating in %{org}</strong> with my own citizen proposal and only if you also add you can I achieve the necessary support to make the city we all want.
```

### Expected Behavior & Definition of done
Have a correct source string easier to understand at original application language which is English.

PR #4122, opened 2020-09-04
Fix dashboard poster intro text

Senen:
## References
Related Issues: #4119 

## Objectives
Fix wrong translation.

Replace

```<strong>I am participating in %{org}</strong> with my own citizen proposal and only if you also add you can I achieve the necessary support to make the city we all want.```

with:

```"<strong>I am participating in %{org}</strong> with my own citizen proposal and only with your support can we make the city we all want"```

which has more sense.

### DIFF
--- a/config/locales/en/general.yml
+++ b/config/locales/en/general.yml
@@ -572,9 +572,7 @@ en:
         title: Poster preview
         poster_title: "Do not keep looking,"
         poster_subtitle: "back me up! ;)"
-        intro_text:
-          "<strong>I am participating in %{org}</strong> with my own citizen proposal and only if you also add you can I achieve the
-          necessary support to make the city we all want."
+        intro_text: "<strong>I am participating in %{org}</strong> with my own citizen proposal and only with your support can we make the city we all want"
         proposal_code: "Code of the proposal: %{code}"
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: config/locales/en/general.yml [hunk 1]
```diff
@@ -573,8 +573,7 @@ en:
         poster_title: "Do not keep looking,"
         poster_subtitle: "back me up! ;)"
         intro_text:
-          "<strong>I am participating in %{org}</strong> with my own citizen proposal and only if you also add you can I achieve the
-          necessary support to make the city we all want."
+          "<strong>I am participating in %{org}</strong> with my own citizen proposal and only if you add to it i will achieve the necessary support to make the city we all want."
```
javierm, line 576: "only if you add to it" sounds a bit strange to me :thinking:. How about:

```
"<strong>I am participating in %{org}</strong> with my own citizen proposal and only with your support can we make the city we all want"
```

Not sure about using the word "city", since CONSUL can be used in other contexts :thinking:, but I guess we can leave it the way it was :relieved:.

Senen: Much better @javierm! Thanks ☺️

javierm (approved):

PR #4143, opened 2020-09-25
Update Travis badge to use travis-ci.com 

javierm:
## Background

The domain travis-ci.org will stop working on December 31st, 2020, so we've now using travis-ci.com for continuous integration.

## Objectives

* Update the Travis badge so it points to the URL we now use to run our test suite
* Remove unused slack notifications in Travis configuration file


Senen (approved):
👍

PR #4192, opened 2020-10-02
Use Harmony mode in uglifier for ES2015+ support

javierm:
## References

* This bug was introduced in pull request #3288
* This problem is mentioned in rmosolgo/graphiql-rails#44 and rmosolgo/graphiql-rails#70
* How to enable this mode is described in the [Uglifier README](https://github.com/lautis/uglifier/tree/v4.2.0#es6--es2015--harmony-mode)

## Objectives

Make it possible to compile the assets included in `graphql-rails` with Uglifier on production environments.


Senen (approved):
I was able to reproduce the error trying to deploy `master` branch to a production environment then i deployed successfully the `harmony` branch! 👍

PR #4196, opened 2020-10-11
Hide related content proposals when these proposals are retired by their authors #3559

matisnape:
## References

#3559

## Objectives

Hide retired proposals from "related content" section for show view for: debate, budget investment, proposal

## Visual Changes

No changes

## Notes

Not applicable

### DIFF
--- a/app/controllers/budgets/investments_controller.rb
+++ b/app/controllers/budgets/investments_controller.rb
@@ -56,10 +56,7 @@ module Budgets
     def show
       @commentable = @investment
       @comment_tree = CommentTree.new(@commentable, params[:page], @current_order)
-      related_without_retired_proposals = @investment.relationed_contents
-                                                      .select { |rel| rel[:retired_at] == nil }
-      @related_contents = Kaminari.paginate_array(related_without_retired_proposals)
-                                  .page(params[:page]).per(5)
+      @related_contents = Kaminari.paginate_array(@investment.relationed_contents).page(params[:page]).per(5)
       set_comment_flags(@comment_tree.comments)
       load_investment_votes(@investment)
       @investment_ids = [@investment.id]
--- a/app/controllers/budgets/investments_controller.rb
+++ b/app/controllers/budgets/investments_controller.rb
@@ -56,7 +56,8 @@ module Budgets
     def show
       @commentable = @investment
       @comment_tree = CommentTree.new(@commentable, params[:page], @current_order)
-      @related_contents = Kaminari.paginate_array(@investment.relationed_contents).page(params[:page]).per(5)
+      related_contents_without_retired_proposals = @investment.relationed_contents.select { |rel| rel[:retired_at] == nil }
+      @related_contents = Kaminari.paginate_array(related_contents_without_retired_proposals).page(params[:page]).per(5)
       set_comment_flags(@comment_tree.comments)
       load_investment_votes(@investment)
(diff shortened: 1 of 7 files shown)


javierm:
@matisnape Thanks for the pull request. Before reviewing, could you have a look at the failing tests?

matisnape:
> @matisnape Thanks for the pull request. Before reviewing, could you have a look at the failing tests?

Will do, of course. 
However, when I run tests on my local master, I get 47 errors. Do you know what I might be missing in config? It's almost all Capybara errors or stuff like didn't find some text (from screenshots it looks like sleep is required). But there are some like this:
``` 
47) Budget Investments Search Advanced search Search by author type Public employee
      Failure/Error: investments = investments.filter(params[:advanced_search]) if params[:advanced_search].present?

      ArgumentError:
        wrong number of arguments (given 1, expected 0)
```

javierm:
@matisnape That's strange :thinking:. Which Ruby version are you using? What's the content of the `.ruby-version` file?

matisnape:
@javierm Yeah, I have 2.5.8 like in the readme. I think last year I had the same issue and just watched the tests touching the area I was working with. 
Anyways, I fixed the tests I definitely broke and the build is green :)

javierm:
That's strange, because the error you mentioned happens when using Ruby 2.6+ but doesn't happen with Ruby 2.5 or below. Anyway, I'll review the changes now that the tests pass.

app/controllers/budgets/investments_controller.rb
```diff
@@ -56,7 +56,10 @@ def new
     def show
       @commentable = @investment
       @comment_tree = CommentTree.new(@commentable, params[:page], @current_order)
-      @related_contents = Kaminari.paginate_array(@investment.relationed_contents).page(params[:page]).per(5)
+      related_without_retired_proposals = @investment.relationed_contents
+                                                      .select { |rel| rel[:retired_at] == nil }
```
javierm, line 60: In order to avoid duplication, do you think we could move this line to the `Relationable` concern? Maybe in the `relationed_contents` itself :thinking:; that way we won't have to change every place using it (like the `DashboardController`) and probably the change in the view won't be necessary either :thinking:.

javierm: One more thing: using `rel[:retired_at]` is certainly an interesting idea :ok_hand:. However, it isn't something we usually use. We usually write something like: `.reject { |related| related.respond_to?(:retired?) && related.retired? }` like we do in the `Notifiable` concern.

javierm (commented):
@matisnape Nice! I completely misunderstood what the issue was about even though I read it several times :sweat_smile:, but you got it right :+1:.

Regarding the tests you mentioned, I think a test in the `spec/models/related_content_spec.rb` file should do the trick once we've moved the code to the `Relationable` concern.

app/controllers/budgets/investments_controller.rb
javierm: Comments edited after remembering you had added a comment to the issue mentioning the same things I was mentioning.

matisnape: all fixed :)
(hopefully with Hound too :D)

javierm (approved):
Looks good! I'll do some final tests on Monday and merge if everything is OK.

Thanks a lot!

PR #4200, opened 2020-10-15
Use labels for web_section_ids checkboxes texts

Senen:
# Objectives
Use labels for banners `web_section_ids` checkboxes intead of plain text so when a user clicks on the checkbox label the checkbox is checked/unchecked. 

Use fieldset to group the checkboxes collection and replace label with a legend.

# Screenshots

**Before**

Captura de pantalla 2020-10-28 a las 11 25 12

**After**
Captura de pantalla 2020-10-28 a las 11 20 04

### DIFF
--- a/app/views/admin/banners/_form.html.erb
+++ b/app/views/admin/banners/_form.html.erb
@@ -37,7 +37,9 @@
     <div class="small-12 column">
       <%= f.label :sections, t("admin.banners.banner.sections_label") %>
       <%= f.collection_check_boxes(:web_section_ids, @banner_sections, :id, :name) do |b| %>
-        <%= b.check_box %> <%= t("admin.banners.banner.sections.#{b.text}") %><br>
+        <%= b.label do %>
+          <%= b.check_box + t("admin.banners.banner.sections.#{b.text}") %>
+        <% end %>
       <% end %>
     </div>
   </div>
--- a/app/views/admin/banners/_form.html.erb
+++ b/app/views/admin/banners/_form.html.erb
@@ -34,14 +34,14 @@
   </div>
 
   <div class="row">
-    <div class="small-12 column">
-      <%= f.label :sections, t("admin.banners.banner.sections_label") %>
+    <fieldset class="small-12 column">
+      <legend><%= t("admin.banners.banner.sections_label") %></legend>
       <%= f.collection_check_boxes(:web_section_ids, @banner_sections, :id, :name) do |b| %>
         <%= b.label do %>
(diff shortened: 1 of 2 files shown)


app/views/admin/banners/_form.html.erb
```diff
@@ -34,12 +34,14 @@
   </div>
 
   <div class="row">
-    <div class="small-12 column">
-      <%= f.label :sections, t("admin.banners.banner.sections_label") %>
+    <fieldset class="small-12 column">
+      <legend><%= t("admin.banners.banner.legend") %></legend>
```
javierm, line 38: Since we might eventually use other legends in the form, IMHO using `sections_legend` (or other word starting with `sections_`) makes it less likely to have conflicts in the future :thinking:.

Leaving it as `sections_label` is also fine with me because crowdin doesn't automatically rename keys for all languages and so translators will have to translate this text again if we rename the key.

Both solutions are fine with me, so choose the one you prefer :wink:.

spec/system/admin/banners_spec.rb
```diff
@@ -73,8 +73,6 @@
   end
 
   scenario "Publish a banner" do
-    section = create(:web_section, name: "proposals")
```
javierm, line 76: I know this line isn't necessary, but IMHO it makes the test easier to follow. Otherwise there might be a few seconds where the developer is wondering why the section already exists :thinking:.

If you think it's confusing to create a section that already exists, we can also create a section with a different name. What do you think?

app/views/admin/banners/_form.html.erb
Senen: >Since we might eventually use other legends in the form, IMHO using sections_legend (or other word starting with sections_) makes it less likely to have conflicts in the future 🤔.

Nice catch 👍🏼

>Leaving it as sections_label is also fine with me because crowdin doesn't automatically rename keys for all languages and so translators will have to translate this text again if we rename the key.

I think Crowdin's translation memory would take care of this automaticallly but always better to reduce the changes so finally i will keep the previous translation key to not to bother ourselves with more Crowdin's commits.

Thanks @javierm

spec/system/admin/banners_spec.rb
Senen: Good idea! 💡

Senen: Here i decide to replace the `create` with  `section = WebSection.find_by(name: "proposals")` so latest test expectations keep working without changing anyting else.

javierm: :ok_hand:

javierm (approved):
Cool! :smile:

PR #4206, opened 2020-10-21
Allow loading icons from SVG files

javierm:
## References

* [Death to Icon Fonts](https://speakerdeck.com/ninjanails/death-to-icon-fonts)

## Objectives

* Improve tne user experience for users who customize their fonts for accessibility reasons
* Avoid any issues due to devices/operating systems using custom emojis where we intend to display font awesome icons

## Notes

* Deciding the right solution isn't trivial; check the ultra-long commit messages to know the reasons behind this change and the options we've considered.
* This pull request doesn't change the way existing icons behave; while sooner or later we'll use SVG icons everywhere, for now we're only changing the icons to expand/collapse comments and the "comments" icon in the legislation process summary
* An obvious downside of this approach is we'll have to remember to update the `vendor/assets/images/fontawesome` folder every time we update the font-awesome-sass gem


Senen (approved):
If someday we decide to move from Asset Pipeline to Webpacker probably we will be able to use a Node.js package with SVG files already included so we will be able to remove all these SVG files from the `vendor` folder.

Tested successfully in the production environment.

PR #3797, opened 2019-10-24
Use a block to travel in time in specs

javierm:
## Background

Since 15 hours ago a test is failing on Travis, for unknown causes. It's making many tests fail because it travels in time but never travels back.

## Objectives

Make tests travelling in time go back even if they fail.

## Notes

The following test is failing on every Travis build in the last hours, no matter if the build is based on an old branch:

```
1) Proposal Notifications Use time traveling to make sure notifications can be sent after time interval
     Failure/Error: expect(page).to have_content "Your message has been sent correctly."
       expected to find text "Your message has been sent correctly." in "CONSUL Language: Deutsch English Español Français Nederlands Português brasileiro 中文 Menu Notifications You don't have new notifications My content My account Sign out Debates Proposals Voting Collaborative legislation Participatory budgeting Help Debates Proposals Voting Collaborative legislation Participatory budgeting Help Go back Send message This message will be sent to 0 people and it will be visible in the proposal's page. Messages are not sent immediately, users will receive periodically an email with all proposal notifications. × 1 error prevented this Notification from being saved. Please check the marked fields to know how to correct them: TitleYou have to wait a minimum of 3 days between notifications Message Please share it again with others so we can make it happen! Open government This portal uses the CONSUL application which is open-source software. For technical assistance visit Participation Decide how to shape the city you want to live in. CONSUL, 2019 | Privacy Policy | Terms and conditions of use | Accessibility"
     # ./spec/features/proposal_notifications_spec.rb:23:in `block (3 levels) in <top (required)>'
     # ./spec/features/proposal_notifications_spec.rb:17:in `block (2 levels) in <top (required)>'
```

PR #4208, opened 2020-10-22
Avoid seasonal clock changes issues in specs

javierm:
## References

* This issue happened last year as well and we mitigated it a bit in pull request #3797

## Background

Using `travel` we go to `Time.now + interval`. If the application's time zone changes due to seasonal clock changes during that interval, we might travel to a time which is not the time we intended to travel to.

Example:

On a system using the UTC time zone, it's 9AM on October 25 (Friday). Since by default CONSUL uses the Madrid time zone, it means the application's time is 11AM.

We use `travel` to advance three days. That means we go to 9AM UTC on October 28 (Monday). The application's time will be 10AM due to the seasonal clock change, so we haven't fully traveled three days in application's time.

To reproduce locally, run:

```bash
TZ=UTC rspec spec/system/proposal_notifications_spec.rb:410
```

## Objectives

Using `travel_to` with `3.days.from_now`, which uses the application's time zone and so it will travel to October 28 at 11AM, solves the problem.

### DIFF
--- a/spec/system/proposal_notifications_spec.rb
+++ b/spec/system/proposal_notifications_spec.rb
@@ -405,7 +405,7 @@ describe "Proposal Notifications" do
 
       expect(page).to have_content "Your message has been sent correctly."
 
-      travel(3.days + 1.second) do
+      travel_to(3.days.from_now + 1.second) do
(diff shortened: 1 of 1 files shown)


Senen (approved):
Nice catch! 👏

Issue #3551, opened 2019-05-28
Restore blocked users content


decabeza:
## Background
Users can be blocked through the Moderation interface `/moderation/users` hiding also all their content

## What we need
When a blocked user is restored by an administrator through `/admin/hidden_users` all his/her previously hidden content should be also unhidden.

carlosiniesta:
👋 

I make something here https://github.com/consul/consul/pull/3770

Hope, you like it

carlosiniesta:
After one year... I'm back and this time 
I added some specs, hopefully is useful.

PR #4213, opened 2020-10-25
Restore all related content along when a user is restored

carlosiniesta:
## References

https://github.com/consul/consul/issues/3551

## Objectives

Restore all related user content after restore user itself.

## Visual Changes

Nope

## Notes

As far as I understand is not needed to restore proposal notification, but nort 100% about this part.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -250,7 +250,9 @@ class User < ApplicationRecord
       Comment.restore_all comments.where("hidden_at >= ?", hidden_at)
       Proposal.restore_all proposals.where("hidden_at >= ?", hidden_at)
       Budget::Investment.restore_all budget_investments.where("hidden_at >= ?", hidden_at)
-      ProposalNotification.restore_all ProposalNotification.where("hidden_at >= ?", hidden_at).where(author_id: id).unscope(:where)
+      ProposalNotification.restore_all(
+        ProposalNotification.where("hidden_at >= ?", hidden_at).where(author_id: id).unscope(:where)
+      )
 
       restore
     end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -235,19 +235,20 @@ class User < ApplicationRecord
   end
 
   def block
-    debates_ids = Debate.where(author_id: id).pluck(:id)
-    comments_ids = Comment.where(user_id: id).pluck(:id)
-    proposal_ids = Proposal.where(author_id: id).pluck(:id)
-    investment_ids = Budget::Investment.where(author_id: id).pluck(:id)
-    proposal_notification_ids = ProposalNotification.where(author_id: id).pluck(:id)
-
     hide
 
-    Debate.hide_all debates_ids
-    Comment.hide_all comments_ids
+    Debate.hide_all debate_ids
+    Comment.hide_all comment_ids
     Proposal.hide_all proposal_ids
-    Budget::Investment.hide_all investment_ids
-    ProposalNotification.hide_all proposal_notification_ids
+    Budget::Investment.hide_all budget_investment_ids
+    ProposalNotification.hide_all ProposalNotification.where(author_id: id).pluck(:id)
+  end
+
+  def after_restore
+    Debate.restore_all debate_ids
+    Comment.restore_all comment_ids
+    Proposal.restore_all proposal_ids
+    Budget::Investment.restore_all budget_investment_ids
   end
 
   def erase(erase_reason = nil)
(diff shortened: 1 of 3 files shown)


spec/models/user_spec.rb
```diff
@@ -709,4 +709,21 @@
...
     end
   end
+
+  describe "#after_restore" do
+    it "restore all previous hidden user content" do
+      user                  = create(:user, :hidden)
+      debate                = create(:debate, :hidden, author: user)
+      comment               = create(:comment, :hidden, author: user)
+      proposal              = create(:proposal, :hidden, author: user)
+      budget_investment     = create(:budget_investment, :hidden, author: user)
```
javierm, line 719: Although we don't do it very often, aligning code is OK. If possible, align it in a way so there's only one space between the longest column on the left and the equals sign.

I guess if you add a `proposal_notification` variable all these lines will remain exactly the way they are now, though :smile:.

spec/models/user_spec.rb
```diff
@@ -709,4 +709,21 @@
...
+
+      user.restore
+
+      expect(debate.reload.hidden?).to be_falsey
```
javierm, line 723: For consistency with other specs, what do you think about this code?

```
  expect(debate.reload).not_to be_hidden
  # .. same with the rest
```

javierm (commented):
Thanks once again for the pull request! :smile:

app/models/user.rb
```diff
@@ -235,19 +235,20 @@ def display_official_position_badge?
...
 
-    Debate.hide_all debates_ids
-    Comment.hide_all comments_ids
+    Debate.hide_all debate_ids
+    Comment.hide_all comment_ids
     Proposal.hide_all proposal_ids
-    Budget::Investment.hide_all investment_ids
-    ProposalNotification.hide_all proposal_notification_ids
+    Budget::Investment.hide_all budget_investment_ids
+    ProposalNotification.hide_all ProposalNotification.where(author_id: id).pluck(:id)
```
javierm, line 244: @carlosiniesta Sorry, I was editing in two places in parallel and messed up with this comment :sweat_smile:.

IMHO, if we hide proposal notifications when hiding a user, we need to restore them when restoring the user :thinking:. This way accidentally hiding and immediately restoring a user has no side effects.

app/models/user.rb
```diff
@@ -235,19 +235,20 @@ def display_official_position_badge?
...
+  end
+
+  def after_restore
+    Debate.restore_all debate_ids
```
javierm, line 248: @carlosiniesta Sorry, I was editing the messages in two places at the same time and messed up with this comment pray.

IMHO we shouldn't restore records which were already hidden before hiding the user :thinking:. This is tricky, though :thinking:.

carlosiniesta: That make a lot of sense. I did it, now only related content hidden at the same time will be restored.

app/models/user.rb
```diff
@@ -235,19 +235,27 @@ def display_official_position_badge?
...
+  end
+
+  def full_restore
+    ActiveRecord::Base.transaction do
+      Debate.restore_all debates.where("hidden_at >= ?", hidden_at)
+      Comment.restore_all comments.where("hidden_at >= ?", hidden_at)
+      Proposal.restore_all proposals.where("hidden_at >= ?", hidden_at)
+      Budget::Investment.restore_all budget_investments.where("hidden_at >= ?", hidden_at)
+      ProposalNotification.restore_all(
+        ProposalNotification.where("hidden_at >= ?", hidden_at).where(author_id: id).unscope(:where)
```
javierm, line 254: I think we need to "unscope" before the other filters :thinking:. Otherwise we would restore all proposal notifications, even if they've got a different author.

How about: `ProposalNotification.only_hidden.where("hidden_at >= ?", hidden_at).where(author_id: id)`

app/models/user.rb
```diff
@@ -235,19 +235,27 @@ def display_official_position_badge?
...
+    Comment.hide_all comment_ids
     Proposal.hide_all proposal_ids
-    Budget::Investment.hide_all investment_ids
-    ProposalNotification.hide_all proposal_notification_ids
+    Budget::Investment.hide_all budget_investment_ids
+    ProposalNotification.hide_all ProposalNotification.where(author_id: id).pluck(:id)
+  end
+
+  def full_restore
+    ActiveRecord::Base.transaction do
```
javierm, line 248: IMHO a transaction here is a very good option. Nice! :+1:

javierm (commented):
@carlosiniesta Very nice :+1:. It looks like there's a small bug related to proposal notifications. Once it's fixed, I think we're good to go.

javierm:
@carlosiniesta Are you still working on this pull request? :thinking: As mentioned, I think once we replace the `unscope`, we're good to merge.

javierm:
Continuing in #4465 due to lack of response.

PR #4226, opened 2020-10-29
Allow users to delete their own comments

decabeza:
## Objectives

Sometimes a user makes a typo when writing a comment and cannot change it. Allow to edit is tricky because it could affect the context if the comment has some answers.

Other times users want to be able to delete the content they have created.

This PR allows users to delete their comments. A confirmation alert is displayed to avoid deleting content by mistake.

## Visual Changes
delete_comment_1

delete_comment_2

### DIFF
--- a/app/assets/stylesheets/layout.scss
+++ b/app/assets/stylesheets/layout.scss
@@ -1940,6 +1940,20 @@ table {
     }
   }
 
+  .delete-comment {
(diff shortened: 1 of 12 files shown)


taitus (approved):

javierm (approved):
Awesome! Thanks!

PR #4228, opened 2020-10-30
Send informative email for already confirmed users

decabeza:
## Objectives

When a user requests the confirmation email, a message appears indicating that they will receive an email. The same message appears if their account is already confirmed, but CONSUL doesn't send any email.

This PR maintains the behavior to don't resend confirmation instructions email for already confirmed users.
But, we will send an email to the user informing them that their account is now registered.

This way no one can know if someone else's account is confirmed and we don't have to worry about GDPR either.

## Notes
One of the problems with adding this text could be that it provides information that the account already exists to a potential "hacker". But the application already gives this information when registering a new user if you enter an existing email "has already been taken".

Perhaps the more correct text would be:
"If your email address exists in our database **and you have not received a confirmation email in the past,** you will receive an email with instructions for how to confirm your email address in a few minutes." The problem with this text is that it is too long and not concise for a notification. 

Conclusion: between the fact that the information disclosed with the proposed PR message ("You have already confirmed your email account.") is already available from the registration page, and that the more correct text in my opinion is too long and not concise, I think we can include this change.

Edit: 
As the two previously proposed solutions had some drawbacks, a new commit has been added which directly sends an email to the user informing him that his account is now validated and a link to reset his password. This way we avoid a notice with too much information and unclear or a notice that could somehow breach the GDPR.

### DIFF
--- a/config/locales/en/mailers.yml
+++ b/config/locales/en/mailers.yml
@@ -2,6 +2,10 @@ en:
   mailers:
     title: "Open Government"
     no_reply: "This message was sent from an email address that does not accept replies."
+    already_confirmed:
+      info: "We've received a request to send you instructions to confirm your account. However, your account is already confirmed, so there's no need to do so again."
+      new_password: "If you've forgotten your password, you can reset it at the following link:"
+      subject: Your account is already confirmed
     comment:
       hi: Hi
       new_comment_by: There is a new comment from <strong>%{commenter}</strong>
--- a/app/controllers/users/confirmations_controller.rb
+++ b/app/controllers/users/confirmations_controller.rb
@@ -1,4 +1,17 @@
 class Users::ConfirmationsController < Devise::ConfirmationsController
+  # POST /resource/confirmation
+  def create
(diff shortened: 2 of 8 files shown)


javierm:
This one's tricky :thinking:, since with these changes anyone could check whether a certain email is confirmed in the application. We need to check whether we're following data protection guidelines here.

Senen (approved):

config/locales/en/mailers.yml
```diff
@@ -2,6 +2,9 @@ en:
   mailers:
     title: "Open Government"
     no_reply: "This message was sent from an email address that does not accept replies."
+    already_confirmed:
+      new_password: "You can reset your password at the following link:"
```
javierm, line 6: I think this sentence isn't very clear in this context :thinking:. It might imply that resetting the password is necessary after confirming an email. Not to mention I might receive this message after someone else has asked to confirm my email :sweat_smile:.

How about something like (just an example):

> We've received a request to send you instructions to confirm your account. However, your account is already confirmed, so there's no need to do so again.
>
> If you've forgotten your password, you can reset it at (...)

PR #4237, opened 2020-11-07
Remove blank space after footer

javierm:
## References

* This issue was introduced in commit a8537f7e1

## Background

We added a `height: 100%` rule on links inside cards, which is great for cards in the "Featured" section of the homepage. However, the card in the "Open processes" section of the homepage has as many links as open processes inside, causing its height to be 300% on some browsers (Firefox at the very least) if there are three open processes and so expanding below the footer.

## Objectives

Remove blank space after footer when there are open processes.

### Before these changes

There's a blank space below the footer

### After these changes
There is no blank space below the footer

### DIFF
--- a/app/assets/stylesheets/layout.scss
+++ b/app/assets/stylesheets/layout.scss
@@ -2820,6 +2820,10 @@ table {
   .cards-container {
     display: flex;
     flex-wrap: wrap;
+
+    .card a {
+      height: 100%;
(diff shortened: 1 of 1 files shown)


Senen (approved):
This error is not happening on most of latest version browsers, but it happens with older versions like Firefox 70 among others.

PR #4260, opened 2020-11-17
Add SDG targets

Senen:
## References

* [WAI table with multi-level headers example](https://www.w3.org/WAI/tutorials/tables/multi-level/#table-with-three-headers-related-to-each-data-cell)

## Objectives

Pending

## Visual Changes

Pending

## Notes

Pending


app/components/admin/sdg/targets/index_component.html.erb
```diff
@@ -0,0 +1,38 @@
...
+
+<h2><%= title %></h2>
+
+<%= link_to t("admin.sdg.goals.index.title"), admin_sdg_goals_path %>
+<%= link_to t("admin.sdg.targets.index.title"), admin_sdg_targets_path %>
+
+<table>
+  <thead>
+    <tr>
+      <th></th>
```
javierm, line 13: I think the [WAI example](https://www.w3.org/WAI/tutorials/tables/examples/threeheaders/) used an empty `<td>` here, right?

app/components/admin/sdg/targets/index_component.html.erb
```diff
@@ -0,0 +1,38 @@
...
+      <th id="target-title" scope="col"><%= attribute_name(:title) %></th>
+    </tr>
+  </thead>
+
+  <tbody>
+    <% targets.group_by(&:goal).map do |goal, targets| %>
+      <tr class="goal-header">
+        <th id="<%= header_id(goal) %>" colspan="2" scope="colgroup">
+          <%= goal.code %>. <%= goal.title %>
```
javierm, line 22: `erb-lint` is giving a warning because the `.` is not being translated; not sure about the best approach here :thinking:.

app/controllers/admin/sdg/targets_controller.rb
```diff
@@ -0,0 +1,7 @@
+class Admin::SDG::TargetsController < Admin::BaseController
+  Target = ::SDG::Target
+
+  def index
+    @targets = Target.all
```
javierm, line 5: I forgot to add an `order` clause in the `Goal` model. We should add it here as well.

I've tried:

```ruby
Target.joins(:goal).merge(Goal.order(:code)).order("code")
```

But "1.2" is shown before "1.10" :relieved:.

app/models/sdg/target.rb
```diff
@@ -0,0 +1,16 @@
+class SDG::Target < ApplicationRecord
+  validates :code, presence: true, uniqueness: { scope: :goal_id }
```
javierm, line 2: I don't have a strong preference about scoping the unique index. However, I think we should do the same thing in both the model and the database; right now the model is scoped but the database index is not :thinking:.

javierm (commented):
I've left a few comments. Let me know what you think :wink:.

app/components/admin/sdg/targets/index_component.html.erb
Senen: Maybe we can add a method to the Goal model. Probably this will be handy to call it from some different component views. What do you think?

```
def full_title
  "#{code}. #{title}"
end
```

app/models/sdg/target.rb
Senen: As the first part of the code of any target is the code of the parent goal, maybe we can rely on the current unique index on code and do not use the scope. 🤔

javierm: Sounds good to me :+1:.

app/components/admin/sdg/targets/index_component.html.erb
javierm: We'd still have a potential translation issue where some languages use the `.` and some do not... but for now I guess it's OK :smile:. Either leaving it in the view or moving it to the model is fine with me :ok_hand:.

Senen: I kept this version finally!

app/controllers/admin/sdg/targets_controller.rb
Senen: It seems that the sorting of alphanumeric strings is a difficult thing to achieve through both Ruby and SQL. I finally used Ruby to convert targets codes into numbers after a little bit of magic.

app/models/sdg/target.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  belongs_to :goal
+
+  def title
+    I18n.t("sdg.goals.goal_#{goal.code}.targets.target_#{code_key}.title")
+  end
+
+  def numeric_code
+    code.downcase.gsub(".", "").gsub(/\D+/) do |letter|
+      ("a".."z").to_a.index(letter) + 100
```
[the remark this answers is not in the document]
javierm: How about using Ruby's `Comparable` module? Here's a possible implementation:

```diff
diff --git a/app/controllers/admin/sdg/targets_controller.rb b/app/controllers/admin/sdg/targets_controller.rb
index 92f0399411..9228d0f620 100644
--- a/app/controllers/admin/sdg/targets_controller.rb
+++ b/app/controllers/admin/sdg/targets_controller.rb
@@ -2,6 +2,6 @@ class Admin::SDG::TargetsController < Admin::BaseController
   Target = ::SDG::Target
 
   def index
-    @targets = Target.all.sort_by(&:numeric_code)
+    @targets = Target.all.sort
   end
 end
diff --git a/app/models/sdg/target.rb b/app/models/sdg/target.rb
index 7463d9cc38..44fefc2db2 100644
--- a/app/models/sdg/target.rb
+++ b/app/models/sdg/target.rb
@@ -1,4 +1,6 @@
 class SDG::Target < ApplicationRecord
+  include Comparable
+
   validates :code, presence: true, uniqueness: true
   validates :goal, presence: true
 
@@ -8,15 +10,27 @@ class SDG::Target < ApplicationRecord
     I18n.t("sdg.goals.goal_#{goal.code}.targets.target_#{code_key}.title")
   end
 
-  def numeric_code
-    code.downcase.gsub(".", "").gsub(/\D+/) do |letter|
-      ("a".."z").to_a.index(letter) + 100
-    end.to_i
+  def <=>(target)
+    [goal.code, numeric_subcode] <=> [target.goal.code, target.numeric_subcode]
   end
 
+  protected
+
+    def numeric_subcode
+      if subcode.to_i > 0
+        subcode.to_i
+      else
+        subcode.to_i(36) * 1000
+      end
+    end
+
   private
 
     def code_key
       code.gsub(".", "_")
     end
+
+    def subcode
+      code.split(".").last
+    end
 end
diff --git a/spec/models/sdg/target_spec.rb b/spec/models/sdg/target_spec.rb
index 59c783ab9b..7aeb3ee142 100644
--- a/spec/models/sdg/target_spec.rb
+++ b/spec/models/sdg/target_spec.rb
@@ -40,33 +40,24 @@ describe SDG::Target do
     end
   end
 
-  describe "#numeric_code" do
-    let(:goal) { SDG::Goal.find_by!(code: "1") }
+  describe "#<=>" do
+    let(:goal)   { build(:sdg_goal, code: 10) }
+    let(:target) { build(:sdg_target, code: "10.19", goal: goal) }
 
-    it "converts code from string to integer" do
-      target = build(:sdg_target, code: "1.1", goal: goal)
+    it "compares using the goal first" do
+      lesser_target = build(:sdg_target, code: "2.14", goal: build(:sdg_goal, code: 2))
+      greater_target = build(:sdg_target, code: "11.1", goal: build(:sdg_goal, code: 11))
 
-      expect(target.numeric_code).to be_a(Integer)
+      expect(target).to be > lesser_target
+      expect(target).to be < greater_target
     end
 
-    it "converts code with downcase letters from string to integer" do
-      target = build(:sdg_target, code: "1.a", goal: goal)
+    it "compares using the target code when the goal is the same" do
+      lesser_target = build(:sdg_target, code: "10.2", goal: goal)
+      greater_target = build(:sdg_target, code: "10.A", goal: goal)
 
-      expect(target.numeric_code).to eq(1100)
-
-      target = build(:sdg_target, code: "1.z", goal: goal)
-
-      expect(target.numeric_code).to eq(1125)
-    end
-
-    it "converts code with upcase letters from string to integer" do
-      target = build(:sdg_target, code: "1.A", goal: goal)
-
-      expect(target.numeric_code).to eq(1100)
-
-      target = build(:sdg_target, code: "1.Z", goal: goal)
-
-      expect(target.numeric_code).to eq(1125)
+      expect(target).to be > lesser_target
+      expect(target).to be < greater_target
     end
   end
 end

```

spec/models/sdg/target_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+    expect(build(:sdg_target, code: nil)).not_to be_valid
+  end
+
+  it "is not valid without a goal" do
+    target = build(:sdg_target, goal: nil)
+
+    expect(target).not_to be_valid
+  end
+
+  it "is not valid when code is duplicated for the same parent goal" do
```
javierm, line 18: I think we have to update this test: it is not valid even if the parent goal is not the same.

javierm (commented):
Nice improvements. I've left a couple of comments, mainly about the way to sort targets. Let me know what you think, since it's pretty tricky :thinking:.

app/models/sdg/target.rb
```diff
@@ -0,0 +1,36 @@
...
+  validates :code, presence: true, uniqueness: true
+  validates :goal, presence: true
+
+  belongs_to :goal
+
+  def title
+    I18n.t("sdg.goals.goal_#{goal.code}.targets.target_#{code_key}.title")
+  end
+
+  def <=>(target)
```
javierm, line 13: I forgot to add a line here:

```
    return unless target.class == self.class
```

And then we should add a test where we compare it with another object.

This bug was found when trying to use `match_array` in a spec when the array contained both targets and goals.

app/models/sdg/target.rb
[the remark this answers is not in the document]
Senen: Oh, nice! Patch applied.

app/controllers/admin/sdg/targets_controller.rb
Senen: Finally solved adding Comparable module to Target model.

javierm (approved):

app/components/sdg_management/subnavigation_component.rb
```diff
@@ -0,0 +1,26 @@
...
+  private
+
+    def sections
+      %i[goals targets]
+    end
+
+    def link_to_section(section)
+      link_to "SDG::#{section.to_s.classify}".constantize.model_name.human(count: 2).titleize,
+              path_for(section),
+              "aria-selected": section == current
```
javierm, line 17: It looks like this attribute is causing invalid HTML:

`Element a is missing one or more of the following attributes: aria-checked, role.`

Senen: Finally, `aria-selected` attribute was replaced with the HTML class `is-active` which provides the same styles, and resultant markup is also valid.

javierm (approved):

PR #4263, opened 2020-11-20
Simplify tests requiring admin login

javierm:
## Objectives

* Simplify tests requiring admin login
* Reduce duplication

### DIFF
--- a/spec/system/admin/settings_spec.rb
+++ b/spec/system/admin/settings_spec.rb
@@ -1,10 +1,6 @@
 require "rails_helper"
 
-describe "Admin settings" do
-  before do
-    login_as(create(:administrator).user)
-  end
-
+describe "Admin settings", :admin do
   scenario "Index" do
     create(:setting, key: "super.users.first")
     create(:setting, key: "super.users.second")
@@ -36,16 +32,12 @@ describe "Admin settings" do
     end
 
     scenario "When `Map settings` tab content is hidden map should not be initialized" do
-      admin = create(:administrator).user
-      login_as(admin)
       visit admin_settings_path
 
       expect(page).not_to have_css("#admin-map.leaflet-container", visible: :all)
     end
 
     scenario "When `Map settings` tab content is shown map should be initialized" do
-      admin = create(:administrator).user
-      login_as(admin)
       visit admin_settings_path
 
       find("#map-tab").click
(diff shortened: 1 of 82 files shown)


spec/system/admin/settings_spec.rb
```diff
@@ -1,10 +1,6 @@
 require "rails_helper"
 
-describe "Admin settings" do
-  before do
-    login_as(create(:administrator).user)
-  end
-
+describe "Admin settings", :admin do
```
Senen, line 3: As `:admin` is defined at the main `describe`block we can remove other `:admin` ocurrences from the inner `describe` blocks. Here is a patch to remove all of them the except the main one:

```
diff --git a/spec/system/admin/settings_spec.rb b/spec/system/admin/settings_spec.rb
index 32770534aa..50af581105 100644
--- a/spec/system/admin/settings_spec.rb
+++ b/spec/system/admin/settings_spec.rb
@@ -26,7 +26,7 @@ describe "Admin settings", :admin do
     expect(page).to have_content "Value updated"
   end
 
-  describe "Map settings initialization", :js, :admin do
+  describe "Map settings initialization", :js do
     before do
       Setting["feature.map"] = true
     end
@@ -46,7 +46,7 @@ describe "Admin settings", :admin do
     end
   end
 
-  describe "Update map", :admin do
+  describe "Update map" do
     scenario "Should not be able when map feature deactivated" do
       Setting["feature.map"] = false
 
@@ -107,7 +107,7 @@ describe "Admin settings", :admin do
     end
   end
 
-  describe "Update content types", :admin do
+  describe "Update content types" do
     scenario "stores the correct mime types" do
       setting = Setting.create!(key: "upload.images.content_types", value: "image/png")
       visit admin_settings_path
@@ -137,7 +137,7 @@ describe "Admin settings", :admin do
     end
   end
 
-  describe "Update Remote Census Configuration", :admin do
+  describe "Update Remote Census Configuration" do
     before do
       Setting["feature.remote_census"] = true
     end
@@ -165,7 +165,7 @@ describe "Admin settings", :admin do
     end
   end
 
-  describe "Should redirect to same tab after update setting", :admin do
+  describe "Should redirect to same tab after update setting" do
     context "remote census" do
       before do
         Setting["feature.remote_census"] = true
```

Senen (changes requested):
Hi @javierm, great simplification! 👏🏼

I only found some unneeded occurrences of the `:admin` tag at `spec/system/admin/settings_spec.rb` file.

spec/system/admin/settings_spec.rb
javierm: Fixed, thanks! :smile:

Senen (approved):

PR #4284, opened 2020-12-13
Fix flaky annotation specs

javierm:
## Background
We were selecting a word using a double click. However, the test failed sometimes when styles such as font family, font size or the element's width changed, since the double click takes place in the middle of the element and it's possible that the middle of the element contains a space between two words, meaning no word would be selected.

## Objectives

* Select text in specs a way that does not fail when certain styles are used

## Notes

Without these changes, the failure in the "Create" test can be reproduced by changing the `$base-font-size` variable to `2rem` or changing the default font family to Roboto.


javierm:
Closing since it doesn't fix the issue and we haven't found a solution. Opening issue #4713 instead.

PR #4294, opened 2020-12-22
Extend SDG managment section to allow content reviews

Senen:
## References

Extends #4269.

## Objectives
Extend the SDG content management section to allow users to review user-provided content. Polls and Legislation Processes are marked as reviewed during creation as administrators users always create them. 

Only an administrator or an SDG manager user can mark the SDG content as reviewed.

Split index into three sections:

1. Pending: List only records pending to be reviewed.
2. All: List all records no matter the review status
3. Marked as reviewed: List only the reviewed records

The search form works within the three sections.

When a user with permission updates the relatable, it is marked as reviewed and a notice is shown to the user "Debate updated successfully and marked as reviewed". The users are always redirected to the "Pending" tab to continue reviewing of others relatables.

## Visual Changes

Captura de pantalla 2021-01-19 a las 15 01 08
Captura de pantalla 2021-01-19 a las 15 01 21
Captura de pantalla 2021-01-19 a las 15 01 40

### DIFF
--- a/app/controllers/sdg_management/relations_controller.rb
+++ b/app/controllers/sdg_management/relations_controller.rb
@@ -2,8 +2,12 @@ class SDGManagement::RelationsController < SDGManagement::BaseController
   before_action :check_feature_flags
   before_action :load_record, only: [:edit, :update]
 
+  FILTERS = %w[pending_sdg_review all sdg_reviewed].freeze
+  has_filters FILTERS, only: :index
+
   def index
     @records = relatable_class
+               .send(@current_filter)
                .accessible_by(current_ability)
                .by_goal(params[:goal_code])
                .by_target(params[:target_code])
--- a/app/controllers/sdg_management/relations_controller.rb
+++ b/app/controllers/sdg_management/relations_controller.rb
@@ -23,7 +23,7 @@ class SDGManagement::RelationsController < SDGManagement::BaseController
   def update
     @record.sdg_target_list = params[@record.class.table_name.singularize][:sdg_target_list]
 
-    redirect_to action: :index
+    redirect_to({ action: :index }, notice: update_notice)
   end
 
   private
@@ -42,5 +42,14 @@ class SDGManagement::RelationsController < SDGManagement::BaseController
 
       check_feature_flag(process_name)
       raise FeatureDisabled, process_name unless Setting["sdg.process.#{process_name}"]
+    end
+
+    def update_notice
+      if @record.sdg_review.present?
+        t("sdg_management.relations.update.notice", relatable: relatable_class.model_name.human)
+      else
+        @record.create_sdg_review!
+        t("sdg_management.relations.update_and_review.notice", relatable: relatable_class.model_name.human)
+      end
     end
 end
--- a/app/models/concerns/sdg/relatable.rb
+++ b/app/models/concerns/sdg/relatable.rb
@@ -10,6 +10,8 @@ module SDG::Relatable
                source: :related_sdg,
                source_type: sdg_type
     end
+
+    has_one :sdg_review, as: :relatable, dependent: :destroy, class_name: "SDG::Review"
   end
 
   class_methods do
--- a/app/models/concerns/sdg/relatable.rb
+++ b/app/models/concerns/sdg/relatable.rb
@@ -28,6 +28,14 @@ module SDG::Relatable
 
       joins(sdg_class.table_name.to_sym).merge(sdg_class.where(code: code))
     end
+
+    def sdg_reviewed
+      joins(:sdg_review)
+    end
+
+    def pending_sdg_review
+      left_joins(:sdg_review).merge(SDG::Review.where(id: nil))
+    end
   end
 
   def related_sdgs
--- a/config/locales/es/sdg_management.yml
+++ b/config/locales/es/sdg_management.yml
@@ -32,3 +32,10 @@ es:
         title: "Nueva meta localizada"
       update:
         notice: "Meta localizada asctualizada correctamente"
+    relations:
+      index:
+        filter: "Filtro"
+        filters:
+          all: "Todos"
+          pending_sdg_review: "Pendientes"
+          sdg_reviewed: "Marcados como revisados"
--- a/config/locales/es/sdg_management.yml
+++ b/config/locales/es/sdg_management.yml
@@ -39,3 +39,7 @@ es:
           all: "Todos"
           pending_sdg_review: "Pendientes"
           sdg_reviewed: "Marcados como revisados"
+      update:
+        notice: "%{relatable} actualizado correctamente"
+      update_and_review:
+        notice: "%{relatable} actualizado correctamente y marcado como revisado"
--- a/spec/factories/sdg.rb
+++ b/spec/factories/sdg.rb
@@ -18,4 +18,13 @@ FactoryBot.define do
   factory :sdg_phase, class: "SDG::Phase" do
     kind { :sensitization }
   end
+
+  factory :sdg_review, class: "SDG::Review" do
+    SDG::Related::RELATABLE_TYPES.map { |relatable_type| relatable_type.downcase.gsub("::", "_") }
+    .each do |relatable|
+      trait :"#{relatable}_review" do
+        association :relatable, factory: relatable
+      end
+    end
+  end
 end
--- a/spec/models/sdg/relatable_spec.rb
+++ b/spec/models/sdg/relatable_spec.rb
@@ -117,6 +117,18 @@ describe SDG::Relatable do
     end
   end
 
+  describe "#sdg_review" do
+    it "returns nil when relatable is not reviewed" do
+      expect(relatable.sdg_review).to be_blank
+    end
+
+    it "returns the review when relatable is reviewed" do
+      review = create(:sdg_review, relatable: relatable)
+
+      expect(relatable.sdg_review).to eq(review)
+    end
+  end
+
   describe ".by_goal" do
     it "returns everything if no code is provided" do
       expect(relatable.class.by_goal("")).to eq [relatable]
--- a/spec/models/sdg/relatable_spec.rb
+++ b/spec/models/sdg/relatable_spec.rb
@@ -170,4 +170,23 @@ describe SDG::Relatable do
       expect(relatable.class.by_target(target.code)).to be_empty
     end
   end
+
+  describe ".pending_sdg_review" do
+    let!(:relatable) { create(:proposal) }
+
+    it "returns records not reviewed yet" do
+      create(:sdg_review, relatable: create(:proposal))
+
+      expect(relatable.class.pending_sdg_review).to match_array [relatable]
+    end
+  end
+
+  describe ".sdg_reviewed" do
+    let!(:relatable) { create(:proposal) }
+    let!(:reviewed_relatable) { create(:sdg_review, relatable: create(:proposal)).relatable }
+
+    it "returns records already reviewed" do
+      expect(relatable.class.sdg_reviewed).to match_array [reviewed_relatable]
+    end
+  end
 end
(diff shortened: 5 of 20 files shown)


spec/factories/sdg.rb
```diff
@@ -14,4 +14,12 @@
 
     target { SDG::Target[code.rpartition(".").first] }
   end
+
+  factory :sdg_revision, class: "SDG::Revision" do
+    %i[budget_investment debate legislation_process poll proposal].each do |model|
+      trait :"#{model}_as_relatable" do
```
javierm, line 20: We use `#{model}_comment` for comments and `for_#{model}` for notifications. Maybe we can follow one of those patterns here as well :thinking:.

spec/factories/sdg.rb
```diff
@@ -14,4 +14,12 @@
 
     target { SDG::Target[code.rpartition(".").first] }
   end
+
+  factory :sdg_revision, class: "SDG::Revision" do
+    %i[budget_investment debate legislation_process poll proposal].each do |model|
```
javierm, line 19: Would it make sense to use `SDG::Related::RELATABLE_TYPES` here?

app/controllers/sdg_management/relations_controller.rb
```diff
@@ -7,6 +7,7 @@ def index
                .accessible_by(current_ability)
                .by_goal(params[:goal_code])
                .by_target(params[:target_code])
+               .by_revision_status(params[:revision_status])
```
javierm, line 10: We usually use `has_filters` in controllers when dealing with this interface :thinking:. Are there complications preventing us from doing so here? Then we could add a couple of scopes to the `SDG::Relatable` concern.

app/controllers/sdg_management/relations_controller.rb
```diff
@@ -19,7 +20,11 @@ def edit
   def update
     @record.sdg_target_list = params[@record.class.table_name.singularize][:sdg_target_list]
 
-    redirect_to action: :index
+    if @record.sdg_revision.blank? && @record.create_sdg_revision
+      flash[:notice] = t("sdg_management.relations.update.notice", relatable: relatable_class.model_name.human)
+    end
```
javierm, line 25: I actually forgot to add the `flash[:notice]` in this action :sweat_smile:. So how about:

```ruby
@record.create_sdg_revision! unless @record.sdg_revision
flash[:notice] = t("sdg_management.relations.update.notice", relatable: relatable_class.model_name.human)
```

And I guess we should update the flash message to be more generic :thinking:.

javierm (commented):
Here's the initial review.

Since changing the controller to use `has_filters` and the views to use the `filter_subnav` partial (as done in may places) would require a few changes, I'll continue after we've decided what to do here.

app/models/concerns/sdg/relatable.rb
```diff
@@ -26,6 +28,18 @@ def by_sdg_related(sdg_class, code)
 
       joins(sdg_class.table_name.to_sym).merge(sdg_class.where(code: code))
     end
+
+    def by_revision_status(revision_status)
+      case revision_status
+      when "pending", nil
+        revised_ids = SDG::Revision.where(relatable_type: self.name).pluck(:relatable_id)
+        where.not(id: revised_ids)
```
javierm, line 36:
```suggestion
        left_joins(:sdg_revision).merge(SDG::Revision.where(id: nil))
```

Or something similar?

db/migrate/20201221100246_create_sdg_revisions.rb
```diff
@@ -0,0 +1,10 @@
+class CreateSDGRevisions < ActiveRecord::Migration[5.2]
+  def change
+    create_table :sdg_revisions do |t|
+      t.references :relatable, polymorphic: true, index: false
+      t.timestamps
+
+      t.index [:relatable_type, :relatable_id], name: "sdg_revisions_unique", unique: true
```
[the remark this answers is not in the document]
Senen: Much better! 👌🏼

app/controllers/sdg_management/relations_controller.rb
Senen: I did not realize we have the `HasFilter` option, never used it before, I think we could use here too. Thanks

app/controllers/sdg_management/relations_controller.rb
Senen: As I am redirecting the user always to the "Pending" tab after the update action I intended to show a more precise notice when the SDG manager is marking the relatable as revised (this happens when the user updates for the first time the relatable from SDG management section) so the user knows that the relatable was marked as revised and moved into "Marked as revised" tab. What do you think about having two notices?

1º Notice for the first relatable update: Proposal was marked as revised and its SDG relations were updated successfully!
2º Notice for subsequent updates: SDGrelations were updated successfully!

The action code could look like:
```
  def update
    @record.sdg_target_list = params[@record.class.table_name.singularize][:sdg_target_list]

    if @record.sdg_revision.blank?
      @record.create_sdg_revision!
      flash[:notice] = t("sdg_management.relations.update_and_revise.notice",
        relatable: relatable_class.model_name.human)
    else
      flash[:notice] = t("sdg_management.relations.update.notice",
        relatable: relatable_class.model_name.human)
    end
    redirect_to action: :index
  end
```

app/models/concerns/sdg/relatable.rb
```diff
@@ -26,6 +28,14 @@ def by_sdg_related(sdg_class, code)
 
       joins(sdg_class.table_name.to_sym).merge(sdg_class.where(code: code))
     end
+
+    def revised
+      joins(:sdg_revision)
+    end
+
+    def pending
```
javierm, line 36: This method might be called as `Proposal.pending`, which might be a bit confusing since we might not know what `pending` means, depending on the context. How about `pending_sdg_review`, so it's similar to other scopes like `pending_review` and `pending_flag_review`?

Senen: I completely agree I did not realize about this!

app/components/sdg_management/relations/subnavigation_component.html.erb
```diff
@@ -0,0 +1,13 @@
+<ul class="tabs">
```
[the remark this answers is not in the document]
Senen: Finally used inheritance. How it looks like to you?

app/models/concerns/sdg/relatable.rb
Senen: Done!

app/controllers/sdg_management/relations_controller.rb
Senen: Ping @javierm, just in case you missed it.

spec/models/sdg/relatable_spec.rb
```diff
@@ -158,4 +170,23 @@
       expect(relatable.class.by_target(target.code)).to be_empty
     end
   end
+
+  describe ".pending_sdg_review" do
+    let!(:relatable) { create(:proposal) }
+
+    it "returns records not reviewed yet" do
+      create(:sdg_review, relatable: create(:proposal)).relatable
```
javierm, line 178: I guess the final `.relatable` is a typo and can be deleted :thinking:.

spec/system/sdg_management/relations_spec.rb
```diff
@@ -88,58 +93,215 @@
...
-      select "4.1", from: "target_code"
-      click_button "Search"
+      expect(page).not_to have_text "I'm a debate"
+      expect(page).to have_text "I'm a reviewed debate"
+    end
 
-      expect(page).to have_content "School"
-      expect(page).not_to have_content "Preschool"
+    describe "search" do
+      describe "on 'pending' tab" do
```
javierm, line 134: Since system specs are very slow, how about adding one case checking the search respects the current filter instead of adding the same cases for each filter?

javierm (commented):
@Senen For some reasons I don't find the option to reply to your comments :thinking:.

Inheritance is fine with me, and having a different flash message is also fine with me.

spec/system/sdg_management/relations_spec.rb
```diff
@@ -88,6 +93,43 @@
...
+    scenario "list records pending to review for the current model by default" do
+      create(:debate, title: "I'm a debate")
+      create(:sdg_review, relatable: create(:debate, title: "I'm a reviewed debate"))
+
+      visit sdg_management_debates_path
+
+      expect(page).to have_text "I'm a debate"
+      expect(page).not_to have_text "I'm a reviewed debate"
+
+      click_link "Pending"
```
javierm, line 105: How about checking the "pending" filter has the active HTML class instead? `expect(page).to have_link "Pending", class: "is-active"`

The "problem" with clicking the link is that the texts we check are already present before clicking the link, and so there might be a potential race condition between the request and the expectations.

app/controllers/sdg_management/relations_controller.rb
```diff
@@ -39,4 +43,15 @@ def check_feature_flags
...
       raise FeatureDisabled, process_name unless Setting["sdg.process.#{process_name}"]
     end
+
+    def update_notice
+      if @record.sdg_review.present?
+        flash[:notice] = t("sdg_management.relations.update.notice",
+          relatable: relatable_class.model_name.human)
+      else
+        @record.create_sdg_review!
+        flash[:notice] = t("sdg_management.relations.update_and_review.notice",
```
javierm, line 53: Is the `flash[:notice]` assignment necessary? :thinking: I mean, what this method returns will be used in the `notice:` param of the `redirect_to` method.

spec/system/sdg_management/relations_spec.rb
Senen: Good catch! Done

config/locales/es/sdg_management.yml
```diff
@@ -27,3 +27,12 @@ es:
...
         notice: "Meta localizada asctualizada correctamente"
+    relations:
+      review_status:
+        all: "Todos"
+        pending_sdg_review: "Pendientes"
+        sdg_reviewed: "Marcados como revisados"
+      update:
+        notice: "%{relatable} marcado como revisado"
+      update_and_review:
+        notice: "%{relatable} marcado como revisado y relaciones ODS actualizadas"
```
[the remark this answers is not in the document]
Senen: Done

config/locales/es/sdg_management.yml
```diff
@@ -27,3 +27,12 @@ es:
         title: "Nueva meta localizada"
       update:
         notice: "Meta localizada asctualizada correctamente"
+    relations:
+      review_status:
+        all: "Todos"
+        pending_sdg_review: "Pendientes"
+        sdg_reviewed: "Marcados como revisados"
+      update:
+        notice: "%{relatable} marcado como revisado"
```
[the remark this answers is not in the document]
Senen: Done

app/controllers/sdg_management/relations_controller.rb
```diff
@@ -39,4 +43,15 @@ def check_feature_flags
       check_feature_flag(process_name)
       raise FeatureDisabled, process_name unless Setting["sdg.process.#{process_name}"]
     end
+
+    def update_notice
+      if @record.sdg_review.present?
+        flash[:notice] = t("sdg_management.relations.update.notice",
```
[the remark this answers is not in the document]
Senen: You are right, it is not necessary! Thanks.

app/controllers/sdg_management/relations_controller.rb
Senen: Fixed!

config/locales/es/sdg_management.yml
```diff
@@ -32,3 +32,12 @@ es:
         title: "Nueva meta localizada"
       update:
         notice: "Meta localizada asctualizada correctamente"
+    relations:
+      review_status:
+        all: "Todos"
+        pending_sdg_review: "Pendientes"
+        sdg_reviewed: "Marcados como revisados"
+      update:
+        notice: "%{relatable} actualizado correctamentes"
```
javierm, line 41: Typo: "correctamente" instead of "correctamentes".

javierm (commented):
Like Columbo used to say: "just one more thing, sir" :wink:.

javierm (approved):

PR #4271, opened 2020-11-25
Add SDG local targets

Senen:
## References

Pending:

1. Check max_length for description attribute

## Objectives
Pending

## Visual Changes
Captura de pantalla 2020-12-06 a las 16 37 40
Captura de pantalla 2020-12-06 a las 16 37 50
Captura de pantalla 2020-12-06 a las 16 38 01

## Notes
Pending

PR #4296, opened 2020-12-23
Allow SDG managers to manage SDG local targets records

Senen:
## References

This PR extends #4271 

## Objectives

Allow SDG managers to manage local target records.

Load local target records by using cancancan `load_and_authorize_resource` method.

### DIFF
--- a/app/controllers/sdg_management/local_targets_controller.rb
+++ b/app/controllers/sdg_management/local_targets_controller.rb
@@ -1,19 +1,16 @@
 class SDGManagement::LocalTargetsController < SDGManagement::BaseController
   include Translatable
 
-  LocalTarget = ::SDG::LocalTarget
+  load_and_authorize_resource class: "SDG::LocalTarget"
 
   def index
-    @local_targets = LocalTarget.all.sort
+    @local_targets = @local_targets.sort
   end
 
   def new
-    @local_target = LocalTarget.new
   end
 
   def create
-    @local_target = LocalTarget.new(local_target_params)
-
     if @local_target.save
       redirect_to sdg_management_local_targets_path, notice: t("sdg_management.local_targets.create.notice")
     else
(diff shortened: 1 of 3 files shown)


spec/models/abilities/sdg/manager.rb
```diff
@@ -9,6 +9,7 @@
 
   it { should be_able_to(:read, SDG::Goal) }
   it { should be_able_to(:read, SDG::Target) }
+  it { should be_able_to(:read, SDG::LocalTarget) }
```
javierm, line 12: Maybe in this case permissions to create, update and destroy are more important :thinking:.
```suggestion
  it { should be_able_to(:manage, SDG::LocalTarget) }
```

app/controllers/sdg_management/local_targets_controller.rb
```diff
@@ -3,12 +3,13 @@ class SDGManagement::LocalTargetsController < SDGManagement::BaseController
 
   LocalTarget = ::SDG::LocalTarget
```
javierm, line 4: This line isn't that useful anymore :thinking:. How about removing it and using `SDG::LocalTarget` in the `create` action?

app/controllers/sdg_management/local_targets_controller.rb
```diff
@@ -1,19 +1,16 @@
...
-    @local_targets = LocalTarget.all.sort
+    @local_targets = @local_targets.sort
   end
 
   def new
-    @local_target = LocalTarget.new
   end
 
   def create
-    @local_target = LocalTarget.new(local_target_params)
```
javierm, line 15: I think we need to keep this line :thinking:; otherwise we are not filtering `params` and so are vulnerable to mass assignment.

Correct me if I'm wrong!

javierm: I was wrong :smile:.

https://github.com/CanCanCommunity/cancancan/wiki/Strong-Parameters

javierm (approved):

PR #4301, opened 2020-12-29
Include link to unsubscribe in email notifications

taitus:
## Objectives
Allow users to unsubscribe from email notifications through an "Unsubscribe" link in the footer of the email without the need to log in.
This link will take them to a unsubscribe page where they can manage the notifications they want to receive.

## Visual Changes
Captura de pantalla 2021-02-02 a las 11 02 17

### DIFF
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -1,0 +1,17 @@
+class SubscriptionsController < ApplicationController
+  before_action :set_user
+  skip_authorization_check
+
+  def edit
+  end
+
+  private
+
+    def set_user
+      @user = if params[:token].present?
+                User.find_by!(subscriptions_token: params[:token])
+              else
+                current_user || raise(CanCan::AccessDenied)
+              end
+    end
+end
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -1,5 +1,5 @@
 class SubscriptionsController < ApplicationController
-  before_action :set_user
+  before_action :set_user, :set_user_locale
   skip_authorization_check
 
   def edit
@@ -27,5 +27,12 @@ class SubscriptionsController < ApplicationController
 
     def allowed_params
       [:email_on_comment, :email_on_comment_reply, :email_on_direct_message, :email_digest, :newsletter]
+    end
+
+    def set_user_locale
+      if params[:locale].blank?
+        I18n.locale = I18n.available_locales.find { |locale| locale == @user.locale&.to_sym }
+        session[:locale] = I18n.locale
+      end
     end
 end
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -5,6 +5,12 @@ class SubscriptionsController < ApplicationController
   def edit
   end
 
+  def update
+    @user.update!(subscriptions_params)
+    redirect_to edit_subscriptions_path(token: @user.subscriptions_token),
+                notice: t("flash.actions.save_changes.notice")
+  end
+
   private
 
     def set_user
@@ -13,5 +19,13 @@ class SubscriptionsController < ApplicationController
               else
                 current_user || raise(CanCan::AccessDenied)
               end
+    end
+
+    def subscriptions_params
+      params.require(:user).permit(allowed_params)
+    end
+
+    def allowed_params
+      [:email_on_comment, :email_on_comment_reply, :email_on_direct_message, :email_digest, :newsletter]
     end
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -415,6 +415,10 @@ class User < ApplicationRecord
     devise_mailer.send(notification, self, *args).deliver_later
   end
 
+  def add_subscriptions_token
+    update!(subscriptions_token: SecureRandom.base58(32)) if subscriptions_token.blank?
+  end
+
   private
 
     def clean_document_number
--- a/app/views/mailer/comment.html.erb
+++ b/app/views/mailer/comment.html.erb
@@ -18,6 +18,13 @@
   </p>
 
   <p style="font-family: 'Open Sans','Helvetica Neue',arial,sans-serif;font-size: 12px;font-weight: normal;line-height: 20px;">
-    <%= t("mailers.config.manage_email_subscriptions") %> <%= link_to t("account.show.title"), account_url, style: "color: #2895F1; text-decoration:none;" %>
+    <%= sanitize(t("mailers.config.unsubscribe_text",
+      notifications: link_to(
+        t("mailers.config.notifications_link"),
+        edit_subscriptions_url(token: @token),
+        style: "color: #2895F1; text-decoration: none;"
+      ),
+      notification: User.human_attribute_name(:email_on_comment)
+    )) %>
   </p>
 </td>
--- a/app/views/mailer/newsletter.html.erb
+++ b/app/views/mailer/newsletter.html.erb
@@ -2,4 +2,15 @@
   <p style="font-family: 'Open Sans','Helvetica Neue',arial,sans-serif;font-size: 14px;line-height: 24px;">
     <%= auto_link_already_sanitized_html wysiwyg(@newsletter.body) %>
   </p>
+
+  <p style="font-family: 'Open Sans','Helvetica Neue',arial,sans-serif;font-size: 12px;font-weight: normal;line-height: 20px;">
+    <%= sanitize(t("mailers.config.unsubscribe_text",
+      notifications: link_to(
+        t("mailers.config.notifications_link"),
+        edit_subscriptions_url(token: @token),
+        style: "color: #2895F1; text-decoration: none;"
+      ),
+      notification: User.human_attribute_name(:newsletter)
+    )) %>
+  </p>
 </td>
--- a/app/views/mailer/reply.html.erb
+++ b/app/views/mailer/reply.html.erb
@@ -18,6 +18,13 @@
   </div>
 
   <p style="font-family: 'Open Sans','Helvetica Neue',arial,sans-serif;font-size: 12px;font-weight: normal;line-height: 20px;">
-    <%= t("mailers.config.manage_email_subscriptions") %> <%= link_to t("account.show.title"), account_url, style: "color: #2895F1; text-decoration:none;" %>
+    <%= sanitize(t("mailers.config.unsubscribe_text",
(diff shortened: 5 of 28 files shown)


app/models/user.rb
```diff
@@ -427,4 +428,8 @@ def validate_username_length
         maximum: User.username_max_length)
       validator.validate(self)
     end
+
+    def add_unsubscribe_hash
+      self.unsubscribe_hash = SecureRandom.hex
```
javierm, line 433: While this is secure, hexadecimal strings look like MD5 hashes and don't feel so secure :thinking:. So, just in order to make the string look more random, how about:

```suggestion
      self.unsubscribe_hash = SecureRandom.base58(32) 
```

I think we can't use `SecureRandom.base64` in this case because it might contain a slash, which is a special character in a URL.

app/views/mailer/comment.html.erb
```diff
@@ -18,6 +18,10 @@
   </p>
 
   <p style="font-family: 'Open Sans','Helvetica Neue',arial,sans-serif;font-size: 12px;font-weight: normal;line-height: 20px;">
-    <%= t("mailers.config.manage_email_subscriptions") %> <%= link_to t("account.show.title"), account_url, style: "color: #2895F1; text-decoration:none;" %>
+    <%= t("mailers.config.manage_email_subscriptions") %>
+    <%= link_to t("mailers.comment.unsubscribe"),
+                edit_unsubscribe_url(@receiver.unsubscribe_hash),
+                style: "color: #2895F1; text-decoration:none;",
+                target: :_blank %>
```
javierm, line 25: Why `target: :_blank`? We almost never use it in mailer views :thinking:.

The same applies to the other links added in this pull request.

javierm (changes requested):
@taitus Here's an initial review. Let me know what you think! :wink:

app/controllers/subscriptions_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    def set_user
+      @user = User.find_by(subscriptions_token: params[:token])
+    end
+
+    def subscriptions_params
+      attributes = [:email_on_comment, :email_on_comment_reply, :email_on_direct_message,
+                    :email_digest, :newsletter]
+      params.require(:user).permit(*attributes)
+    end
```
javierm, line 24: Using an additional method would make customization easier:

```suggestion
    def subscriptions_params
      params.require(:user).permit(allowed_params)
    end
    
    def allowed_params
       [:email_on_comment, :email_on_comment_reply, :email_on_direct_message, :email_digest, :newsletter]
     end
```

This way the custom code could say:

```
  alias_method :consul_allowed_params, :allowed_params

  def allowed_params
    consul_allowed_params + [:my_custom_parameter]
  end
```

app/views/mailer/comment.html.erb
```diff
@@ -18,6 +18,9 @@
   </p>
 
   <p style="font-family: 'Open Sans','Helvetica Neue',arial,sans-serif;font-size: 12px;font-weight: normal;line-height: 20px;">
-    <%= t("mailers.config.manage_email_subscriptions") %> <%= link_to t("account.show.title"), account_url, style: "color: #2895F1; text-decoration:none;" %>
+    <%= t("mailers.config.manage_email_subscriptions") %>
+    <%= link_to t("mailers.comment.unsubscribe"),
+                edit_subscriptions_url(token: @receiver.subscriptions_token),
+                style: "color: #2895F1; text-decoration:none;" %>
```
javierm, line 24: Since we're changing these lines, how about using just one translation, like in other places?

```suggestion
    <%= sanitize(t("...", unsubscribe: link_to(...))) %>
```

This way it'll be easier to translate this sentece to languages where the syntax has nothing to do with the one used in English.

app/views/mailer/newsletter.html.erb
```diff
@@ -2,4 +2,13 @@
...
     <%= auto_link_already_sanitized_html wysiwyg(@newsletter.body) %>
   </p>
+
+  <% if @receiver.present? %>
+    <p style="font-family: 'Open Sans','Helvetica Neue',arial,sans-serif;font-size: 12px;font-weight: normal;line-height: 20px;">
+      <%= t("mailers.config.manage_email_subscriptions") %>
+      <%= link_to t("mailers.newsletter.unsubscribe"),
+                  edit_subscriptions_url(token: @receiver.subscriptions_token),
+                  style: "color: #2895F1; text-decoration:none;" %>
+    </p>
```
javierm, line 12: Here we can also use just one translation, like in other places:

```suggestion
    <%= sanitize(t("...", unsubscribe: link_to(...))) %>
```

app/views/mailer/reply.html.erb
```diff
@@ -18,6 +18,9 @@
   </div>
 
   <p style="font-family: 'Open Sans','Helvetica Neue',arial,sans-serif;font-size: 12px;font-weight: normal;line-height: 20px;">
-    <%= t("mailers.config.manage_email_subscriptions") %> <%= link_to t("account.show.title"), account_url, style: "color: #2895F1; text-decoration:none;" %>
+    <%= t("mailers.config.manage_email_subscriptions") %>
+    <%= link_to t("mailers.reply.unsubscribe"),
+                edit_subscriptions_url(token: @receiver.subscriptions_token),
+                style: "color: #2895F1; text-decoration:none;" %>
```
javierm, line 24: Since we're changing these lines, how about using just one translation, like in other places?

```suggestion
    <%= sanitize(t("...", unsubscribe: link_to(...))) %>
```

This way it'll be easier to translate this sentece to languages where the syntax has nothing to do with the one used in English.

javierm (commented):
Took me a while to get back to this pull request, but here's the second review :smile:.

spec/system/subscriptions_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    scenario "Render content in the user's preferred locale" do
+      user.update!(locale: "es")
+      visit edit_subscriptions_path(token: user.subscriptions_token)
+
+      expect(page).to have_content "Notificaciones"
+      expect(page).to have_content "Recibir un email cuando alguien comenta en mis propuestas o debates"
+      expect(page).to have_content "Recibir un email cuando alguien contesta a mis comentarios"
+      expect(page).to have_content "Recibir emails con información interesante sobre la web"
+      expect(page).to have_content "Recibir resumen de notificaciones sobre propuestas"
+      expect(page).to have_content "Recibir emails con mensajes privados"
```
javierm, line 16: We could use `have_field`, just like we did in the component spec.

spec/system/subscriptions_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    end
+  end
+
+  context "Update" do
+    scenario "Allow update the status notification" do
```
javierm, line 21: Typo:

```suggestion
    scenario "Allow updating the status notification" do
```

spec/system/subscriptions_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      user.update!(email_on_comment: true,
+                   email_on_comment_reply: true,
+                   email_on_direct_message: true,
+                   email_digest: true,
+                   newsletter: true)
```
javierm, line 26: Maybe we could change a couple of them to `false`, and update just 4 of them, so we also check what happens when changing from `false` to `true` and what happens when the option isn't changed :thinking:.

javierm (commented):
Getting there! :+1: I haven't reviewed my comments, but since we wrote them together, I hope that's fine :smile:.

spec/models/user_spec.rb
```diff
@@ -823,4 +823,22 @@
       expect(other_user_legislation_proposal.reload).to be_hidden
     end
   end
+
+  describe "#add_subscriptions_token" do
+    let(:user) { build(:user, subscriptions_token: nil) }
+
+    it "generate subscriptions_token when user has not related token" do
```
javierm, line 830:
```suggestion
    it "generates a subscriptions token when the user doesn't have one" do
```

spec/models/user_spec.rb
```diff
@@ -823,4 +823,22 @@
...
+      user.add_subscriptions_token
+
+      expect(user.subscriptions_token).to be_present
+    end
+
+    it "keep the existing subscriptions_token when a user already has a related token" do
```
javierm, line 836:
```suggestion
    it "keeps the existing subscriptions token when the user already has one" do
```

spec/system/subscriptions_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+      check "Receive a summary of proposal notifications"
+      uncheck "Receive emails about direct messages"
+      click_button "Save changes"
+
+      expect(page).to have_content "Changes saved"
+      expect(user.reload.email_on_comment).to eq true
+      expect(user.reload.email_on_comment_reply).to eq false
+      expect(user.reload.newsletter).to eq false
+      expect(user.reload.email_digest).to eq true
+      expect(user.reload.email_on_direct_message).to eq false
```
javierm, line 48: Using `user.reload` means we're accessing the database after the process running the browser has already started, which might result in some `PG::ProtocolViolation` exceptions when running the tests.

How about checking the checkboxes are present with the updated values?

spec/mailers/mailer_spec.rb
```diff
@@ -29,4 +29,26 @@
       expect(email.subject).to include("commented on your proposal")
     end
   end
+
+  describe "#manage_subscriptions_token" do
+    let(:user) { create(:user) }
+    let(:proposal) { create(:proposal, author: user) }
+    let(:comment) { create(:comment, commentable: proposal) }
+
+    it "when receiver has not subscriptions_token we generate subscriptions_token for it" do
```
javierm, line 38:
```suggestion
    it "generates a subscriptions token when the receiver doesn't have one" do
```

spec/mailers/mailer_spec.rb
```diff
@@ -29,4 +29,26 @@
...
+      user.update!(subscriptions_token: nil)
+
+      Mailer.comment(comment).deliver_now
+
+      expect(user.reload.subscriptions_token).to be_present
+    end
+
+    it "when the receiver already has subscriptions_token, we keep it " do
```
javierm, line 46:
```suggestion
    it "uses the existing subscriptions token when the receivesr already has one " do
```

javierm (changes requested):
Probably my final review here; just found a couple of typos and a possible cause of flaky tests. I'm gonna miss this pull request :smile:.

javierm (approved):

PR #4311, opened 2021-01-11
Add SDG cards management section

javierm:
## References

* Depends on pull request #4310

## Objectives

* Make it possible to add cards which will be assigned to SDG phases

## Visual changes

In the SDG content management there's a new homepage section listing and allowing to edit cards for each phase

## Notes

In a future pull request, we'll add the cards to the public SDG homepage


db/sdg.rb
```diff
@@ -23,3 +23,5 @@
 ].each do |code|
   SDG::Target.where(code: code, goal: SDG::Goal.find_by!(code: code.split(".").first)).first_or_create!
 end
+
+3.times { |kind| SDG::Phase.where(kind: kind).first_or_create! }
```
Senen, line 27: What about looping over `SDG::Phase.kinds.values`? By doing this way, if someone customizes the kinds array definition for any reason won't have to update this line too.

```
SDG::Phase.kinds.values.each { |kind| SDG::Phase.where(kind: kind).first_or_create! }
```

Senen (commented):
Hi @javierm, i have to do a manual testing yet but here are a few comments.

What do you think about creating card/s for each phase?

app/models/sdg/phase.rb
```diff
@@ -0,0 +1,8 @@
+class SDG::Phase < ApplicationRecord
+  enum kind: %w[sensitization planning monitoring]
```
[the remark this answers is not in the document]
javierm: There's no way to access this array (it isn't stored in a variable/constant), so it's not needed. And the `enum` method converts this array to a hash and freezes it automatically :smile:.

db/sdg.rb
javierm: Done!

javierm: Updated the `dev_seeds` as well :+1:.

app/models/sdg/phase.rb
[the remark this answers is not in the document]
Senen: Didn't know that. Thanks for the explanation.

app/models/abilities/everyone.rb
```diff
@@ -29,6 +29,7 @@ def initialize(user)
       can [:search, :comments, :read, :create, :new_comment], Legislation::Annotation
 
       can :read, ::SDG::Goal
+      can :read, ::SDG::Phase
```
Senen, line 32: I noticed there a missing spec for this change.

javierm: Added :+1:.

Senen (approved):

PR #4314, opened 2021-01-13
Add icons and cards to SDG homepage

javierm:
## References

* Depends on pull request #4311

## Objectives

* Add content to the SDG homepage

## Visual changes

SDG homepage with the SDG icons and cards for each phase

## Notes

We've still need to decide what to do about this page main header.


Senen (approved):
Tested locally on responsive screens and seems to work fine!

PR #4315, opened 2021-01-15
Add SDG goal/target selector

taitus:
## References
Related PR: #4262 

## Objectives
- Adding a new component to be able to relate a resource to Goals and Targets with the same field.
- This component will also have a list of icons of all the Goals to be able to add or remove tags by clicking on them.
- This field has to be able to autocomplete both Goals and Targets.
- We will show tags with for each Goal and Target related to the following format:
  - Goals:  "SDG" prefix + Goal code. Example "SDG1", "SDG2", ...
  - Targets: Target code. Example "1.1", "2.A", ...
- Add a help section to display the title of the selected Goals and Targets/

## Visual Changes
Captura de pantalla 2021-01-20 a las 13 59 28

## Notes
Add JQuery library for input tags: https://github.com/amsify42/jquery.amsify.suggestags

### DIFF
--- a/app/components/sdg/related_list_selector_component.rb
+++ b/app/components/sdg/related_list_selector_component.rb
@@ -1,0 +1,7 @@
+class SDG::RelatedListSelectorComponent < ApplicationComponent
+  attr_reader :f
+
+  def initialize(form)
+    @f = form
+  end
+end
--- a/app/components/sdg/related_list_selector_component.rb
+++ b/app/components/sdg/related_list_selector_component.rb
@@ -5,6 +5,10 @@ class SDG::RelatedListSelectorComponent < ApplicationComponent
     @f = form
   end
 
+  def checked?(code)
+    f.object.sdg_goals.find_by(code: code).present?
+  end
+
   def sdg_related_suggestions
     goals_and_targets.map { |goal_or_target| suggestion_tag_for(goal_or_target) }
   end
--- a/app/components/sdg/related_list_selector_component.rb
+++ b/app/components/sdg/related_list_selector_component.rb
@@ -22,6 +22,7 @@ class SDG::RelatedListSelectorComponent < ApplicationComponent
   def suggestion_tag_for(goal_or_target)
     {
       tag: "#{goal_or_target.code}. #{goal_or_target.title.gsub(",", "")}",
+      display_text: text_for(goal_or_target),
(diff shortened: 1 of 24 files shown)


javierm (commented):
The interface is looking great. Thanks a lot!

I've left a few comments regarding possible improvements and simplifications. Let me know what you think!

javierm (commented):
I always forget to check why we don't have Hound anymore... I think it happened when we moved to github actions :thinking:.

app/components/sdg/related_list_selector_component.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def checked?(code)
+    f.object.sdg_goals.find_by(code: code).present?
+  end
+
+  def sdg_related_suggestions
+    goals.map do |goal|
+      [goal, *goal.targets.sort]
+    end.flatten.map do |goal_or_target|
+      display_text = goal_or_target.class.name == "SDG::Goal" ? "SDG#{goal_or_target.code}" : "#{goal_or_target.code}"
```
javierm, line 16: Forgot one thing :relieved:.

Should we use internationalization here? So, instead of "SDG1", we get "ODS1" in Spanish (for instance).

config/locales/en/activerecord.yml
```diff
@@ -2,6 +2,7 @@ en:
   attributes:
     geozone_id: "Scope of operation"
     results_enabled: "Show results"
+    sdg_related_list: "Sustainable Development Goals and Targets"
     sdg_target_list: "Targets"
```
javierm, line 6: I think we can now remove this line, since it's replaced by `sdg_related_list`.

config/locales/es/activerecord.yml
```diff
@@ -2,6 +2,7 @@ es:
   attributes:
     geozone_id: "Ámbito de actuación"
     results_enabled: "Mostrar resultados"
+    sdg_related_list: "Objetivos de Desarrollo Sostenible y Metas"
     sdg_target_list: "Metas"
```
javierm, line 6: I think we can now remove this line, since it's replaced by `sdg_related_list`.

javierm (commented):
Just a couple of small things I've noticed. Let me know what you think!

javierm (approved):
Super cool!

PR #4320, opened 2021-01-19
Make it possible to add banners to the SDG section

javierm:
## References

* We added a new SDG section in pull request #4292

## Objectives

* Simplify code related to obtain and render banners
* Make it possible to add banners to the SDG section, just like it's possible to add banners to any other section.

### DIFF
--- a/app/components/shared/banner_component.html.erb
+++ b/app/components/shared/banner_component.html.erb
@@ -1,3 +1,5 @@
-<div class="banner" style="background-color:<%= banner.background_color %>;">
-  <%= sanitize link, attributes: %w[href style] %>
-</div>
+<% if banner %>
+  <div class="banner" style="background-color:<%= banner.background_color %>;">
+    <%= sanitize link, attributes: %w[href style] %>
+  </div>
+<% end %>
--- a/app/components/shared/banner_component.html.erb
+++ b/app/components/shared/banner_component.html.erb
@@ -1,0 +1,3 @@
+<div class="banner" style="background-color:<%= banner.background_color %>;">
+  <%= sanitize link, attributes: %w[href style] %>
+</div>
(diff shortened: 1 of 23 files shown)


Senen (commented):
Hi @javierm, I left only one suggestion.

app/components/shared/banner_component.html.erb
```diff
@@ -1,3 +1,5 @@
-<div class="banner" style="background-color:<%= banner.background_color %>;">
-  <%= sanitize link, attributes: %w[href style] %>
-</div>
+<% if banner %>
```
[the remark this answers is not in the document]
javierm: That's a good idea! However, some CONSUL users/developers might have a harder time figuring out why a view is not rendered :thinking:. I know I introduced the `render?` method myself a couple of times, but been wondering since then whether we should remove it :thinking:.

So, when in doubt, I think our criteria is to write code which is similar to existing code :thinking:.

Senen (approved):

spec/components/shared/banner_component_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    create(:banner, web_sections: [WebSection.find_by!(name: "debates")], title: "Debate banner")
+    create(:banner, web_sections: [WebSection.find_by!(name: "proposals")], title: "Proposal banner")
+
+    render_inline Shared::BannerComponent.new("debates")
+
+    expect(page.find(".banner")).to have_content "Debate banner"
+    expect(page).not_to have_content "Proposal banner"
+  end
+
+  it "consitently renders one banner" do
```
Senen, line 20: Small typo here: 
```suggestion
  it "consistently renders one banner" do
```

javierm: Fixed, thanks!

PR #4322, opened 2021-01-20
Allow filtering by local target in SDG management

javierm:
## References

* Filters by goal and target were added in pull request #4269

## Objectives

* Allow selecting a local target in the SDG management target filter

## Visual Changes

The target selector lists both targets and local targets

### DIFF
--- a/app/models/sdg/local_target.rb
+++ b/app/models/sdg/local_target.rb
@@ -17,10 +17,12 @@ class SDG::LocalTarget < ApplicationRecord
 
   belongs_to :target
 
-  def <=>(local_target)
-    return unless local_target.class == self.class
-
-    [target, numeric_subcode] <=> [local_target.target, local_target.numeric_subcode]
+  def <=>(any_target)
+    if any_target.class == self.class
+      [target, numeric_subcode] <=> [any_target.target, any_target.numeric_subcode]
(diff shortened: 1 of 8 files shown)


app/models/sdg/local_target.rb
```diff
@@ -17,6 +17,10 @@ class SDG::LocalTarget < ApplicationRecord
 
   belongs_to :target
 
+  def self.[](code)
```
Senen, line 20: I think we are missing the spec for this method.

```
  describe ".[]" do
    it "finds existing local targets by code" do
      create(:sdg_local_target, code: "1.1.1")

      expect(SDG::LocalTarget["1.1.1"].code).to eq "1.1.1"
    end

    it "raises an exception for non-existing codes" do
      expect { SDG::LocalTarget["Z.j3"] }.to raise_exception ActiveRecord::RecordNotFound
    end
  end
```

Senen (changes requested):
Hi @javierm, the only thing I detected is a new model method without specs.

app/models/sdg/local_target.rb
javierm: I've removed the method; it was only used in one spec.

taitus (approved):

PR #4323, opened 2021-01-21
Add SDG stats page to admin section

Senen:
## Objectives

Add new page to admin stats to show each goal stats:

* Number of proposals
* Number of polls
* Number of debates
* Investment projects:
  * Number of submitted projects
  * Number of winner projects

## Visual Changes

**Desktop version**
Captura de pantalla 2021-01-21 a las 16 16 50
Captura de pantalla 2021-01-21 a las 16 17 01

**Mobile version**
Captura de pantalla 2021-01-21 a las 16 21 26

### DIFF
--- a/app/components/admin/stats/sdg_component.html.erb
+++ b/app/components/admin/stats/sdg_component.html.erb
@@ -1,0 +1,5 @@
+<%= back_link_to %>
+
+<%= header %>
+
+<%= render Admin::Stats::SDG::GoalComponent.with_collection(goals) %>
--- a/app/controllers/admin/stats_controller.rb
+++ b/app/controllers/admin/stats_controller.rb
@@ -90,6 +90,10 @@ class Admin::StatsController < Admin::BaseController
     @participants = ::Poll::Voter.where(poll: @polls)
   end
 
+  def sdg
+    @goals = SDG::Goal.order(:code)
+  end
+
   private
 
     def voters_in_heading(heading)
--- a/spec/system/admin/stats_spec.rb
+++ b/spec/system/admin/stats_spec.rb
@@ -395,4 +395,29 @@ describe "Stats", :admin do
       end
     end
   end
+
+  context "SDG", :js do
+    scenario "Shows SDG stats link when SDG feature is enabled" do
+      Setting["feature.sdg"] = true
+
+      visit admin_stats_path
+
+      expect(page).to have_link "SDG", href: sdg_admin_stats_path
+    end
+
+    scenario "Does not show SDG stats link when SDG feature is disbled" do
+      Setting["feature.sdg"] = false
+
+      visit admin_stats_path
+
+      expect(page).not_to have_link "SDG"
+    end
+
+    scenario "Renders all goals stats" do
+      visit sdg_admin_stats_path
+
+      expect(page).to have_css "h3", count: SDG::Goal.count
+      expect(page).to have_css ".sdg-goal-stats", count: SDG::Goal.count
+    end
+  end
 end
--- a/app/components/admin/stats/sdg/goal_component.html.erb
+++ b/app/components/admin/stats/sdg/goal_component.html.erb
@@ -1,0 +1,14 @@
+<div class="sdg-goal-stats">
+
+  <h3><%= goal.code_and_title %></h3>
+
+  <div class="stats-numbers">
+    <% stats.each do |text, amount, options = {}| %>
+      <div class="small-12 medium-4 column">
+        <%= tag.p(options) do %>
+          <%= text %> <br><span class="number"><%= amount %></span>
+        <% end %>
+      </div>
+    <% end %>
+  </div>
+</div>
--- a/app/components/admin/stats/sdg/goal_component.rb
+++ b/app/components/admin/stats/sdg/goal_component.rb
@@ -1,0 +1,30 @@
+class Admin::Stats::SDG::GoalComponent < ApplicationComponent
+  with_collection_parameter :goal
+
+  attr_reader :goal
+
+  def initialize(goal:)
+    @goal = goal
+  end
+
+  private
+
+    def stats
+      [
+        [t("admin.stats.sdg.polls"), goal.polls.count],
+        [t("admin.stats.sdg.proposals"), goal.proposals.count],
+        [t("admin.stats.sdg.debates"), goal.debates.count],
+        [t("admin.stats.sdg.budget_investments.sent"), goal.budget_investments.count],
+        [t("admin.stats.sdg.budget_investments.winners"), goal.budget_investments.winners.count, featured],
+        [t("admin.stats.sdg.budget_investments.amount"), amount, featured]
+      ]
+    end
+
+    def amount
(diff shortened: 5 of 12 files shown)


app/components/shared/sdg/goal_stats_component.html.erb
```diff
@@ -0,0 +1,43 @@
...
+
+    <div class="small-12 medium-4 column">
+      <p>
+        <%= t("sdg.stats.proposals") %> <br><span class="number"><%= goal.proposals.count %></span>
+      </p>
+    </div>
+
+    <div class="small-12 medium-4 column">
+      <p>
+        <%= t("sdg.stats.debates") %> <br><span class="number"><%= goal.debates.count %></span>
```
Senen, line 21: Here, I wanted to use the next format:
```
<%= t("sdg.stats.debates") %> <br>
<span class="number"><%= goal.debates.count %></span>
```
the same we are using at `admin/stats/show.html.erb`, but the [test](https://github.com/consul/consul/pull/4323/files#diff-945a664114cd9aafbbb3111efcc100260c21aca2ae6ee85c861ffc4e106fad14R16) failed because the `page.text` has the text `Proposals         \n3` so the following expectations fails:

```
expect(page).to have_text "Proposals 3"
```
@javierm, @taitus do you know what I am missing here? Maybe this is something related to the type of the spec, which is `:component`.

javierm: The `system/admi/stats_spec.rb` tests are passing because they've got JavaScript disabled. It looks like component specs follow similar rules as JavaScript tests :thinking:.

So, yeah, probably the test should expect the test to have the "\n" :thinking:.

app/components/admin/stats/sdg_component.html.erb
```diff
@@ -0,0 +1,5 @@
+<%= back_link_to %>
+
+<%= header %>
+
+<%= render Shared::SDG::GoalStatsComponent.with_collection(goals) %>
```
javierm, line 5: Why `Shared::` instead of `Admin::` :thinking:? How about `Admin::Stats::GoalComponent` or `Admin::Stats::SDG::GoalComponent`?

app/components/shared/sdg/goal_stats_component.rb
```diff
@@ -0,0 +1,15 @@
...
+
+  attr_reader :goal
+
+  def initialize(goal:)
+    @goal = goal
+  end
+
+  private
+
+    def approved_budget
```
javierm, line 12: When I read this name I thought it would return something like `Budget.approved.first`. How about naming it `approved_amount`?

app/components/shared/sdg/goal_stats_component.html.erb
```diff
@@ -0,0 +1,43 @@
+<div class="sdg-goal-stats">
+
+  <h3><%= goal.code_and_title %></h3>
+
+  <div class="stats-numbers">
+
+    <div class="small-12 medium-4 column">
```
javierm, line 7: Maybe we could remove some duplication :thinking:. Here's one option:

```erb
  <div class="stats-numbers">
    <% stats.each do |text, amount, options = {}| %>
      <div class="small-12 medium-4 column">
        <%= tag.p(options) do %>
          <%= text %> <br><span class="number"><%= amount %></span>
        <% end %>
      </div>
    <% end %>
```

And then, in the component:

```ruby
  def stats
    [
      [t("sdg.stats.polls"), goal.polls.count],
      # ...
      [t("sdg.stats.budget_investments.sent"), goal.budget_investments.count, class: "sent"]
      # ...
    ]
  end
```

app/controllers/admin/stats_controller.rb
```diff
@@ -90,6 +90,10 @@ def polls
     @participants = ::Poll::Voter.where(poll: @polls)
   end
 
+  def sdg
+    @goals = SDG::Goal.all.order(:code)
```
javierm, line 94:
```suggestion
    @goals = SDG::Goal.order(:code)
```

spec/system/admin/stats_spec.rb
```diff
@@ -395,4 +395,24 @@
       end
     end
   end
+
+  context "SDG" do
```
javierm, line 399:
```suggestion
  context "SDG", :js do
```

spec/system/admin/stats_spec.rb
```diff
@@ -395,4 +395,24 @@
...
+    scenario "Shows SDG stats link when SDG feature is enabled" do
+      visit admin_stats_path
+
+      expect(page).not_to have_link "SDG"
+
+      Setting["feature.sdg"] = "t"
```
javierm, line 405: IMHO splitting it in two tests we don't lose any performance (there's no common setup for these tests) and we gain a bit of clarity:

```ruby
    scenario "Shows SDG stats link when SDG feature is enabled" do
      Setting["feature.sdg"] = true

      visit admin_stats_path

      expect(page).to have_link "SDG", href: sdg_admin_stats_path
    end

    scenario "Does not show SDG stats link when SDG feature is disbled" do
      Setting["feature.sdg"] = false

      visit admin_stats_path

      expect(page).not_to have_link "SDG"
    end
```

Also note we set the setting both times, so these tests do not depend on the default option (which we might change in the future).

config/locales/es/sdg.yml
```diff
@@ -436,6 +436,15 @@ es:
...
       remove_tag: "Eliminar"
+    stats:
+      budget_investments:
+        amount: "Presupuesto"
+        approved: "Proyectos de inversión ganadores"
+        sent: "Proyectos de inversión enviados"
+      debates: "Debates"
+      polls: "Votaciones"
+      proposals: "Propuestas"
+      title: "Objetivos del Desarrollo Sostenible - Estadísticas"
```
javierm, line 450:
```suggestion
      title: "Objetivos de Desarrollo Sostenible - Estadísticas"
```

config/locales/en/sdg.yml
```diff
@@ -439,3 +439,12 @@ en:
       hint: "You can introduce the code of a specific goal/target or a text to find one"
       placeholder: "Write a goal or target code or description"
       remove_tag: "Remove"
+    stats:
```
javierm, line 442: I wonder whether we should move these lines so they're next to the rest of the `stats` translations :thinking:. Right now it makes sense to me, since translations for budgets stats are there and not in the `budgets.yml` file. What do you think?

app/components/admin/stats/sdg_component.html.erb
Senen: Initially, I thought that probably we would want to render SDG stats at public pages too, but of course, currently, this is potentially a YAGNI thing.

config/locales/en/sdg.yml
Senen: Ok, as I am going to change the component name as you suggested, `admin.yml` would be a better location for the component translations and is consistent with other stats translations.

Senen:
Hi @javierm, I think everything is fixed now.

javierm (approved):

PR #4324, opened 2021-01-21
Add SDG target tags and filter

javierm:
## References

* Goal tags and filters were added in pull requests #4305 and #4300

## Objectives

* Make it possible to filter debates, proposals and budget investments by SDG target
* Display associated SDG targets alongside associated SDGs

### Before these changes

There's no filter by target and no targets are displayed alongside SDG icons

### After these changes

There's a filter by target in the debates index and only debates tagged with that target are displayed

### DIFF
--- a/app/assets/stylesheets/advanced_search.scss
+++ b/app/assets/stylesheets/advanced_search.scss
@@ -22,10 +22,10 @@
 
     .date-filters {
       float: left;
-      width: 50%;
+      width: 25%;
 
       .filter {
-        width: 50%;
+        width: 100%;
       }
 
       .custom-date-filters {
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -29,6 +29,7 @@
 @import "jquery-ui/sortable";
 @import "leaflet";
 @import "sticky_overrides";
+@import "tags";
 @import "admin/*";
 @import "sdg/**/*";
 @import "sdg_management/*";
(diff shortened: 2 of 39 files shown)


Senen (changes requested):
Hi Javi, here are the issues I found:

1º There are some missing translations. See attached patch: `TagListComponenetLocalTargetTitles.txt`

2º I think I found a case when the see more link of SDG::Goal::TagListComponent is not working properly. To reproduce the issue you can assign more targets or local targets to a relatable but do not save the direct relation with their goals, this happens because the method `count_out_of_limit` only considers the relatable goals number. If we are always going to ensure to save the direct relation with the goal we probably can let it as it is.
Check the attached patch called: `TagListComponenetSeeMoreLimit.txt`

And the last visual thing is the text color of the "See more" link. Check the image below:
Captura de pantalla 2021-01-22 a las 14 31 01

**Patches**
[TagListComponentLocalTargetsTitles.txt](https://github.com/consul/consul/files/5856145/TagListComponentLocalTargetsTitles.txt)
[TagListComponentSeeMoreLimit.txt](https://github.com/consul/consul/files/5856151/TagListComponentSeeMoreLimit.txt)


I left two more comments at code.

javierm:
Thanks for the review! I've added the changes you mentioned. We might need to review the "limit" thing in the future, but I guess using the same interface and limit we use for tags is OK for now :thinking:.

Senen (approved):
👍🏼

Senen (approved):

PR #4331, opened 2021-01-25
Add targets to SDG view

javierm:
## Objectives

* Add lists of global and localized targets related to a goal

## Visual Changes
Captura de pantalla 2021-01-27 a las 10 15 12

### DIFF
--- a/app/components/sdg/goals/show_component.html.erb
+++ b/app/components/sdg/goals/show_component.html.erb
@@ -34,4 +34,6 @@
       <%= render ::Widget::Feeds::FeedComponent.new(processes_feed) %>
     </div>
   <% end %>
+
+  <%= render SDG::Goals::TargetsComponent.new(goal) %>
 </div>
--- a/app/components/sdg/goals/targets_component.html.erb
+++ b/app/components/sdg/goals/targets_component.html.erb
@@ -10,12 +10,18 @@
   <div class="tabs-content" data-tabs-content="target_tabs">
     <% [global_targets, local_targets].each do |targets| %>
       <div class="tabs-panel <%= active(targets) %>" id="tab_<%= type(targets) %>_targets">
-        <dl>
-          <% targets.sort.each do |target| %>
-            <dt><%= target.code %><dt>
-            <dd><%= target.title %><dd>
-          <% end %>
-        </dl>
+        <% if targets.blank? %>
+          <div class="callout primary">
+            <%= t("sdg.goals.show.targets.no_local_targets") %>
+          </div>
+        <% else %>
+          <dl>
+            <% targets.sort.each do |target| %>
+              <dt><%= target.code %><dt>
+              <dd><%= target.title %><dd>
+            <% end %>
+          </dl>
+        <% end %>
       </div>
     <% end %>
   </div>
--- a/app/components/sdg/goals/targets_component.html.erb
+++ b/app/components/sdg/goals/targets_component.html.erb
@@ -1,0 +1,22 @@
+<section class="targets">
+  <ul class="tabs" id="target_tabs" data-deep-link="true" data-tabs>
+    <% [global_targets, local_targets].each do |targets| %>
+      <li class="tabs-title <%= active(targets) %>">
+        <%= link_to title(targets), "#tab_#{type(targets)}_targets" %>
+      </li>
+    <% end %>
+  </ul>
+
+  <div class="tabs-content" data-tabs-content="target_tabs">
+    <% [global_targets, local_targets].each do |targets| %>
+      <div class="tabs-panel <%= active(targets) %>" id="tab_<%= type(targets) %>_targets">
+        <dl>
+          <% targets.sort.each do |target| %>
+            <dt><%= target.code %><dt>
+            <dd><%= target.title %><dd>
+          <% end %>
+        </dl>
+      </div>
+    <% end %>
+  </div>
+</section>
--- a/app/components/sdg/goals/targets_component.rb
+++ b/app/components/sdg/goals/targets_component.rb
@@ -1,0 +1,37 @@
+class SDG::Goals::TargetsComponent < ApplicationComponent
+  attr_reader :goal
+
+  def initialize(goal)
+    @goal = goal
+  end
+
+  def render?
+    feature?("sdg")
+  end
+
+  private
+
+    def global_targets
+      goal.targets
+    end
+
+    def local_targets
+      goal.local_targets
+    end
+
+    def type(targets)
+      if targets.model.name == "SDG::Target"
+        "global"
+      else
+        "local"
+      end
+    end
+
+    def active(targets)
+      "is-active" if targets.model.name == "SDG::Target"
+    end
+
+    def title(targets)
+      targets.model.model_name.human(count: :other).upcase_first
+    end
+end
--- a/config/locales/en/sdg.yml
+++ b/config/locales/en/sdg.yml
@@ -444,6 +444,8 @@ en:
       show:
         read_more: "Read more about %{goal}"
         read_less: "Read less about %{goal}"
+        targets:
+          no_local_targets: This goal has no local targets
       title: "Sustainable Development Goals"
       filter:
         heading: "Filters by SDG"
--- a/config/locales/es/sdg.yml
+++ b/config/locales/es/sdg.yml
@@ -444,6 +444,8 @@ es:
       show:
         read_more: "Leer más sobre %{goal}"
         read_less: "Leer menos sobre %{goal}"
+        targets:
+          no_local_targets: Este objetivo no tiene metas localizadas
       title: "Objetivos de Desarrollo Sostenible"
       filter:
         heading: "Filtros por ODS"
--- a/spec/components/sdg/goals/targets_component_spec.rb
+++ b/spec/components/sdg/goals/targets_component_spec.rb
@@ -1,0 +1,39 @@
+require "rails_helper"
+
+describe SDG::Goals::TargetsComponent, type: :component do
+  let(:goal) { SDG::Goal[1] }
+  let(:component) { SDG::Goals::TargetsComponent.new(goal) }
+
+  before do
+    Setting["feature.sdg"] = true
+  end
+
+  it "does not render when the feature is disabled" do
+    Setting["feature.sdg"] = false
+
+    render_inline component
+
+    expect(page).not_to have_css ".targets"
+    expect(page).not_to have_css "#target_tabs"
+    expect(page).not_to have_css ".tabs-content"
+  end
+
+  it "renders tabs panel" do
+    render_inline component
+
+    expect(page).to have_css ".targets"
+    expect(page).to have_css "#target_tabs"
+    expect(page).to have_css "li", count: 2
+    expect(page).to have_content "Targets"
+    expect(page).to have_content "Local targets"
+    expect(page).to have_css ".tabs-content"
+    expect(page).to have_css "#tab_global_targets"
+  end
+
+  it "renders code and title for each target" do
+    render_inline component
+
+    expect(page).to have_content "1.1"
+    expect(page).to have_content "By 2030, eradicate extreme poverty for all people everywhere"
+  end
+end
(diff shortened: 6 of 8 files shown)


app/components/sdg/goals/show_component.html.erb
```diff
@@ -34,4 +34,8 @@
       <%= render ::Widget::Feeds::FeedComponent.new(processes_feed) %>
     </div>
   <% end %>
+
+  <section class="targets">
```
javierm, line 38: To be consistent with the location of the CSS file (under `sdg/goals/targets.scss`), maybe this `<section class="targets">` should be inside the `sdg/goals/targets` component :thinking:.

app/components/sdg/goals/target_component.html.erb
```diff
@@ -0,0 +1,10 @@
+<div class="tabs-panel <%= active %>" id="tab_<%= type %>_targets">
```
javierm, line 1: IMHO the `tabs-panel` is easier to understand if written inside the `TargetsComponent`. This would make this component (potentially) more reusable, since it would be just a definition list which could be inserted anywhere.

The code in the `TargetsComponent` view could be something like:

```erb
<ul class="tabs" ...>
...
</ul>

<div class="tabs-content" data-tabs-content="target_tabs">
  <div class="tabs-panel ...">
    <%= render SDG::Goals::TargetComponent.new(global_targets) %>
  </div>
  <div class="tabs-panel ...">
    <%= render SDG::Goals::TargetComponent.new(local_targets) %>
  </div>
</div>
```

If you'd like to avoid the duplication this change introduces, we could do something like:

```erb
<% [global_targets, local_targets].each do |targets| %>
  <div class="tabs-panel ...">
    <%= render SDG::Goals::TargetComponent.new(targets) %>
  </div>
<% end %>
```

Let me know what you think :thinking:. One inconvenient could be the logic behind the `active` method.

app/components/sdg/goals/target_component.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  def render?
+    feature?("sdg")
+  end
+
+  private
+
+    def type
+      return "global" if targets.model.name == "SDG::Target"
+      "local"
```
javierm, line 16:
```suggestion
      if targets.model.name == "SDG::Target"
        "global"
      else
        "local"
      end
```

It looks the same, but IMHO using `if` has a subliminal message meaning both routes are similar, while using `return` means something special/different happens when that condition is true.

app/components/sdg/goals/targets_component.html.erb
```diff
@@ -0,0 +1,13 @@
+<ul class="tabs" id="target_tabs" data-deep-link="true" data-tabs>
+  <li class="tabs-title is-active">
+    <%= link_to SDG::Target.model_name.human(count: 2).upcase_first, "#tab_global_targets" %>
```
javierm, line 3: How about extracting `SDG::Target.model_name.human(count: 2).upcase_first` to a method in the component?

By the way, use `count: :other` to avoid possible conflicts in languages with different plurals for different amounts.

app/components/sdg/goals/targets_component.html.erb
```diff
@@ -0,0 +1,13 @@
...
+  </li>
+  <li class="tabs-title">
+    <%= link_to SDG::LocalTarget.model_name.human(count: 2).upcase_first, "#tab_local_targets" %>
```
javierm, line 6: How about extracting `SDG::LocalTarget.model_name.human(count: 2).upcase_first` to a method in the component?

By the way, use `count: :other` to avoid possible conflicts in languages with different plurals for different amounts.

spec/components/sdg/goals/target_component_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    expect(page).to have_css "#tab_global_targets"
+    expect(page).to have_css ".global-targets"
+  end
+
+  it "renders code and title for each target" do
+    render_inline component
+
+    expect(page).to have_content "1.1"
+    expect(page).to have_content "By 2030, eradicate extreme poverty for all people everywhere, currently measured as people living on less than $1.25 a day"
```
javierm, line 31: This proves changing line length limit to 140 will only result in requests to change it to 160 (only way this line wouldn't be an offense) :smirk:.

```suggestion
    expect(page).to have_content "By 2030, eradicate extreme poverty for all people everywhere"
```

spec/components/sdg/goals/target_component_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+    expect(page).to have_content "1.B"
+    expect(page).to have_content "Create sound policy frameworks at the national, regional and international levels, based on pro-poor and gender-sensitive development strategies, to support accelerated investment in poverty eradication actions"
```
javierm, line 33: This proves changing line length limit to 160 will only result in requests to change it to 250 (only way this line wouldn't be an offense) :sweat_smile:.

IMHO we can remove the last two lines :thinking:.

spec/system/sdg/goals_spec.rb
```diff
@@ -117,5 +117,27 @@
...
+      create(:sdg_local_target, code: "15.1.1", title: "SDG local target sample text")
+      visit sdg_goal_path(15)
+
+      within "#target_tabs" do
+        expect(page).to have_content "Targets"
+        expect(page).to have_content "Local targets"
+      end
+
+      within ".tabs-content" do
+        expect(page).to have_content "15.1 By 2020, ensure the conservation, restoration and sustainable use of terrestrial and inland freshwater ecosystems and their services, in particular forests, wetlands, mountains and drylands, in line with obligations under international agreements."
```
javierm, line 131: This proves changing line length limit to 250 will only result in requests to change it to 300 (only way this line wouldn't be an offense) :rofl:.

How about using the first sentence of a different target (this one seems to be long even if we cut the sentence after "ecosystems").

javierm (commented):
Here's an initial review. Let me know what you think!

app/components/sdg/goals/targets_component.html.erb
```diff
@@ -0,0 +1,23 @@
...
+    </li>
+    <li class="tabs-title">
+      <%= link_to SDG::LocalTarget.model_name.human(count: :other).upcase_first, "#tab_local_targets" %>
+    </li>
+  </ul>
+
+  <div class="tabs-content" data-tabs-content="target_tabs">
+    <% [global_targets, local_targets].each do |targets| %>
+      <div class="tabs-panel <%= active(targets) %>" id="tab_<%= type(targets) %>_targets">
+        <dl>
```
javierm, line 14: Right now, when there are no local targets, we just get blank content when clicking on the "local targets" link. How about displaying a message in this case, like we do in other sections (milestones and proposal notifications in a proposal's view come to mind).

javierm (commented):
I think once we deal with one possible usability issue, we're good to go.

app/components/sdg/goals/targets_component.html.erb
```diff
@@ -0,0 +1,29 @@
...
+      <%= link_to SDG::LocalTarget.model_name.human(count: :other).upcase_first, "#tab_local_targets" %>
+    </li>
+  </ul>
+
+  <div class="tabs-content" data-tabs-content="target_tabs">
+    <% [global_targets, local_targets].each do |targets| %>
+      <div class="tabs-panel <%= active(targets) %>" id="tab_<%= type(targets) %>_targets">
+        <% if targets.blank? %>
+          <div class="callout primary text-center">
+            <%= t("sdg.goals.show.targets.empty_local_targets") %>
```
javierm, line 16: While we sometimes use the `empty_` prefix, I think the `no_` prefix is far more common in the code :thinking:.

config/locales/en/sdg.yml
```diff
@@ -444,6 +444,8 @@ en:
       show:
         read_more: "Read more about %{goal}"
         read_less: "Read less about %{goal}"
+        targets:
+          empty_local_targets: This Goal has no local targets
```
javierm, line 448: Typo.
```suggestion
          empty_local_targets: This goal has no local targets
```

config/locales/es/sdg.yml
```diff
@@ -444,6 +444,8 @@ es:
       show:
         read_more: "Leer más sobre %{goal}"
         read_less: "Leer menos sobre %{goal}"
+        targets:
+          empty_local_targets: Este Objetivo no tiene metas localizadas
```
javierm, line 448: Typo.
```suggestion
          empty_local_targets: Este objetivo no tiene metas localizadas
```

app/components/sdg/goals/targets_component.html.erb
```diff
@@ -0,0 +1,29 @@
...
+    <li class="tabs-title">
+      <%= link_to SDG::LocalTarget.model_name.human(count: :other).upcase_first, "#tab_local_targets" %>
+    </li>
+  </ul>
+
+  <div class="tabs-content" data-tabs-content="target_tabs">
+    <% [global_targets, local_targets].each do |targets| %>
+      <div class="tabs-panel <%= active(targets) %>" id="tab_<%= type(targets) %>_targets">
+        <% if targets.blank? %>
+          <div class="callout primary text-center">
```
javierm, line 15: Use CSS instead of the `text-center` class :pray:. Otherwise this could lead to hilarous code in custom CSS files, like `.text-center { text-align: left }` :sweat_smile:.

I'm fine with either aligning the text to the right or in the center; I'm not sure we are consistent everywhere regarding the alignment of callout items :thinking:.

app/components/sdg/goals/targets_component.html.erb
```diff
@@ -0,0 +1,29 @@
+<section class="targets">
+  <ul class="tabs" id="target_tabs" data-deep-link="true" data-tabs>
+    <li class="tabs-title is-active">
+      <%= link_to SDG::Target.model_name.human(count: :other).upcase_first, "#tab_global_targets" %>
```
javierm, line 4: It looks like we must have lost some changes here :thinking:. Here's the original comment:

> How about extracting SDG::Target.model_name.human(count: 2).upcase_first to a method in the component?
> 
> By the way, use count: :other to avoid possible conflicts in languages with different plurals for different amounts.

app/components/sdg/goals/targets_component.html.erb
```diff
@@ -0,0 +1,29 @@
...
+    </li>
+    <li class="tabs-title">
+      <%= link_to SDG::LocalTarget.model_name.human(count: :other).upcase_first, "#tab_local_targets" %>
```
javierm, line 7: It looks like we must have lost some changes here :thinking:. Here's the original comment:

> How about extracting SDG::Target.model_name.human(count: 2).upcase_first to a method in the component?
> 
> By the way, use count: :other to avoid possible conflicts in languages with different plurals for different amounts.

spec/components/sdg/goals/targets_component_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+
+    expect(page).not_to have_css ".targets"
+    expect(page).not_to have_css "#target_tabs"
+    expect(page).not_to have_css ".tabs-content"
+  end
+
+  it "renders tabs panel" do
+    render_inline component
+
+    expect(page).to have_css ".targets"
```
gabrielbaldao, line 24: You could avoid these expect and use de `and` method.
like 
```
expect(page).to have_css ("".targets").and(have_content("Targets")).and(have_content(""Local targets"))
```

javierm: Thanks for the comment! However, since the "one expectation per line" pattern is used in the rest of the code, here we're using it for consistency.

app/components/sdg/goals/targets_component.rb
```diff
@@ -0,0 +1,41 @@
...
+    def active(targets)
+      "is-active" if targets.model.name == "SDG::Target"
+    end
+
+    def title(targets)
+      if targets.model.name == "SDG::Target"
+        SDG::Target.model_name.human(count: :other).upcase_first
+      else
+        SDG::LocalTarget.model_name.human(count: :other).upcase_first
+      end
```
javierm, line 39:
```suggestion
    targets.model.model_name.human(count: :other).upcase_first
```

taitus (approved):

PR #4334, opened 2021-01-27
New header section on SDG Goals Index

taitus:
## References
Related PR:  #4292

## Objectives
- Add new header section on SDG Goals index page

## Visual Changes
Captura de pantalla 2021-01-28 a las 14 04 16

### DIFF
--- a/spec/components/sdg/goals/index_component_spec.rb
+++ b/spec/components/sdg/goals/index_component_spec.rb
@@ -1,0 +1,25 @@
+require "rails_helper"
+
+describe SDG::Goals::IndexComponent, type: :component do
+  let!(:goals) { SDG::Goal.all }
+  let!(:phases) { SDG::Phase.all }
+  let!(:component) { SDG::Goals::IndexComponent.new(goals, phases) }
+
+  before do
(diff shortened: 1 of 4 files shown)


spec/components/sdg/goals/index_component_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+    render_inline component
+
+    expect(page).to have_css "h1", exact_text: "Sustainable Development Goals"
+  end
+
+  it "renders phases" do
+    render_inline component
+
+    expect(page).to have_content "Sensitization"
+    expect(page).to have_content "Planning"
```
gabrielbaldao, line 22: You could avoid 3 expect and use de `and` method.
like 
```
expect(page).to have_content("Sensitization").and(have_content("Planning")).and(have_content("Monitoring"))
```

javierm: Thanks for the comment! However, since the "one expectation per line" pattern is used in the rest of the code, here we're using it for consistency.

javierm (approved):

PR #4337, opened 2021-01-28
Make target filter change when goal filter changes

javierm:
## Objectives

* Improve the usability of the goal/target filters

### Before these changes

After selecting a goal, the target filter shows all targets

### After these changes

After selecting a goal, the target filter only shows targets for that goal

## Notes

If users want to get records related to, let's say, goal 1 and target 2.2, they will no longer be able to do so. However, we think it's an unlikely use case.


app/assets/javascripts/sdg/sync_goal_and_target_filters.js
```diff
@@ -0,0 +1,28 @@
+(function() {
+  "use strict";
+  App.SDGSyncGoalAndTargetFilters = {
+    sync: function(form) {
+      var goal_filter = form.find("[name*=goal]");
+      var target_filter = form.find("[name*=target]");
+
```
Senen, line 7: I found a case where we are showing all targets when a goal is selected.

1º Go to debates index page
2º Filter by any goal and submit
3º The target selector shows all targets instead of the selected goal ones

We can fix this case by hiding them during initialization:

```
      if (goal_filter.val() > 0) {
        App.SDGSyncGoalAndTargetFilters.enable_goal_options(target_filter, goal_filter.val());
      }
```

Senen (changes requested):
I am not done with the review yet, but in the meantime I found a use case doing manual testing. I left a comment below.

app/assets/javascripts/sdg/sync_goal_and_target_filters.js
javierm: Nice catch! I've added a fix and tests.

Senen (approved):

PR #4325, opened 2021-01-21
Add related list selector component to forms

taitus:
## References
Related PRs: #4305 #4315 

## Objectives
Add SDG::RelatedListSelectorComponent to allow related SDGs and Targets to:
  - public forms: Proposals, Debates, Investments
  - admin forms: Polls, Legislation Processes

## Visual Changes
Captura de pantalla 2021-01-22 a las 10 58 56

## Notes
Pending create a new PR to render all icons from RelatedListSelectorComponent in one line on frontend forms.

PR #4353, opened 2021-02-10
Add SDG help page

taitus:
## References
Related PR: #4325 

## Objectives
Create new SDG help page to render all targets and local targets related by goal.

## Visual Changes
- New help page
Captura de pantalla 2021-02-22 a las 12 32 35

- Add link to related list selector component
Captura de pantalla 2021-02-22 a las 13 07 41

### DIFF
--- a/app/assets/stylesheets/mixins.scss
+++ b/app/assets/stylesheets/mixins.scss
@@ -244,10 +244,12 @@
   width: calc(100% + #{$spacing});
   width: calc(100% + #{$max-spacing});
 
-  > * {
+  > *,
+  > .tabs-title {
     margin-bottom: 0;
     margin-left: $spacing;
     margin-left: $max-spacing;
+    margin-right: 0;
     margin-top: $spacing;
     margin-top: $max-spacing;
   }
@@ -259,6 +261,33 @@
 
 %sdg-goal-list {
   @include sdg-goal-list;
+}
+
+%sdg-goal-list-expanded {
+  $spacing: 0.5%;
+  @include sdg-goal-list($spacing);
+
+  label,
+  li {
+    min-width: $sdg-icon-min-width;
+    width: calc(100% / 17 - #{$spacing});
+
+    &:hover {
+      cursor: pointer;
+    }
+
+    .sdg-goal-icon {
+      width: 100%;
+    }
+
+    a:focus {
+      outline: none;
+
+      img {
+        outline: $outline-focus;
+      }
+    }
+  }
 }
 
 %tags {
--- a/app/assets/stylesheets/sdg/help.scss
+++ b/app/assets/stylesheets/sdg/help.scss
@@ -1,0 +1,14 @@
+.sdg-help-content {
+  @include grid-column-gutter;
+
+  > .tabs {
+    @extend %sdg-goal-list-expanded;
+
+    li {
+
+      &:focus {
+        outline: $outline-focus;
+      }
+    }
+  }
+}
--- a/app/components/sdg/related_list_selector_component.html.erb
+++ b/app/components/sdg/related_list_selector_component.html.erb
@@ -12,7 +12,11 @@
     <%= f.text_field :related_sdg_list,
                       class: "input",
                       placeholder: t("sdg.related_list_selector.placeholder"),
-                      label: t("sdg.related_list_selector.hint"),
+                      hint: sanitize(t("sdg.related_list_selector.hint",
+                                      link: link_to(t("sdg.related_list_selector.help.text"),
+                                                      sdg_help_path,
+                                                      title: t("shared.target_blank"),
+                                                      target: "_blank"))),
                       data: { "suggestions-list": sdg_related_suggestions,
                               "remove-tag-text": t("sdg.related_list_selector.remove_tag") } %>
   </div>
--- a/app/components/sdg/related_list_selector_component.html.erb
+++ b/app/components/sdg/related_list_selector_component.html.erb
@@ -16,7 +16,8 @@
                                       link: link_to(t("sdg.related_list_selector.help.text"),
                                                       sdg_help_path,
                                                       title: t("shared.target_blank"),
-                                                      target: "_blank"))),
+                                                      target: "_blank")),
+                                     attributes: %w[href title target]),
                       data: { "suggestions-list": sdg_related_suggestions,
                               "remove-tag-text": t("sdg.related_list_selector.remove_tag") } %>
   </div>
--- a/app/views/sdg/goals/help.html.erb
+++ b/app/views/sdg/goals/help.html.erb
@@ -1,0 +1,1 @@
+<%= render SDG::Goals::HelpPageComponent.new(@goals) %>
--- a/app/views/shared/_subnavigation.html.erb
+++ b/app/views/shared/_subnavigation.html.erb
@@ -46,7 +46,7 @@
       <li>
         <%= layout_menu_link_to t("layouts.header.sdg"),
                                 sdg_goals_path,
-                                controller_path.split("/").first == "sdg",
+                                controller_path == "sdg/goals" && action_name != "help",
                                 accesskey: "6" %>
       </li>
     <% end %>
--- a/config/locales/en/sdg.yml
+++ b/config/locales/en/sdg.yml
@@ -24,7 +24,7 @@ en:
       goal_2:
         title: "Zero Hunger"
         title_in_two_lines: "Zero\nHunger"
-        description: "Zero Hunger"
+        description: "Zero Hunger."
         long_description: '<p>After decades of steady decline, the number of people who suffer from hunger – as measured by the prevalence of undernourishment – began to slowly increase again in 2015. Current estimates show that <a href="http://www.fao.org/publications/sofi/2020/en/">nearly 690 million people are hungry, or 8.9 percent of the world population</a> – up by 10 million people in one year and by nearly 60 million in five years.</p> <p>The world is not on track to achieve Zero Hunger by 2030. If recent trends continue, the number of people affected by hunger would surpass 840 million by 2030.</p> <p>According to the World Food Programme, <a href="https://www.wfp.org/publications/2020-global-report-food-crises">135 million suffer from acute hunger</a> largely due to man-made conflicts, climate change and economic downturns. The COVID-19 pandemic could now double that number, putting an additional 130 million people at risk of suffering acute hunger by the end of 2020.</p> <p>With more than <a href="https://insight.wfp.org/covid-19-will-almost-double-people-in-acute-hunger-by-end-of-2020-59df0c4a8072">a quarter of a billion people potentially at the brink of starvation</a>, swift action needs to be taken to provide food and humanitarian relief to the most at-risk regions.</p> <p>At the same time,&nbsp;a profound change of the global food and agriculture system is needed if we are to nourish the more than 690 million people who are hungry today – and the&nbsp;<a href="https://www.un.org/development/desa/en/news/population/world-population-prospects-2019.html">additional 2 billion people</a>&nbsp;the world will have by 2050. Increasing agricultural productivity and sustainable food production are crucial to help alleviate the perils of hunger.</p>'
         targets:
           target_2_1:
@@ -441,6 +441,9 @@ en:
             title: "By 2020, enhance capacity-building support to developing countries, including for least developed countries and small island developing States, to increase significantly the availability of high-quality, timely and reliable data disaggregated by income, gender, age, race, ethnicity, migratory status, disability, geographic location and other characteristics relevant in national contexts."
           target_17_19:
             title: "By 2030, build on existing initiatives to develop measurements of progress on sustainable development that complement gross domestic product, and support statistical capacity-building in developing countries."
+      help:
+        title: "Sustainable Development Goals help"
+        description: "You can align your contributions to the community (debates, citizen proposals and investment projects) with the Sustaniable Development Goals 2030 Agenda, this will allow us to have an overview of our contribution to each of the Sustainable Development Goals. Below you can find out about all the goals, their targets and localized targets."
       show:
         read_more: "Read more about %{goal}"
         read_less: "Read less about %{goal}"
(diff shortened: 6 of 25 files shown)


app/views/sdg/goals/help.html.erb
```diff
@@ -0,0 +1,30 @@
...
+        <%= link_to "#goal_#{goal.code}_targets_tab", "aria-selected": goal.code == 1 do %>
+          <% render SDG::Goals::IconComponent.new(goal) %>
+        <% end %>
+      </li>
+    <% end %>
+  </ul>
+
+  <div class="tabs-content" data-tabs-content="help_tabs">
+    <% @goals.each do |goal| %>
+      <div id="goal_<%= goal.code %>_targets_tab" class="tabs-panel tabs-icons <%= is_active?(goal) %>">
```
javierm, line 23: Why the `_targets` in the ID, when it also contains title and description? :thinking:

Is the `tabs-icons` class necessary? :thinking:

app/views/sdg/goals/help.html.erb
```diff
@@ -0,0 +1,30 @@
...
+        <h2><%= goal.code_and_title %></h2>
+        <%= sanitize t("sdg.goals.goal_#{goal.code}.long_description") %>
```
javierm, line 25: How about using just `description` instead of `long_description`? People accessing the help page are usually in the middle of filling in a form, and IMHO the long description is too long and might distract users from the task they're doing.

app/views/shared/_subnavigation.html.erb
```diff
@@ -46,7 +46,7 @@
       <li>
         <%= layout_menu_link_to t("layouts.header.sdg"),
                                 sdg_goals_path,
-                                controller_path.split("/").first == "sdg",
+                                controller_path == "sdg/goals" && action_name != "help",
```
javierm, line 49: Could you expand on the reasons for this change? Should there be an active section in the subnavigation when we're in the SDG help page? If so, why? If not, why not? :smile:

config/locales/en/sdg.yml
```diff
@@ -441,6 +441,9 @@ en:
             title: "By 2020, enhance capacity-building support to developing countries, including for least developed countries and small island developing States, to increase significantly the availability of high-quality, timely and reliable data disaggregated by income, gender, age, race, ethnicity, migratory status, disability, geographic location and other characteristics relevant in national contexts."
           target_17_19:
             title: "By 2030, build on existing initiatives to develop measurements of progress on sustainable development that complement gross domestic product, and support statistical capacity-building in developing countries."
+      help:
+        title: "Sustainable Development Goals help"
+        description: "You can align your contributions to the community (debates, citizen proposals and investment projects) with the Sustaniable Development Goals 2030 Agenda, this will allow us to have an overview of our contribution to each of the Sustainable Development Goals. Below you can find out about all the goals, their targets and localised targets."
```
javierm, line 446: I think we use "localize" instead of "localise" in other places :thinking:, so I'd write it here with a "z" as well for consistency.

spec/system/sdg/goals_spec.rb
```diff
@@ -140,4 +140,25 @@
       end
     end
   end
+
+  describe "Help" do
+    scenario "shows all SDGs targets" do
+      create(:sdg_local_target, code: "15.1.1", title: "SDG local target sample text")
+      visit sdg_help_path
+
+      click_link "7. Affordable and Clean Energy"
```
javierm, line 149: How about an expectation about the initial text, before users click any tab? :thinking:

app/components/sdg/related_list_selector_component.html.erb
```diff
@@ -12,7 +12,12 @@
     <%= f.text_field :related_sdg_list,
                       class: "input",
                       placeholder: t("sdg.related_list_selector.placeholder"),
-                      label: t("sdg.related_list_selector.hint"),
+                      label: sanitize(t("sdg.related_list_selector.hint",
+                                      link: link_to(t("sdg.related_list_selector.help.text"),
+                                                      sdg_help_path,
+                                                      title: t("shared.target_blank"),
+                                                      target: "_blank")),
+                                     attributes: %w(href title target)),
```
javierm, line 20: This doesn't seem to work :thinking:; the link opens in the same window. Could you add a test case to confirm/refute this behavior?

I think the cause is the call to `sanitize` in `ConsulFormBuilder#custom_label`.

By the way, Pronto reports:

> Style/PercentLiteralDelimiters: `%w`-literals should be delimited by `[` and `]`. (https://rubystyle.guide#percent-literal-braces)

app/assets/stylesheets/sdg/help.scss
```diff
@@ -0,0 +1,29 @@
...
+
+      &:focus {
+        outline: $outline-focus;
+      }
+
+      &:hover {
+        cursor: pointer;
+      }
+
+      a[aria-selected=true] img {
```
javierm, line 20: Pronto reports:

> Avoid qualifying attribute selectors with an element

Translation: don't mix `a` with `[aria-selected]` on the same selector.

I'm not so fond of this rule, but since it's there, we might as well follow it until (if) we decide to remove it :thinking:

spec/system/help_page_spec.rb
```diff
@@ -25,4 +25,24 @@
...
+
+    Setting["feature.sdg"] = true
+    visit root_path
+    within("#navigation_bar") do
+      click_link "Help"
+    end
+
+    expect(page).to have_link "Sustainable Development Goals help", href: sdg_help_path
+
+    Setting["feature.sdg"] = nil
```
javierm, line 40: We'we recently had many strange test failures on github actions which might (or might not :sweat_smile:) be associated to accessing the database in the middle of a browser test. How about splitting this test in two in order to avoid this issue?

javierm (commented):
Here's the initial review. Let me know what you think!

config/locales/en/sdg.yml
```diff
@@ -458,7 +461,9 @@ en:
       goal_list: "You can choose one or several SDGs aligned with your %{record}"
       help:
         title: "Which SDGs and targets are aligned with my %{record}?"
-      hint: "You can introduce the code of a specific goal/target or a text to find one"
+        text: "help"
+      hint: You can introduce the code of a specific goal/target or a text to find one. For more information visit the %{link} page.
+      label: You can introduce the code of a specific goal/target or a text to find one
```
javierm, line 466: With the current text, a screen reader would say: "You can introduce the code of a specific goal/target or a text to find one, Edit, You can introduce the code of a specific goal/target or a text to find one. For more information visit the help page.", which IMHO is a bit redundant.

```suggestion
      label: Targets
```

Not sure using just "Targets" is right, though, :thinking:, since this field allows selecting goals as well. Any suggestions?

app/components/sdg/related_list_selector_component.html.erb
```diff
@@ -12,9 +12,16 @@
     <%= f.text_field :related_sdg_list,
                       class: "input",
                       placeholder: t("sdg.related_list_selector.placeholder"),
-                      label: t("sdg.related_list_selector.hint"),
+                      label:  t("sdg.related_list_selector.label"),
```
javierm, line 15: Maybe we can move this translation to `activerecord.yml` (again :sweat_smile:), like it was before [commit ab2b3a267](https://github.com/consul/consul/commit/ab2b3a26716d4ce5987310205dd2b5e1f1290912#diff-0f1a8cdb0736d2fd3a4b2e34d155932946101748715cd2b5b1b5cb5c1dc39488L5) :thinking:. That way we can omit this line.

app/assets/stylesheets/sdg/help.scss
```diff
@@ -0,0 +1,29 @@
+.sdg-help-content {
+  @include grid-column-gutter;
+
+  > .tabs {
+    $spacing: 0.5%;
+
+    li {
+      min-width: $sdg-icon-min-width;
+      width: calc(100% / 17 - #{$spacing});
```
javierm, line 9: This code adds some duplication with the related list selector but does not add the minimum margin rules :thinking:, which is more apparent on mobile phones.

Here's a possible patch:

[sdg_selector.txt](https://github.com/consul/consul/files/6050217/sdg_selector.txt)

Note we need to add the `.tabs-title` selector in order to override rules set by Foundation :relieved:.

app/components/sdg/related_list_selector_component.html.erb
```diff
@@ -12,9 +12,16 @@
...
+                      hint: sanitize(t("sdg.related_list_selector.hint",
+                                      link: link_to(t("sdg.related_list_selector.help.text"),
+                                                      sdg_help_path,
+                                                      title: t("shared.target_blank"),
+                                                      target: "_blank")),
+                                     attributes: %w[href title target]),
                       data: { "suggestions-list": sdg_related_suggestions,
-                              "remove-tag-text": t("sdg.related_list_selector.remove_tag") } %>
+                              "remove-tag-text": t("sdg.related_list_selector.remove_tag") },
+                      "aria-label": t("sdg.related_list_selector.label") %>
```
javierm, line 24: `aria-label` is redundant since we provide a real `label` tag :wink:. Use either one or the other.

spec/system/sdg_management/relations_spec.rb
```diff
@@ -407,6 +407,14 @@
           expect(find("input[data-code='1']")).to be_checked
         end
       end
+
+      scenario "Help page link opens in new window" do
+        process = create(:legislation_process, title: "SDG process")
+
+        visit sdg_management_edit_legislation_process_path(process)
+
+        expect(find_link("help")[:target]).to eq("_blank")
```
javierm, line 416: How about testing it from the user's point of view?

```ruby
        within_window(window_opened_by { click_link "SDG help page" }) do
          expect(page).to have_content "Sustainable Development Goals help"
        end
```
Writing tests from the user's point of view in this case made me realize that maybe the "help" text in the link is not specific enough, so I added a small change:

```diff
diff --git a/config/locales/en/sdg.yml b/config/locales/en/sdg.yml
index 36ac548062..06b7559e65 100644
--- a/config/locales/en/sdg.yml
+++ b/config/locales/en/sdg.yml
@@ -461,8 +461,8 @@ en:
       goal_list: "You can choose one or several SDGs aligned with your %{record}"
       help:
         title: "Which SDGs and targets are aligned with my %{record}?"
-        text: "help"
-      hint: You can introduce the code of a specific goal/target or a text to find one. For more information visit the %{link} page.
+        text: "SDG help page"
+      hint: You can introduce the code of a specific goal/target or a text to find one. For more information visit the %{link}.
       label: You can introduce the code of a specific goal/target or a text to find one
       placeholder: "Write a goal or target code or description"
       remove_tag: "Remove"
diff --git a/config/locales/es/sdg.yml b/config/locales/es/sdg.yml
index 304a4d8af4..8945e4acb3 100644
--- a/config/locales/es/sdg.yml
+++ b/config/locales/es/sdg.yml
@@ -461,8 +461,8 @@ es:
       goal_list: "Puedes seleccionar uno o varios ODS con los que se alinea tu %{record}"
       help:
         title: "¿Qué ODS y metas se alinean con mi %{record}?"
-        text: "ayuda"
-      hint: Puedes introducir el código de un objetivo/meta específico o un texto para encontrar uno. Para más información visita la página de %{link}.
+        text: "página de ayuda de ODS"
+      hint: Puedes introducir el código de un objetivo/meta específico o un texto para encontrar uno. Para más información visita la %{link}.
       label: Puedes introducir el código de un objetivo/meta específico o un texto para encontrar uno
       placeholder: "Escribe las etiquetas que desees"
       remove_tag: "Eliminar"
```

javierm (commented):
Second review :smile:. Let me know what you think!

spec/system/help_page_spec.rb
```diff
@@ -25,4 +25,26 @@
       expect(page).not_to have_link "Help"
     end
   end
+
+  scenario "renders SDG help page link when feature is enabled" do
+    Setting["feature.help_page"] = true
+
+    Setting["feature.sdg"] = true
+    visit root_path
```
javierm, line 33: Maybe we can group the setup together and leave a blank line after it:

```suggestion
    Setting["feature.help_page"] = true
    Setting["feature.sdg"] = true
    
    visit root_path
```

spec/system/help_page_spec.rb
```diff
@@ -25,4 +25,26 @@
       expect(page).not_to have_link "Help"
     end
   end
+
+  scenario "renders SDG help page link when feature is enabled" do
```
javierm, line 29: Typo.

```suggestion
  scenario "renders the SDG help page link when the feature is enabled" do
```

spec/system/sdg/goals_spec.rb
javierm: Unresolving; what I meant was an expectation about the initial `h2` :wink:.

By the way, the test would pass if we displayed the information for all 17 goals all the time, right? IMHO we should also test that after clicking a link, the previous `h2` is not available anymore :thinking:.

spec/system/sdg/goals_spec.rb
```diff
@@ -140,4 +140,27 @@
...
+      expect(page).to have_content "Ensure access to affordable, reliable, sustainable and modern energy."
+      expect(page).to have_content "7.1 By 2030, ensure universal access to affordable"
+
+      click_link "15. Life on Land"
+      expect(page).to have_css "h2", exact_text: "15. Life on Land"
+      expect(page).to have_content "Sustainably manage forests, combat desertification, halt and reverse"
+      expect(page).to have_content "15.1 By 2020, ensure the conservation, restoration and sustainable use"
+
+      click_link "Local targets"
+      expect(page).to have_content "SDG local target sample text"
```
javierm, line 163:
```suggestion
      expect(page).to have_content "SDG local target sample text"
      expect(page).not_to have_content "15.1 By 2020, ensure the conservation, restoration and sustainable use"
```

app/assets/stylesheets/mixins.scss
```diff
@@ -244,10 +244,12 @@
...
-  > * {
+  > *,
+  > .tabs-title {
     margin-bottom: 0;
     margin-left: $spacing;
     margin-left: $max-spacing;
+    margin-right: 0;
     margin-top: $spacing;
     margin-top: $max-spacing;
   }
```
javierm, line 255: There seems to be a padding added by Foundation to link tabs which is a bit annoying on `:focus` :relieved:.

```suggestion
  }
  
  a {
    padding: 0;
  }
```

javierm (commented):
@taitus Back to you :wink:.

javierm (approved):

PR #4378, opened 2021-02-27
Fix docker

iceraluk:
Switched to ruby v 2.6.6. as it was raising incompatibility issues with capybara


efgalvao (commented):
I have the same issue, needed to fix this locally.

PR #4383, opened 2021-03-02
Setup mailer default settings on tests

Senen:
So when a fork changes the default mailer settings, the affected specs keep passing.

## References

* Similar to pull request #4281
* Backport from https://github.com/AyuntamientoLorca/consul/pull/9

## Objectives

Allow forks to customize the default mailer settings without changing the specs.

Before this change when a fork overrides the default mailer settings some specs crash.


Senen:
Closing this PR as this work was already done through df280ffbfe1.

PR #4441, opened 2021-03-24
Fix active banners in non-UTC time zones

taitus:
## Objectives
Since the `:post_started_at` and `:post_ended_at` fields are of type `Date`, we check with `Date.current` and not with `Time.current`.

This change has been caused because some test suites were failing when run at midnight. https://github.com/consul/consul/runs/2170798218?check_suite_focus=true

### DIFF
--- a/app/models/banner.rb
+++ b/app/models/banner.rb
@@ -19,9 +19,7 @@ class Banner < ApplicationRecord
   has_many :sections
   has_many :web_sections, through: :sections
 
-  scope :with_active,   -> { where("post_started_at <= ?", Time.current).where("post_ended_at >= ?", Time.current) }
-
-  scope :with_inactive, -> { where("post_started_at > ? or post_ended_at < ?", Time.current, Time.current) }
-
+  scope :with_active, -> { where("post_started_at <= :date and post_ended_at >= :date", date: Date.current) }
+  scope :with_inactive, -> { where.not(id: with_active) }
   scope :in_section, ->(section_name) { joins(:web_sections, :sections).where("web_sections.name ilike ?", section_name) }
 end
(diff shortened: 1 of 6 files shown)


app/models/banner.rb
```diff
@@ -19,9 +19,7 @@ class Banner < ApplicationRecord
   has_many :sections
   has_many :web_sections, through: :sections
 
-  scope :with_active,   -> { where("post_started_at <= ?", Time.current).where("post_ended_at >= ?", Time.current) }
-
-  scope :with_inactive, -> { where("post_started_at > ? or post_ended_at < ?", Time.current, Time.current) }
-
+  scope :with_active, -> { where("post_started_at <= ? and post_ended_at >= ?", Date.current, Date.current) }
```
Senen, line 22:
```suggestion
  scope :with_active, -> { where("post_started_at <= :date and post_ended_at >= :date", date: Date.current) }
```

app/models/banner.rb
```diff
@@ -19,9 +19,7 @@ class Banner < ApplicationRecord
...
+  scope :with_inactive, -> { where("post_started_at > ? or post_ended_at < ?", Date.current, Date.current) }
```
Senen, line 23:
```suggestion
  scope :with_inactive, -> { where("post_started_at > :date or post_ended_at < :date", date: Date.current) }
```

Senen (commented):
Hi @taitus, nice catch! 👏🏼 

I added a minor update suggestion. It is not a big deal. We can keep as you wrote it if you do not like the suggestion.

Senen (approved):

javierm (approved):

PR #4442, opened 2021-03-25
Improve user search by email/name

### LINKED ISSUE #2149: Improve user search by email/name
# Current Behaviour
At the Admin panel, inside any Profile section submenu, searching by email/name doesn't find records if we add a trailing or leading space. 

For example on a clean local installation with `rake db:dev_seed` ran to generate dummy data, if you go to http://localhost:3000/admin/moderators and search ` mod@consul.dev` or  `mod@consul.dev ` you won't find any user, only if you use  `mod@consul.dev`

search_spaces

# Expected Behaviour

Searching by email should ignore leading/trailing blank spaces on the text used to search (either email or name)

taitus:
## References

Closes #2149 

## Objectives
Searching by email/name should ignore leading/trailing blank spaces on the text used to search.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -113,8 +113,8 @@ class User < ApplicationRecord
     joins(:comments).where("comments.commentable": commentables).distinct
   end
   scope :by_username_email_or_document_number, ->(search_string) do
-    string = "%#{search_string}%"
-    where("username ILIKE ? OR email ILIKE ? OR document_number ILIKE ?", string, string, string)
+    search = "%#{search_string.strip}%"
+    where("username ILIKE ? OR email ILIKE ? OR document_number ILIKE ?", search, search, search)
   end
   scope :between_ages, ->(from, to) do
     where(
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -302,7 +302,7 @@ class User < ApplicationRecord
   end
 
   def self.search(term)
-    term.present? ? where("email = ? OR username ILIKE ?", term, "%#{term}%") : none
+    term.present? ? where("email = ? OR username ILIKE ?", term.strip, "%#{term}%") : none
   end
 
   def self.username_max_length
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -302,7 +302,10 @@ class User < ApplicationRecord
   end
 
   def self.search(term)
-    term.present? ? where("email = ? OR username ILIKE ?", term.strip, "%#{term}%") : none
+    return none if term.blank?
+
+    search = term.strip
+    where("email = ? OR username ILIKE ?", search, "%#{search}%")
   end
 
   def self.username_max_length
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -440,6 +440,14 @@ describe User do
       expect(search).to eq [user1]
     end
 
+    it "find users by name with whitespaces" do
+      user1 = create(:user, username: "Larry Bird")
+      create(:user, username: "Robert Parish")
+      search = User.search(" larry ")
+
+      expect(search).to eq [user1]
+    end
+
     it "returns no results if no search term provided" do
       expect(User.search("    ")).to be_empty
     end
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -416,36 +416,26 @@ describe User do
   end
 
   describe "self.search" do
-    it "find users by email" do
-      user1 = create(:user, email: "larry@consul.dev")
-      create(:user, email: "bird@consul.dev")
-      search = User.search("larry@consul.dev")
+    describe "find users" do
+      let!(:larry) { create(:user, email: "larry@consul.dev", username: "Larry Bird") }
 
-      expect(search).to eq [user1]
-    end
+      before { create(:user, email: "robert@consul.dev", username: "Robert Parish") }
 
-    it "find users by email with whitespaces" do
-      user1 = create(:user, email: "larry@consul.dev")
-      create(:user, email: "bird@consul.dev")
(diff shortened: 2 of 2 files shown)


app/models/user.rb
```diff
@@ -302,7 +302,7 @@ def locked?
   end
 
   def self.search(term)
-    term.present? ? where("email = ? OR username ILIKE ?", term, "%#{term}%") : none
+    term.present? ? where("email = ? OR username ILIKE ?", term.strip, "%#{term.strip}%") : none
```
Senen, line 305: I know it is not a big deal but I was wondering how to call strip only once. This is the only way a managed to do it:
```
  def self.search(term)
    return none if term.blank?

    term.strip!
    where("email = ? OR username ILIKE ?", term, "%#{term}%")
  end
```

Senen (commented):
I only found a minor thing! Let me know what you think 🙏🏼

app/models/user.rb
javierm: I'm not fond of modifying parameters inside a method call :thinking:. Maybe in this case it's OK, but the scenario could be error-prone in some situations. Consider this console session:

```ruby
term = " abc "
=> " abc "
User.search(term)
term
=> "abc"
```

Very unintuitive IMHO.

Senen: Yes, you are totally right @javierm ☺️ ! Thanks

So, what about this version?
```
  def self.search(term)
    return none if term.blank?

    search = term.strip
    where("email = ? OR username ILIKE ?", search, "%#{search}%")
  end
```

Maybe to keep the duplication is the best option here keeping the one line implementation. 🤔

javierm: No strong preference.

spec/models/user_spec.rb
```diff
@@ -416,20 +416,25 @@
   end
 
   describe "self.search" do
-    it "find users by email" do
-      user1 = create(:user, email: "larry@consul.dev")
-      create(:user, email: "bird@consul.dev")
-      search = User.search("larry@consul.dev")
+    describe "find users" do
+      let(:larry) { create(:user, email: "larry@consul.dev", username: "Larry Bird") }
```
javierm, line 420: Note this test passes because the user is only created when the `larry` is accessed.

```ruby
      it "by email" do
        expect(User.search("larry@consul.dev")).to eq []
        expect(User.search("larry@consul.dev")).to eq [larry]
      end
```

Here's a fix:

```suggestion
      let!(:larry) { create(:user, email: "larry@consul.dev", username: "Larry Bird") }
```

spec/models/user_spec.rb
```diff
@@ -416,20 +416,25 @@
...
+      let(:robert) { create(:user, email: "robert@consul.dev", username: "Robert Parish") }
```
javierm, line 421: Since `robert` is never used:

```suggestion
      before { create(:user, email: "robert@consul.dev", username: "Robert Parish") }
```

spec/models/user_spec.rb
```diff
@@ -416,20 +416,25 @@
...
-      user1 = create(:user, username: "Larry Bird")
-      create(:user, username: "Robert Parish")
-      search = User.search("larry")
+      it "by name" do
+        expect(User.search("larry")).to eq [larry]
+      end
 
-      expect(search).to eq [user1]
+      it "by name with whitespaces" do
+        expect(User.search(" larry")).to eq [larry]
```
javierm, line 436: How about testing trailing whitespace as well?

```suggestion
        expect(User.search(" larry ")).to eq [larry]
```

javierm (changes requested):
Just realized this works properly on most sections:

Searching admnistrators named "admin " with trailing finds the "admin" user

However, it doesn't work in the users section:

Searching users named "admin " with trailing whitespaces doesn't find any users

I think the reason is the user section uses the `by_username_email_or_document_number` method instead of the `search` method :thinking:.

spec/models/user_spec.rb
```diff
@@ -416,27 +416,67 @@
...
+      end
     end
 
     it "returns no results if no search term provided" do
       expect(User.search("    ")).to be_empty
     end
   end
 
+  describe "scope" do
+    describe ".by_username_email_or_document_number" do
```
javierm, line 447: How about moving this part inside the `describe "scopes" do` (line 384)?

app/models/user.rb
```diff
@@ -113,8 +113,8 @@ class User < ApplicationRecord
     joins(:comments).where("comments.commentable": commentables).distinct
   end
   scope :by_username_email_or_document_number, ->(search_string) do
-    string = "%#{search_string}%"
-    where("username ILIKE ? OR email ILIKE ? OR document_number ILIKE ?", string, string, string)
+    search = "%#{search_string.strip}%"
+    where("username ILIKE :search OR email ILIKE :search OR document_number ILIKE :search", search: search)
```
javierm, line 117: Since we leave the `?` style in the `.search` method, how about using the `?` style here as well? I'm suggesting this style because it's the one used everywhere else; I don't have a strong prefence for one style or the other, and when in doubt I prefer consistency.

spec/models/user_spec.rb
```diff
@@ -416,27 +416,67 @@
...
+      let!(:larry) do
+        create(:user, email: "larry@consul.dev", username: "Larry Bird", document_number: "12345678Z")
+      end
+
+      before { create(:user, email: "robert@consul.dev", username: "Robert Parish") }
+
+      it "find users by email" do
```
javierm, line 454: Typo: use the third person after an `it` :wink:. Otherwise reading this line feels strange.

```suggestion
      it "finds users by email" do
```

Same in the other tests.

javierm (approved):

PR #4446, opened 2021-03-28
Update Gemfile.lock

paolafedou:
## References

> Related Issues/Pull Requests/errors/etc...

## Objectives

## Visual Changes

## Notes


javierm:
@paolafedou I think you made a request against the wrong repository :thinking:.

By the way, you've uploaded a SSH private key. If you use the public key for anything, now everybody who has downloaded the private key can impersonate you, so if that's the case I suggest you delete the public key from every server where it's installed.

paolafedou:
@javierm Yes!! Sorry for that !!! And thank you for the info.

PR #4450, opened 2021-03-29
Fix admin notification with relative URLs

javierm:
## Background

We had a test checking admin notifications worked with relative URLs. Howeter, the test was passing because it was using the rack driver, but on real browsers it wasn't generating the expected URL.

## Objectives

Fix wrong URL being generated on admin notifications with relative URLs.


Senen (approved):
Nice catch! I was able to reproduce de error locally and check the commit fixes it! Although, it seems we have two flakes here.

We also have to rebase with the base branch as it changed recently.

PR #4463, opened 2021-04-05
Expire cache when users follow/unfollow

### LINKED ISSUE #4224: Follow proposal button not indicating actual follow state
### Description

On Staging and Production, the "Follow citizen proposal" button seems to be bugged. When a user clicks the button and refreshes the page, the buttons shows as if the user is not following the proposal. The "follow" does get created and stored in the database, though, and the js refresh as well as the delete work properly. The same is true for the "Follow investment" button.
Also, this does not occur on my local development environment, even though the code is the same, so I assume the `def find_or_build_follow(user, followable)` method always searches the development database or something like that.

### Current Behavior
When a user follows a proposal and refreshes the page, the button looks like the user is not following the proposal.

### Expected Behavior
When a user follows a proposal and refreshes the page, the button should look like the user is following the proposal.

### Steps to reproduce
1. Go to any proposal or investment on a production environment, for example: https://demo.consulproject.org/proposals/5-prevent-supermarkets-from-wasting-food
2. Follow the proposal by clicking "Follow citizen proposal".
3. Refresh the page.

javierm:
## References

* Closes #4224

## Objectives

* Make sure we display the right button to follow/unfollow proposals or budget investments after users reload the page
* Make sure users cannot delete other people's follows
* Reduce the number of security warnings reported by brakeman

### DIFF
--- a/app/controllers/follows_controller.rb
+++ b/app/controllers/follows_controller.rb
@@ -3,13 +3,12 @@ class FollowsController < ApplicationController
   load_and_authorize_resource
 
   def create
-    @follow = current_user.follows.create!(follow_params)
+    @follow.save!
     flash.now[:notice] = t("shared.followable.#{followable_translation_key(@follow.followable)}.create.notice")
     render :refresh_follow_button
   end
 
   def destroy
-    @follow = Follow.find(params[:id])
     @follow.destroy!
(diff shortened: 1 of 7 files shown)


Senen (approved):
The PR #4456 fixed the failing spec, so I think we are good to go.

PR #4465, opened 2021-04-05
Restore all related content along when a user is restored

### LINKED ISSUE #3551: Restore blocked users content

javierm:
## References

* Closes #3551
* Original pull request was #4213


taitus (approved):

javierm:
Thanks @carlosiniesta :smile:.

PR #4488, opened 2021-04-22
Test Pronto changes for Dependabot

javierm:
## Objectives

Test what happens with Pronto after changes to make it compatible with Dependabot pull requests.


javierm:
Everything seems to be working now; closing in favor of #4489.

PR #4501, opened 2021-05-06
Improve budget header

taitus:
## References
- This is part of pull request #4198

## Objectives
- Add title in the budget header
- Not showing the phase description in the budget header
- Allow to define a link (text and url) on budget form for render on the budget header

## Visual Changes
- Frontend
Captura de pantalla 2021-05-06 a las 12 28 41

- Backend
Captura de pantalla 2021-05-06 a las 13 13 37

### DIFF
--- a/app/assets/stylesheets/admin.scss
+++ b/app/assets/stylesheets/admin.scss
@@ -293,6 +293,11 @@ $table-header:   #ecf1f6;
     }
   }
 
+  .form-label {
+    font-weight: bold;
+    margin-bottom: 0;
+  }
+
   .menu.simple {
     margin-bottom: $line-height / 2;
 
--- a/app/assets/stylesheets/participation.scss
+++ b/app/assets/stylesheets/participation.scss
@@ -1132,8 +1132,9 @@
 .budget-header {
   @extend %brand-background;
   margin-top: -$line-height;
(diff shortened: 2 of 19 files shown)


Senen (commented):
Hi @taitus,

I found that the `main_link_text` field is not translatable.

Once I made it translatable, I cannot render it at the desired position because it's not possible to render budget translation fields within different form `translation_fields` blocks. Now, all the fields for a translation have to be rendered within the same `translatable_fields` form block. 

I tried to solve the problem to allow to use of multiple `translations_fields` form blocks for the same translation, and surprisingly it works but is still too experimental, furthermore, there are more design problems derived from this. So, I think the faster option is to move the `main_link_text` to the current `translation_fields` form block. See screenshot below:

Captura de pantalla 2021-05-26 a las 16 19 01

As you can see I had to move the `voting style` and `currency` fields below `main_link` fields.

I uploaded a copy of this branch [here](/rockandror/consul/pull/63) with a commit with possible fixes.

[Here](/rockandror/consul/pull/62) is the experimental PR to allow to have multiple translations_fields blocks for the same object.

How it looks like keeping original field positions
Captura de pantalla 2021-05-26 a las 14 55 22

PR #4507, opened 2021-05-10
Preview list of investments

taitus:
## References
- This is part of pull request #4198

## Objectives
Show a preview list of investments in the budget landing page.

## Visual Changes
Captura de pantalla 2021-05-10 a las 14 06 14


Senen (commented):
Hi @taitus, trying this locally I saw that the investment images does not have the same height in this version. I tried to search for a solution and compared it with the original PR version, but did not find a proper way to achieve this. IMO, I prefer how this version behaves to the original one, as images are responsive and the only disadvantage is that sometimes some investments image and title are not perfectly aligned with the side by side ones. 🤔

Senen (approved):

PR #4516, opened 2021-05-16
Limit Font Awesome imports to reduce CSS size

javierm:
## References

* We added Font Awesome in pull request #3606
* We started using a mixin to add Font Awesome icons in commit a8a79fe97f6

## Background

We were using Font Awesome fonts and selectors to support the browsers which don't support mask images (as the time of writing, about 5% of the browsers). However, we were only importing the selectors in order to extend them. This resulted in our compiled CSS including styles for every Font Awesome icon (currenty, more than a thousand), even if we only use a few of them.

## Objectives

* Make the page load faster by removing unused selectors in the compiled CSS

## Notes

With these changes, the size of the application.css file is reduced from 539KB to 475KB.

## Release notes

:warning: CONSUL installations using Font Awesome HTML classes or extending Font Awesome selectors (like `.far` or `.fa-*`) will now have to add extra `@import` clauses in ther `custom.scss` file to import files like `font-awesome/icons`. Another option for these installations would be to remove these classes/selectors and use our `has-fa-icon` mixin.


Senen (approved):
Nice @javierm!

I checked the difference between the `v1.3.1` and `reduce_font_awesome` branch, and in my case, I found a difference of ~38KB on a production server.

**Before**
Application CSS size before

**After**
Application CSS size after

PR #4533, opened 2021-05-31
Simplify creating single heading budgets

taitus:
## References

* This is part of pull request #4198
* Depends on pull request #4531

## Background

CONSUL offers a flexible system so institutions can split budgets among several budget groups, each one having different headings.

However, most institutions use budgets for only one heading, so they don't need the options to create more groups and headings when creating a budget.

## Objectives

* Make it easier to create budgets with only one heading

## Visual Changes
Captura de pantalla 2021-06-10 a las 22 04 44

Captura de pantalla 2021-06-10 a las 21 59 33

Captura de pantalla 2021-06-10 a las 22 00 04


app/components/admin/budgets/duration_component.rb
```diff
@@ -24,6 +24,6 @@ def formatted_end_date
     end
 
     def formatted_date(time)
-      time_tag(time, format: :datetime)
+      time_tag(time, format: :duration)
```
javierm, line 27: The name `duration` is a bit strange because it could be anything :thinking:. "3 days" is a duration, and so is "from January the 27th to February the 13th".

Any ideas for a different name?

PR #4539, opened 2021-06-07
Fix custom system emails preview

decabeza:
## Objectives
Now if you add a view in `/custom/mailer/` it will not show up in the `admin/system_emails` preview.

This PR:

- Hide the border on system emails preview tables.
- Render the custom mailer views if exists.

### Before
before

### After
after

## Notes

I don't know any other way to check if a custom file exists than using `File.exist?`, or if it's possible to check this with a capybara spec. 😅  

So feel free to modify or refactor my commit. 🙏🏻 😌

### DIFF
--- a/app/assets/stylesheets/admin.scss
+++ b/app/assets/stylesheets/admin.scss
@@ -1205,7 +1205,7 @@ table {
 // 11. Newsletters
 // -----------------
 
-.newsletter-body-content {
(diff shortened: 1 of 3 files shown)


javierm (approved):
Super cool, thanks! :tada:

P.S. I've applied the refactor you mentioned. I don't know how to write a test dealing with custom files either :sweat_smile:.

PR #4542, opened 2021-06-10
Remove Capybara `default_set_options` for non JS system tests

Senen:
## References

* We're getting warnings since pull request #4456

## Background

The next warning is flooding the tests log:

`Options passed to Node#set but the RackTest driver doesn't support any - ignoring`

As the rack test driver does not support any option for the Node#set method, we get rid of all `default_set_options` before each system spec that uses this driver.

From Capybara's docs:
`default_set_options (Hash = {}) - The default options passed to Element#set.`

## Objectives

Do not pass options to rack test driver to get rid of the warning.

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -47,7 +47,6 @@ end
 
 Capybara.exact = true
 Capybara.enable_aria_label = true
-Capybara.default_set_options = { clear: :backspace }
 Capybara.disable_animation = true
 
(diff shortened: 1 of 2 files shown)


javierm (approved):
Awesome!

Funny how I tested this approach while working on pull request #4456 and discarded it because I got an "unknown option" error.

Who knows, maybe when testing it I wrote `clear: backpsace` or something :sweat_smile:.

Seems to be working fine :+1:.

PR #4543, opened 2021-06-10
Make confirm alerts show the triggering action

### LINKED ISSUE #4015: Detailed confirmation texts
### Description
The confirmation alert for any action should specify what the action does in detail. If it just asks "Are you sure?", I will always click yes, because I have no reason to change my mind so suddenly, unless I misclicked. But I would not be able to realize a misclick, because the alert does not tell me what I clicked.

### Current Behavior
Currently, all confirmation alerts just ask "Are you sure?".

### Expected Behavior
Each confirmation alert should have a detailed description about what is going to happen upon confirmation.

### Steps to reproduce
Click on "Hide" and "Hide author" on a comment on a discussion.

rhian-cs:
## References

> Related Issues/Pull Requests/errors/etc...

Resolves #4015

## Objectives

When the user clicks on certain buttons in the app, a confirmation alert is shown. However, this alert only asks the user "Are you sure?", but that doesn't give much information about what he should be sure of.

In this PR, I made the confirmation alerts show the triggering action, with the format:

`Are you sure? %{action}`

For instance, if the user clicks on a 'Hide author' button, the alert now displays 'Hide author: Are you sure?'.

System specs were also updated to check if the confirmation alerts actually showed the triggering action.

## Visual Changes

For example, in a page that has comments:

01-confirmation-screenshot

I didn't include a mobile screenshot because the dev tools still show the alerts as if it was in a computer, but since this is the standard alert, not much has changed.

## Notes

Since this PR is already pretty big, I plan to open another one (if/after this one is merged) to update the other locales too.

### DIFF
--- a/spec/system/admin/valuator_groups_spec.rb
+++ b/spec/system/admin/valuator_groups_spec.rb
@@ -64,10 +64,12 @@ describe "Valuator groups", :admin do
   end
 
   scenario "Delete" do
-    create(:valuator_group)
+    create(:valuator_group, name: "Economy")
 
     visit admin_valuator_groups_path
-    accept_confirm { click_button "Delete" }
+
+    confirmation = "Are you sure? This action will delete \"Economy\" and can't be undone."
+    accept_confirm(confirmation) { click_button "Delete" }
 
     expect(page).to have_content "Valuator group deleted successfully"
     expect(page).to have_content "There are no valuator groups"
--- a/app/components/moderation/shared/index_component.html.erb
+++ b/app/components/moderation/shared/index_component.html.erb
@@ -21,18 +21,18 @@
   <%= submit_tag t("moderation.#{i18n_namespace}.index.block_authors"),
                  name: "block_authors",
                  class: "button hollow alert",
-                 data: { confirm: t("moderation.#{i18n_namespace}.index.confirm") } %>
+                 data: { confirm: t("moderation.actions.confirm_action", action: t("moderation.#{i18n_namespace}.index.block_authors")) } %>
 
   <div class="float-right">
     <%= submit_tag t("moderation.#{i18n_namespace}.index.hide"),
                    name: "hide_#{table_name}",
                    class: "button hollow alert",
-                   data: { confirm: t("moderation.#{i18n_namespace}.index.confirm") } %>
+                   data: { confirm: t("moderation.actions.confirm_action", action: t("moderation.#{i18n_namespace}.index.hide")) } %>
 
     <%= submit_tag t("moderation.#{i18n_namespace}.index.ignore_flags"),
                    name: "ignore_flags",
                    class: "button hollow",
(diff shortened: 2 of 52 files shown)


javierm:
Hi, @rhian-cs :smile:.

Thank you so much for this pull request, and sorry for the late reply! :pray:.

As part of a list of accessibility improvements regarding the links and buttons in the admin section, we implemented the detailed confirmation dialog in the admin section so it includes the name of the thing it acts on. The general message format is: `"Are you sure? %{action} \"%{name}\""`, so for example it would read like `Are you sure? Hide "Send an expedition to the Sun"` for a proposal titled "Send an expedition to the Sun". See commit 6510cb9615db4.

Would it be possible  to update this pull request to use that format in the public sections? If you can, great! :tada:. If you can't, it's OK; we understand we've kept you waiting for a review far too long.

Thanks!

rhian-cs:
Hello @javierm! I'll take another look, this PR was pretty big too.

I hope it doesn't take too long since this branch is pretty outdated as well.

We could perhaps try to redo it or split it into smaller PRs, WDYT?

javierm:
@rhian-cs Thanks! :smile: One pull request is OK, if that's fine with you.

Regarding whether it's better to redo it or not, whatever suits you better works for us :wink:.

javierm:
(I clicked on "close with comment" instead of "comment" by accident; please ignore :relieved:)

rhian-cs:
Hello @javierm!

I've rebased the branch and fixed the failing specs. I also ran all the specs locally. Could you take a look at this and let the CI run the tests?

javierm:
Hi, @rhian-cs :smile:.

This is great, thanks! We'll have a look at this pull request before the end of the year :+1:.

Oh, and thank you so much for taking the time to update the tests :wink:.

spec/system/admin/valuator_groups_spec.rb
```diff
@@ -67,7 +67,10 @@
     create(:valuator_group)
 
     visit admin_valuator_groups_path
-    accept_confirm { click_button "Delete" }
+
+    confirmation = "Are you sure? This action will delete "\
+      "\"#{ValuatorGroup.last.name}\" and can't be undone."
```
javierm, line 72: We run into so-called "flaky tests" (meaning sometimes tests fail with no apparent reason) when running database queries after the first `visit` in the test (see pull request #4475). Here `ValuatorGroup.last` would generate a database query which has a low probability (maybe about 1%?) to generate a `PG::ProtocolViolation` error when running the tests.

How about changing line 67 so we give the group a name like "Early birds"? Then we could write:

```suggestion
      "\Early birds\" and can't be undone."
```

javierm (commented):
@rhian-cs Thank you so much, this is great! :smile: :tada:

I've left a few comments; let me know what you think! And, by all means,  if you don't agree with something I mention, say so! :wink:

javierm (approved):
Thanks! :+1: Really appreciate it :smile:.

PR #4574, opened 2021-06-30
Improve keyboard accessibility in language selector

javierm:
## References

* We started avoiding location changes when selecting an option in pull request #4566
* Depends on pull request #4573
* [WebAIM: JavaScript Jump Menus](https://webaim.org/techniques/forms/controls#javascript)
* [Understanding Success Criterion 3.2.2: On Input](https://www.w3.org/WAI/WCAG21/Understanding/on-input.html)

## Objectives

* Make it easier to switch languages using a keyboard

### Before these changes

The dropdown to select language is active and shows select options using the native browser styles

### After these changes

The dropdown to select language is active and shows select options using custom styles, with a check next to the selected item


javierm:
These changes worsen the experience on mobile browsers, since these browsers use modal dialogs to offer the options available in a `<select>` field. Closing for now. We might use a modal dialog in the future :thinking:.

PR #4577, opened 2021-07-02
Hide SDG columns on admin polls if SDG feature is disabled

decabeza:
## Objectives

Now in `/admin/polls/` and `/admin/legislation/processes` the **Goals** and **Targets** columns appear although the SDG feature is disabled. It can be confusing for administrators because when they edit the poll no SDGs related fields appear.

This PR hide SDG columns on admin polls and processes index if SDG feature is disabled.

### Before 
_(with SDG feature disabled)_
before

### After
_(with SDG feature disabled)_
after


Senen (approved):
👍🏼 ☺️

javierm (approved):
Cool, thanks! :pray:

PR #4587, opened 2021-07-14
Do not show confirm message if user can vote in all headings

decabeza:
## Objectives

In the Participatory Budgets during the "Selecting projects" phase, an alert appears warning the user of the number of headings in which he can support projects. The problem is that this message appears whenever there is more than 1 heading, although the user can select projects in all headings.

support_alert

This PR changes that the message only appears when the user can select in less headings than the maximum number of headings. 

For a group of 4 headings, if the Maximum number of headings in which a user can select projects:

`= 4: the message don't appear.`
`< 4: the message appears.`


javierm (approved):
Makes sense to me; definitely an improvement. Thanks! :tada:

PR #4588, opened 2021-07-14
Fix outline position for header logo

decabeza:
## Objectives

For accessibility issues when a link is in the focus state an outline is shown helping the user to know where he has clicked and also to help when navigating using the keyboard.

This PR fixes the alignment of the main CONSUL logo outline in the header.

### Before
before

### After
after


decabeza:
Already fixed on https://github.com/consul/consul/commit/b4203c07b427f5bc! 😉

PR #4605, opened 2021-08-10
Fix inconsistent order in system email preview

javierm:
## References

* This bug caused a test to fail sometimes; for example in our [test run 1879, job 3](https://github.com/consul/consul/runs/2954630259) or our [test run 2207, job 1](https://github.com/consul/consul/runs/3290963815)

## Objectives

* Fix inconsistent order in system email preview
* Fix invalid HTML in system email preview


Senen (approved):
Nice catch! 👏🏼

Issue #4659, opened 2021-08-13
Error when combining investments advanced search with filtering


microweb10:
### Description

An error is raised when combining advanced search and filtering for budget investments.

### Steps to reproduce

1. Visit the [list of investments](https://demo.consulproject.org/budgets/1/investments?heading_id=1) for a participatory budget.
2. Click on advanced search
3. Fill the field **With the text** with any text
4. Select **Last 24 hours** from the selector **By date**
5. Click button **Filter**
6. Select **Not unfeasible** from the selector **Filtering projects by**

### Screenshot
![Captura de pantalla 2021-08-13 a las 10 54 22](https://user-images.githubusercontent.com/942995/129302852-29ab0123-7e77-4f20-8d82-86820e058bfc.png)

javierm:
Was wondering about how to get this exception when we saw it yesterday in the demo. Thanks for reporting! :pray:

microweb10:
That was actually me testing in the demo 😛

PR #4660, opened 2021-08-13
Avoid error when combining investments advanced search with filters

microweb10:
## References

closes issue https://github.com/consul/consul/issues/4659

## Objectives

Avoid raising an error when combining advanced search with filtering for budget investments.


javierm (approved):
@microweb10 Looks good to me :smile:. Thanks! :pray:

P.S. You might have received a notification about changes done in this pull request; I've only rebased it to the current master branch to make sure we run the latest version of the tests.

PR #4662, opened 2021-08-15
Upgrade Ruby to version 2.7.4

javierm:
## Objectives

* Update Ruby to a more recent version
* Use a version of Ruby which includes version 2 of the Bundler gem so it's compatible with more development/production environments

## Notes

Ruby 2.7 introduced a separation of positional and keyword arguments, which caused warnings on almost every existing gem. This was solved by silencing those warnings in Ruby 2.7.2. However, before merging this pull request we're still going to update as many gems as possible and check the warnings in order to study how hard it will be to upgrade to Ruby 3.


Senen (commented):
Hi, @javierm

As ruby 2.7 already includes bundler version 2 I wonder if we may delete the [line 38](https://github.com/consul/consul/pull/4662/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557L38)  from the Dockerfile. Same with [/bin/setup#L17](https://github.com/consul/consul/blob/master/bin/setup#L17) and [/bin/update#L17](https://github.com/consul/consul/blob/master/bin/update#L17) 🤔 

The Dockerfile seems to work fine when running `docker build -t consul .` without the mentioned line.

javierm:
@Senen It looks like we could have actually done so after upgrading to Ruby 2.6 :thinking:. But, yes, now it's a good moment :+1:.

I've updated the pull request changing the Dockerfile. I'm not so sure about the other two, since those are scripts automatically generated by Rails and use the `--conservative` flag, which means that line won't do anything:

```bash
time gem install bundler --conservative
=> gem install bundler --conservative  0,42s user 0,16s system 114% cpu 0,508 total
```

I'm usually fond of deleting code that does nothing, but since this is an external script and every time we upgrade to a new Rails version we'd have to remember to remove this line, I'd vote for leaving it like that.

Senen (approved):
💎

PR #4538, opened 2021-06-04
replace the word message with notification on proposal notification s…

andyjdavis:
…creens

This PR attempts to resolve issue https://github.com/consul/consul/issues/2709

There is a closed PR on that issue. I have attempted to incorporate the feedback that was provided on that PR.

## References

https://github.com/consul/consul/issues/2709

https://github.com/consul/consul/pull/2721

## Objectives

Clarify language in the UI to highlight the difference between sending a message and a notification.

## Visual Changes

notification1

notification2

I have left the text "Message" above the large text box. That could be "notification text" if we want to entirely eliminate the user of the word message on this screen.

PR #4663, opened 2021-08-17
Fix tests after replacing message with notification

javierm:
## References

* Continues #4538

## Background

It looks like github actions wasn't executed on pull request #4538 and we thought tests had passed.


javierm:
The Docker image failure is not related to this pull request.

PR #4684, opened 2021-08-31
Remove a potentially pre-existing server.pid in entrypoint.sh

raoulkramer:
This command will tigger the script to remove a potentially pre-existing server.pid for Rails. With this change the ops does not have to manually `rm tmp/pids/server.pid` remove the server.pid file.

### DIFF
--- a/scripts/entrypoint.sh
+++ b/scripts/entrypoint.sh
@@ -15,3 +15,6 @@ chgrp -R -h "$USER_GID" "$BUNDLE_PATH"
 
 /usr/bin/sudo -EH -u consul "$@"
 
+# Remove a potentially pre-existing server.pid for Rails.
+rm -f /var/www/consul/tmp/pids/server.pid
+
--- a/scripts/entrypoint.sh
+++ b/scripts/entrypoint.sh
@@ -16,5 +16,12 @@ chgrp -R -h "$USER_GID" "$BUNDLE_PATH"
 /usr/bin/sudo -EH -u consul "$@"
 
 # Remove a potentially pre-existing server.pid for Rails.
-rm -f /var/www/consul/tmp/pids/server.pid
+set -e
+
+if [ -f tmp/pids/server.pid ]; then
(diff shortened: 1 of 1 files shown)


javierm (commented):
@raoulkramer Thanks for this pull request! I've left a couple of questions.

scripts/entrypoint.sh
```diff
@@ -13,5 +13,13 @@ usermod -g "$USER_GID" consul 2> /dev/null
 chown -R -h "$USER_UID" "$BUNDLE_PATH"
 chgrp -R -h "$USER_GID" "$BUNDLE_PATH"
 
+# Remove a potentially pre-existing server.pid for Rails.
+set -e
```
javierm, line 17: Just one more thing (as Columbo used to say :smile:): Is this line necessary? :thinking:

raoulkramer:
mmm :thinking: `set -e` causes the shell to exit if any subcommand or pipeline returns a non-zero status.... I took it over from the docker docs concerning the entrypoint.sh file ... 

`rm -f /var/www/consul/tmp/pids/server.pid` is used **without** `set -e`

`if [ -f tmp/pids/server.pid ]; then rm tmp/pids/server.pid fi` is used **with** `set -e`

So I am not sure ... need to test it . :test_tube:

javierm (approved):

javierm:
OK, I've approved the pull request. If your tests go well, say so and we'll merge it :ok_hand:.

raoulkramer:
> mmm thinking `set -e` causes the shell to exit if any subcommand or pipeline returns a non-zero status.... I took it over from the docker docs concerning the entrypoint.sh file ...
> 
> `rm -f /var/www/consul/tmp/pids/server.pid` is used **without** `set -e`
> 
> `if [ -f tmp/pids/server.pid ]; then rm tmp/pids/server.pid fi` is used **with** `set -e`
> 

@javierm I finished testing. It looks that we need the `set -e` rake will be aborted without it. :boom:

PR #4699, opened 2021-09-14
Remove unused attachments code

javierm:
## References

* Depends on pull request #4598

## Objectives

* Make it easier to migrate to Active Storage
* Remove unused code

### DIFF
--- a/app/models/concerns/galleryable.rb
+++ b/app/models/concerns/galleryable.rb
@@ -4,9 +4,5 @@ module Galleryable
   included do
     has_many :images, as: :imageable, inverse_of: :imageable, dependent: :destroy
     accepts_nested_attributes_for :images, allow_destroy: true, update_only: true
-
-    def image_url(style)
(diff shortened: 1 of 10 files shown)


app/helpers/images_helper.rb
```diff
@@ -3,7 +3,7 @@ def image_absolute_url(image, version)
     return "" unless image
 
     if Paperclip::Attachment.default_options[:storage] == :filesystem
-      URI(request.url) + image.attachment.url(version)
+      URI(request.url) + image.url(version)
     else
       investment.image_url(version)
```
Senen, line 8: This line is broken as there is no investment defined within this context. This throws always an exception when using a paperclip remote storage.

```suggestion
      image.url(version)
```

Senen (commented):
Hi @javierm, I placed a comment in the code. Not sure if we would prefer to include this fix in a separate PR.

Senen (approved):

PR #4701, opened 2021-09-20
Fix social share image URL with external storage

### LINKED ISSUE #2850: Error 500 when creating a budget investment with S3

javierm:
## References

* Closes #2850

## Objectives

* Fix a bug causing a 500 error when storing images using external service and rendering the link to share an investment with an image in social media

### DIFF
--- a/app/helpers/images_helper.rb
+++ b/app/helpers/images_helper.rb
@@ -5,7 +5,7 @@ module ImagesHelper
     if Paperclip::Attachment.default_options[:storage] == :filesystem
       URI(request.url) + image.attachment.url(version)
     else
-      investment.image_url(version)
+      image.attachment.url(version)
     end
   end
 


javierm (commented):
@Senen I can't approve the pull request because I've opened it (even if the commit is yours). Can you approve it yourself?

Senen (approved):

PR #4702, opened 2021-09-21
Add sanitize to description on moderator activity

taitus:
## Objectives
Currently html tags were being displayed in the description.
We used the sanitize method to not display them.

## Visual Changes
- Before
Captura de pantalla 2021-09-21 a las 12 08 41

- After
Captura de pantalla 2021-09-21 a las 12 09 52

### DIFF
--- a/spec/system/admin/activity_spec.rb
+++ b/spec/system/admin/activity_spec.rb
@@ -9,7 +9,7 @@ describe "Admin activity" do
 
   context "Proposals" do
     scenario "Shows moderation activity on proposals" do
-      proposal = create(:proposal)
+      proposal = create(:proposal, description: "<p>Description with html tag</p>")
 
       visit proposal_path(proposal)
 
@@ -24,6 +24,7 @@ describe "Admin activity" do
         expect(page).to have_content(proposal.title)
         expect(page).to have_content("Hidden")
         expect(page).to have_content(admin.user.username)
+        expect(page).to have_css("p", exact_text: "Description with html tag")
       end
     end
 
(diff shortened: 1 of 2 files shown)


javierm (changes requested):
@taitus Add a test, please :pray:.

spec/system/admin/activity_spec.rb
```diff
@@ -373,4 +373,24 @@
...
+    within("#proposal_#{proposal.id}") do
+      accept_confirm { click_link "Hide" }
+    end
+    expect(page).to have_css("#proposal_#{proposal.id}.faded")
+
+    visit admin_activity_path
+
+    within first("tbody tr") do
+      expect(page).to have_content("Make Pluto a planet again")
+      expect(page).to have_content("Description with html tag")
```
javierm, line 392: This test passes with either `sanitize` or `strip_tags` in the view :thinking:. How about:

```suggestion
      expect(page).to have_css("p", exact_text: "Description with html tag")
```

spec/system/admin/activity_spec.rb
```diff
@@ -373,4 +373,24 @@
       end
     end
   end
+
+  scenario "Shows moderation activity without html tags in description" do
+    proposal = create(:proposal, title: "Make Pluto a planet again",
+                                 description: "<p>Description with html tag</p>")
+
+    visit proposal_path(proposal)
```
javierm, line 381: Since this test is almost identical to an existing one, what do you think about merging them together? :thinking: That way we'd make the tests faster.

If you'd rather keep the tests separate, then how about removing the part with the first `vistit` and clicking on "Hide"? That part is already tested in the other test, and so in this one we could setup the database and visit the `admin_activity_path` directly, saving a couple of requests.

javierm (commented):
@taitus I've left a couple of comments about the test. Let me know what you think! :wink:

javierm (approved):

Issue #2756, opened 2018-07-11
Adds proposals images on dev_seeds


decabeza:
Background
==
The user can upload an image when creates a proposal and this image is shown on the index list in `/proposals`. 

What we need
==

Include some random images on proposals created on `dev_seeds/proposals`

Notes
==

You can use any random image from https://unsplash.com/ (please avoid using photos of people faces or any inappropriate image)

Screenshots
==
![proposals_image](https://user-images.githubusercontent.com/631897/42566743-dac4bc1e-8507-11e8-8edd-dd62222dc680.png)

nando:
@decabeza i'll take this issue, is that ok?

voodoorai2000:
Awesome @nando!
All yours! 👍

decabeza:
Closed by https://github.com/consul/consul/pull/2768

PR #4703, opened 2021-09-22
Wait for suggestions to finish loading in tests

javierm:
## References

* Some tests were hanging indefinitely in pull request #4598
* There were some timeouts in Github Actions due to this, like [test run #2741, build 3](https://github.com/consul/consul/runs/3656948132), [test run #2752, build 1](https://github.com/consul/consul/runs/3664024045), [test run #2754, build 3](https://github.com/consul/consul/runs/3668786453), [test run #2755, build 1](https://github.com/consul/consul/runs/3668791332) or [test run #2756, build 1](https://github.com/consul/consul/runs/3669379951).

## Background

Sometimes tests were hanging indefinitely. Debugging shows that in some cases it's due to submitting a form before the AJAX request to get proposals, debates or investments suggestions is finished, since having an AJAX and a non-AJAX request at the same time when running the tests sometimes leads to unexpected results.

In our case, we were having many timeouts in Github Actions in the branches where we use both ActiveStorage and Paperclip to store files (based on pull request #4598). I can reproduce it in those branches running the following test ("Should show new image after successful creation with one uploaded file"), although only when my laptop isn't plugged (!!):

```
rspec './spec/system/proposals_spec.rb[1:33:1:14]'
```

## Objectives

* Avoid having one AJAX and one regular request in tests at the same time, which might cause the test to hang forever

### DIFF
--- a/app/assets/javascripts/suggest.js
+++ b/app/assets/javascripts/suggest.js
@@ -6,6 +6,14 @@
         var $this, callback, timer;
         $this = $(this);
         callback = function() {
+          var js_suggest_selector, locale;
+          js_suggest_selector = $this.data("js-suggest");
+
+          if (js_suggest_selector.startsWith(".")) {
+            locale = $this.closest(".translatable-fields").data("locale");
+            js_suggest_selector += "[data-locale=" + locale + "]";
+          }
+
           $.ajax({
             url: $this.data("js-url"),
             data: {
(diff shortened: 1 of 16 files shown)


app/assets/javascripts/suggest.js
```diff
@@ -6,22 +6,27 @@
...
+
           $.ajax({
             url: $this.data("js-url"),
             data: {
               search: $this.val()
             },
             type: "GET",
             dataType: "html",
+            beforeSend: function() {
+              $(js_suggest_selector).removeClass("success").addClass("loading");
```
Senen, line 25: Hi @javierm, as the CSS classes `success` and `loading` are for testing purposes and are quite generic class names. What do you think about adding the following prefixes `js-success`, `suggest-success` or maybe `test-success`? I can imagine forks with custom code where CSS styles are defined for the `.success` or `.loading` classes, which could break the functionality or change the suggestions styles unintentionally.

Is the `loading` class necessary, or maybe it is added for javascript clarity? Seems fine to me, just curious.

Senen (commented):
Hi @javierm, I can't reproduce the error locally even with my laptop unplugged after some tryouts. It works nice. I left a comment about the CSS class names.

Senen (approved):

app/assets/javascripts/suggest.js
javierm: I've changed them to `suggest-succes` and `suggest-loading` :+1:.

> Is the loading class necessary, or maybe it was added for javascript clarity?

Just for clarity. I thought it was a bit strange to keep the suggest class while new suggestions were loading (would happen in a second request if we didn't remove the class en `beforeSend`), and then I thought we could differentiate the case :thinking:.

Senen: Thanks for the clarification and for the commit message update 🎖️ !

PR #4720, opened 2021-10-25
Release version 1.4.0

javierm:
##  Objectives

Release CONSUL version 1.4.0

## Notes

* [Changelog](https://github.com/consul/consul/blob/master/CHANGELOG.md#140-2021-11-24)
* [Release Notes](https://github.com/consul/consul/releases/tag/1.4.0)
* Community Post TODO

Release Notes for when we create the new tag 1.4.0:

- Limit Font Awesome imports to reduce CSS size #4516
:warning: CONSUL installations using Font Awesome HTML icon classes or extending Font Awesome selectors (like `.fa-icon_name`) will now have to add extra `@import` clauses in ther `custom.scss` file to import files like `font-awesome/icons`. Another option for these installations would be to remove these classes/selectors and use our `has-fa-icon` mixin.

- Store files with both Paperclip and ActiveStorage #4598
:warning: If you've added custom attachments, make sure to change these attachments so they use  with `has_attachment` instead of `has_attached_file`. Otherwise you might run into missing files when upgrading to version 1.5.0.

  In version 1.5.0 we'll change the way we store attachments from Paperclip to Active Storage. In order to make the migration painless, version 1.4.0 stores new files using both systems. If you've changed anything related to the way attachments are handled, there's a chance you might run into some issues when executing `consul:execute_release_tasks` (which includes a task to copy existing files to Active Storage). In version 1.5.0 we'll drop Paperclip and use just Active Storage, so make sure to fix the issues or report them; otherwise you might run into missing files when upgrading to version 1.5.0.

  After copying the existing files to Active Storage, **keep both the Active Storage files and the Paperclip files**. The Paperclip files are needed in version 1.4.0, and the Active Storage files will be needed in version 1.5.0.

  If you're storing files using an external service (such as S3), make sure to edit the `config/storage.yml` file and configure the service. Sample configurations are provided for S3, GCS and Azure; uncomment them where appropriate and then edit your `config/secrets.yml` file with your credentials.


Senen (approved):
Hi @javierm, I checked the CHANGELOG to see if we have included all the merged PR's since the 1.3.1 tag, and it seems to be okay. We have 189 new PR for the new release 👏🏼 .

PR #4510, opened 2021-05-11
Improve budget phases

taitus:
## References
- This is part of pull request #4198

## Objectives
- Allow attaching an image to budget phases
- Add main link to each phase of the budget

## Visual Changes
- Attaching an image Backend
Captura de pantalla 2021-05-11 a las 16 49 26

- Attaching an image Frontend
Captura de pantalla 2021-05-11 a las 16 50 09

- Add main link Backend
Captura de pantalla 2021-05-11 a las 16 51 04

- Add main link Frontend
Captura de pantalla 2021-05-11 a las 16 51 33

## Notes
Merge after #4502

PR #4726, opened 2021-11-03
Translate budget and budget phase main link url

Senen:
## References

Related PR's: 
 * Improve budget header #4501
 * Improve budget phases #4510 

## Objectives

Make the `main_link_url` field translatable to store different destinations for each language, particularly useful when the destination page is available in many languages.

## Visual Changes

**Before: Budget edition page**
Captura de pantalla 2021-11-03 a las 16 20 39-fullpage

**Before: Budget phase edition page**
Captura de pantalla 2021-11-03 a las 16 20 49-fullpage

**After: Budget edition page**
Captura de pantalla 2021-11-04 a las 11 18 15-fullpage

**After: Budget phase edition page**
Captura de pantalla 2021-11-04 a las 11 20 47-fullpage

## Notes

I omitted the data migration process from non-translatable fields to the new translatable ones because the related PR's were not released yet.

### DIFF
--- a/app/components/admin/budget_phases/form_component.html.erb
+++ b/app/components/admin/budget_phases/form_component.html.erb
@@ -49,11 +49,11 @@
       <p class="help-text"> <%= t("admin.budget_phases.edit.main_call_to_action_description") %></p>
       <%= translations_form.text_field :main_link_text %>
     </div>
-  <% end %>
 
-  <div class="small-12 column">
-    <%= f.text_field :main_link_url, placeholder: t("admin.shared.example_url") %>
-  </div>
+    <div class="small-12 column">
+      <%= translations_form.text_field :main_link_url, placeholder: t("admin.shared.example_url") %>
+    </div>
+  <% end %>
 
   <% if feature?(:allow_images) %>
     <div class="images small-12 column">
--- a/app/components/admin/budget_phases/form_component.html.erb
+++ b/app/components/admin/budget_phases/form_component.html.erb
@@ -51,7 +51,7 @@
     </div>
 
     <div class="small-12 column">
-      <%= translations_form.text_field :main_link_url, placeholder: t("admin.shared.example_url") %>
+      <%= translations_form.text_field :main_link_url, hint: t("admin.shared.example_url") %>
     </div>
   <% end %>
 
--- a/app/components/admin/budgets/form_component.html.erb
+++ b/app/components/admin/budgets/form_component.html.erb
@@ -21,13 +21,12 @@
           <%= translations_form.text_field :main_link_text %>
         </div>
       </div>
-    <% end %>
-
-    <div class="row expanded">
-      <div class="small-12 medium-9 large-6 column end">
-        <%= f.text_field :main_link_url, placeholder: t("admin.shared.example_url") %>
+      <div class="row expanded">
+        <div class="small-12 medium-9 large-6 column end">
+          <%= translations_form.text_field :main_link_url, placeholder: t("admin.shared.example_url") %>
+        </div>
       </div>
-    </div>
+    <% end %>
 
     <div class="row expanded">
       <div class="small-12 medium-4 column">
--- a/app/components/admin/budgets/form_component.html.erb
+++ b/app/components/admin/budgets/form_component.html.erb
@@ -21,9 +21,10 @@
           <%= translations_form.text_field :main_link_text %>
         </div>
       </div>
+
       <div class="row expanded">
         <div class="small-12 medium-9 large-6 column end">
-          <%= translations_form.text_field :main_link_url, placeholder: t("admin.shared.example_url") %>
+          <%= translations_form.text_field :main_link_url, hint: t("admin.shared.example_url") %>
         </div>
       </div>
     <% end %>
--- a/app/models/budget.rb
+++ b/app/models/budget.rb
@@ -5,7 +5,7 @@ class Budget < ApplicationRecord
   include Reportable
   include Imageable
 
-  translates :name, :main_link_text, touch: true
+  translates :name, :main_link_text, :main_link_url, touch: true
   include Globalizable
 
   class Translation
@@ -24,11 +24,11 @@ class Budget < ApplicationRecord
   VOTING_STYLES = %w[knapsack approval].freeze
 
   validates_translation :name, presence: true
+  validates_translation :main_link_url, presence: true, unless: -> { main_link_text.blank? }
   validates :phase, inclusion: { in: Budget::Phase::PHASE_KINDS }
   validates :currency_symbol, presence: true
   validates :slug, presence: true, format: /\A[a-z0-9\-_]+\z/
   validates :voting_style, inclusion: { in: VOTING_STYLES }
-  validates :main_link_url, presence: true, if: -> { main_link_text.present? }
 
   has_many :investments, dependent: :destroy
   has_many :ballots, dependent: :destroy
--- a/app/models/budget/phase.rb
+++ b/app/models/budget/phase.rb
@@ -9,6 +9,7 @@ class Budget
     translates :summary, touch: true
     translates :description, touch: true
     translates :main_link_text, touch: true
+    translates :main_link_url, touch: true
     include Globalizable
     include Sanitizable
     include Imageable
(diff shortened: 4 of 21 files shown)


app/models/budget/phase.rb
```diff
@@ -19,9 +20,9 @@ class Phase < ApplicationRecord
 
     validates_translation :name, presence: true
     validates_translation :description, length: { maximum: DESCRIPTION_MAX_LENGTH }
+    validates_translation :main_link_url, presence: { if: -> { main_link_text.present? }}
```
javierm, line 23:
```suggestion
    validates_translation :main_link_url, presence: true, unless: -> { main_link_text.blank? }
```

For the same reasons mentioned in the comment about the budget model.

config/locales/en/activerecord.yml
```diff
@@ -160,9 +160,9 @@ en:
         voting_style: "Final voting style"
```
javierm, line 160: We can apply the changes on this file to `es/activerecord.yml` as well :+1:.

db/migrate/20211103090940_translate_budget_main_link_url.rb
```diff
@@ -0,0 +1,6 @@
+class TranslateBudgetMainLinkUrl < ActiveRecord::Migration[5.2]
+  def change
+    add_column :budget_translations, :main_link_url, :string
+    remove_column :budgets, :main_link_url
```
javierm, line 4:
```suggestion
    remove_column :budgets, :main_link_url, :string
```

This way we make the migration reversible :wink:.

db/migrate/20211103112944_translate_budget_phase_main_link_url.rb
```diff
@@ -0,0 +1,6 @@
+class TranslateBudgetPhaseMainLinkUrl < ActiveRecord::Migration[5.2]
+  def change
+    add_column :budget_phase_translations, :main_link_url, :string
+    remove_column :budget_phases, :main_link_url
```
javierm, line 4:
```suggestion
    remove_column :budget_phases, :main_link_url, :string
```

This way we make the migration reversible :wink:.

spec/models/concerns/globalizable.rb
```diff
@@ -11,6 +11,15 @@
...
   let(:attribute) { required_fields.sample || fields.sample }
+  let(:invalid_translation_attributes) do
+    Hash.new.tap do |translation_attributes|
+      translation_attributes[:id] = record.translations.find_by(locale: :es).id
+      translation_attributes[attribute] = ""
+      if attribute == :main_link_url && [:budget, :budget_phase].include?(factory_name)
+        translation_attributes[:main_link_text] = "Enlace"
+      end
+    end
+  end
```
javierm, line 22: Why are these changes necessary? :thinking:

javierm (changes requested):
Awesome! :tada: I've left some comments; let me know what you think!

app/models/budget.rb
```diff
@@ -24,11 +24,11 @@ def name_uniqueness_by_budget
   VOTING_STYLES = %w[knapsack approval].freeze
 
   validates_translation :name, presence: true
+  validates_translation :main_link_url, presence: { if: -> { main_link_text.present? }}
```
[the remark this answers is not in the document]
Senen: Yes, this way to define validator options was weird for me too, but [it seems to be a valid way to define validators] particularly useful if you want to define more than one validator for a field in one line. (https://github.com/rails/rails/blob/v5.2.6/activemodel/lib/active_model/validations/validates.rb#L101)

app/components/admin/budgets/form_component.html.erb
```diff
@@ -21,13 +21,12 @@
...
         </div>
       </div>
-    <% end %>
-
-    <div class="row expanded">
-      <div class="small-12 medium-9 large-6 column end">
-        <%= f.text_field :main_link_url, placeholder: t("admin.shared.example_url") %>
+      <div class="row expanded">
+        <div class="small-12 medium-9 large-6 column end">
+          <%= translations_form.text_field :main_link_url, placeholder: t("admin.shared.example_url") %>
```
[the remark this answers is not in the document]
Senen: Done!

spec/models/concerns/globalizable.rb
Senen: Because `globalizable` specs do not take into account conditioned validations. The `main_link_url` field validation only takes place when the field `main_link_text` is fulfilled. The globalizable spec method that fulfils the required fields does not fulfil the `main_link_text`as it is not mandatory and then some spec crashes when executed for the `Budget` or `Budget::Phase` model. This is the fastest way I found to solve it. Maybe there is a better approach 🤔 .

app/components/admin/budget_phases/form_component.html.erb
```diff
@@ -49,11 +49,11 @@
       <p class="help-text"> <%= t("admin.budget_phases.edit.main_call_to_action_description") %></p>
       <%= translations_form.text_field :main_link_text %>
     </div>
-  <% end %>
 
-  <div class="small-12 column">
-    <%= f.text_field :main_link_url, placeholder: t("admin.shared.example_url") %>
-  </div>
+    <div class="small-12 column">
+      <%= translations_form.text_field :main_link_url, placeholder: t("admin.shared.example_url") %>
```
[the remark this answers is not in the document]
Senen: Done!

config/locales/en/activerecord.yml
Senen: Done!

db/migrate/20211103090940_translate_budget_main_link_url.rb
Senen: Done!

db/migrate/20211103112944_translate_budget_phase_main_link_url.rb
Senen: Done!

app/models/budget.rb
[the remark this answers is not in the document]
Senen: > Awesome! 🎉 I've left some comments; let me know what you think!

Wow! It seems this time I wanted to run too much.

Senen:
> Awesome! 🎉 I've left some comments; let me know what you think!

Wow! It seems this time I wanted to run too much.

spec/models/concerns/globalizable.rb
javierm: Two possible solutions come to mind :thinking:.

One would be defining an extra factory, like it was done for other models (see commit 66334b5757).

```ruby
  factory :budget do
  # ...
    factory :budget_with_link do
      main_link_text { "Link" }
      main_link_url { "/" }
    end
  end
 
  # Then, in the budget spec:
  it_behaves_like "globalizable", :budget_with_link
```

The other would be changing the way we mark translations as required in order to handle the case where translatable fields are required but only under certain conditions. Here's the patch:

```diff
diff --git a/app/models/concerns/globalizable.rb b/app/models/concerns/globalizable.rb
index 98613b5a7..4c47249bd 100644
--- a/app/models/concerns/globalizable.rb
+++ b/app/models/concerns/globalizable.rb
@@ -32,14 +32,9 @@ module Globalizable
     private
 
       def required_attribute?(attribute)
-        presence_validators = [ActiveModel::Validations::PresenceValidator,
-          ActiveRecord::Validations::PresenceValidator]
-
-        attribute_validators(attribute).any? { |validator| presence_validators.include? validator }
-      end
-
-      def attribute_validators(attribute)
-        self.class.validators_on(attribute).map(&:class)
+        self.class.validators_on(attribute).any? do |validator|
+          validator.kind == :presence && !validator.options[:unless]
+        end
       end
 
       def check_translations_number
```

And here's the test case for this patch :sweat_smile:. I had to add the `validate :check_translations_number, if: :translations_required?` because I didn't manage to add a test using the `on: :update` condition. Due to this, I'm not even sure we're testing the right scenario here :thinking:.

```ruby
require "rails_helper"

describe Globalizable do
  it "doesn't require a translation when translatable fields require conditional presence" do
    dummy_translation = Class.new do
      include ActiveModel::Model
      attr_accessor :title, :locale
      validates :locale, presence: true
    end

    stub_const("DummyTranslation", dummy_translation)

    dummy_model = Class.new do
      include ActiveModel::Model
      attr_accessor :title, :translations

      def self.globalize_accessors; end
      def self.accepts_nested_attributes_for(...); end
      def self.after_validation(...); end
      def self.paranoid?; end

      def self.translation_class
        DummyTranslation
      end

      def translated_attribute_names
        [:title]
      end

      include Globalizable
      validate :check_translations_number, if: :translations_required?
      validates_translation :title, presence: true, unless: -> { true }

      def translations
        @translations ||= []
      end
    end

    stub_const("DummyModel", dummy_model)

    expect(DummyModel.new).to be_valid
  end
end
```

The advantage of the second solution is we could remove the factories defined in commit 66334b5757. But even so, I'm not sure which one I like better, so up to you!

javierm (approved):

PR #4727, opened 2021-11-03
Unify spec components expectations

taitus:
## Objectives
Make components specs clearer creating a matcher `be_rendered`.

This matcher can be called with argument `with` to check rendered content or without argument to check whether the content is empty or not.


spec/support/matchers/be_rendered.rb
```diff
@@ -0,0 +1,21 @@
...
+  failure_message do |page|
+    if page.has_css?("body")
+      "expected page to be rendered with #{with}, but was rendered with #{page.find("body").native.inner_html}"
+    else
+      "expected page to be rendered with #{with}, but was not rendered"
+    end
+  end
+
+  failure_message_when_negated do |page|
+    "expected page not to be rendered, but was rendered with #{page.native.inner_html}"
```
javierm, line 19: In order to have more concise error messages in our usual scenario, how about:

```suggestion
    if page.has_css?("body")
      "expected page not to be rendered, but was rendered with #{page.find("body").native.inner_html}"
    else
      "expected page not to be rendered, but was rendered with #{page.native.inner_html}"
    end
```

Maybe the `else` part isn't strictly necessary; it's there just because there might be edge cases we haven't considered and because without it the `!page.native.inner_html.empty?` condition used in the matcher seems a bit strange :thinking:.

javierm (commented):
Great! I've left a commet about a possible improvement.

Would it be possible to rebase against the latest `master`? It'd make it easier for me to compare against the old version of this pull request :smile:.

javierm (approved):

PR #4729, opened 2021-11-05
Add consistency between deploy-secrets and installer

taitus:
## Objectives
Add the same path in the `deploy-secrets.yml.example` file where the default application is installed with the installer.

The installer creates the application in the following path:
`home_dir: "/home/{{{ deploy_user }}"`
We update the variable "deploy_to" in the deploy-secrets file to be consistent with the installer.


javierm (approved):
Nice! :smile:

PR #4759, opened 2021-12-16
Fix newsletter actions

### LINKED ISSUE #2481: Remove "edit" and "remove" buttons from newsletters already sent
In the Newsletters admin section, remove the "edit" and "remove" buttons for the newsletters that have been sent.

gabrielcosta42:
## References

- Closes #2481

## Objectives

Since the newsletter has already been sent and cannot be resent, there is no reason for the edit button to be displayed.

## Visual Changes
image

### DIFF
--- a/app/helpers/newsletter_helper.rb
+++ b/app/helpers/newsletter_helper.rb
@@ -1,0 +1,9 @@
+module NewsletterHelper
+  def available_actions(newsletter)
+    if newsletter.draft? 
+      { actions: [:edit, :destroy] }
+    else
+      { actions: [:destroy] }
+    end
+  end
+end
--- a/app/views/admin/newsletters/index.html.erb
+++ b/app/views/admin/newsletters/index.html.erb
@@ -29,7 +29,7 @@
           <% end %>
         </td>
         <td>
-          <%= render Admin::TableActionsComponent.new(newsletter) do |actions| %>
+          <%= render Admin::TableActionsComponent.new(newsletter, available_actions(newsletter)) do |actions| %>
             <%= actions.action :preview, text: t("admin.newsletters.index.preview") %>
           <% end %>
         </td>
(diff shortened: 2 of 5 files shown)


app/helpers/newsletter_helper.rb
```diff
@@ -0,0 +1,9 @@
+module NewsletterHelper
```
wenderjean, line 1: Should we test this helper?

javierm: > Should we test this helper?

Definitely. I'm not sure I understand the comment; the tests are already there, aren't they? :thinking:

By the way, we no longer write helper methods unless strictly necessary but use components instead. However, @gabrielcosta42, if you aren't familiar with components and don't feel like looking into them, you can leave the helper method and we'll move it to a component :ok_hand:.

app/helpers/newsletter_helper.rb
```diff
@@ -0,0 +1,9 @@
...
+  def available_actions(newsletter)
+    if newsletter.draft? 
+      { actions: [:edit, :destroy] }
+    else
+      { actions: [:destroy] }
+    end
+  end
```
javierm, line 8: With these changes we remove the link to the `edit` action, but the action is still accessible by manually typing the URL on the browser :thinking:.

How about defining abilities in `app/models/abilities/administrator.rb`, changing the `Admin::NewslettersController` so it uses the `load_and_authorize_resource!` method, and using `can?` here? Maybe something like:

```suggestion
  def available_actions(newsletter)
    [:edit, :destroy].select { |action| can?(action, newsletter) }
  end
```

javierm (commented):
Hi, @gabrielcosta42. Thank you very much for this pull request! :smile:

Correct me if I'm wrong: the original issue says we should remove the "delete" (although due to a typo it says "remove") button as well when the newsletter has already been sent, but you're only removing the "edit" button :thinking:. Is there a reason for this?

javierm:
@gabrielcosta42 Hello again! :smile:

Any comments about the things I mentioned in the review? :thinking:

javierm:
Closing due to lack of response. @gabrielcosta42 feel free to reopen if you continue working on this pull request :wink:.

PR #4766, opened 2022-01-05
Add new GraphQL types & schema

tonekk:
### Problems with the current GraphQL API

The current consul GraphQL API has two problems.

1) It uses some unnecessary complicated magic to automatically create the GraphQL types and querys using an `api.yml` file. This approach is over-engineered, complex and has no benefits. It's just harder to understand the code for people which are not familiar with the project (like me, lol).

2) It uses a [deprecated DSL](https://graphql-ruby.org/schema/class_based_api.html) that is soon going to be removed from `graphql-ruby` completely. We are already seeing deprecation warning because of this (see References).

## References

Deprecation warnings, see #4632

## Objectives

The objective is to make the GraphQL Api more compliant with the way that the doc of `graphql-ruby` suggests, making it more maintainable.

## Notes

~~There is one problem. I wanted to create the API so that it is fully backwards compatible with the old one, BUT the old one uses field names which are directly derived from the ruby code, which results in `snake_case` field names - not the GraphQL way.~~
~~When I'm using the `graphql-ruby` Class-based syntax, it automatically creates the fields in `camelCase`, which breaks backwards-compatibility.~~
~~So we could either:~~

~~* Create a documentation that tells people how to migrate to the new API (my recommendation here)~~
~~* Mount the new API on another mountpoint, so that people can use both and then later remove the old API~~

~~I labelled this PR as a WIP, because once we decide which approach we are going to take, there is still some cleanup needed.~~

See below, the API is now backwards compatible.
 
## Release notes

:warning: We've updated the GraphQL API so it no longer uses a deprecated DSL. If you've added custom fields or models to your GraphQL API, you'll need to update those changes to the new DSL.


javierm:
@tonekk Thanks a lot! :pray:

I know this is just a WIP pull request, so when you continue working on it, would you mind splitting the changes dealing with mutations so they're in a different commit? So we'd get one commit to change the existing code while keeping the same functionality, and another one to add new functionality. That'd make the review a lot easier for us, particularly since our knowledge of GraphQL is really, really basic :relieved:.

tonekk:
Actually, most of it is Boilerplate code.
Here is a list of files that were completely generated by `rails g graphql:install`:

* `app/controllers/graphql_controller.rb`

* `app/graphql/types/base_argument.rb`
* `app/graphql/types/base_enum.rb`
* `app/graphql/types/base_field.rb`
* `app/graphql/types/base_input_object.rb`
* `app/graphql/types/base_interface.rb`
* `app/graphql/types/base_object.rb`
* `app/graphql/types/base_scalar.rb`
* `app/graphql/types/base_union.rb``

* `app/graphql/consul_schema.rb`

* `app/graphql/mutations/base_mutation.rb`


The only thing that I did, was creating the types (e.g. `Types::UserType` which inherit from `Types::BaseObject`. There, I added the fields, which you can currently find in `config/api.yml`.
Then, I added the root queries to the `Types::QueryType` (`app/graphql/types/query_type.rb`) to make them accessible.

Hope this helps :)

tonekk:
@javierm I now fixed the Controller spec and made the API backwards-compatible.
The old snake-case queries like `public_debates` on the `User` will throw a deprecation warning when used with a GraphQL client, that suggests to use the new `publicDebates` camelCase instead.

I consider this PR ready2merge now and have thus removed the WIP tag.

app/graphql/types/debate_type.rb
```diff
@@ -2,17 +2,18 @@ module Types
...
+
+    field_with_deprecated_sibling :public_created_at, String, null: true
+    field_with_deprecated_sibling :cached_votes_total, Integer, null: true
+    field_with_deprecated_sibling :cached_votes_up, Integer, null: true
+    field_with_deprecated_sibling :cached_votes_down, Integer, null: true
+    field_with_deprecated_sibling :comments_count, Integer, null: true
+    field_with_deprecated_sibling :hot_score, Integer, null: true
+    field_with_deprecated_sibling :confidence_score, Integer, null: true
+    field_with_deprecated_sibling :public_author, Types::UserType, null: true
+    field_with_deprecated_sibling :votes_for, Types::VoteType.connection_type, null: true
```
donni106, line 16: maybe putting this block below `field :tags,` would result in a better readability with quicker overview of fields and fields with deprecated siblings?

app/graphql/types/tag_type.rb
```diff
@@ -2,7 +2,7 @@ module Types
   class TagType < Types::BaseObject
     field :id, ID, null: false
     field :name, String, null: true
-    field :taggings_count, Integer, null: true
+    field_with_deprecated_sibling :taggings_count, Integer, null: true
```
donni106, line 5: maybe putting this below `field :kind`, would result in a better readability with quicker overview of fields and fields with deprecated siblings?

app/graphql/types/user_type.rb
```diff
@@ -2,8 +2,8 @@ module Types
   class UserType < Types::BaseObject
     field :id, ID, null: false
     field :username, String, null: true
-    field :public_debates, Types::DebateType.connection_type, null: true
-    field :public_proposals, Types::ProposalType.connection_type, null: true
-    field :public_comments, Types::CommentType.connection_type, null: true
+    field_with_deprecated_sibling :public_debates, Types::DebateType.connection_type, null: true
```
donni106, line 5: could a line break like in the files with more fields also be helpful here and vote_type and tag_type for readability?

config/routes/graphql.rb
```diff
@@ -1,3 +1,3 @@
-get "/graphql", to: "graphql#query"
-post "/graphql", to: "graphql#query"
+post "/graphql", to: "graphql#execute"
+get "/graphql", to: "graphql#execute"
```
donni106, line 2: could the removal of the `get` be a problem regarding backwards compatibility somewhere in the world?

tonekk: `get` wasn't removed as you can see above.
I removed it in one commit, then added it back again

donni106: sure! my fault

app/graphql/types/query_type.rb
```diff
@@ -0,0 +1,108 @@
...
+
+    field :proposals, Types::ProposalType.connection_type, 'Returns all proposals', null: false
+    field :proposal, Types::ProposalType, 'Returns proposal for ID', null: false do
+      argument :id, ID, required: true, default_value: false
+    end
+
+    field :proposal_notifications, Types::ProposalNotificationType.connection_type, 'Returns all proposal notifications', null: false
+    field :proposal_notification, Types::ProposalNotificationType, 'Returns proposal notification for ID', null: false do
+      argument :id, ID, required: true, default_value: false
+    end
```
zcei, line 27: Should those also have a deprecated sibling?

javierm:
@tonekk There are some restrictions in Github Actions we haven't managed to overcome, resulting in our CI not running the syntax check on pull request opened by people without write access to the repository :relieved:. Would you mind running `pronto run` on your machine and applying the suggestions? You may skip the ones regarding `Layout/LineLength`, since that rule uses a lower severity. We apply the rest of the rules even to files generated automatically, as you mentioned was the case here, so it would be great if you could modify them :pray:

Regarding the tests, there were [some failures](https://github.com/consul/consul/runs/4762466895?check_suite_focus=true) when running the test suite which seem related to this pull request :thinking:. Any ideas as to why these tests failed?

There was also [another failure](https://github.com/consul/consul/runs/4762466562?check_suite_focus=true) which was unrelated to this pull request. Sometimes we have [flaky bugs](https://github.com/consul/consul/issues?q=label%3A%22Flaky+bug%22+is%3Aclosed) and [flaky specs](https://github.com/consul/consul/issues?q=label%3A%22Flaky+spec%22+is%3Aclosed), which means there are issues with either the code or the specs which cause the test suite to fail sometimes for reasons we haven't found out yet.

Thanks!

tonekk:
@javierm I did the `pronto run` fixes and also fixed the `grapql_spec`. Plus I removed the old `ApiTypesCreator` and `QueryTypesCreator`

javierm:
@tonekk Sorry for the delay! Didn't have much time in January. This pull request is currently second on our TODO list (the first one being finishing #4764).

javierm (commented):
@tonekk Thank you very much for this pull request :pray: and sorry once again for the delay in the review.

I've left a few comments. Please note I might be wrong in some of the things I say, so feel free to correct me or disagree with anything :wink:.

Other than those comments, I've got a couple of general questions.

Now that we've got classes like `Types::DebateType`, is it necessary to keep the `config/api.yml` file? Right now it looks like the same attributes are defined in two different places :thinking:.

And, finally, would it make sense to squash the first two commits, "Add new GraphQL types, schema (with fields) & base mutation" and "Fix controller spec, re-add old GraphQL error handling", together? If so, it'd make the review easier, since we usually review commit by commit before doing a final "global" review.

app/controllers/graphql_controller.rb
```diff
@@ -6,52 +6,62 @@ class GraphqlController < ApplicationController
...
+      raise GraphqlController::QueryStringError if query_string.nil?
 
-      response = consul_schema.execute query_string, variables: query_variables
-      render json: response, status: :ok
+      variables = prepare_variables(params[:variables])
+      operation_name = params[:operationName]
+      context = {
+        # Query context goes here, for example:
+        # current_user: current_user,
+      }
```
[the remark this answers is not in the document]
tonekk: This is actually `graphql-ruby` boilerplate code.
The comment can be removed and we can just pass an empty hash to the `context:` keyword argument below.

app/controllers/graphql_controller.rb
```diff
@@ -6,52 +6,62 @@ class GraphqlController < ApplicationController
...
+      when String
+        if variables_param.present?
+          JSON.parse(variables_param) || {}
+        else
+          {}
+        end
+      when Hash
+        variables_param
+      when ActionController::Parameters
+        variables_param.to_unsafe_hash # GraphQL-Ruby will validate name and type of incoming variables.
```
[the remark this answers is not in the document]
tonekk: Actually I'm also kind of clueless here, this is all boilerplate code from `graphql-ruby`... 😅

app/controllers/graphql_controller.rb
```diff
@@ -6,52 +6,62 @@ class GraphqlController < ApplicationController
...
+      operation_name = params[:operationName]
+      context = {
+        # Query context goes here, for example:
+        # current_user: current_user,
+      }
+      result = ConsulSchema.execute(query_string,
+        variables: variables,
+        context: context,
+        operation_name: operation_name
+      )
```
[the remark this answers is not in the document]
tonekk: ok

app/graphql/types/base_object.rb
```diff
@@ -0,0 +1,12 @@
+module Types
+  class BaseObject < GraphQL::Schema::Object
+    field_class Types::BaseField
+
+    def self.field_with_deprecated_sibling(field_name, type, null:)
+      reason = "Snake case fields are deprecated. Please use #{field_name.to_s.camelize(:lower)}."
+
+      field field_name, type, null: null
+      field field_name, type, null: null, camelize: false, deprecation_reason: reason
+    end
```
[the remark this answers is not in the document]
tonekk: Actually that's a pretty good idea

app/controllers/graphql_controller.rb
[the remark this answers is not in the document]
tonekk: I did research on this.

Actually it is possible to send GraphQL in two ways:

1) The _standard_ way:
** Use `application/json` as request Content type and send the whole thing (graphql query and variables) in the body as valid JSON:

```json
{
  "query":"query GetProposal($id: ID!) {\n\tproposal(id: $id) {\n\t\tid\n\t\ttitle\n\t\tsummary\n\t\tcachedVotesUp\n\t\tpublicAuthor {\n\t\t\tusername\n\t\t}\n\t}\n}",
  "variables":{
    "id":"30"
  },
  "operationName":"GetProposal"
}
```

Whole HTTP example:

```
POST /graphql HTTP/1.1
Content-Type: application/json; charset=utf-8
Host: localhost:3000
Content-Length: 220

{"query":"query GetProposal($id: ID!) {\n\tproposal(id: $id) {\n\t\tid\n\t\ttitle\n\t\tsummary\n\t\tcachedVotesUp\n\t\tpublicAuthor {\n\t\t\tusername\n\t\t}\n\t}\n}","variables":{"id":"30"},"operationName":"GetProposal"}
```

2) The _non-standard_ way
** Use `application/graphql` as content type.
** Send the graphql query in the body as graphql (not json!)
** Variables can be send in as query parameters

Full HTTP example:

```
POST /graphql?variables=%7B%20%22id%22%3A%20%2230%22%20%7D HTTP/1.1
Content-Type: application/graphql
Host: localhost:3000
Content-Length: 127

query GetProposal($id: ID!) {
	proposal(id: $id) {
		id
		title
		summary
		cachedVotesUp
		publicAuthor {
			username
		}
	}
}
```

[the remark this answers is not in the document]
tonekk: IMO we could even argue about just supporting `application/json` requests, which would make the `query_string` method obsolete as well.

app/controllers/graphql_controller.rb
```diff
@@ -6,52 +6,62 @@ class GraphqlController < ApplicationController
...
     rescue GraphqlController::QueryStringError
       render json: { message: "Query string not present" }, status: :bad_request
     rescue JSON::ParserError
       render json: { message: "Error parsing JSON" }, status: :bad_request
     rescue GraphQL::ParseError
       render json: { message: "Query string is not valid JSON" }, status: :bad_request
-    rescue
-      unless Rails.env.production? then raise end
+    rescue ArgumentError => e
+      render json: { message: e.message }, status: :bad_request
```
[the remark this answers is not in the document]
tonekk: Catching the `ArgumentError` here that is raised in `prepare_variables`.

Also I think it wasn't a good idea to suppress Errors in production.

app/graphql/types/base_scalar.rb
```diff
@@ -0,0 +1,4 @@
+module Types
+  class BaseScalar < GraphQL::Schema::Scalar
+  end
+end
```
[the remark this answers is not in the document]
tonekk: Actually, we can remove:

```
app/graphql/types/base_argument.rb
app/graphql/types/base_enum.rb
app/graphql/types/base_field.rb
app/graphql/types/base_input_object.rb
app/graphql/types/base_interface.rb
app/graphql/types/base_union.rb
```

This is all boilerplate code and since we added no custom logic there, we don't need these files.
I would keep `BaseMutation` though, in case we later need to add custom Mutation logic (kind of like having an `ApplicationController` instead of inheriting from `ActionController::Base`.

app/graphql/types/base_union.rb
```diff
@@ -0,0 +1,4 @@
+module Types
+  class BaseUnion < GraphQL::Schema::Union
```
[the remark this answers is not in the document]
tonekk: See above.

app/graphql/types/base_scalar.rb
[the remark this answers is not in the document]
tonekk: Done in https://github.com/consul/consul/pull/4766/commits/59ce9a85bce016a8faa845fd890a5f2fb40caabf

tonekk:
> Now that we've got classes like `Types::DebateType`, is it necessary to keep the `config/api.yml` file? Right now it looks like the same attributes are defined in two different places 🤔.

No, you are right.
Removed in https://github.com/consul/consul/pull/4766/commits/9b19e1ebe0e38596696890157b76781dec7943aa

app/graphql/types/base_object.rb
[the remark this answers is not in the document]
tonekk: Done in https://github.com/consul/consul/pull/4766/commits/36db8ee9e39bdbbcb3ad72f203e096eb4da5cdf3

tonekk:
Now how do we proceed?
I would suggest that you review the changes I made and then I squash the commits.

app/graphql/types/query_type.rb
```diff
@@ -0,0 +1,128 @@
+module Types
+  class QueryType < Types::BaseObject
+    field :comments, Types::CommentType.connection_type, "Returns all comments", null: false
+    field :comment, Types::CommentType, "Returns comment for ID", null: false do
+      argument :id, ID, required: true, default_value: false
```
javierm, line 5: Do you know whether the `required: true, default_value: false` are the default options? If they are, maybe we can remove them :thinking:.

app/graphql/types/query_type.rb
```diff
@@ -0,0 +1,128 @@
...
+    field(
+      :proposal_notification,
+      Types::ProposalNotificationType,
+      "Returns proposal notification for ID",
+      null: false,
+      camelize: false,
+      deprecation_reason: "Snake case fields are deprecated. Please use proposalNotification."
+    ) do
+      argument :id, ID, required: true, default_value: false
+    end
```
javierm, line 47: Are these lines necessary now that we've changed the `field` method to generate the snak case field as well? :thinking:

javierm (commented):
@tonekk I've left a few comments regarding possible simplifications. Let me know what you think!

Thank you very much for your comments and your patience! And, of course, for your work on this pull request :smile:.

app/graphql/types/comment_type.rb
```diff
@@ -0,0 +1,16 @@
+module Types
+  class CommentType < Types::BaseObject
+    field :ancestry, String, null: true
```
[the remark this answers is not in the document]
javierm: @tonekk My bad :pray:. I read the [field documentation](https://graphql-ruby.org/fields/introduction.html#field-return-type) and since it used examples without `null` and said that `true` was the default value, I assumed it would work.

But I've just tested it and it doesn't, so feel free to ignore this comment.

app/graphql/types/query_type.rb
javierm: Same here; it doesn't work :relieved:. Feel free to ignore this comment too.

app/graphql/types/query_type.rb
```diff
@@ -0,0 +1,128 @@
...
+    field :vote, Types::VoteType, "Returns vote for ID", null: false do
+      argument :id, ID, required: true, default_value: false
+    end
+
+    def comments
+      Comment.public_for_api
+    end
+
+    def comment(id:)
+      Comment.find(id)
```
javierm, line 69: Originally, we had this line to find a record:

```ruby
model.public_for_api.find_by(id: ...)
```

There are two differences here:

1. (Most important) We used the `public_for_api` scope to guarantee only public records could be accessed
2. We returned `nil` instead of raising an exception

I guess we can change this line:

```suggestion
      Comment.public_for_api.find_by(id)
```

And replace `null: false` with `null: true` in the `field` method :thinking:.

Since we have to make this change for every model, maybe we could refactor the code a little bit:

```ruby
    {
      comment: Types::CommentType,
      debate: Types::DebateType,
      geozone: Types::GeozoneType,
      proposal: Types::ProposalType,
      proposal_notification: Types::ProposalNotificationType,
      tag: Types::TagType,
      user: Types::UserType,
      vote: Types::VoteType
    }.each do |field_name, type_class|
      plural_name = field_name.to_s.pluralize
      model_class = field_name.to_s.classify.constantize

      field plural_name, type_class.connection_type, "Returns all #{plural_name.gsub("_", " ")}", null: false
      field field_name, type_class, "Returns #{field_name} for a given ID", null: true do
        argument :id, ID, required: true, default_value: false
      end

      define_method(field_name) { |id:| model_class.public_for_api.find_by(id: id) }
      define_method(plural_name) { model_class.public_for_api }
    end
```

Not sure whether this is an improvement, though :thinking:. I'll leave it up to you whether to apply this refactoring :wink:.

lib/graph_ql/query_type_creator.rb
```diff
@@ -1,28 +0,0 @@
-require "graphql"
-
-module GraphQL
-  class QueryTypeCreator
```
javierm, line 4: Since we're removing this class, I think we can also remove all the `graphql_` methods in the `Graphqlable` concern and in the `ActsAsTaggableOn` initializer.

javierm (commented):
Very sorry to add another review :pray:; didn't realize a couple of things earlier :sweat_smile:.

app/graphql/types/query_type.rb
tonekk: Actually, if you would have asked me a few years ago, I would probably said: This is a really good idea.
But recent projects have taught me that very often, code that is a bit more verbose, but 'standard' according to the doc of whatever library you are using, is a better choice.

This change would re-introduce some of the complexity that I removed with the 'weird' `ApiTypesCreator`, so I think I wouldn't want to do it.
Also: You never know what's going to happen in the future. You might end up with edge cases for some models and than you would have to introduce more complexity here.

javierm: Yes, that's why I wasn't sure about it. As mentioned, up to you! :wink:

javierm:
@tonekk I was a bit worried about the fact that the tests were passing even if we changed the behavior a little bit; it looks like our test coverage for Graphql isn't that great :sweat_smile:. Sorry for that!

I've written a test which passes in the master branch but fails on this one due to the `public_for_api` behavior, which is associations:

```ruby
    it "only links public comments" do
      user = create(:administrator).user
      create(:comment, author: user, body: "Public")
      create(:budget_investment_comment, author: user, valuation: true, body: "Valuation")

      response = execute("{ user(id: #{user.id}) { public_comments { edges { node { body } } } } }")
      received_comments = dig(response, "data.user.public_comments.edges")

      expect(received_comments).to eq [{ "node" => { "body" => "Public" }}]
    end
```

At least three possible solutions come to mind.

One option could be adding a new method `collection_field`:

```ruby
    def self.collection_field(*args, **kwargs, &block)
      field(*args, **kwargs, &block)

      field_name = args[0]

      define_method(field_name) { object.send(field_name).public_for_api }
    end
```

And then we'd have to replace `field` with `collection_field` where appropriate (and maybe putting those lines at the end of the type definition so the code is still easy to read).

Another option could be modifying the `field` method, adding:

```ruby
      type_class = args[1]

      if type_class.is_a?(Class) && type_class.ancestors.include?(GraphQL::Types::Relay::BaseConnection)
        define_method(field_name) { object.send(field_name).public_for_api }
      end
```

In both cases, we could remove the `tags` methods in `proposal_type.rb` and `debate_type.rb`.

The third option would be to add collection methods everywhere where it's needed, adding `.public_for_api`, as done with the `tags` method.

I've tried to, but so far not succeeded to, write a similar test for a `belongs_to` association. This case was previously handled with:

```ruby
  association_target = object.send(field_name)
  association_target.present? ? field_type.public_for_api.find_by(id: association_target.id) : nil
```

How do you think we should handle these cases? :thinking:

app/graphql/types/comment_type.rb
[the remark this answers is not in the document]
tonekk: All good :)

app/graphql/types/mutation_type.rb
```diff
@@ -0,0 +1,10 @@
+module Types
+  class MutationType < Types::BaseObject
+    # TODO: remove me
+    field :test_field, String, null: false,
+      description: "An example field added by the generator"
+    def test_field
+      "Hello World"
+    end
```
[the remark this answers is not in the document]
tonekk: Resolved in [9e07772](https://github.com/consul/consul/pull/4766/commits/9e07772f498b88906680bf43c5556f3123b2529f)

app/graphql/types/base_object.rb
```diff
@@ -0,0 +1,31 @@
...
+  class BaseObject < GraphQL::Schema::Object
+    # def self.field(field_name, type, null:)
+    #   super
+
+    #   if field_name.include?("_")
+    #     reason = "Snake case fields are deprecated. Please use #{field_name.to_s.camelize(:lower)}."
+    #     super field_name, type, null: null, camelize: false, deprecation_reason: reason
+    #   end
+    # end
+
```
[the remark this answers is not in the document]
tonekk: Resolved in [2eb3886](https://github.com/consul/consul/pull/4766/commits/2eb3886056c1409bc4edc1909010494aacc07be8)

app/graphql/types/base_object.rb
```diff
@@ -0,0 +1,31 @@
...
+        super(*args, **kwargs, &block)
+      end
+    end
+
+    # def self.field_with_deprecated_sibling(field_name, type, null:)
+    #   reason = "Snake case fields are deprecated. Please use #{field_name.to_s.camelize(:lower)}."
+
+    #   field field_name, type, null: null
+    #   field field_name, type, null: null, camelize: false, deprecation_reason: reason
+    # end
```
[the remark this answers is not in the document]
tonekk: Resolved in [2eb3886](https://github.com/consul/consul/pull/4766/commits/2eb3886056c1409bc4edc1909010494aacc07be8)

app/graphql/types/query_type.rb
tonekk: Yup, can confirm that this works!

tonekk: Removed in [e155403](https://github.com/consul/consul/pull/4766/commits/e155403aa55b9b0cf2a2ce208bbaacbaedd17566)

lib/graph_ql/query_type_creator.rb
tonekk: I think the `Graphqlable` concern should be removed completely, but out of scope for this PR.
I have no idea why `public_created_at` was created...

tonekk: [63f78d3](https://github.com/consul/consul/pull/4766/commits/63f78d30aad8e37137f88c7878e77e305fdf31b7)

tonekk:
While reading your last comment about "test coverage" of the GraphQL suite, I had a look at some of the specs which are currently there and I got a little bit... startled. Some of the specs are plain useless...

Take these specs:

```ruby
   it "does not include comments of debates that are not public" do                                                       
~       not_public_debate = create(:debate)                                                                                  
        not_public_debate_comment = create(:comment, commentable: not_public_debate)                                         
        allow(Comment).to receive(:public_for_api).and_return([])                                                            
                                                                                                                             
        response = execute("{ comments { edges { node { body } } } }")                                                       
        received_comments = extract_fields(response, "comments", "body")                                                     
                                                                                                                             
        expect(received_comments).not_to include(not_public_debate_comment.body)                                             
      end                                                                                                                    
                                                                                                                                   it "does not include comments of proposals that are not public" do                                                     
        not_public_proposal = create(:proposal)                                                                              
        not_public_proposal_comment = create(:comment, commentable: not_public_proposal)                                     
        allow(Comment).to receive(:public_for_api).and_return([])                                                            
                                                                                                                             
        response = execute("{ comments { edges { node { body } } } }")                                                       
        received_comments = extract_fields(response, "comments", "body")                                                     
                                                                                                                             
        expect(received_comments).not_to include(not_public_proposal_comment.body)                                           
      end                                                                                                                    
                                                                                                                             
      it "does not include comments of polls that are not public" do                                                         
        not_public_poll = create(:poll)                                                                                      
        not_public_poll_comment = create(:comment, commentable: not_public_poll)                                             
        allow(Comment).to receive(:public_for_api).and_return([])                                                            
                                                                                                                             
        response = execute("{ comments { edges { node { body } } } }")                                                       
        received_comments = extract_fields(response, "comments", "body")                                                     
                                                                                                                             
        expect(received_comments).not_to include(not_public_poll_comment.body)                                               
      end
      ```

tonekk:
Regarding your last comment, I like the 2nd option and I have included it here: [89f40aa](https://github.com/consul/consul/pull/4766/commits/89f40aa8e30534aa2345610b4bdc7a8fae69573a)

Regarding `belongs_to`-associations: There is only one `belongs_to` association here and that is `ProposalNotification belongs_to Proposal`. `Proposal.public_for_api` is currently just returning `all`...

As I wrote above, we should not overcomplicate things and, in the future, question what this api should really accomplish...

javierm:
> As you can see, public_for_api is mocked everywhere... These specs will always succeed.
> (...)
> Definitely out of scope for this PR, but I think we should completely rethink what the GraphQL Api should do and question the concept of public_for_api and delete useless specs...

Thank you for the comment! This is very interesting.

My best guess is the tests are done this way so these tests check that `public_for_api` is actually called when requesting data to the GraphQL API, and the actual behavior of the `public_for_api` methods is tested in `spec/models`. And so, these tests would fail if, for instance, in the `QueryType` we called `Comment.all` instead of `Comment.public_for_api`, while the model tests would fail if we changed the `public_for_api` method itself.

I like the concept of replacing all the `public_for_api` methods with methods in the `*Type` classes and replacing the model tests with tests for these `*Type` classes, so feel free to open a pull request with these changes in the future :+1:.

lib/graph_ql/query_type_creator.rb
javierm: > I have no idea why public_created_at was created...

That makes two of us :sweat_smile:. No idea why it's important that records return the creation date only showing the hour but not the minute or the second when they were created.

javierm:
Hi, @tonekk :smile:.

Just in case you're interested, we've just merged pull request #4807, simplifying some of the API-related scopes. Thanks for inspiring this refactoring! Even though we might end up removing all the `public_for_api` scopes, in the meantime, these methods are now easier to understand.

javierm (approved):
@tonekk Thank you so much! It's been a pleasure to work with you on this pull request :smile:.

I've taken the liberty to rebase against the latest `master` branch and squash the commits. Hope you don't mind!

PR #4767, opened 2022-01-11
Fix generating dashboard poster on production

javierm:
## References

* We introduced this bug in commit 905ac48bb from pull request #4141

## Objectives

Fix a bug which made it impossible to generate the proposals dashboard poster on production environments.

## Notes

We aren't adding a test case because this bug is only present on production environments when assets have been precompiled.


Senen (approved):
Nice catch and solution!

PR #4768, opened 2022-01-11
Avoid mousewheel/finger trap in the map

Senen:
## References

This fixes usability issues when scrolling pages with a map and more content below. [card-74165431](https://github.com/orgs/consul/projects/1#card-74165431)

## Objectives

Make it easier for users to scroll the page to the bottom by:

1. Disable the map mouse wheel event to zoom the map on desktop devices
2. Disable map dragging and tap events on mobile devices
3. Add controls to the map so desktop and mobile users can move around the map

## Visual Changes

**Before**
map-pan-control-before

**After**
map-pan-control-after-mobile
map-pan-control-after

### Alternative solutions
The first solution I tried was to install a Leaflet plugin that adds a similar behaviour to [Google Maps Gesture Handling plugin](https://developers.google.com/maps/documentation/javascript/examples/interaction-cooperative). During mobile testing, I found it works nicely in Chrome, Edge, Opera and Safari, while the user experience when using two fingers to move or zoom the map is annoying in Firefox. Unfortunately, it seems to conflict in some way with the page scroll, so I finally chose to use controls to allow users to move the map. Here is a [work in progress branch](https://github.com/rockandror/consul/tree/map-usability) with the Leatflet.GestureHandling plugin installed in CONSUL.


Senen:
Closing in favor of #5096.

PR #4787, opened 2022-03-16
Make it easier to customise model constants

taitus:
## Objectives
Make it easier to customise the model constants when overwriting the model in the custom folder.
Add `lambda` to the validations that use these constants so that if we modify the value in the `model/custom` folder it can load the new values.

An example of how to customise these constants in the `models/custom` folder would be:

```
require_dependency Rails.root.join("app", "models", "budget", "phase").to_s

class Budget::Phase
  consul_phase_kinds = PHASE_KINDS
  remove_const("PHASE_KINDS")
  PHASE_KINDS = consul_phase_kinds + %w[whatever]
end
```

### DIFF
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -4,7 +4,7 @@ class Activity < ApplicationRecord
 
   VALID_ACTIONS = %w[hide block restore valuate email].freeze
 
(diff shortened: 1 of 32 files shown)


javierm (approved):
@taitus Awesome! This was a really painful point so many forks were suffering.

PR #4794, opened 2022-03-29
Fix confirmation instructions texts

taitus:
## Objectives
The texts for the confirmation instructions referred to "reset the password".
We have updated the texts to refer to confirmation instructions.

## Visual Changes
Display updated text


taitus:
These changes have been added in the PR #4228

PR #4769, opened 2022-01-17
Show assigned heading on investment show

microweb10:
## Objectives

Show assigned heading on investment show

PR #4796, opened 2022-03-29
Refactor investments controller

taitus:
## References
Related PR: #4769 

## Objectives
Unify management for the`@assigned_heading` in the new callback `before_action :load_assigned_heading` added in related PR. 

## Development details
Some minor improvements that will help us to unify the management of the `:load_assigned_heading`:
- Extract the logic of `load_map` from `load_heading` callback. Create a new callback `load_map`
- Split the `:investments` method to avoid duplication of lines as much as possible without losing clarity. Besides that splitting this method and having created the `:investments_with_filters` method, will help us to have a list of investments without the need to have a `@current_order`. This new method will help us/will be used in the main objective of this PR which is to unify the management of the `@assigned_heading`.
- Load a `@heading` whenever possible, so you can delegate the `assigned_heading` logic to his callback. Right now we have an `if/elsif` in the `load_heading` callback that is called from the `:index` and the `:show` actions, I thought it would be interesting to convert it to an `if/else` so that we have an `@heading` whenever possible, so that we can end up delegating the logic of `@assigned_heading` to its `:load_assigned_heading callback`.

## Notes
This PR need merged after the Related PR.


taitus:
Some of these changes have been added in the related PR #4769

PR #4800, opened 2022-03-31
Fix checkbox label styles

taitus:
## References

Reference to Pending notes: https://github.com/orgs/consul/projects/1#card-74165527

## Background

Currently we use a `display: block` style for labels containing check boxes inside them, and so the label has a width of 100%.

Since, on most browsers, clicking on a label has a similar effect as clicking on its associated input, this means that clicking on the blank space on the right of the label text will check/uncheck the checkbox. In usability tests we've seen people accidentally activating and deactivating options without noticing, due to this issue.

## Objectives

Check checkbox labels styles in the admin section in order to prevent accidental clicks.

### DIFF
--- a/app/assets/stylesheets/layout.scss
+++ b/app/assets/stylesheets/layout.scss
@@ -1063,6 +1063,10 @@ form {
     line-height: $line-height;
   }
 
+  .checkbox-label {
+    display: table;
+  }
+
   fieldset legend {
     font-weight: bold;
   }
--- a/spec/lib/consul_form_builder_spec.rb
+++ b/spec/lib/consul_form_builder_spec.rb
@@ -4,7 +4,7 @@ describe ConsulFormBuilder do
   before do
     dummy_model = Class.new do
       include ActiveModel::Model
-      attr_accessor :title, :quality
+      attr_accessor :title, :quality, :published
     end
 
     stub_const("DummyModel", dummy_model)
@@ -72,6 +72,15 @@ describe ConsulFormBuilder do
 
       expect(page).to have_css ".help-text", text: "Ugly is neither good nor bad"
       expect(page).to have_css "select[aria-describedby='dummy_quality-help-text']"
+    end
+  end
+
+  describe "#check_box" do
+    it "adds a checkbox-label class to the label by default" do
+      render builder.check_box(:published)
+
+      expect(page).to have_css "label", count: 1
+      expect(page).to have_css ".checkbox-label"
     end
   end
 
--- a/app/assets/stylesheets/mixins/forms.scss
+++ b/app/assets/stylesheets/mixins/forms.scss
@@ -170,6 +170,11 @@
     @include breakpoint(medium) {
       width: 75%;
     }
+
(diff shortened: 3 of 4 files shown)


app/assets/stylesheets/admin.scss
```diff
@@ -315,6 +315,16 @@ $table-header:   #ecf1f6;
     .proposal-new {
       padding-top: 0;
     }
+
+    form {
```
javierm, line 319: I've just realized this also happens in public sections such as "My account". Maybe we can use a more generic solution, and apply it just to check boxes :thinking:.

```diff
diff --git a/app/assets/stylesheets/layout.scss b/app/assets/stylesheets/layout.scss
index d26617846..b4eecfd5f 100644
--- a/app/assets/stylesheets/layout.scss
+++ b/app/assets/stylesheets/layout.scss
@@ -1063,6 +1063,10 @@ form {
     line-height: $line-height;
   }
 
+  .checkbox-label {
+    display: table;
+  }
+
   fieldset legend {
     font-weight: bold;
   }
diff --git a/lib/consul_form_builder.rb b/lib/consul_form_builder.rb
index 5eb292079..9d2453458 100644
--- a/lib/consul_form_builder.rb
+++ b/lib/consul_form_builder.rb
@@ -25,7 +25,11 @@ class ConsulFormBuilder < FoundationRailsHelper::FormBuilder
     else
       label = tag.span sanitize(label_text(attribute, options[:label])), class: "checkbox"
 
-      super(attribute, options.merge(label: label, label_options: label_options_for(options)))
+      # TODO: add test; note adding options[:label_options][:class] would overwrite it; is it OK like this?
+      super(attribute, options.merge(
+        label: label,
+        label_options: { class: "checkbox-label" }.merge(label_options_for(options))
+      ))
     end
   end
```

javierm (commented):
@taitus The `display: table` solution looks great :+1:. The only issue I see is there might be side effects if we apply it everywhere :thinking:. I've left an idea about doing it only for checkboxes.

spec/lib/consul_form_builder_spec.rb
```diff
@@ -75,6 +75,15 @@
     end
   end
 
+  describe "#check_box" do
+    it "renders the label with new checkbox-label class by default" do
```
javierm, line 79: I think there's a typo here :thinking:. An article is missing and the class won't be "new" when we read the test in a few months :smile:.

```suggestion
    it "adds a checkbox-label class to the label by default" do
```

app/assets/stylesheets/layout.scss
```diff
@@ -1063,6 +1063,17 @@ form {
     line-height: $line-height;
   }
 
+  .actions {
+    label {
+      margin-left: auto;
+      margin-right: auto;
+    }
```
javierm, line 1070: How about moving this code to the `public-form` mixin, which contains the rest ofthe styles for the `.actions` selector?

javierm (approved):
@taitus Nice! :tada:

PR #4801, opened 2022-03-31
Allow to customize tests common action modules

Senen:
## Objectives

Provide an easy way to customize `spec/support/common_actions` modules. 

With this change custom common_actions modules can be placed in the `spec/support/custom/common_actions/` folder. 

Let's say a fork added a new mandatory field estimated_price to the budget investments table. This new field requires updating some helper methods that automatically fulfil the mandatory investment fields. Now we can create a new file to override the implementation of the original method, so it fulfils the newly added field:

**spec/support/custom/common_action/maps.rb**
```
module Maps
  def fill_in_budget_investment_form
    fill_in "Title", with: "Budget investment title"
    fill_in_ckeditor "Description", with: "Budget investment description"
    fill_in "Estimated price", with: "10000"
    check :budget_investment_terms_of_service
  end
end
```


spec/support/common_actions.rb
```diff
@@ -1,10 +1,14 @@
 Dir["./spec/support/common_actions/*.rb"].each { |f| require f }
+Dir["./spec/support/custom/**/**.rb"].each { |f| require f }
```
javierm, line 2: It looks a bit strange to have only `support/common_actions` for CONSUL support files but everything under `support/custom` (whether common actions or not) under custom files :thinking:. We're inside the `CommonActions` module, after all.

Two possibilities come to mind. The first one:

```suggestion
Dir["./spec/support/common_actions/custom/*.rb"].each { |f| require f }
```

The second one:

```suggestion
Dir["./spec/support/custom/common_actions/*.rb"].each { |f| require f }
```

Not sure which one I prefer (assuming they even work :smile:).

javierm (commented):
@Senen In general, I'm fine with these changes :ok_hand:. I'm not a big fan of our `CommonActions` module, but I'm not a big fan of having methods defined in a shared examples file outside any class/module/block either. If now these methods can be customized, at least that's an improvement :ok_hand:.

I've left two comments about minor things; let me know what you think!

spec/support/common_actions/documents.rb
```diff
@@ -0,0 +1,52 @@
+module Documents
+  def do_login_for(user)
+    common_do_login_for(user, management: management)
+  end
```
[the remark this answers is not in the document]
Senen: Nice catch! 

> By the way, I've no idea why this method but it's magically using the management variable even if we're not passing it as an argument, although it seems to work properly ❓.

After some debugging, it seems the `management` variable gets its value from the caller context.

> It's very possible that we already had this very same problem, since previously we were defining the methods inside the default object...

Yes, I think so. I finally left one method within the `common_actions/users` module and added to all the method calls the management argument, so we do not have to wonder where it comes from.

spec/support/common_actions.rb
javierm: Adding that line to the `spec_helper.rb` file is fine with me :ok_hand:.

javierm: Oh, sorry, my bad :pray:.

The files under `spec/support/custom/**/*.rb` are already included in `spec/support/**/*.rb`, so the new line wouldn't do anything :thinking:.

I think we'd still need to change the `common_actions.rb` file in order to ensure all the custom common actions files are loaded _before_ the content of the `common_actions.rb` file.

Senen: Apologies not accepted! I should have tested it before suggesting it! 🙏🏼

javierm: I'm reading the last paragraph of my previous message and I'm not sure I understand it :rofl:. In summary, my initial suggestions regarding this line still apply.

javierm (approved):

PR #4806, opened 2022-04-11
Add and improve hint information related to phases duration

taitus:
## Objectives
Add a hint telling administrators that the dates for budget phases are informative and the phase has to be manually changed.

## Visual Changes
- Updated text on wizard phases step
Updated text on wizard phases step

- New text above phases table on admin budget show page
New text above phases table on admin budget show page

- Updated hint text for Duration field and new hint texts for Dates fields
Updated hint text for Duration field and new hint texts for Dates fields

- Do not render wizard component for edit phase page when we are not in the wizard
Do not render wizard component for edit phase page when we are not in the wizard

### DIFF
--- a/config/locales/en/admin.yml
+++ b/config/locales/en/admin.yml
@@ -113,6 +113,7 @@ en:
         delete: Delete budget
         phase: Phase
         phases_caption: "Phases"
+        phases_table_help_text: "The configuration of these phases is used for information purposes only. Its function is to define the phases information displayed on the public page of the participatory budget."
         duration: "Duration"
         enabled: Enabled
         actions: Actions
@@ -153,7 +154,7 @@ en:
         headings:
           multiple: "Headings are meant to divide the money of the participatory budget. Here you can add headings for this group and assign the amount of money that will be used for each heading."
           single: "Headings are meant to divide the money of the participatory budget. Since this budget will only contain one heading, this is the place where you stablish the money that will be spent in this participaroty budget."
-        phases: "Participatory budgets have different phases. Here you can enable or disable phases and also customize each individual phase."
+        phases: "Participatory budgets have different phases. Here you can enable or disable phases and also customize each individual phase. The configuration of these phases is used for information purposes only and will be displayed on the public page of the participatory budget."
       show:
         add_group: "Add group"
         add_heading: "Add heading"
--- a/config/locales/en/admin.yml
+++ b/config/locales/en/admin.yml
@@ -215,8 +215,8 @@ en:
         title: "Edit phase"
         description_help_text: This text will appear in the header when the phase is active
         duration: "Phase's duration"
-        duration_description: "The period of time this phase will be active."
-        enabled_help_text: This phase will be public in the budget's phases timeline, as well as active for any other purpose
+        duration_description: "These fields are used for information purposes only and <strong>do not trigger an automatic update of the active phase.</strong> In order to update it, edit the budget and select the active phase."
+        enabled_help_text: This phase will be public in the budget's phases timeline.
         image_description: "If an image is uploaded it will be displayed next to the description of this phase."
         main_call_to_action: "Main call to action (optional)"
         main_call_to_action_description: "This link will appear on main banner of this participatory budget when this phase is enabled and encourages your user to perform a specific action like creating a proposal, voting for existing ones, or learn more about the process."
--- a/config/locales/en/admin.yml
+++ b/config/locales/en/admin.yml
@@ -213,6 +213,7 @@ en:
     budget_phases:
       edit:
         title: "Edit phase"
+        dates_help_text: "For information purposes only"
         description_help_text: This text will appear in the header when the phase is active
         duration: "Phase's duration"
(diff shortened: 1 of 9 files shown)


config/locales/en/admin.yml
```diff
@@ -215,7 +216,7 @@ en:
         title: "Edit phase"
         description_help_text: This text will appear in the header when the phase is active
         duration: "Phase's duration"
-        duration_description: "The period of time this phase will be active."
+        duration_description: "The period of time this phase will be active. Remember that this field is for information only and does not trigger an automatic update of the active phase."
```
javierm, line 219: I was using the form and I read "Phase's duration, blah, blah, blah, start date, end date" :thinking:.

So maybe we could add hints in three places: the one we're using now, under the "start date" label, and under the "end date" label.

Since the first sentence isn't really true and is also a bit redundant because administrators will already expect that when they see the date fields, maybe we could focus on the second one:

```suggestion
        duration_description: "These fields are used for information purposes only and do not trigger an automatic update of the active phase. In order to update it, edit the budget and select the active phase."
```

And then, in both "start date" and "end date", we could add a hint saying "For information purposes only".

We could also add a `<strong>` tag for the "do not trigger an automatic update of the active phase"  text :thinking:. I'm wondering whether we should then modify the styles slightly:

```diff
diff --git a/app/assets/stylesheets/layout.scss b/app/assets/stylesheets/layout.scss
index b4eecfd5f..46036acb7 100644
--- a/app/assets/stylesheets/layout.scss
+++ b/app/assets/stylesheets/layout.scss
@@ -450,6 +450,10 @@ button,
 .help-text {
   line-height: rem-calc(20);
   margin-top: 0;
+
+  strong {
+    font-style: normal;
+  }
 }
 
 .menu-and-content {
```

If you think this is overkill, say so! :wink:

config/locales/en/admin.yml
```diff
@@ -114,6 +114,7 @@ en:
         phase: Phase
         phases_caption: "Phases"
         duration: "Duration"
+        duration_help_text: "(Informative field)"
```
javierm, line 117: I think we can remove the parenthesis since we don't use them in other hints in the application :thinking:.

"Informative field" sound a bit funny; I was thinking of "For information purposes only", but I'm not sure whether that's too  long :thinking:.

javierm (commented):
@taitus I've left a couple of comments; let me know what you think!

While it isn't really related to the hint text, I've noticed something while editing the phases that maybe could fix as part of this pull request :thinking:.

When editing a phase, the "go back" link goes to the wizard step of editing the headings. It should go there only if we're inside the wizard; otherwise it should go to the budget page.

config/locales/en/admin.yml
javierm: I've just realized the whole phases table is used for information purposes only, since it's possible to disable the active phase or select a disabled phase as the active one. So maybe we could add an explanatory text before the table and then we wouldn't have to make the text particularly short :thinking:.

Just an idea; I'm not sure at all whether it's a good one.

I've also noticed that the hint for the "phase enabled" field says "This phase will be public in the budget's phases timeline, as well as active for any other purpose". I haven't found any other purpose :sweat_smile:; is there anything I'm missing? If there isn't any other purpose, we could remove this part of the sentence.

config/locales/es/admin.yml
```diff
@@ -153,7 +154,7 @@ es:
         headings:
           multiple: "Las partidas sirven para dividir el dinero del presupuesto participativo. Aquí puedes ir añadiendo partidas para cada grupo y establecer la cantidad de dinero que se gastará en cada partida."
           single: "Las partidas sirven para dividir el dinero del presupuesto participativo. Como este presupuesto solo tendrá una partida aquí podrás establecer la cantidad de dinero que se gastará en este presupuesto participativo."
-        phases: "Los presupuestos participativos tienen distintas fases. Aquí puedes habilitar o deshabilitar fases y también personalizar cada una de las fases."
+        phases: "Los presupuestos participativos tienen distintas fases. Aquí puedes habilitar o deshabilitar fases y también personalizar cada una de las fases. Este contenido es sólo informativo y es el que se mostrará en la página pública del presupuesto participativo."
```
javierm, line 157: Just to be consistent with the English text:

```suggestion
        phases: "Los presupuestos participativos tienen distintas fases. Aquí puedes habilitar o deshabilitar fases y también personalizar cada una de las fases. La configuración de estas fases se utiliza únicamente con fines informativos y se mostrará en la página pública del presupuesto participativo."
```

config/locales/es/admin.yml
```diff
@@ -212,10 +213,11 @@ es:
         title: "Partidas de %{budget} / %{group}"
     budget_phases:
       edit:
+        dates_help_text: "Sólo con fines informativos"
```
javierm, line 216: Typo: use either "solo" (no accent) or "solamente" or "únicamente".

javierm (commented):
@taitus I've added comments about some texts. I'm not sure about the exact texts we should use, so if you think they aren't that great, just say so :wink:.

javierm (approved):
Great! Let's hope we've increased the number of people who don't assume the current phase is based on the dates of the phases :smile:.

Issue #3988, opened 2020-04-27
Error when delete a user with admin role


decabeza:
### Description

*Type*: Bug

The site returns an **`ArgumentError in Admin::BudgetInvestments#index`** when trying to visit `/admin/budgets/N/budget_investments` and one user with admin role was erased.

The reason is in `/helpers/admin_budget_investments_helper.rb:16`

```rails
  def admin_select_options(budget)
    budget.administrators.with_user.map { |v| [v.description_or_name, v.id] }.sort_by { |a| a[0] }
  end
````
This method is calling the list of Administrators list but the erased one is nil:

`[["admin", 1], [nil, 2]]`

This is unusual because normally a user with the role of admin will rarely delete his account.

### Steps to reproduce
1. Assign admin role to a user from `admin/administrators`.
2. Assign that admin to a budget
3. Login as that user and erase your account in `/users/registrations/delete_form`.
4. Visit `/admin/budgets/N/budget_investments`.

### Current Behavior
When a user deletes their account the relationship with their roles keeps existing.

### Expected Behavior
When a user deletes their account also their roles should be deleted.

## Note

This same error may be occurring in more sections or with different roles.

PR #4809, opened 2022-04-12
Remove roles when block or delete users

decabeza:
## References

* This PR resolves https://github.com/consul/consul/issues/3988
* Depends on pull request #4823

## Objectives

With these changes, when a user is blocked or delete their account, all their roles (Administrator, Valuator, Moderator, Manager and SDG Manager) are removed.

### DIFF
--- a/spec/system/moderation/users_spec.rb
+++ b/spec/system/moderation/users_spec.rb
@@ -94,4 +94,26 @@ describe "Moderate users" do
       expect(page).to have_content "Hidden"
     end
   end
+
+  scenario "Block a user removes all their roles" do
+    admin = create(:administrator).user
+    user = create(:user, username: "Budget administrator")
+    budget = create(:budget, administrators: [create(:administrator, user: user)])
+    debate = create(:debate, author: user)
+    login_as(admin)
+    visit admin_budget_budget_investments_path(budget)
+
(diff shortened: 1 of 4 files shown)


spec/system/moderation/users_spec.rb
```diff
@@ -94,4 +94,28 @@
...
+    admin = create(:administrator).user
+    user = create(:user, username: "Budget administrator")
+    budget = create(:budget, administrators: [create(:administrator, user: user)])
+    debate = create(:debate, author: user)
+    login_as(admin)
+    visit admin_budget_budget_investments_path(budget)
+
+    expect(page).to have_select options: ["All administrators", "Budget administrator"]
+
+    login_as(create(:moderator).user)
```
javierm, line 108: @Senen We're creating a record after starting the browser, which might cause concurrency issues in the database. Maybe we can create the record before starting the browser, or perhaps we can remove this line and block the user while being logged in as an admin :thinking:.

```suggestion
```

javierm: If we remove this line, we can also remove the line `login_as(admin)` which appears later in the test.

Senen: Without those line works fine 👍🏼

javierm (approved):

PR #4814, opened 2022-04-27
Fix background image from Participatory Budgets page

### LINKED ISSUE #4786: Full background images in Participatory Budgets
### User Story

It seems that after the changes introduced in https://github.com/consul/consul/commit/9cc4d5272356b4acef4f46b55345a1e0af3440c2 a margin is added to all elements, including those that should fill the entire background width.

I think CONSUL could keep the Foundation `.row` and `.row.expanded` classes that already allowed to do this.

_Foundation classes already allow this effect._

foundation

This change affects the Participatory Budget header, but also any other section that a fork wants to customize with this same full background image "effect".

### Why is this Needed?

Allow to use a full background image in Participatory Budgeting as before, and display it correctly.

### Steps to reproduce

1. Add a new image from https://demo.consulproject.org/admin/budgets/1/edit.
2. Visit https://demo.consulproject.org/budgets where image is duplicated, and not use full width.

### Current Behavior
before

### Expected Behavior
after

iraline:
## References

Closes #4786

## Objectives

Allow to use a full background image in Participatory Budgeting

## Visual Changes

image

## Notes

In the issue description, it is informed about the _Foundation class_ already allowing this effect. However, in the file _/app/assets/stylesheets/layout.scss_ it is possible to notice that there is a limitation in the _body_ with the value of _$body-margin_ .This value is set in the file  app/assets/stylesheets/_consul_settings.scss , thus preventing the image from expanding to the entire background width. To fix it, tests were done and the best way found was to add the _z-index_ to the _budget-header_

### DIFF
--- a/app/assets/stylesheets/participation.scss
+++ b/app/assets/stylesheets/participation.scss
@@ -989,6 +989,7 @@
   min-height: $line-height * 25;
   padding-bottom: $line-height;
   padding-top: $line-height * 4;
+  z-index: -5;
 
   &.with-background-image {
     background-position: center;


iraline:
I don't understand why the [check](https://github.com/consul/consul/runs/6193727777?check_suite_focus=true#step:5:24) is falling 🤔

javierm:
@iraline Very sorry for that! :pray: At a certain point, there was a change in GitHub Actions on pull requests done by people who don't have write access to the repository. Since then, we haven't been able to run the linters in these cases. We still haven't found how to fix the issue (ideas are welcome! :wink:)

You can manually run the linters on your machine and check everything is OK with `pronto run`.

javierm (approved):
@iraline Great, thanks! :tada: Creating a new stacking context so the background is behind the background of the `::before` pseudoelement works indeed :+1:.

javierm:
@iraline Very sorry! :pray: I noticed an issue right after merging the pull request. I've opened pull request #4825 to fix it.

PR #4818, opened 2022-04-28
System emails design

decabeza:
## References

In the [System Emails section](https://demo.consulproject.org/admin/system_emails) there are 12 emails that are sent from CONSUL. 

As the styles for these emails must be inline, it's complicated to modify them to add custom styles. Also, as the emails were created at different times there were some inconsistencies in the design. 

## Objectives

This PR adds the following changes:

- **Add mailer helpers to simplify customization**: all styles can now be changed from the helper and are common to all emails, maintaining the design. 

- **Remove icon images on mailers**: these images do not contribute anything to the user and are not used on the rest of the website.

- **Add missing heading title on mailers**: for consistency, titles have been added to all emails that did not have one.

## Visual Changes 👨‍🎨 
email_1

email_2


decabeza:
The [failing spec](https://github.com/consul/consul/runs/6218125836?check_suite_focus=true#step:9:74) it seems not related with these changes. 🤔

javierm:
> The [failing spec](https://github.com/consul/consul/runs/6218125836?check_suite_focus=true#step:9:74) it seems not related with these changes

That's right, that spec has been failing all day... maybe there's an update in chromedrive causing it. We'll take a look next week!

Senen:
> > That's right, that spec has been failing all day... maybe there's an update in chrome/chromedriver causing it. We'll take a look next week!
> 
> Hi! We have the same spec failing in our repo that is synced with the CONSUL version 1.4.1. We have added the commits on the [PR 4784](https://github.com/consul/consul/pull/4784) and [PR 4824](https://github.com/consul/consul/pull/4824) but the spec is still failing. Is there any other commit we should add to avoid that spec to fail? 🤔 Thanks in advance!

Hi @microweb10,

I think we fixed it in:
* #4822 😌.

app/views/mailer/budget_investment_unfeasible.html.erb
```diff
@@ -1,24 +1,33 @@
...
-  <p style="font-family: 'Open Sans','Helvetica Neue',arial,sans-serif;font-size: 14px;font-weight: normal;line-height: 24px;">
-  <%= sanitize(t("mailers.budget_investment_unfeasible.new",
-      url: link_to(t("mailers.budget_investment_unfeasible.new_href"),
-      new_budget_investment_url(@investment.budget), style: "color: #2895F1; text-decoration: underline;"))) %>
+  <p style="<%= css_for_mailer_text %>">
+    <%= sanitize(
+      t("mailers.budget_investment_unfeasible.new",
+        url: link_to(t("mailers.budget_investment_unfeasible.new_href"),
+        new_budget_investment_url(@investment.budget),
+        style: css_for_mailer_link)
```
javierm, line 20: I know the original version wasn't easy to read either :relieved:, but how about indenting the last couple of lines so it's obvious that they're part of the `link_to`?

app/views/mailer/proposal_notification_digest.html.erb
```diff
@@ -1,59 +1,56 @@
...
+  </p>
 
   <% @notifications.each do |notification| %>
     <table style="width: 100%; border-top: 1px solid #DEE0E3; padding-top: 12px;">
       <tbody>
         <tr>
           <td style="padding-bottom: 20px; padding-left: 10px;">
-            <p style="font-family: 'Open Sans','Helvetica Neue',arial,sans-serif;font-size: 20px;font-weight: bold;line-height: 24px; margin: 0;">
-              <%= link_to notification.notifiable.title, proposal_url(notification.notifiable.proposal, anchor: "tab-notifications"), style: "color: #2895F1; text-decoration: none;" %>
+            <p style="<%= css_for_mailer_text %>">
```
javierm, line 18: Note that after these changes, it's harder to know where a notification starts :thinking:.

Before these changes:

The notification title looks like a regular link

After these changes:

The notification title looks like a heading

I'm fine with some of the changes, like the links/buttons, and also in the original version, the text with information about the date and author was barely readable and so changing the color is great :+1:.

However, maybe we can keep the font-size of the title bigger, with the information about the date and author immediately below it?

I also wonder whether we should use an `<h2>` tag instead of a `<p>` :thinking:.

javierm (approved):

Issue #3535, opened 2019-05-24
Poll must not be visible in website unless it be marked to appear published


tiagozini:
### What and How

As a administrator 
As a administrator 
I want to a poll could not appear on website unless I have marked this poll in its edition or creation section allowing this poll to be published. Although, always this poll must be visible for administrator for preview how the poll will be formatted in the website.

### Why

As a administrator is necessary that a poll not finished with its questions in the building process doesn´t appear neither in coming section or in the current section of polls.

PR #4819, opened 2022-04-29
Giving the administrator the option to hide polls

arthur1041:
## References

> Related to issue #3535

## Objectives

> Make it possible to the administrator set a poll as published or not at its creation/update. When the poll is set published it will be shown at the polls page **/polls**. If the poll is not published it will not be shown.

## Visual Changes

> Addition of a new checkbox in the create/update poll form:
screenshot_poll_form

## Notes

> Currently the "published" attribute in the poll model is not used to filter polls. I saw that it was not set to true at any moment, so it was not being used. With the new feature brought by this PR, the "published" atribute is now being used. If a poll has the "published" attribute set to true it will be exhibited and vice versa. All the already existing polls need to have it set to true in order for them to be shown, so the following SQL script needs to be executed: `update polls set published = true`

### DIFF
--- a/config/locales/en/activerecord.yml
+++ b/config/locales/en/activerecord.yml
@@ -300,6 +300,7 @@ en:
         geozone_restricted: "Restricted by geozone"
         summary: "Summary"
         description: "Description"
+        published: "Publish poll"
       active_poll/translation:
         description: "Description"
       poll/booth:
--- a/config/locales/en/activerecord.yml
+++ b/config/locales/en/activerecord.yml
@@ -300,7 +300,7 @@ en:
         geozone_restricted: "Restricted by geozone"
         summary: "Summary"
         description: "Description"
-        published: "Publish poll"
+        published: "Published"
       active_poll/translation:
         description: "Description"
       poll/booth:
--- a/spec/system/admin/poll/polls_spec.rb
+++ b/spec/system/admin/poll/polls_spec.rb
@@ -48,6 +48,39 @@ describe "Admin polls", :admin do
     expect(page).not_to have_content "There are no polls"
   end
 
+  scenario "Checkbox is disabled when a poll receive votes" do
+    poll = create(:poll, :published)
+    create(:poll_question, poll: poll)
+    create(:poll_voter, :from_booth, :valid_document, poll: poll)
+
+    visit admin_poll_path(poll)
+    click_link "Edit poll"
+
+    expect(page).to have_css "#poll_published[disabled='disabled']"
+  end
+
+  scenario "System won't allow to update a poll if someone inspect elements and enable the 'published' button" do
+    poll = create(:poll, :published)
+    create(:poll_question, poll: poll)
+    create(:poll_voter, :from_booth, :valid_document, poll: poll)
+
+    visit admin_poll_path(poll)
+    click_link "Edit poll"
+
+    expect(page).to have_css "#poll_published[disabled='disabled']"
+
+    # enable inputs
+    page.execute_script("document.querySelectorAll(\"label[for='poll_published'] input\")[0].disabled=false")
+    page.execute_script("document.querySelectorAll(\"label[for='poll_published'] input\")[1].disabled=false")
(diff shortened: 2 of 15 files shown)


config/locales/en/activerecord.yml
```diff
@@ -300,6 +300,7 @@ en:
         geozone_restricted: "Restricted by geozone"
         summary: "Summary"
         description: "Description"
+        published: "Publish poll"
```
wenderjean, line 303: Do you think something like "Published?" would describe better this attribute?

spec/system/admin/poll/polls_spec.rb
```diff
@@ -33,6 +33,21 @@
     expect(page).not_to have_content "There are no polls"
   end
 
+  scenario "Can see all polls despite it is with the published attribute set to true or not" do
+    create(:poll)
```
wenderjean, line 37: What do you think about giving theses polls a name and use it to make the assertion? There is a good example in the line `12`, "Index show polls list order by starts at date".

config/locales/en/activerecord.yml
arthur1041: Yeah, it makes more sense!

spec/system/admin/poll/polls_spec.rb
arthur1041: I'll give it a try.

arthur1041: Since this test is only to check the poll's quantity and not order, names aren't needed.

config/locales/en/activerecord.yml
arthur1041: Label name was updated.

javierm:
Hi, @arthur1041 :smile:. Thanks for your contribution! :pray:

If I understand it correctly, the original issue was about adding a "preview" feature to polls, similar to what we've got in budgets, so administrators could access non-published polls in the public page. Is that correct?

As far as I can see, with the changes in this pull request, administrators still can't access non-published polls in the public page :thinking:. They can of course access them in the admin area, but that's something they can already do without this pull request by setting the publication date of the poll in the future.

Is there something I'm missing? Please let me know if that's the case!

Another thing to consider is that I think it shouldn't be possible to "unpublish" a poll once it's received votes, just like it isn't possible to delete a poll once it's received votes. Do you think this makes sense?

arthur1041:
> Hi, @arthur1041 smile. Thanks for your contribution! pray
> 
> If I understand it correctly, the original issue was about adding a "preview" feature to polls, similar to what we've got in budgets, so administrators could access non-published polls in the public page. Is that correct?
> 
> As far as I can see, with the changes in this pull request, administrators still can't access non-published polls in the public page thinking. They can of course access them in the admin area, but that's something they can already do without this pull request by setting the publication date of the poll in the future.
> 
> Is there something I'm missing? Please let me know if that's the case!
> 
> Another thing to consider is that I think it shouldn't be possible to "unpublish" a poll once it's received votes, just like it isn't possible to delete a poll once it's received votes. Do you think this makes sense?

Yes. I read the issue again and I saw the admistrator needs to see the poll in the public page even if it's not published. I'll work on a solution for it.

Yeah, it also makes sense not to be able to unpublish a poll once it received votes, I'll add this restriction.

javierm:
@arthur1041 Thanks once again! :pray:

We're currently preparing version 1.5; we'll review this pull request after that. Sorry for the delay! :pray: And thank you for your patience :relieved:.

javierm (commented):
@arthur1041 Sorry for the late review!

I'm also very sorry because apparently there was another team working on a similar issue and due to miscommunication we didn't know about it.

So I've left a few comments regarding this pull request but, at the same time, I don't know whether this pull request could be merged if you apply the changes :facepalm:.

I apologize for this whole situation :pray:.

javierm:
Closing as incompatible because the changes in #4904 aren't compatible with this approach. @arthur1041 Sorry about that!

PR #4820, opened 2022-04-30
Remove unnecessary alert on legislation processes

decabeza:
## Objectives

An alert with the message "Participate in the debate" appears on the [collaborative legislation debate](https://demo.democrateam.com/legislation/processes/1/debate) page.

This text was included more than [5 years ago](https://github.com/consul/consul/commit/36399277ec) but there is no similar text on the platform.

It can also cause several UX problems reported by some forks:

- The message "Participate in the debate" appears if there is a question, but still appears even if the phase is closed and user can't participate.
- Because of it's size it's looks like a button.
- It creates incoherence because there are no similar messages in the rest of the CONSUL sections.

As it's something that I think doesn't contribute anything to the user, I think it is a good idea to remove it. 🧹 

### Before
before

### After
after


javierm (approved):
Great, thanks! :tada:

I love removing useless things :smiling_imp:.

taitus (approved):

PR #4821, opened 2022-04-30
Add census terms page by default

decabeza:
## Objectives

On new installations there is a link `link_to(t("verification.residence.new.terms"), "/census_terms"` on [verification/residence/new](https://github.com/consul/consul/blob/master/app/views/verification/residence/new.html.erb#L79) view, but `census_terms` page doesn't exists by default.

This PR creates the **Census terms** page by default.


javierm (approved):
Nice catch! :clap:

Hopefully, since most installations customize the verification process and nobody reads the terms conditions anyway :smile:, not so many people have clicked on the broken link during the last three years :crossed_fingers:.

PR #4823, opened 2022-05-03
Fix crash deleting admin with assigned budgets

Senen:
## References

While reviewing other PR I found this bug.

## Objectives

Remove dependent budget administrator records when an administrator record is removed.

### DIFF
--- a/spec/models/administrator_spec.rb
+++ b/spec/models/administrator_spec.rb
@@ -38,4 +38,13 @@ describe Administrator do
       expect(administrator.description_or_name_and_email).to eq("Billy Wilder (test@test.com)")
     end
   end
+
+  describe "#destroy" do
+    it "removes dependent budget administrator records" do
+      administrator = create(:administrator)
+      create_list(:budget, 2, administrators: [administrator])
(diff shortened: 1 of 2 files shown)


spec/models/administrator_spec.rb
```diff
@@ -38,4 +38,13 @@
       expect(administrator.description_or_name_and_email).to eq("Billy Wilder (test@test.com)")
     end
   end
+
+  describe "#destroy" do
+    it "removes dependent budget administrator records" do
+      administrator = create(:administrator)
+      create_list(:budget_administrator, 2, administrator: administrator)
```
javierm, line 45: Maybe we can make the code more similar to the code we use in the `spec/models/budget_spec.rb` file :thinking:. This way, we don't have to create a new factory just for this test.

```suggestion
      create_list(:budget, 2, administrators: [administrator])
```

There are other things in this test which could be more similar to the test in `spec/models/budget_spec.rb`, but I don't really care about them :ok_hand:.

javierm (approved):

PR #4827, opened 2022-05-04
Include machine learning settings type

decabeza:
### Description

On the [Configuration settings page](https://demo.democrateam.com/admin/settings) three settings appear without description:

- **Comments Summary**: No description.
- **Related Content**: No description.
- **Tags**: No description.

These settings are related with the **AI / Machine learning** feature. They only should appear on [AI / Machine learning setting page](https://demo.democrateam.com/admin/machine_learning#settings) when the feature is enabled. 🤖 

This PR include machine learning settings type.

### Notes
Actually adding the `machine_learning` type prevents them from being added to the `configuration` list in [models/setting.rb](https://github.com/consul/consul/blob/master/app/models/setting.rb#L16), but as that type is not being used, maybe there is another way to exclude them. 🤔 

#### _Settings removed from Configuration settings list_
settings

### DIFF
--- a/app/models/setting.rb
+++ b/app/models/setting.rb
@@ -8,7 +8,7 @@ class Setting < ApplicationRecord
   end
 
   def type
-    if %w[feature process proposals map html homepage uploads sdg].include? prefix
+    if %w[feature process proposals map html homepage uploads sdg machine_learning].include? prefix
       prefix
     elsif %w[remote_census].include? prefix
       key.rpartition(".").first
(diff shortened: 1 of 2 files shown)


javierm (approved):
Great! :tada:

I'm fine with adding `machine_learning` as a new type :ok_hand:, even if we don't use it. Maybe we could use it instead of the `MachineLearning#script_kinds` method :thinking:, but that's a different story :smile:.

taitus:
It works fine in the testing server and the spec that fails is not related to PR and looks like a flaky.

Issue #4829, opened 2022-05-12
Error 422 when creating admin Tags


microweb10:
### Description
If trying to create a Tag from the admin panel without a name, an error 422 is raised.

### Current Behavior
Error 422 raised.

### Expected Behavior
Form validation saying the name cannot be blank 🤔 

### Steps to reproduce

1. Visit the [CONSUL demo](https://demo.consulproject.org/?locale=en)
2. [Log in](https://demo.consulproject.org/users/sign_in) with admin user
3. Visit the [admin Tags page](https://demo.consulproject.org/admin/tags)
4. Click button/link "Create topic"

joaoGabriel55:
I can take a look @microweb10

joaoGabriel55:
I got to simulate this error, I will develop a solution for it

joaoGabriel55:
Waiting for review

PR #4837, opened 2022-05-25
Fix bug when creating admin tags

joaoGabriel55:
## References

PR from Issue #4829

## Objectives

Fix Error 422 when creating admin Tags

## Visual Changes

image

### DIFF
--- a/app/controllers/admin/tags_controller.rb
+++ b/app/controllers/admin/tags_controller.rb
@@ -1,17 +1,23 @@
 class Admin::TagsController < Admin::BaseController
   before_action :find_tag, only: [:update, :destroy]
+  before_action :tags, only: [:index, :create]
 
   respond_to :html, :js
 
   def index
-    @tags = Tag.category.page(params[:page])
-    @tag  = Tag.category.new
+    @tag = Tag.category.new
   end
 
   def create
-    Tag.find_or_create_by!(name: tag_params["name"]).update!(kind: "category")
+    @tag = Tag.find_or_initialize_by(tag_params)
 
-    redirect_to admin_tags_path
+    save_and_update = @tag.save && @tag.update!(kind: "category")
+
+    if save_and_update
+      redirect_to admin_tags_path
+    else
+      render :index
+    end
   end
 
   def destroy
@@ -20,6 +26,9 @@ class Admin::TagsController < Admin::BaseController
   end
 
   private
+    def tags
+      @tags ||= Tag.category.page(params[:page])
+    end
 
     def tag_params
       params.require(:tag).permit(:name)
--- a/spec/system/admin/tags_spec.rb
+++ b/spec/system/admin/tags_spec.rb
@@ -105,4 +105,15 @@ describe "Admin tags", :admin do
 
     expect(page).to have_content "Soon a category"
   end
+
+  scenario "Create shows validation error when tag name is empty" do
+    visit admin_tags_path
+
+    within("form.new_tag") do
+      fill_in "tag_name", with: ""
+      click_button "Create topic"
+    end
+
+    expect(page).to have_content "can't be blank"
+  end
 end


Gemfile
```diff
@@ -90,6 +90,7 @@ group :test do
   gem "rspec-rails", "~> 5.0"
   gem "selenium-webdriver", "~> 3.142"
   gem "webdrivers", "~> 4.6.0"
+  gem "rails-controller-testing", "~> 0.0.3"
```
Senen, line 93: Using system tests instead of controller tests will allow us to avoid installing additional dependencies for controller testing.

app/models/tag.rb
```diff
@@ -1,4 +1,8 @@
 class Tag < ActsAsTaggableOn::Tag
+  include ActiveModel::Validations
+
+  validates :name, presence: true
+
```
Senen, line 5: The Tag model inherits from ActAsTagabbleOn::Tag [1] class, which inherits from ActiveRecord. Therefore, there is no need to include the ActiveModel::Validations. The presence validation of the name attribute is defined the same in the ActAsTagabbleOn::Tag class so that we can remove it [and the related test](https://github.com/consul/consul/pull/4837/files#diff-933427331ee759197c461d6677d7dd021f4505c2177f1bd70b60c950e5aa2f8fR34).

[1] https://github.com/mbleigh/acts-as-taggable-on/blob/v8.1.0/lib/acts_as_taggable_on/tag.rb#L3

spec/controllers/admin/tags_controller_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    context "without tag name" do
+      it "tag model returns error message" do
+        post :create, params: { tag: { name: "" } }
+        
+        expect(assigns(:tag).errors.messages[:name]).to eq ["can't be blank"]
+      end
+    end
+  end
+end
```
Senen, line 21: We do not usually use controller tests to check user workflows. Instead, we use system tests which emulate users' behaviour. Probably we can replace the controller tests by adding a new system spec to `spec/system/admin/tags_spec.rb` that checks it's not possible to create a new tag when the name is blank.

Here is an example of a system spec that emulates the user workflow:

```
  scenario "Create shows validation error when tag name is empty" do
    visit admin_tags_path

    within("form.new_tag") do
      fill_in "tag_name", with: ""
      click_button "Create topic"
    end

    expect(page).to have_content "can't be blank"
  end
```

spec/models/tag_spec.rb
```diff
@@ -30,6 +30,13 @@
       tag = build(:tag, name: Faker::Lorem.characters(number: 160))
       expect(tag).to be_valid
     end
+
+    context "when name is blank" do
+      it "is not valid" do
+        tag = build(:tag, name: "")
+        expect(tag).not_to be_valid
+      end
+    end
```
Senen, line 39: Probably this test is already covered in the `ActAsTaggableOn::Tag` tests. I think we can remove it 🤔 .

app/controllers/admin/tags_controller.rb
```diff
@@ -1,17 +1,23 @@
...
-    redirect_to admin_tags_path
+    @tag = Tag.category.new(tag_params)
+
+    if @tag.valid?
+      Tag.find_or_create_by!(name: @tag.name).update!(kind: "category")
+      
+      redirect_to admin_tags_path
+    else
+      render :index
+    end
```
Senen, line 20: Here is just another way of doing this. I think it'd cover the same cases 🤔 .

```
  def create
    @tag = Tag.find_or_initialize_by(tag_params)
    if @tag.save && @tag.update!(kind: "category")
      redirect_to admin_tags_path
    else
      render :index
    end
  end
```

or 


```
  def create
    @tag = Tag.find_or_initialize_by(tag_params)
    if @tag.save
      @tag.update!(kind: "category")
      redirect_to admin_tags_path
    else
      render :index
    end
  end
```

Senen (changes requested):
Hi @joaoGabriel55 , thanks for the PR! 🙏🏼 

I left a few comments and examples in the code. Let me know what you think!

app/controllers/admin/tags_controller.rb
joaoGabriel55: Very nice suggestion.

app/models/tag.rb
joaoGabriel55: I did not know that. Very good to know

spec/controllers/admin/tags_controller_spec.rb
joaoGabriel55: I think that both tests are valid, cause `spec/system/admin/tags_spec.rb` I think is a e2e test and the `spec/controllers/admin/tags_controller_spec.rb` is an simple unit test for controller. What do you think?

Senen:
Hi @joaoGabriel55, thank you for taking the time to apply the suggestions. :pray::skin-tone-3:

As all the commits share the same objective and we are changing just two files, I think it'd be nice if you merge all the changes into a single one. You can use `git rebase -i 4f1f7eaa40e2b712edc8e7cc6f8beb42a57ada13` to fix up the last three commits of four into the first one, which has the most meaningful commit message.

Senen:
Hi @joaoGabriel55, thanks for rebasing 🙏🏼.

Just in case you are interested, at a certain point, there was a change in GitHub Actions on pull requests done by people who don't have write access to the repository. Since then, we haven't been able to run the linters in these cases. We still haven't found how to fix the issue. Sorry for that!

I just downloaded this branch and executed Pronto, and there are no issues. 👌🏼

Senen (approved):

PR #4839, opened 2022-06-01
Add comments count on budget investments

decabeza:
## Objectives

As with the debates and proposals, this PR adds the number of comments to the budget investments projects.

## Visual Changes
investment_comments_1
investment_comments_2

## Before releasing version 1.6 (note for maintainers)

Make sure the `shared.comments` text is available in every language where similar texts are available now. Also, delete the "Release notes" label in this pull request after doing so (we're only adding it so we don't forget about this task).


javierm (commented):
@taitus I've left a few comments. Let me know what you think!

javierm (approved):

PR #4843, opened 2022-06-02
Make sure SMTP settings use symbols as keys

javierm:
## References

* Continues pull request #3871

## Background

It's possible that, while editing the `secrets.yml` file, people accidentally delete the starting colon in the SMTP settings, which would result in these settings being ignored.

## Objectives

* Make sure the SMTP settings work even if the starting colon is accidentally deleted


javierm:
My bad, this was fixed in version 1.2, when we upgrade to Rails 5.1. The CONSUL installation where I saw this happening was using an older version. Rails 5.1 [symbolizes these keys automatically](https://github.com/rails/rails/pull/26929).

PR #4850, opened 2022-06-06
Disable Turbolinks for languages with different text direction

Senen:
## References

Not a long time ago, we changed the language selection interface when the application uses four languages or less. 

* Check #4573  for more information

**Steps to reproduce:**
1. Configure the application to use four languages or less. You can add the following to the `application_custom.rb` file
`config.i18n.available_locales = %w[ar en es fr] ` to set it up.

2. Visit the page in any language with a left to right text direction. English an example.
3. Click on the right to left language link at the top of the page. In the example, the Arabic language link.
4. The layout is not loaded correctly as it still uses the `application.css` file instead of the `application-rtl.scss` file.
 
The same happens when you first visit an RTL language and then click on any LTR language link.

**Problem description**
1. We have [separate stylesheets for LTR and RTL languages](https://github.com/consul/consul/blob/master/app/views/layouts/_common_head.html.erb#L6). 
2. As with all the application links the language links were using `Turbolinks` so only the page body changes across requests.
3. When changing from an LTR language to an RTL language or vice-versa we need to use a different stylesheet and change [some other layout attributes](https://github.com/consul/consul/blob/master/app/views/layouts/application.html.erb#L2).

## Objectives
Do a full page reload (disable Turbolinks) when the target language has a different text direction than the current language, so the application layout loads the correct stylesheet and replaces the lang attributes of the `<html>` element.

## Visual changes

**Arabic page before this patch**
Captura de Pantalla 2022-06-06 a las 17 18 09

**Arabic page after this patch**

Captura de Pantalla 2022-06-06 a las 17 18 48

### DIFF
--- a/spec/components/layout/locale_switcher_component_spec.rb
+++ b/spec/components/layout/locale_switcher_component_spec.rb
@@ -74,4 +74,27 @@ describe Layout::LocaleSwitcherComponent do
       expect(page).to have_css "[aria-current]", exact_text: "English"
     end
   end
+
+  context "when the target language has a different text direction than the current language" do
+    let!(:default_locales) { I18n.available_locales }
+
+    before do
+      I18n.available_locales = %i[ar en]
+      I18n.reload!
+    end
+
+    after do
+      I18n.available_locales = default_locales
+      I18n.reload!
+    end
+
+    it "disables Turbolinks for language links" do
+      render_inline component
+
+      expect(page).to have_link "عربى", href: "/?locale=ar"
+      expect(page).to have_css "[href='/?locale=ar'][data-turbolinks=false]"
+      expect(page).to have_link "English", href: "/?locale=en"
+      expect(page).to have_css "[href='/?locale=en'][data-turbolinks=true]"
+    end
+  end
 end
--- a/app/components/layout/locale_switcher_component.rb
+++ b/app/components/layout/locale_switcher_component.rb
@@ -1,5 +1,5 @@
 class Layout::LocaleSwitcherComponent < ApplicationComponent
-  delegate :name_for_locale, :link_list, :current_path_with_query_params, to: :helpers
+  delegate :name_for_locale, :link_list, :current_path_with_query_params, :rtl?, to: :helpers
 
   def render?
     locales.size > 1
@@ -29,7 +29,8 @@ class Layout::LocaleSwitcherComponent < ApplicationComponent
           name_for_locale(locale),
           current_path_with_query_params(locale: locale),
           locale == I18n.locale,
-          lang: locale
+          lang: locale,
+          data: { turbolinks: rtl?(I18n.locale) == rtl?(locale) }
         ]
       end
     end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -6,8 +6,8 @@ module ApplicationHelper
     url_for(request.query_parameters.merge(query_parameters).merge(only_path: true))
   end
 
-  def rtl?
-    %i[ar fa he].include?(I18n.locale)
+  def rtl?(locale = I18n.locale)
+    %i[ar fa he].include?(locale)
   end
 
   def markdown(text)


config/environments/test.rb
```diff
@@ -4,7 +4,7 @@
   # Some tests require the following languages (en, es, fr, nl, pt-BR)
   # so we override the available languages for the test environment.
   config.i18n.default_locale = :en
-  config.i18n.available_locales = %w[de en es fr nl pt-BR zh-CN]
+  config.i18n.available_locales = %w[ar de en es fr nl pt-BR zh-CN]
```
javierm, line 7: Including extra locales in the tests is tricky because the pages loading translatable fields become slower and so does the test suite :thinking:. Some of the tests even fail sometimes because loading the page takes too long.

How about removing this change and, in the test, doing:

```ruby
    let!(:default_enforce) { I18n.enforce_available_locales }

    before do
      I18n.enforce_available_locales = false
      allow(I18n).to receive(:available_locales).and_return(%i[ar en es fr])
    end

    after { I18n.enforce_available_locales = default_enforce }
```

However, this would require changing the tests so we write `expect(page).to have_link href: "/?locale=ar"` instead of `expect(page).to have_link "عربى", href: "/?locale=ar"`. If we want to keep the test as it is, we'd have to write something like:

```ruby
    let!(:default_locales) { I18n.available_locales }

    before do
      I18n.available_locales = %i[ar en es fr]
      I18n.reload!
    end

    after do
      I18n.available_locales = default_locales
      I18n.reload!
    end
```

Since reloading the I18n backend takes a second or two, if we follow this approach, I'd suggest doing this just for the test we've added in this pull request, and not for the "renders a list of links" test.

What do you think? :thinking:

Come to think about it, there might be a couple of languages we could remove from this array and then enable them only when needed and that might speed up some tests :thinking:. I'll take that into consideration :wiggum:.

spec/components/layout/locale_switcher_component_spec.rb
```diff
@@ -54,19 +54,37 @@
...
       expect(page).to have_link "Español", href: "/?locale=es"
       expect(page).to have_link "Français", href: "/?locale=fr"
       expect(page).not_to have_css "form"
     end
 
+    it "disables Turbolinks for languages with different text direction than the current one" do
+      render_inline component
+
+      expect(page).to have_link "عربى", href: "/?locale=ar" do |link|
+        link.has_xpath?("//a[@data-turbolinks='false']")
```
javierm, line 75: I'm not that  familiar with xpath, but I've found something funny:

```ruby
         link.has_xpath?("//a[@data-turbolinks='false']") # true
         link.has_xpath?("//a[@data-turbolinks='true']") # true
```

So it looks like this assertion isn't working as expected and the test would pass even if we change the method and introduce a bug :thinking:.

I've tried a way to test this and haven't found an elegant one since it looks like we're a bit limited in component tests due to `page` being a `Capybara::Node::Simple`. This is the best I've found so far:

```ruby
      expect(page).to have_link "عربى", href: "/?locale=ar"
      expect(page).to have_css "[href='/?locale=ar'][data-turbolinks=false]"
```

javierm (commented):
@Senen Great catch! Yes, Turbolinks needs to be reloaded when we remove CSS from the page, or it might still use it.

I've left a couple of comments regarding the tests. Let me know what you think!

config/environments/test.rb
Senen: Hi @javierm, I agree with you! I finally opted for the second option and created a new context in the component test, so it does not affect other tests. 

>Come to think about it, there might be a couple of languages we could remove from this array and then enable them only when needed and that might speed up some tests 🤔. I'll take that into consideration :wiggum:.

Are you talking about the `available_locales` defined at `config/environments/test.rb` file? 🤔 Probably we can remove some from there and speed up the suite a bit. I annotated that into my Wigum typing machine 😉 .

spec/components/layout/locale_switcher_component_spec.rb
Senen: Ups, I did not notice this. Applied your suggestion which fails and succeeds as expected.

javierm (approved):

PR #4859, opened 2022-06-14
Allow prioritizing multiple answers in polls

decabeza:
## References

* Depends on pull request #4904

## Objectives

Add Poll Votation types. Now administrators can select between three types of votation when creates a new question.

- **Unique answer**: it's only possible to answer one time to the question.
- **Multiple answers**: allows to choose multiple answers. It's possible to set the maximum number of answers.
- **Multiple prioritized answer**: allows to choose more than one answer and they will be prioritized. It's possible to set the maximum number of answers chosen.

### Administration panel
admin_1

admin_2

admin_3

### User's views
polls_user


javierm:
Closing since we've decided to make open answers a priority and the changes we're doing in #5540 and #6061 aren't compatible with these changes.

Issue #4834, opened 2022-05-20
Allow include images on newsletters 


decabeza:
### Description

It would be nice to allow include images on [newsletters sent from admin](https://demo.consulproject.org/admin/newsletters/new).

I've started a PR on another fork with the following changes, but I'm stuck on the last one: 😓 

1. Change the [newsletters form](https://github.com/consul/consul/blob/master/app/views/admin/newsletters/_form.html.erb#L8) adding the class "admin" on  `class: "html-area admin" %>` to show the images button.

2. Change the [newsletter view](https://github.com/consul/consul/blob/master/app/views/mailer/newsletter.html.erb#L3) so the images renders correctly, replacing: 

```erb
<%= auto_link_already_sanitized_html wysiwyg(@newsletter.body) %> 
```
to:
```erb
<%= auto_link_already_sanitized_html AdminWYSIWYGSanitizer.new.sanitize(@newsletter.body) %> 
```

3. Finally it would be necessary to change the path in **ckEditor** so that the image use an absolute path instead of a relative path so that they work from an external site, an email in this case.

In version 1.4.1. I could change the file [ckeditor/picture.rb](https://github.com/consul/consul/blob/master/app/models/ckeditor/picture.rb) and add in URL the `Setting["url"]`, something like:

```erb
url: "#{Setting["url"]}/ckeditor_assets/pictures/:id/:style_:basename.:extension
```

But I've seen that currently that file [ckeditor/picture.rb](https://github.com/consul/consul/blob/master/app/models/ckeditor/picture.rb) is very different. 😅 

Where could we change that URL value for ckEditor images?  🤔 

If someone helps me to solve my doubt 🤞 I can send a PR to add this to CONSUL. 😌

lucialuzuriaga:
This issue should be worked on once #5458 is complete, to avoid redoing once CKEditor4 is replaced.

PR #4865, opened 2022-06-29
Allow including images on newsletters

decabeza:
## References

https://github.com/consul/consul/issues/4834

## Objectives

Allow include images on newsletters `admin/newsletters/new`.

## Visual Changes
newsletter_image

## Notes

I have added some specs to test that the image works, but I had to add a `Setting["url"] = ""` in `spec/system/ckeditor_spec.rb` because I have not been able to test an image with a full external URL 😅, and I think it should be added before merge to properly test these changes. 🙏🏻

### DIFF
--- a/app/models/ckeditor/picture.rb
+++ b/app/models/ckeditor/picture.rb
@@ -5,7 +5,8 @@ class Ckeditor::Picture < Ckeditor::Asset
   validates :storage_data, file_content_type: { allow: /^image\/.*/ }, file_size: { less_than: 2.megabytes }
 
   def url_content
-    rails_representation_url(storage_data.variant(resize: "800>"), only_path: true)
+    url = root_url(ActionMailer::Base.default_url_options)
+    "#{url}#{rails_representation_url(storage_data.variant(resize: "800>"), only_path: true)}"
   end
 
   def url_thumb
--- a/app/views/admin/newsletters/_form.html.erb
+++ b/app/views/admin/newsletters/_form.html.erb
@@ -5,7 +5,7 @@
                                                       @newsletter[:segment_recipient]) %>
   <%= f.text_field :subject %>
   <%= f.text_field :from %>
-  <%= f.text_area :body, class: "html-area" %>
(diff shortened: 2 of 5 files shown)


javierm:
@decabeza Thanks for the pull request!

Adding a note so we don't forget about it: we need to re-check how CKEditor works when ActiveStorage is configured to use external services (S3, Azure, ...) and what would happen if we added `Setting["url"]` in this case :thinking:.

In the configuration to send emails we use `Rails.application.secrets.server_name` instead of `Setting["url"]` in order to have links with absolute URLs, but this is a different case because links are generated when the email is sent and CKEditor content is stored in the database :thinking:. I'm wondering which one we should use here (I also wonder sometimes why people need to define the same URL in two different places :sweat_smile:, but that's another story).

javierm:
I've opened pull request #5000 to remove the `Setting["url"]`, so I guess we'll use the `server_name` or directly `ActionMailer::Base.default_url_options[:host]` (or a method in the `Tenant` class once we add multitenancy) to generate the absolute URL. I wonder if the `ActiveStorage::Current.host` method could help in this case :thinking:.

javierm:
Testing shows that using a service like S3 also generates a relative URL, so adding the host seems to work in this case too :+1:.

So we might just need to update this pull request so it uses the host defined in `default_url_options` instead of using the (already removed) `Setting["url"]` :thinking:.

lucialuzuriaga (commented):
Hi @decabeza, thank you for your pull request. This is indeed a gap in Newsletters and a feature we would like to have.

That said, I don't think this specific implementation is the right path to merge at this point. It predates several major Rails upgrades, and more importantly, it's built entirely around CKEditor 4's toolbar config, which is already on our radar for replacement. Reviving CKEditor-specific code now means redoing it again once that migration happens.

I'd like to close this, and carry https://github.com/consuldemocracy/consuldemocracy/issues/4834 as an upcoming feature for whichever editor replaces CKEditor 4, rather than losing the feature request. Thank you!

PR #4872, opened 2022-07-11
GraphQL mutations API

tonekk:
## References

https://github.com/consul/consul/pull/4766

## Objectives

Supporting a subset of operations that are possible using consul.
This pull request does not aim to provide a complete API.
Currently these mutations are supported:

```javascript
// Commenting
addCommentToDebate
addCommentToPoll
addCommentToProposal
addReplyToComment
deleteComment

// Voting
castVoteOnComment
castVoteOnDebate

// Authentication
login
register

// Attachment handling
// NOTE: Uploading attachments is done via normal HTTP
deleteDocument
deleteImage

// Creating and answering polls
startPollForProposal
provideAnswerToPollQuestion

// Creating and updating proposals
publishProposal
submitProposal
supportProposal
updateProposal

// Creating and updating debates
startDebate
updateDebate
```

## Visual Changes

No.

## Notes

To test the API, I'm providing a collection file use with [insomnia](https://github.com/Kong/insomnia): https://gist.github.com/tonekk/6846ba6f60609ca2faf052837d91f407

### Example implementation

The open source [Smart Village App](https://github.com/ikuseiGmbH/smart-village-app-app) is a modular system to build apps for villages and small cities. Its in-app citizen participation service is based on consul and includes the modules for debates, proposals and polls.

#### Proposals
image (2)

#### Debates

image (3)

#### Polls

image (4)


tonekk:
@javierm any idea why all of these tests are failing that seem to have nothing to do with the contents of my PR?

These lines in `community.rb`:
```ruby
users_participants = users_who_commented +
                            users_who_topics_author +
                            author_from_community
```

also prevent me from accessing the Proposal dashboard locally...

javierm:
Hi, @tonekk :smile:.

I've just had a first look at the pull request. You've certainly done a lot of work; thanks for that!

Before doing a review in detail, here are some general things I've seen:

1. First, as you mentioned, we need to change the column type to jsonb. I know the default is json but, in our case, jsonb seems to be the right one (see lynndylanhurley/devise_token_auth#841)
2. It would be really cool to have some tests showing that only users with the proper permissions can perform write operations
3. Related to the previous one, how about adding tests checking each mutation works as expected? Are you familiar with git interactive rebase? If so, it would be great if you could add a test on the same commit that the respective mutation was added. If not, just add the tests in new commits (one commit per mutation) and we'll rebase them right before merging :ok_hand:.
4. Pronto is reporting some issues. Sorry, we still haven't found how to automate GitHub Actions to report it automatically; could you execute `pronto run` on your machine and fix the issues (you may ignore the LineLength ones)? :pray:
5. Are we following any naming conventions for the mutation classes :thinking:? If so, could you point them out? If not, how about naming them like `ActionnameControllername` or `Controllername::Actionname`? For example, if the mutation does a similar thing than what is done in the DebatesController#create action, the mutation could be named `CreateDebate` or `Debates::Create`. Some other names that would come up if we followed this convention would be `DestroyComment` (or `Comments::Destroy`), `AnswerPollQuestion` (or `Polls::Questions::Answer`), ... Feel free to suggest other possible conventions :wink:
6. We're copying some authorization rules. For instance, we use `image.imageable&.author&.id == user.id` in the mutation and `can [:destroy], Image, imageable: { author_id: user.id }` in the abilities. Would it be possible to use the `can?` method in the mutations?
7. Sometimes we're copying the controller logic. For example, in the `ProvideAnswerToPollQuestion` the logic is very similar to the one in the `Polls::QuestionsController`. I think there's about a 100% chance that in the future someone (probably me :sweat_smile:) changes the logic in the controller and forgets to do the same in the mutation, so it would be awesome if we could find ways to reduce the duplication as much as we can

As usual, feel free to disagree with anything I've mentioned :wink:.

And, once again, thank you for taking care of this huge task. If you need some help or have some questions, just ask :wink:.

tonekk:
Hi @javierm , finally got some time to move forward on this.
I agree with everything you say and yes, I know how to utilize git interactive rebase (including the fixup feature).
Regarding the naming: I asked a friend who is really proficient in GraphQL and he suggested to make the naming more natural. Apparently that is the supposed to be the "GraphQL way".
In a Rails context though, where there is a maximum amount of convetion over configuration, it might be better to stick to the naming that's already in place.
I'm going to do everything else first and then think about the naming again.

- [ ] Change column to `jsonb`
- [ ] Add tests
- [ ] Satisfy linter
- [ ] Reconsider naming of Mutations
- [ ] Use CanCan rules instead of rewriting them
- [ ] Extract Controller logic to make it reusable in mutations

tonekk:
### Naming

I think I'm going to stick to the naming used in the docs of [graphql-ruby](https://graphql-ruby.org/mutations/mutation_classes) - `ActionModel` (e.g. `CreateComment`).
The only problem here are the mutations like `AddCommentToDebate`. One suggestion here would be to use `CreateCommentForDebate`.

And what to do when there is no Model? `CastVoteOnComment` is an example. We could just omit the extra verb and name it `VoteForComment` - which sounds a little odd in my ears. Maybe `VoteOnComment` is better. But I don't know how to derive a convention from this...

javierm:
Hi, @tonekk :smile:.

Naming is super-hard, indeed, particularly if we take into account that naming in this project is already inconsistent in some places :sweat_smile:.

Regarding the question about voting a comment, in the `CommentsController` there's an action named `vote`, so in this case the mutation could be named `VoteComment` :thinking:.

javierm:
@tonekk By the way, I see there are some tests failing. The reason seems to be the `include GraphqlDevise::Concerns::Model` in the users model, which probably adds a validation making email mandatory.

In CONSUL, the email is *not* required. While it is required when you sign up using the web interface, some public administrations provide a place where people can show their ID and a public worker signs them up manually and so they can participate in public processes without having an email address.

So I guess we need to find a way not to add the extra validation here :thinking:.

javierm:
A possible workaround could be disabling the default callbacks (`config.default_callbacks = false` in the devise auth token initializer). Not sure about the possible implications, since that would also mean that the `sync_uid` method isn't called when saving a record, but worst case scenario, I guess we could add that callback to our `User` model :thinking:.

javierm:
@tonekk We've got an open pull request to [Bump graphql from 1.12.14 to 1.13.15](https://github.com/consul/consul/pull/4955). Since we don't want to interfere with the task you're currently doing, what do you prefer? Shall we update the gem before merging this pull request or after doing so?

pwilimzig:
@tonekk Do you still plan to finish this topic?

javierm:
Hi, @tonekk :smile:.

As mentioned earlier, we're going to merge a pull request to upgrade this gem to version 1.13.15. Hope that's OK!

By the way, are you still working on this pull request? :thinking:

pwilimzig:
Hi @javierm Unfortunately @tonekk has left our team. I will check with my other teammates to see who can take over.

pwilimzig:
Yes, the API is integrated with the Consul module within the Smart Village App framework. Currently, the "Bad Belzig App" is the sole application leveraging this module. To access it, navigate to the "Services" section and select "Beteiligung." Up to this point, we have not received any feedback indicating issues or complaints.

lucialuzuriaga (commented):
Hi @tonekk! This is genuinely solid groundwork, providing a sensible subset of mutations worth crediting for.

Since several years have passed and we're currently working on a GraphQL to REST API migration and expansion, this would be a useful reference. For now, I'm closing this PR given the current plan and the considerable work involved in making it mergeable four years later. However, if you're able to pick this up again, please let us know, and we can revisit it.

PR #4878, opened 2022-07-21
Show verification info only if verification is enabled

decabeza:
## References

* Depends on pull request #4902

## Objectives

Now, even if the "**Skip user verification**" setting is enabled at `admin/settings#tab-feature-flags` on the "**My account**" `/account` page the text "To perform all the actions verify your account. / Account verified" is displayed which can be a bit confusing for users.

This PR hides this information if the "Skip user verification" setting is active.

### Before
before

before2

### After
after

### DIFF
--- a/app/assets/stylesheets/layout.scss
+++ b/app/assets/stylesheets/layout.scss
@@ -1228,10 +1228,6 @@ form {
     margin-right: $line-height / 2;
   }
 
-  .verify-account {
-    padding-right: $line-height / 2;
-  }
-
   .final-votes-info {
     background: $warning-bg;
     border: 1px solid $warning-border;
@@ -1268,16 +1264,6 @@ form {
   line-height: rem-calc(32);
   height: rem-calc(32);
   vertical-align: top;
-}
-
-.user-permissions {
-
-  p {
-    span {
-      color: $text-medium;
-      font-size: rem-calc(12);
-    }
-  }
 }
 
 .notifications-list {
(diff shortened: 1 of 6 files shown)


javierm:
Great, since we're touching this part of the application, I think we'll add some changes to fix an accessibility issue we've got here. We say "With you account you can (...) icon-x Support proposals". But the icon isn't shown for screen readers, so people using them actually hear that they *can* support proposals when they cannot.

app/assets/stylesheets/layout.scss
```diff
@@ -1270,16 +1266,6 @@ form {
   vertical-align: top;
 }
 
-.user-permissions {
```
Senen, line 1273: Not a big deal, but what do you think about removing the class `user-permissions` from the following views?
* app/views/account/show.html.erb
* app/views/management/_user_permissions.html.erb
* app/views/verification/letter/new.html.erb
* app/views/verification/residence/new.html.erb

It seems now that class is useless 🤔

Senen (commented):
Hi @javierm, I left you a comment regarding a minor thing.

app/assets/stylesheets/layout.scss
javierm: Interesting! I'd vote for keeping the class, at least for now. There might be people using custom styles for this class, and removing it would make the element harder to style :thinking:.

Senen: Ok, nice catch!

PR #4894, opened 2022-08-03
Create user-manual-example.pdf

johnfelipe:


javierm:
Hi, @johnfelipe.

User manuals are part of the [webiste repository](https://github.com/consul/website). This website is publicly accessible through the [CONSUL project page](https://consulproject.org/); you can download these manuals from there.

I'm not sure whether the manuals you're looking for are already available. If they aren't, would you like to open an issue in the mentioned repository and close this one?

By the way, the team handling these manuals will be on holidays for the next 2-3 weeks, so it might be a while before they reply.

Hope this helps!

javierm:
Continuing on consul/website#29. Thanks!

PR #4985, opened 2022-09-02
Show admin heading stats for the current budget

microweb10:
## Objectives

* Do not show information about participants per district from previous budgets in the stats page
* Use the same logic to calculate budget stats in the public section and in the admin section
* Fix invalid HTML in admin budget stats tables

### BEFORE
before

### AFTER
after


app/controllers/admin/stats_controller.rb
```diff
@@ -82,7 +82,7 @@ def budget_balloting
 
     @vote_count_by_heading = @budget.lines.group(:heading_id).count.map { |k, v| [Budget::Heading.find(k).name, v] }.sort
 
-    @user_count_by_district = User.where.not(balloted_heading_id: nil).group(:balloted_heading_id).count.map { |k, v| [Budget::Heading.find(k).name, v] }.sort
+    @user_count_by_district = User.where(balloted_heading_id: @budget.headings.map(&:id)).group(:balloted_heading_id).count.map { |k, v| [Budget::Heading.find(k).name, v] }.sort
```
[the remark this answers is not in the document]
microweb10: I know. 😅 
Feel free to refactor if you wish. 😉

spec/system/admin/stats_spec.rb
```diff
@@ -182,7 +182,7 @@
       let(:budget) { create(:budget, :balloting) }
       let(:group) { create(:budget_group, budget: budget) }
       let(:heading) { create(:budget_heading, group: group) }
-      let!(:investment) { create(:budget_investment, :feasible, :selected, heading: heading) }
+      let(:investment) { create(:budget_investment, :feasible, :selected, heading: heading) }
```
microweb10, line 185: This `let!` was not really needed, but it is for the new spec. ☺️

javierm:
Looks like the code was originally a bit more broken than we thought because the `balloted_heading_id` field is updated when someone votes on any heading, meaning votes for old headings aren't counted if there are votes on new headings :sweat_smile:.

I'll have a look.

app/components/admin/stats/budget_supporting_component.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    def stats
+      @stats ||= Budget::Stats.new(budget)
+    end
+
+    def headings_stats
+      @headings_stats ||= stats.headings
+    end
+
+    def vote_count
```
taitus, line 18: Trying not to think that the supports are in the `Vote` table and the votes in the `Ballot::Line` table 😀 , what would you think if this method `vote_count` inside the Budget**Supporting**Component that calls a stats.total_**supports** we call it "**support**_count" 🤔

javierm: Yes, there's quite a mess with supports, votes and ballots in investments :sweat_smile:.

The main argument I see against this change is that `vote_count` is consistent with the `admin.stats.budget_supporting.vote_count` translation key we're using, and unfortunately renaming the translation key might result in losing the translations :relieved:.

No strong preference, but I vote for keeping `vote_count` for now.

What do you think?

taitus (approved):

PR #4993, opened 2022-09-20
Refactor and simplify public polls views

Senen:
## References

It depends on PR:
* #4904 

## Objectives

Simplify public polls code to make adding or changing feature logic easier.

1. Remove `token` from `Poll::Voter` model as it is no longer shown to users
2. Move partials to components and simplify some logic
3. Make poll work in browsers with disabled Javascript

## Visual Changes

None


app/components/polls/questions/answers_component.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  def question_answers
+    question.question_answers
+  end
+
+  private
+
+    def user_answers
+      @user_answers ||= question.answers.by_author(current_user.id)
```
javierm, line 26:
```suggestion
      @user_answers ||= question.answers.by_author(current_user)
```

No need to provide the ID and, even though right now we only call `user_answers` in methods that are only called if there's a `current_user`, this change will also make this method work with changes that make `current_user` `nil` at this point.

Senen: Done. Thanks

taitus (approved):

PR #4997, opened 2022-09-27
Add description to Legislation questions

### LINKED ISSUE #4907: Legislation debates can't be formatted
## Description

In the admin section, we're currently using a `<textarea>` field to fill in the title of a legislation process debate/question. However, the title is rendered as plain text inside tags such as `<h3>` or `<h4>`, so entering paragraphs here doesn't make any difference.

We should change the type of this field from textarea to a regular text field to avoid confussion.

In order to allow proper formatting of paragraphs, we could add a new `description` column to the `legislation_question_translations` table, add a new field to the form, and then render the description in the `legislation/questions/show.html.erb` file.

decabeza:
## References

This PR resolves #4907.

## Objectives

- Replace Legislation question title textarea to text field.
- Add description to Legislation questions.

### Legislation question admin form _(BEFORE)_
1_before

### Legislation question admin form _(AFTER)_
1_after

### Legislation question admin index
2_questions

### DIFF
--- a/app/views/legislation/questions/show.html.erb
+++ b/app/views/legislation/questions/show.html.erb
@@ -29,6 +29,7 @@
   <div class="row">
     <div class="small-12 medium-9 column">
       <h3 class="quiz-question"><%= @question.title %></h3>
+      <%= AdminWYSIWYGSanitizer.new.sanitize(@question.description) %>
       <div class="debate-questions" id="legislation-answer-form">
         <%= render "answer_form", process: @process, question: @question, answer: @answer %>
       </div>
--- a/app/views/legislation/questions/show.html.erb
+++ b/app/views/legislation/questions/show.html.erb
@@ -5,7 +5,7 @@
     <div class="small-12 medium-9 column">
       <div class="quiz-title">
         <p class="quiz-header-title"><%= t("legislation.questions.show.title") %></p>
-        <h4><%= link_to @process.title, @process %></h4>
+        <strong><%= link_to @process.title, @process %></strong>
       </div>
     </div>
     <div class="small-12 medium-3 column">
@@ -28,7 +28,7 @@
   </div>
   <div class="row">
     <div class="small-12 medium-9 column">
-      <h3 class="quiz-question"><%= @question.title %></h3>
+      <h1 class="quiz-question"><%= @question.title %></h1>
       <%= AdminWYSIWYGSanitizer.new.sanitize(@question.description) %>
       <div class="debate-questions" id="legislation-answer-form">
         <%= render "answer_form", process: @process, question: @question, answer: @answer %>
@@ -37,7 +37,7 @@
 
     <aside class="small-12 medium-3 column">
       <div id="social-share" class="sidebar-divider"></div>
-      <h3><%= t("legislation.questions.show.share") %></h3>
+      <h2><%= t("legislation.questions.show.share") %></h2>
       <%= render "/shared/social_share", title: @question.title, url: legislation_process_question_url(@question.process, @question) %>
     </aside>
   </div>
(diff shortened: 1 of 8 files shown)


app/views/admin/legislation/questions/index.html.erb
```diff
@@ -32,6 +33,7 @@
             <td>
               <%= link_to question.title, edit_admin_legislation_process_question_path(@process, question) %>
             </td>
+            <td><%= sanitize_and_auto_link(question.description) %></td>
```
javierm, line 36: The only place that comes to mind where we show CKEditor descriptions in admin tables are poll question answers, where we use `<%= wysiwyg(answer.description) %>` to display them :thinking:.

So I'd suggest either using `wysiwyg`, removing the description, or changing the table in poll question answers. Personally I'd vote to remove it because descriptions make tables harder to read when they're longer than one or two sentences :thinking:.

app/views/legislation/questions/show.html.erb
```diff
@@ -29,6 +29,7 @@
   <div class="row">
     <div class="small-12 medium-9 column">
       <h3 class="quiz-question"><%= @question.title %></h3>
+      <%= sanitize(@question.description) %>
```
javierm, line 32: Since this is a text edited in a `html-area admin` textarea:

```suggestion
      <%= AdminWYSIWYGSanitizer.new.sanitize(@question.description) %>
```

config/locales/en/activerecord.yml
```diff
@@ -447,6 +447,7 @@ en:
         question_options: Options
       legislation/question/translation:
         title: Question
+        description: "Description"
```
javierm, line 450: I think we don't need this line, since it's already defined in `I18n.t("attributes.description")` :thinking:.

config/locales/en/admin.yml
```diff
@@ -693,6 +693,7 @@ en:
           submit_button: Create question
         table:
           title: Title
+          description: "Description"
```
javierm, line 696: We could also remove this line if we used `human_attribute_name` in the table header, or if we didn't show the description in the questions table. Less work for our translators :smile:.

db/migrate/20220924224924_add_description_to_legislation_question_translations.rb
```diff
@@ -0,0 +1,5 @@
+class AddDescriptionToLegislationQuestionTranslations < ActiveRecord::Migration[6.0]
+  def change
+    add_column :legislation_question_translations, :description, :string
```
javierm, line 3: I think we usually use `text` for descriptions :thinking:.

```suggestion
    add_column :legislation_question_translations, :description, :text
```

spec/system/admin/legislation/questions_spec.rb
```diff
@@ -57,9 +59,11 @@
       click_link "Question 2"
 
       fill_in "Question", with: "Question 2b"
+      fill_in_ckeditor "Description", with: "Description 2b"
       click_button "Save changes"
 
       expect(page).to have_content "Question 2b"
+      expect(page).to have_selector(".html-area.admin", text: "Description 2b", visible: :hidden)
```
javierm, line 66:
```suggestion
      expect(page).to have_ckeditor "Description", with: "Description 2b"
```

app/views/legislation/questions/show.html.erb
```diff
@@ -29,6 +29,7 @@
   <div class="row">
     <div class="small-12 medium-9 column">
       <h3 class="quiz-question"><%= @question.title %></h3>
```
javierm, line 31: Since we're now allowing `<h2>` and `<h3>` tags in the description, it might be a good time to fix this broken `<h3>` tag :thinking:. Since its content is also in the `<title>` element, and in order to be consistent with other `show` actions in the public area, IMHO this tag should be `<h1>`. We might need to adjust the styles accordingly.

Then we can use `<h2>` in the "Share" heading and replace the broken `<h4>` tag above with something that isn't a heading tag :thinking:.

What do you think?

javierm (approved):

PR #5003, opened 2022-10-05
Add SDG images in Galego

DiegoZen:
Images for the translation of the SDG module into Galician have been added.

Text translations are here:
https://github.com/consul/consul/commit/c23a8a1c01e6a6152fc04173fe957e16de50c4cd


javierm (commented):
Hi, @DiegoZen. Thank you for the pull request! :tada:

Would it be possible to convert the images outside the `original` folder (the ones inside the `original` folder are fine as they're now :ok_hand:) to a size of 200x200 pixels? So they're consistent with the icons in other languages.

In the past I've done it with ImageMagick with `magick *.png -resize 200x200>` but of course feel free to use any tool :wink:.

And could you add a link to the source where you got these images in the commit message? :pray:

Thanks!

DiegoZen:
I have two Github users and I mix them a lot. That explains the rebased commit.

I also changed original images 1 and 2 because weren't the same size as the others.

javierm (approved):
Great, thanks! :tada:

PR #5012, opened 2022-10-13
Add poll questions that accept multiple answers per user

Senen:
## References

This PR includes the multiple questions part of the original PR:

* #4859

This PR depends on:
*  #4993

## Objectives
Add a new type of question where the users can select multiple answers.

## Visual Changes
Check original PR screenshots related to multiple answers.

The only difference for questions with multiple answers is that now buttons are disabled when the user reaches the maximum of answers allowed.

Captura de Pantalla 2022-10-14 a las 11 31 28


taitus (approved):

app/models/poll/answer.rb
```diff
@@ -21,4 +22,27 @@ def save_and_record_voter_participation
...
+  private
+
+    def max_votes
+      return if !question || question&.unique? || persisted?
+
+      author.reload
+      author.lock!
+
+      if question.answers.by_author(author).count + 1 > question.max_votes
+        errors.add(:answer, "Maximum number of votes per user exceeded")
```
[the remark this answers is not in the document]
Senen: As this message is not rendered anywhere, I wasn't sure if we should translate it or not 🤔

Senen:
Hi @javierm, thanks for the review. I think I applied all the suggestions 👌🏼 . The only thing I kept the same is the text mentioned in your comment: [#discussion_r996903531](https://github.com/consul/consul/pull/5012#discussion_r996903531). Not sure what to do about it 🤔

app/views/admin/poll/questions/show.html.erb
```diff
@@ -28,6 +28,21 @@
         <%= link_to @question.proposal.title, proposal_path(@question.proposal) %>
       </p>
     <% end %>
+
+    <% if @question.votation_type.present? %>
+      <p>
+        <strong><%= t("admin.polls.votation_type.title") %></strong>
+        <br>
+        <%= VotationType.human_attribute_name("votation_type/vote_type.#{@question.vote_type}") %>
```
javierm, line 36: I think we can simplify this line a bit:

```suggestion
        <%= VotationType.human_attribute_name("vote_type.#{@question.vote_type}") %>
```

app/components/polls/questions/answers_component.rb
```diff
@@ -10,16 +10,18 @@ def already_answered?(answer)
     user_answers.find_by(answer: answer.title).present?
```
javierm, line 10: Maybe we can now reuse the new method :thinking:.

```suggestion
    user_answer(answer).present?
```

app/components/polls/questions/answers_component.rb
```diff
@@ -10,16 +10,18 @@ def already_answered?(answer)
...
-    user_answers.any? do |vote|
-      vote.updated_at < current_user.current_sign_in_at
-    end
-  end
-
   def question_answers
     question.question_answers
   end
 
+  def user_answer(question_answer)
```
javierm, line 17: I see the method `already_answered?` receives a parameter called `answer` but the `user_answer` and `disable_answer?` methods receive a parameter called `question_answer`. It took me a while to realize that in all cases the received parameter was the same object :sweat_smile:; what do you think about using the same parameter name for all these methods? Either name is fine with me.

app/components/polls/questions/answers_component.html.erb
```diff
@@ -1,16 +1,22 @@
...
     <% question_answers.each do |question_answer| %>
-      <% if already_answered?(question_answer) && !voted_before_sign_in? %>
-        <span class="button answered"
-              title="<%= t("poll_questions.show.voted", answer: question_answer.title) %>">
+      <% if already_answered?(question_answer) %>
+        <%= button_to question_answer_path(question, user_answer(question_answer)),
+                      method: :delete,
+                      remote: true,
+                      title: t("poll_questions.show.voted", answer: question_answer.title),
+                      class: "button answered expand",
```
javierm, line 9: Does the `expand` class do anything? :thinking: Since it isn't present in the button to remove the question, I guess it's a typo.

```suggestion
                      class: "button answered",
```

Senen: No. Removed.

PR #5036, opened 2022-11-25
Simplify locking current user when voting

javierm:
## References

* We introduced the lock in commit 36e452437 from pull request #5012

## Objectives

* Simplify code
* Avoid unneeded changes in the `current_user` object


Senen (approved):
Much better 😌

PR #5038, opened 2022-11-28
Fix crash voting on a heading with a content block

### LINKED ISSUE #5035: Bug report: Adding a custom content block in side block of budget investment prevents users from voting.
As in the title, if you add a custom content block to a budget it prevents the voting function from working.

Please check this example.
https://dev.communitychoices.scot/budgets/36/investments
image

When you click on the vote button nothing happens.

This is what we see in the Console using Chrome Dev Tools.
image

If we remove the content block, the voting functions works perfectly.

We have tested this on a direct fork of Consul release 1.5 without any modifications by us.

The custom content block is a list but this happens even if you only add a single line of text.

javierm:
## References

* Closes #5035

## Objectives

* Fix crash when rendering a heading content block after voting for an investment

## Notes

* The ERB Lint error due to using `raw` was already there before this pull request.

### DIFF
--- a/app/components/budgets/investments/content_blocks_component.html.erb
+++ b/app/components/budgets/investments/content_blocks_component.html.erb
@@ -1,0 +1,5 @@
+<ul class="no-bullet categories">
+  <% content_blocks.each do |content_block| %>
+    <%= raw content_block.body %>
+  <% end %>
+</ul>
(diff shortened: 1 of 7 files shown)


javierm:
The ERB error due to using `raw` was already there before this pull request.

taitus (approved):

PR #5048, opened 2022-12-15
Create an administrator in new tenant

taitus:
## References
Related PR: #4030 

## Objectives
When an admin creates a new tenant, we automatically add him as admin of the newly created tenant. The admin that is created is already automatically confirmed as it is a copy of an existing user with the same password.

## Visual Changes
create administrator in new tenant

## Notes
Add this behaviour to the [multitenancy documentation](https://github.com/consul/docs/pull/116).

### DIFF
--- a/app/controllers/admin/tenants_controller.rb
+++ b/app/controllers/admin/tenants_controller.rb
@@ -13,6 +13,7 @@ class Admin::TenantsController < Admin::BaseController
 
   def create
     if @tenant.save
+      update_default_tenant_administrator
       redirect_to admin_tenants_path, notice: t("admin.tenants.create.notice")
     else
       render :new
@@ -49,5 +50,20 @@ class Admin::TenantsController < Admin::BaseController
 
     def tenant_params
       params.require(:tenant).permit(:name, :schema, :schema_type)
+    end
+
+    def update_default_tenant_administrator
+      tenant_administrator_credentials = {
+        email: current_user.email,
+        username: current_user.username,
+        encrypted_password: current_user.encrypted_password
+      }
+
+      Tenant.switch(@tenant.schema) do
+        default_admin_account = User.administrators.first
+        default_admin_account.skip_confirmation_notification!
+        default_admin_account.update!(tenant_administrator_credentials)
+        default_admin_account.confirm
+      end
     end
 end
--- a/config/locales/en/admin.yml
+++ b/config/locales/en/admin.yml
@@ -1642,6 +1642,7 @@ en:
         enable: "Enable tenant %{tenant}"
         enabled: Enabled
       new:
+        admin_information: "When you create a tenant, your current user <strong>\"%{username}\"</strong> with the email <strong>\"%{email}\"</strong> and your password will be copied into the new tenant's database and will be automatically granted administration permissions. Note that these user accounts are stored in completely separate databases, so if you change the password of your user in one tenant, your accounts in other tenants will remain intact."
         title: New tenant
       restore:
         notice: Tenant enabled successfully
--- a/config/locales/es/admin.yml
+++ b/config/locales/es/admin.yml
@@ -1641,6 +1641,7 @@ es:
         enable: "Habilitar entidad %{tenant}"
         enabled: Habilitada
       new:
+        admin_information: 'Al crear una entidad se copiará tu usuario actual <strong>"%{username}"</strong> con el email <strong>"%{email}"</strong> y tu actual contraseña en la base de datos de la nueva entidad y se te otorgarán permisos de administración en la nueva entidad automáticamente. Nótese que serán cuentas de usuarios que están en bases de datos completamente independientes por lo que si cambias la contraseña de tu usuario en una entidad, las cuentas que tengas en otras entidades permanecerán intactas.'
         title: Nueva entidad
       restore:
         notice: Entidad habilitada correctamente
--- a/spec/system/admin/tenants_spec.rb
+++ b/spec/system/admin/tenants_spec.rb
@@ -40,6 +40,33 @@ describe "Tenants", :admin, :seed_tenants do
       expect(page).to have_current_path root_path
       expect(page).to have_link "Sign in"
     end
+
+    scenario "Updates new tenant default user with the main tenant administrator login credentials" do
+      user = create(:user, email: "super@consul.dev", password: "secret_password")
+      create(:administrator, user: user)
+      login_as(user)
+
+      visit new_admin_tenant_path
+
+      expect(page).to have_content "When you create a tenant, your current user"
+
+      fill_in "Name", with: "Earthlings"
+      fill_in "Subdomain", with: "earth"
+      click_button "Create tenant"
+
+      expect(page).to have_content "Tenant created successfully"
+      expect(ActionMailer::Base.deliveries.count).to eq(0)
+
+      click_link "earth.lvh.me"
+
+      click_link "Sign in"
+      fill_in "Email or username", with: "super@consul.dev"
+      fill_in "Password", with: "secret_password"
+      click_button "Enter"
+
+      expect(page).to have_content "You have been signed in successfully."
+      expect(current_host).to eq "http://earth.lvh.me"
+    end
   end
 
   scenario "Update" do
--- a/app/components/admin/tenants/new_component.html.erb
+++ b/app/components/admin/tenants/new_component.html.erb
@@ -1,3 +1,7 @@
 <%= back_link_to admin_tenants_path %>
 <%= header %>
+<div class="callout primary">
+  <%= sanitize(t("admin.tenants.new.admin_information", username: current_user.username,
+                                                        email: current_user.email)) %>
+</div>
 <%= render Admin::Tenants::FormComponent.new(tenant) %>
--- a/app/components/admin/tenants/new_component.rb
+++ b/app/components/admin/tenants/new_component.rb
@@ -1,6 +1,7 @@
 class Admin::Tenants::NewComponent < ApplicationComponent
   include Header
   attr_reader :tenant
+  delegate :current_user, to: :helpers
 
   def initialize(tenant)
     @tenant = tenant
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,5 +1,5 @@
 # Default admin user (change password after first deploy to a server!)
-if Administrator.count == 0 && !Rails.env.test?
+if Administrator.count == 0 && (!Rails.env.test? || !Tenant.default?)
   admin = User.create!(username: "admin", email: "admin@consul.dev", password: "12345678",
                        password_confirmation: "12345678", confirmed_at: Time.current,
                        terms_of_service: "1")


config/locales/en/admin.yml
```diff
@@ -1638,6 +1638,7 @@ en:
       index:
         create: Create tenant
       new:
+        admin_information: 'When you create a tenant, your current user <b>"%{username}"</b> with the email  <b>"%{email}"</b> will be copied into the database of the new tenant and will be automatically granted administration permissions.'
```
javierm, line 1641: I'm not so fond of `<b>` tags (although it's true that we use it sometimes, for whatever reasons :relieved:). We usually use `<strong>` when we'd like to note that a certain element isn't important. If we only want to make the text bold, we use a `<span>` tag and style it with CSS.

By the way, there are two spaces after "the email".

```suggestion
        admin_information: 'When you create a tenant, your current user <strong>"%{username}"</strong> with the email <strong>"%{email}"</strong> will be copied into the database of the new tenant and will be automatically granted administration permissions.'
```

app/controllers/admin/tenants_controller.rb
```diff
@@ -32,4 +33,20 @@ def update
...
+      Tenant.switch(@tenant.schema) do
+        admin = User.create!(username: current_user.username,
+                             email: current_user.email,
+                             password: "password_tmp",
+                             password_confirmation: "password_tmp",
+                             confirmed_at: Time.current,
+                             terms_of_service: "1")
+        admin.update!(password: nil,
+                      password_confirmation: nil,
+                      encrypted_password: current_user.encrypted_password)
```
javierm, line 47: I wonder whether it's clear that if the administrator changes the password in the main tenant, the password in the new tenant remains unchanged :thinking:.

config/locales/es/admin.yml
```diff
@@ -1637,6 +1637,7 @@ es:
       index:
         create: Crear entidad
       new:
+        admin_information: 'Al crear una entidad se copiará su usuario actual <b>"%{username}"</b> con el email <b>"%{email}"</b> en la base de datos de la nueva entidad y se le otorgarán permisos de administración automáticamente.'
```
javierm, line 1640:
```suggestion
        admin_information: 'Al crear una entidad se copiará su usuario actual <strong>"%{username}"</strong> con el email <strong>"%{email}"</strong> en la base de datos de la nueva entidad y se le otorgarán permisos de administración automáticamente.'
```

spec/system/admin/tenants_spec.rb
```diff
@@ -40,6 +40,52 @@
       expect(page).to have_current_path root_path
       expect(page).to have_link "Sign in"
     end
+
+    scenario "Copies the current user as the new tenant administrator" do
+      user = create(:user, password: "secret_password", password_confirmation: "secret_password")
```
javierm, line 45: The `password_confirmation` part doesn't seem necessary :thinking:.

```suggestion
      user = create(:user, password: "secret_password")
```

spec/system/admin/tenants_spec.rb
```diff
@@ -40,6 +40,52 @@
...
+      fill_in "Email or username", with: user.username
+      fill_in "Password", with: "secret_password"
+      click_button "Enter"
+
+      expect(page).to have_content "You have been signed in successfully."
+    end
+
+    scenario "does not create admin by default on seeds when tenant is not default tenant" do
+      allow(Rails.env).to receive(:test?).and_return(false)
+      user = create(:user, password: "secret_password", password_confirmation: "secret_password")
```
javierm, line 69: The `password_confirmation` part doesn't seem necessary :thinking:.

```suggestion
      user = create(:user, password: "secret_password")
```

spec/system/admin/tenants_spec.rb
```diff
@@ -40,6 +40,52 @@
...
+
+      click_link "Sign in"
+      fill_in "Email or username", with: user.username
+      fill_in "Password", with: "secret_password"
+      click_button "Enter"
+
+      expect(page).to have_content "You have been signed in successfully."
+    end
+
+    scenario "does not create admin by default on seeds when tenant is not default tenant" do
```
javierm, line 67: It took me a while to realize what we were testing. 

```suggestion
    scenario "does not create the default admin for the new tenant" do
```

Also, since these tests are sooooo slow, do you think it would make sense to join them? Or would it be harder to understand the tests?

```ruby
      click_link "Sign in"
      fill_in "Email or username", with: "admin@consul.dev"
      fill_in "Password", with: "12345678"
      click_button "Enter"

      expect(page).to have_content "Invalid Email or username or password."
      expect(page).not_to have_content "You have been signed in successfully."

      fill_in "Email or username", with: user.username
      fill_in "Password", with: "secret_password"
      click_button "Enter"

      expect(page).to have_content "You have been signed in successfully."
```

app/controllers/admin/tenants_controller.rb
```diff
@@ -32,4 +33,20 @@ def update
     def tenant_params
       params.require(:tenant).permit(:name, :schema, :schema_type)
     end
+
+    def create_tenant_administrator
+      Tenant.switch(@tenant.schema) do
```
javierm, line 38: Not necessarily bad, although this is the only time where we use `Tenant.switch` inside a request. It scares me a bit but I guess we're implicitly doing the same thing when we run `Apartment::Tenant.create` :relieved:.

So it should be OK, but let's keep an eye on it if we detect unusual behaviors in the future (like some tests failing sometimesbecause they access the wrong tenant).

javierm (commented):
Interesting approach. I've left some comments!

app/controllers/admin/tenants_controller.rb
Senen: I updated the notice text so it better explains the feature behaviour.

spec/system/admin/tenants_spec.rb
Senen: Finally, we kept only one of the tests.

app/controllers/admin/tenants_controller.rb
```diff
@@ -32,4 +33,13 @@ def update
     def tenant_params
       params.require(:tenant).permit(:name, :schema, :schema_type)
     end
+
+    def update_default_tenant_administrator
+      Tenant.switch(@tenant.schema) do
+        default_admin_account = User.administrators.last
```
javierm, line 39: Although it doesn't really matter, using `first` IMHO expresses a bit better the fact that we're changing the default administrator :thinking:.

```suggestion
        default_admin_account = User.administrators.first
```

spec/system/admin/tenants_spec.rb
```diff
@@ -40,6 +40,30 @@
       expect(page).to have_current_path root_path
       expect(page).to have_link "Sign in"
     end
+
+    scenario "Updates new tenant default user with the main tenant administrator login credentials" do
+      allow(Rails.env).to receive(:test?).and_return(false)
```
javierm, line 45: There might side effects if we tell Rails we aren't in the test environment when we actually are :thinking:. There are also a couple of tests in this file which are currently failing for the same reason. What do you think about removing this line and updating the `seeds.rb` instead?

```ruby
if Administrator.count == 0 && (!Rails.env.test? || !Tenant.default?)
```

This would be an option, but if you find a better one, go for that one!

config/locales/en/admin.yml
```diff
@@ -1638,6 +1638,7 @@ en:
       index:
         create: Create tenant
       new:
+        admin_information: 'When you create a tenant, your current user <strong>"%{username}"</strong> with the email <strong>"%{email}"</strong> will be copied into the database of the new tenant and will be automatically granted administration permissions. Note that these will be user accounts that are stored in completely separate databases, so if you change the password of your user in one tenant, your user account passwords you have in other tenants will remain intact.'
```
javierm, line 1641: The end of the sentence sounds a bit funny :thinking:. Maybe we could end it with "your accounts in other tenants will remain intact.", which is similar to the text we use in Spanish :thinking:. What do you think?

config/locales/es/admin.yml
```diff
@@ -1637,6 +1637,7 @@ es:
       index:
         create: Crear entidad
       new:
+        admin_information: 'Al crear una entidad se copiará su usuario actual <strong>"%{username}"</strong> con el email <strong>"%{email}"</strong> y su actual contraseña en la base de datos de la nueva entidad y se le otorgarán permisos de administración automáticamente. Nótese que serán cuentas de usuarios que están en bases de datos completamente independientes por lo que si cambia la contraseña de su usuario en una entidad las cuentas que tengas en otras entidades permanecerán intactas.'
```
javierm, line 1640: I think we usually address people using "tú" instead of "usted" :thinking:.

```suggestion
        admin_information: 'Al crear una entidad se copiará tu usuario actual <strong>"%{username}"</strong> con el email <strong>"%{email}"</strong> y tu actual contraseña en la base de datos de la nueva entidad y se te otorgarán permisos de administración en la nueva entidad automáticamente. Nótese que serán cuentas de usuarios que están en bases de datos completamente independientes por lo que si cambias la contraseña de tu usuario en una entidad, las cuentas que tengas en otras entidades permanecerán intactas.'
```

javierm (commented):
Much better! :tada: I've left a coupe of minor comments.

config/locales/en/admin.yml
Senen: Yes, indeed 😆 . I got your suggestion and tried to simplify the last sentence a bit more.

spec/system/admin/tenants_spec.rb
Senen: Yes, it was a crazy idea, we wanted to confirm that this would break something, and yes, it did. Finally, got your suggestion while we find a better way to deal with default users.

javierm:
Looks like we've got a couple of flaky tests, including the one we've added in this pull request :relieved:. Just a theory: maybe the second one failed because the first one failed and some data leaked between tests :thinking:. No idea why the first one (which is the one we've added) failed, though. The screenshot seems to show that clicking the "sign in" link resulted in a request that took too link, but I don't see a reason why that would happen.

spec/system/admin/tenants_spec.rb
```diff
@@ -40,6 +40,34 @@
...
+      fill_in "Subdomain", with: "earth"
+      click_button "Create tenant"
+
+      expect(page).to have_content "Tenant created successfully"
+
+      click_link "earth.lvh.me"
+
+      expect(page).to have_link "Sign in"
+
+      click_link "Sign in"
```
javierm, line 63: The expectation is redundant since `click_link` expects the link to be there before clicking it. I'm afraid whatever made the test fail wouldn't be solved with this :relieved:, since in the test the link was actually clicked.

```suggestion
      click_link "Sign in"
```

app/controllers/admin/tenants_controller.rb
```diff
@@ -32,4 +33,13 @@ def update
...
       params.require(:tenant).permit(:name, :schema, :schema_type)
     end
+
+    def update_default_tenant_administrator
+      Tenant.switch(@tenant.schema) do
+        default_admin_account = User.administrators.first
+        default_admin_account.update!(email: current_user.email,
+                                      username: current_user.username,
+                                      encrypted_password: current_user.encrypted_password)
+      end
```
javierm, line 43: This is pure conjecture since I haven't reproduced it.

I was jut wondering whether we could have an issue under certain conditions because `current_user` gets the user from the default schema but we're inside a `Tenant.switch` block :thinking:. It might be 100% safe because at this point `current_user` is probably cached, but, in any case, perhaps we could reduce the uncertainty with something like:

```suggestion
      login_params = {
        email: current_user.email,
        username: current_user.username,
        encrypted_password: current_user.encrypted_password
      }

      Tenant.switch(@tenant.schema) do
        default_admin_account = User.administrators.first
        default_admin_account.update!(login_params)
      end
```

Do you think it's worth it? Or am I being paranoid?

Senen: That's good paranoid. I also realized this during development. However, as the current user returned the correct user, I decided to dismiss it. Anyway, it would be better to be sure by doing what you suggest.

javierm (approved):

PR #5050, opened 2022-12-16
Make it possible to disable tenants 

javierm:
## References

* We implemented multitenancy in #4030

## Background

We don't allow removing a tenant because it would remove all its data, which would be a dangerous operation. However, there might be cases where we don't want a tenant to be accessible, and there's currently no way to do so.

## Objectives

* Make it possible to make a tenant unaccessible

### Before these changes

There's no way to disable a tenant

### After these changes

In the tenants index, there's a switch to enable/disable a tenant

### DIFF
--- a/app/models/tenant.rb
+++ b/app/models/tenant.rb
@@ -12,6 +12,8 @@ class Tenant < ApplicationRecord
   after_update :rename_schema
   after_destroy :destroy_schema
 
+  scope :only_hidden, -> { where.not(hidden_at: nil) }
+
   def self.find_by_domain(host)
     domain.find_by(schema: host)
   end
@@ -20,6 +22,16 @@ class Tenant < ApplicationRecord
     return nil unless Rails.application.config.multitenancy.present?
     return nil if host.blank? || host.match?(Resolv::AddressRegex)
 
+    schema = schema_for(host)
+
+    if schema && only_hidden.find_by(schema: schema)
+      raise Apartment::TenantNotFound
+    else
+      schema
+    end
+  end
+
+  def self.schema_for(host)
     host_without_www = host.delete_prefix("www.")
 
     if find_by_domain(host)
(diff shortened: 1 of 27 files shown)


app/models/tenant.rb
```diff
@@ -30,7 +32,13 @@ def self.resolve_host(host)
...
       schema = host_without_www.sub(/\.?#{host_domain}\Z/, "").presence
 
-      schema unless find_by_domain(schema)
+      if schema
+        if only_hidden.or(domain).find_by(schema: schema)
+          raise Apartment::TenantNotFound
+        else
+          schema
+        end
+      end
```
Senen, line 41: I wonder if we could reduce conditional nesting within this method.

```suggestion
      raise Apartment::TenantNotFound if schema && only_hidden.or(domain).find_by(schema: schema)
      
      schema
```

Senen (commented):
Hi, @javierm, I left just a single comment. Let me know what you think!

Senen:
Hi again @javierm, 

While testing locally, I found disabling subdomains tenants works fine but disabling domain tenants is not working. 

I tested it in the development environment by adding the domain I gave to the tenant in my machine `/etc/hosts` file, so the domain resolves to 127.0.0.1. After that, the tenant is always accessible even when disabled.

After digging, I found the cause could be in the `resolve_host` methods call to the `find_by_domain` method, which finds even within the hidden records and returns the host. Probably we want to search only over the not hidden records. 🤔

app/models/tenant.rb
javierm: I've updated the code.

javierm:
@Senen Great! I've updated the pull request so we raise an exception when accessing a hidden tenant with a domain, just like we do when accessing a hidden tenant with a subdomain.

Senen (approved):

PR #5052, opened 2022-12-20
Do not track `Apartment::TenantNotFound` exceptions through `Airbrake`

Senen:
## References

* We implemented multitenancy in #4030

## Objectives

Do not track `Apartment::TenantNotFound` exceptions through `Airbrake` as we do with `ActiveRecord::NotFound` exceptions. Otherwise, we can quickly flood the third-party application in charge of tracking exceptions.


javierm (approved):
Cool!

Issue #5055, opened 2022-12-22
Valuation interface doesn't work with concurrent budgets


coslajohn:
### User Story

When we have more than one Participatory Budget open at the same time, valuators can only see assigned projects from the most recent budget

### Why is this Needed?

We can have more than one budget running at the same time. Valuators need to be able to see all assigned projects

### Description

*Type*: Bug

### Current Behavior

The valuation menu only displays Investment projects assigned with valuation open from the most recently created Budget. If you have more than one budget process running concurrently, investment projects are not shown to the evaluator for projects which belong to older budgets

### Expected Behavior

The valuation page should list projects from all open Participatory Budgets which have valuation assigned

### Steps to reproduce

Create a PB budget A. Submit project A1. Assign to valuator. the project A1 appears on valuation page
Create a PB budget B. Submit project B1. Assign to valuator. the project B1 appears on valuation page. A1 does not appear on the valuation page

### Definition of Done

All projects open for valuation from all budgets are displayed on valuation page

Issue #4770, opened 2022-01-17
Be able to see all budgets on the valuation page


microweb10:
Dear CONSUL developer team,

I think it would be interesting that you consider backporting this commit into CONSUL
https://github.com/democrateam/consul/commit/c916fb7906649fe064de0ddb7cd03a3cbd7aebbb

Cheers!

Senen:
Hi @ microweb10, I'm closing this issue as it's solved via #5061, feel free to re-open it if I'm wrong!

PR #5061, opened 2023-01-04
Show published budgets in the valuation panel

Senen:
## References

Related Issues:
* #5055 
* #4770

## Objectives

Show all **open** and **published** budgets to valuators instead of showing only the latest budget created.


javierm (commented):
@Senen I've left a comment and a question. Let me know what you think!

app/controllers/valuation/budgets_controller.rb
```diff
@@ -5,9 +5,6 @@ class Valuation::BudgetsController < Valuation::BaseController
   load_and_authorize_resource
 
   def index
-    @budget = current_budget
-    if @budget.present?
-      @investments = @budget.investments.by_valuator(current_user.valuator).valuation_open
-    end
+    @budgets = @budgets.published.open.order(created_at: :desc)
```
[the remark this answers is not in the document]
Senen: Hi @javierm, thanks for your comments 🙏🏼 . 

>Note that, before these changes, a finished budget would appear here if there weren't any open budgets. Now we're changing this feature and no budgets will appear if there aren't any open budgets.

You are right, I finally changed the approach. The valuation budgets page in the new version shows all the published budgets. For each budget, the "Evaluate" link is available only when the budget is in `valuating` phase and where the valuator has investments assigned. I also adapted the investments count, so it makes sense with the number of investments shown on the valuation budget investments page.

>Might be fine; unfortunately I don't know enough about the way valuators work to be sure 🤔. What do you think is the right behavior?

Same for me. I finally decided to do some changes that make sense to me 🤔 .

Let me know what you think.

spec/system/valuation/budgets_spec.rb
```diff
@@ -1,29 +1,75 @@
...
+      visit valuation_budgets_path
+
+      within "#budget_#{budget.id}" do
+        expect(page).to have_selector(".investments-count", text: "1")
+      end
+    end
+
+    scenario "Displays zero as investments count when budget is not in valuating phase" do
+      budget = create(:budget, %i[accepting finished].sample, name: "Sports")
+      create(:budget_investment, :visible_to_valuators, budget: budget, valuators: [valuator])
```
javierm, line 34: Since these tests (and the next one) don't test user interaction or controller filters, what do you think about moving them to component tests? :thinking:

Makes sense to keep the "Displays published budgets" test here because it tests controller filters and thanks to it we've got one system test accessing the valuation index.

Senen: Much better! Thanks

javierm (approved):

Issue #3423, opened 2019-04-05
Investment Vote - Cache - Cache not re-rendered on vote


ziyan-junaideen:
### Description

*Type*: Bug

A possible bug was encountered in [NYC](http://pbnyc.participatorybudgeting.org) installation while in voting. Affects only in production environment with caching.

Consider a user visiting a budget investment to vote. When voting, as expected, the voted info shows in the sidebar. How ever if you refresh the page, the the vote link shows up. You click on vote, nothing happens, you cant vote what is already voted.

Upon inspection, I noticed that the `cache` in the `budget/investments/_investment_show.html.erb` using an instance variable `@investment_votes` as an input to generate a cache key. Even if voted, this variable remains an empty hash `{}`. This explains why we see the 'Vote' link as nothing else has changed to generate a new cache key.

![Screenshot from 2019-04-05 11-26-11](https://user-images.githubusercontent.com/762159/55606243-a9154d80-5795-11e9-940b-f3bfdc1fdfd9.png)

I noticed that the vote process doesn't make a record in the `votes` table. But the instance variable is populated using the `votes` table.

If you can confirm, i can work on a fix.

### Current Behavior
In an app running in production environment, voting on a budget investment shoes the vote button after page reload / refresh.

### Expected Behavior

Following the voting process, the budget investment page cache should be invalidated and re-rendered. For this a new cache key should be generated to re-render the view.

### Steps to reproduce

- A consul installation using the installer (with cache and booted in production mode) v0.19
- Budget investment in voting
- Vote on selected project
- Refresh page - vote link shows

javierm:
Closed via #4504.

PR #5077, opened 2023-01-31
Don't cache related actions in investments view

### LINKED ISSUE #5075: View Investment page caching issue
### User Story

Sorry - this one is quite difficult to describe

We have identifed some strange behaviour when viewing investments.  If you view an Investment, then when the Phase of a Participatory Budgeting process changes, the view of the Investment remains the same as the previous phase.  It does not show the correct Buttons on the right sidebar even though they are needed by the new phase.

This feels like caching but we have been unable to force a refresh in the browser to display the updated page

We think it might be on a per user basis but was reported to us as site wide

### Why is this Needed?

This was reported to us by end users who were concerned that their investment was being disadvantaged in a vote because the Vote or Support button was not being displayed in the sidebar. We are still investigating and are not entirely clear whether this affects all users or just individual logged in users.

### Description

*Type*: Bug

### Current Behavior

If you log in as an admin or author and view an investment page then change phase, the investment view remains the same as it was in the previous phase.

For example if you are in the Valuating Phase, view an Investment then change the Active Phase to Voting, the Investment View remains the same as before so does not have the Vote button.  If you view the list of investments they all display the Vote button in the summary list, but when you click in to the Investment, there is no Vote button.

Refreshing/Clearing the cache in the browser seems to have no effect.

Editing the Investment and saving by clicking Update Investment will force a refresh and the correct view for the phase will display

### Expected Behavior
The view which corresponds to the Phase should be displayed

### Steps to reproduce
We have been able to reproduce this in our own site and on the live Consul Demo site. View Investment, change phase, view investment once more.

https://dev.communitychoices.scot/budgets/44/investments?heading_id=71

**The process is in the support phase**

MicrosoftTeams-image (2)

**when I click into the Investment it doesn't show the Support button (this example shows Vote because that was the last phase I tested with)**

MicrosoftTeams-image (3)

**but in a different session using different browser looking at same page I get the correct view**
MicrosoftTeams-image (10)

**Edit the page, click on Update Investment and I get the correct view. The view has been refreshed correctly.**

MicrosoftTeams-image (5)

**Change phase to Voting and in my first session, the browser I changed the phase in and still showing support**  

MicrosoftTeams-image (8)

In my second session it correctly shows Vote 

MicrosoftTeams-image (9)

### Definition of Done

When Phase is changed all users can see updated page view

**Exactly the same behaviour occurs when using the Consul Demo site**

MicrosoftTeams-image (1)
MicrosoftTeams-image

javierm:
## References

* Closes #5075
* We solved similar issues in pull requests #3423 and #4224

## Objectives

* Fix issues with the investment cache not being expired when its budget and/or certain settings are updated
* Reduce the chance of experiencing similar issues in the future

## Notes

As mentioned above, we've experienced some caching issues with this code for years, but we're still running into other issues. In order to really cache this section, we'd need to cache:

* Whether or not the investment should show the aside and vote/support buttons (we could do it by caching its budget)
* Whether its budget is balloting or finished (we could do it by caching its budget)
* Whether the current user is following the investment
* Whether the `remove_investments_supports` feature is enabled
* Whether the `community` feature is enabled
* The value of the `org_name` setting
* The value of the `twitter_handle` setting

Chances are that this list is incomplete.

So, instead of trying to take into account every single possible factor that should make us expire the cache, we're restricting the cache so it only affects the content of the investment. This is similar to what we do in the investments index, where we cache the content of the investment but we don't cache the vote/support buttons.

### DIFF
--- a/app/views/budgets/investments/_investment_show.html.erb
+++ b/app/views/budgets/investments/_investment_show.html.erb
@@ -7,122 +7,118 @@
             og_image_url: (investment.image.present? ? polymorphic_path(investment.image.variant(:thumb)) : nil) %>
 <% end %>
 
-<% cache [locale_and_user_status(investment),
-          investment,
-          investment.author,
-          Flag.flagged?(current_user, investment),
-          investment.followed_by?(current_user),
-          feature?(:remove_investments_supports),
-          current_user&.voted_for?(investment)] do %>
-  <section class="budget-investment-show" id="<%= dom_id(investment) %>">
-
-    <div class="row">
+<section class="budget-investment-show" id="<%= dom_id(investment) %>">
+  <div class="row">
+    <% cache [locale_and_user_status(investment),
+              investment,
+              investment.author,
+              Flag.flagged?(current_user, investment)] do %>
       <div class="small-12 medium-9 column">
         <%= back_link_to budget_investments_path(investment.budget, heading_id: investment.heading) %>
 
         <%= render "/budgets/investments/investment_detail", investment: investment %>
       </div>
+    <% end %>
 
-      <aside class="small-12 medium-3 column">
-        <%= render "/budgets/investments/author_actions", investment: investment %>
+    <aside class="small-12 medium-3 column">
+      <%= render "/budgets/investments/author_actions", investment: investment %>
 
-        <% if investment.should_show_aside? %>
-          <% if investment.should_show_votes? %>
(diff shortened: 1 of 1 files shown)


taitus (approved):

javierm:
Moving it back to reviewing because it was approved by accident :relieved:.

taitus (approved):

PR #5084, opened 2023-02-21
Remove duplicate keys

taitus:
## Objectives

Remove duplicated keys in `:en` locales files.

### DIFF
--- a/app/views/admin/legislation/proposals/_proposals.html.erb
+++ b/app/views/admin/legislation/proposals/_proposals.html.erb
@@ -7,7 +7,7 @@
     <thead>
       <tr>
         <th class="text-center"><%= t("admin.legislation.proposals.index.id") %></th>
-        <th><%= t("admin.legislation.proposals.index.title") %></th>
+        <th><%= Legislation::Proposal.human_attribute_name(:title) %></th>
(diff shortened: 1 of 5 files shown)


config/locales/en/admin.yml
```diff
@@ -619,7 +619,6 @@ en:
           title: Proposals
           back: Back
           id: Id
-          title: Title
```
javierm, line 622: I think in this case the proper thing would be to show "Proposals" as the title of the page and "Title" in the coumn header of the proposals table :thinking:.

What do you think?

app/views/admin/legislation/proposals/_proposals.html.erb
```diff
@@ -7,7 +7,7 @@
     <thead>
       <tr>
         <th class="text-center"><%= t("admin.legislation.proposals.index.id") %></th>
-        <th><%= t("admin.legislation.proposals.index.title") %></th>
+        <th><%= Proposal.human_attribute_name(:title) %></th>
```
javierm, line 10: Since we're in the legislation proposals section:

```suggestion
        <th><%= Legislation::Proposal.human_attribute_name(:title) %></th>
```

javierm (approved):

PR #5074, opened 2023-01-30
Upgrade Ruby to version 3.0.6

javierm:
## References

* Depends on #5089

## Objectives

* Update Ruby to a more recent version

## Notes

These were the gems we're currently using whose compatibility with Ruby 3 we hadn't confirmed yet. They're all confirmed now.

* ~autoprefixer-rails~
* ~byebug~
* ~capistrano-bundler~
* ~capistrano-rails~
* ~capistrano3-delayed-job~ (no updates since January 2019)
* ~capistrano3-puma~
* ~ckeditor~
* ~cocoon~
* ~email_spec~ (no release since April 2018)
* ~erb_lint~ (pull request #5093)
* ~foundation-rails~ (has almost no Ruby code, so it might work)
* ~graphiql-rails~ (has almost no Ruby code)
* ~initialjs-rails~
* ~jquery-ui-rails~ (has almost no Ruby code, so it might work even if there haven been no updates since January 2017)
* ~mdl~ (support in version 0.12 for sure)
* ~omniauth-twitter~ (no updates since February 2017)
* ~pronto-erb_lint~ (pull request #5093)
* ~puma~
* ~rinku~ (no updates since April 2019)
* ~rvm1-capistrano~ (no updates since October 2015)
* ~sassc-rails~ (no updates since December 2019)
* ~scss-lint~
* ~selenium-webdriver~
* ~spring~
* ~spring-commands-rspec~ (no updates since March 2015)
* ~turbolinks~ (unmaintained, but it's mostly written in CoffeeScript and might work)
* ~turnout~ (no updates since December 2018)

PR #5089, opened 2023-03-08
Use Bing Translator instead of TranslatorText

taitus:
## References
Upgrade Ruby to version 3.0.5 #5074 

## Objectives
- TranslatorText isn't compatible with Ruby 3, so we need to use a different gem as example [Bing Translator gem](https://github.com/relrod/bing_translator-gem).
- Fix remote translations for locales `:pt-BR`, `:zh-CN` and `:zh-TW`
- It has been detected that there is already support for locale `:gl` (Galego) and `:sq` (Shqip). The only locale (of all the locales supported by consul) that is currently not supported by this translator is `:val` (Valencià).

## Notes
This PR involves some changes in the documentation: Update remote translation docs [#119](https://github.com/consul/docs/pull/119)


spec/lib/remote_translations/microsoft/available_locales_spec.rb
```diff
@@ -0,0 +1,11 @@
+require "rails_helper"
+
+describe RemoteTranslations::Microsoft::AvailableLocales do
+  describe ".available_locales" do
+    context "app locales that need to be mapping" do
+      it "includes locales with same format that I18n.available_locales" do
```
javierm, line 6: I'm wondering whether we should test that a `RemoteTranslation` record is valid when its locale is `pt-BR`, since that's what was failing before these changes :thinking:. What do you think?

By the way, shall we stub the `remote_available_locales` method? Otherwise the test will try to connect to `api.cognitive.microsofttranslator.com`.

spec/models/remote_translation_spec.rb
```diff
@@ -49,6 +49,14 @@
     expect(remote_translation).to be_valid
   end
 
+  it "checks app locales that need to be mapping" do
+    allow(RemoteTranslations::Microsoft::AvailableLocales).to receive(:remote_available_locales)
+                                                          .and_return(["pt"])
```
javierm, line 54: This test passes even without the changes in this pull request :thinking:. I guess we need to either add `allow(RemoteTranslations::Microsoft::AvailableLocales).to receive(:available_locales).and_call_original` or change the way we handle tests with the `:remote_translations` tag so they stub `remove_available_locales` instead of stubbing `available_locales`.

spec/models/remote_translation_spec.rb
```diff
@@ -49,6 +49,14 @@
     expect(remote_translation).to be_valid
   end
 
+  it "checks app locales that need to be mapping" do
```
javierm, line 52: There seems to be a typo here. Maybe something like:

```suggestion
  it "is valid with a locale that uses a different name in the remote service" do
```

spec/lib/remote_translations/microsoft/available_locales_spec.rb
```diff
@@ -0,0 +1,13 @@
+require "rails_helper"
+
+describe RemoteTranslations::Microsoft::AvailableLocales do
+  describe ".available_locales" do
+    context "app locales that need to be mapping" do
```
javierm, line 5: I guess we could remove this line, since there's only one test here.

If we decide to keep this line, there seems to be a typo:

```suggestion
    context "app locales that need to be mapped" do
```

spec/lib/remote_translations/microsoft/available_locales_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+      it "includes locales with same format that I18n.available_locales" do
```
javierm, line 6: I think there might be some typos here :thinking:.

```suggestion
      it "includes locales with the same format as I18n.available_locales" do
```

spec/lib/remote_translations/microsoft/available_locales_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+        allow(RemoteTranslations::Microsoft::AvailableLocales).to receive(:remote_available_locales)
+                                                              .and_return(%w[de en es fr pt zh-Hans])
+        expect(RemoteTranslations::Microsoft::AvailableLocales.available_locales.include?("pt-BR")).to eq true
```
javierm, line 9: For readability, we could add a blank line before the expectation.

By the way, what do you think about testing the whole mapping?

```suggestion
        expect(RemoteTranslations::Microsoft::AvailableLocales.available_locales).to eq %w[de en es fr pt-BR zh-CN]
```

javierm (approved):

PR #5097, opened 2023-03-28
Disable scroll wheel for zooming in favor of map zoom buttons

Senen:
## References

Replaces:

* #4768

## Objectives

Fix the mousewheel trap when scrolling the page to the bottom when there is more content below the map.

# Manual testing

**Desktop browsers**

Verified successfully in:
* Chrome
* Firefox
* Edge
* Safari
* Opera

This update has no impact in touchable devices as they do not have mouse wheel. It seems the map behaves nicely with all the browsers mentioned above.

## Visual Changes
None


javierm (approved):
What I really want is the computer to read my mind and know whether I'd like to scroll or zoom when I use the mouse wheel :smile:.

But I've tried `scrollWheelZoom: mind_reading` and it didn't work, so I guess this change is an improvement :smile:.

PR #5100, opened 2023-03-30
Codespace

DiegoZen:
This is completely functional configuration for using GitHub Codespaces with Consul.
You can read about it in:
- https://github.com/features/codespaces
- https://code.visualstudio.com/docs/devcontainers/containers
- https://docs.github.com/en/codespaces/getting-started/quickstart
- https://www.youtube.com/watch?v=pmVx7Ic3ITE

## How I made this
_These are the steps I made to create this pull request._

1. Get the project locally. (you can also do it in the browser clicking in Code > Create Codespace)
2. Cmd+Shitf+P to open the command palet and select 'Dev Containers: Add Dev Container Configuration Files'
3. Select 'From a predefined container configuration definition...'
4. Select 'Show All Definitions...'
5. Search for 'Ruby on Rails & Postgres' and select it.
6. Select the Ruby version for the project (e.g. 2.7)
7. You can select additional features but we are going to skip this step. Click OK.

Now you have a `.devcontainer/` folder in the root directory. Let's make some modifications:
1. Uncomment the field 'forwardedPorts'. 3000 will be the port for the web server and 5432 the port for postgres service.
2. Uncomment the field 'postCreateCommand' and add the following value: 
```shell
sudo apt update && sudo apt upgrade -y && sudo apt-get install cmake -y && sudo apt-get install nodejs -y && bundle
```
Now you have everything ready to start your codespace.

## Develop in a codespace
_These are the steps to use it. This should be placed somewhere in the official documentation._

You can use it in three different ways
**Run the containers locally**
1. Start docker
2. Open the command palette in VSCode
3. Select 'Dev Containers: Open Folder in Container'. This run the configuration set up. 

**Run in the browser**
1. Go to the GitHub repository
2. Click *Code > Codespaces > Create Codespace*. It will open VSCode in the browser and start the cofiguration and instalation. It forwards ports 3000 and 5432 to a remote url.

**Run in VSCode Desktop**
1. Install the plugin GitHub Codespaces. 
2. Go to the Remote Explorer tab and select 'GitHub Codespaces'
3. Sign in with your GitHub account (if you encounter problems, restart VSCode). You should see your codespaces listed there (if there are none, you have to create them like in the previous method).
4. Click on the plug an it should open the codespace.

After that, in all cases you need  do the following inside the codespace:
1. Create config/database.yml
```
cp config/database.yml.example config/database.yml
```
2. Change configuration for config/database.yml
```yml
host: db
username: postgres
password: postgres
```
3. Run the command to setup the databases
```shell
bin/rake db:setup
```
4. (Optional) Add development seeds
```shell
bin/rake db:dev_seed
```
5. Run the server
```
rails s
```

#### Note
1. If you have done the commands for creating and migrating tables to the database in the remote codespace via browser you don't have to do it again when connecting from VSCode Desktop app, and viceversa, because it is the same codespace and it keeps the configuration.
2. For pushing changes from docker container running locally with no remote GitHub codespace (method 1) you shoud create a ssh key pair inside the container and add the public key to your GitHub account.
3. With method 1 and 3 you can connect a database client to see the database through a GUI. But running in the browser (method 2) it is not possible. 
4. (bug?) Rails can throw an error because of config.hosts restrictions in config/environments/development.rb. I think this should be removed.


javierm:
Hi, @diegozen :smile:. Thanks for the pull request!

**Disclaimer:** I've never used a devcontainer, codespaces or vscode.

I see Rails will add devcontainers to new applications in Rails 7.2 (rails/rails#50914), so I was wondering whether we should use that :thinking:. Could you check whether the devcontainer contents generated by Rails work in our case? Here's the diff (NOTE: Rails doesn't include the comments describing `config/database.yml`, but I've kept them so Git recognizes the compose file as the same even though it's renamed):

```diff
diff --git a/.devcontainer/Dockerfile b/.devcontainer/Dockerfile
index 8bb84999d4..0e4cec05d6 100644
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,18 +1,3 @@
-FROM mcr.microsoft.com/devcontainers/ruby:0-2.7
-
-# Install Rails
-RUN gem install rails webdrivers 
-
-# Default value to allow debug server to serve content over GitHub Codespace's port forwarding service
-# The value is a comma-separated list of allowed domains 
-ENV RAILS_DEVELOPMENT_HOSTS=".githubpreview.dev,.preview.app.github.dev,.app.github.dev"
-
-# [Optional] Uncomment this section to install additional OS packages.
-# RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
-#     && apt-get -y install --no-install-recommends <your-package-list-here>
-
-# [Optional] Uncomment this line to install additional gems.
-# RUN gem install <your-gem-names-here>
-
-# [Optional] Uncomment this line to install global node packages.
-# RUN su vscode -c "source /usr/local/share/nvm/nvm.sh && npm install -g <your-package-here>" 2>&1
+# Make sure RUBY_VERSION matches the Ruby version in .ruby-version
+ARG RUBY_VERSION=3.2.3
+FROM ghcr.io/rails/devcontainer/images/ruby:$RUBY_VERSION
diff --git a/.devcontainer/docker-compose.yml b/.devcontainer/compose.yaml
similarity index 68%
rename from .devcontainer/docker-compose.yml
rename to .devcontainer/compose.yaml
index 2224d45d27..ecacd27152 100644
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/compose.yaml
@@ -1,7 +1,5 @@
-version: '3'
-
 services:
-  app:
+  rails-app:
     build:
       context: ..
       dockerfile: .devcontainer/Dockerfile
@@ -12,21 +10,35 @@ services:
     # Overrides default command so things don't shut down after the process ends.
     command: sleep infinity
 
-    # Runs app on the same network as the database container, allows "forwardPorts" in devcontainer.json function.
-    network_mode: service:db
+    networks:
+      - default
+
+    # Uncomment the next line to use a non-root user for all processes.
+    # user: vscode
 
     # Use "forwardPorts" in **devcontainer.json** to forward an app port locally.
     # (Adding the "ports" property to this file will not forward from a Codespace.)
+    ports:
+      - 45678:45678
+    depends_on:
+      - selenium
+      - postgres
+
+  selenium:
+    image: seleniarm/standalone-chromium
+    restart: unless-stopped
+    networks:
+      - default
 
-  db:
-    image: postgres:latest
+  postgres:
+    image: postgres:16.1
     restart: unless-stopped
+    networks:
+      - default
     volumes:
       - postgres-data:/var/lib/postgresql/data
-      - ./create-db-user.sql:/docker-entrypoint-initdb.d/create-db-user.sql
     environment:
       POSTGRES_USER: postgres
-      POSTGRES_DB: postgres
       POSTGRES_PASSWORD: postgres
       # Your config/database.yml should use the user and password you set here,
       # and host "db" (as that's the name of this service). You can use whatever
@@ -41,8 +53,5 @@ services:
       #    password: postgres
       #    database: myapp_development
 
-    # Add "forwardPorts": ["5432"] to **devcontainer.json** to forward PostgreSQL locally.
-    # (Adding the "ports" property to this file will not forward from a Codespace.)
-
 volumes:
   postgres-data:
diff --git a/.devcontainer/create-db-user.sql b/.devcontainer/create-db-user.sql
deleted file mode 100644
index 291dfc4272..0000000000
--- a/.devcontainer/create-db-user.sql
+++ /dev/null
@@ -1,2 +0,0 @@
-CREATE USER vscode CREATEDB;
-CREATE DATABASE vscode WITH OWNER vscode;
diff --git a/.devcontainer/devcontainer.json b/.devcontainer/devcontainer.json
index 4ea03ab4ac..5dc9f8311c 100644
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,24 +1,35 @@
 // For format details, see https://aka.ms/devcontainer.json. For config options, see the
-// README at: https://github.com/devcontainers/templates/tree/main/src/ruby-rails-postgres
+// README at: https://github.com/devcontainers/templates/tree/main/src/ruby
 {
-  "name": "Ruby on Rails & Postgres",
-  "dockerComposeFile": "docker-compose.yml",
-  "service": "app",
+  "name": "Consul Democracy app",
+  "dockerComposeFile": "compose.yaml",
+  "service": "rails-app",
   "workspaceFolder": "/workspaces/${localWorkspaceFolderBasename}",
 
   // Features to add to the dev container. More info: https://containers.dev/features.
-  // "features": {},
+  "features": {
+    "ghcr.io/devcontainers/features/github-cli:1": {},
+    "ghcr.io/rails/devcontainer/features/activestorage": {},
+    "ghcr.io/devcontainers/features/node:1": {},
+    "ghcr.io/rails/devcontainer/features/postgres-client": {}
+  },
+
+  "containerEnv": {
+    "CAPYBARA_SERVER_PORT": "45678",
+    "SELENIUM_HOST": "selenium",
+    "DB_HOST": "postgres"
+  },
 
   // Use 'forwardPorts' to make a list of ports inside the container available locally.
-  // This can be used to network with other containers or the host.
   "forwardPorts": [3000, 5432],
 
-  // Use 'postCreateCommand' to run commands after the container is created.
-  "postCreateCommand": "sudo apt update && sudo apt upgrade -y && sudo apt-get install cmake -y && sudo apt-get install nodejs -y && bundle"
-
   // Configure tool-specific properties.
   // "customizations": {},
 
   // Uncomment to connect as root instead. More info: https://aka.ms/dev-containers-non-root.
-  // "remoteUser": "root"
+  // "remoteUser": "root",
+
+
+  // Use 'postCreateCommand' to run commands after the container is created.
+  "postCreateCommand": "bin/setup"
 }
```

Not sure how Codespaces work; if it's possible to run Pronto, I guess we shouldn't remove the `cmake` installation step. Regarding the version of PostgresQL, the latest version would probably be OK :thinking:.

diegozen:
This seems to only apply to new projects via `rails new` (https://github.com/rails/rails/pull/50914#issuecomment-1954718596).

The big difference is that I used `ruby-rails-postgres` template and add extra dependencies in the `postCreateCommand` and they used `ruby` and extended it in the `features` object. I am not sure of the implications of each approach.

I suggest merging this one and have the rails generated one as a reference for future improvements.

javierm:
> This seems to only apply to new projects via rails new

That is correct.

> The big difference is that I used ruby-rails-postgres template and add extra dependencies in the postCreateCommand and they used ruby and extended it in the features object. I am not sure of the implications of each approach.

Me neither :relieved:.

However, there are other differences that I think we should include if possible :thinking:.

First, the Rails Dockerfile specifies the whole Ruby version (like 3.2.3) instead of just the first two digits (like 0-2.7 or 0-3.2). When upgrading Ruby, the easiest way is to look for a text like "3.2.3" in the source code and replace it where appropriate, and we might miss something like "0-3.2". Also, not sure what would happen if we've got a `.ruby-version` file specifying 3.2.3 but the `ruby:0-3.2` image installs Ruby 3.2.4, which was released a few days ago. I guess it wouldn't be possible to start the application :thinking:; if so, that would mean the codespaces integration would break every time a new patch version of Ruby is released or whenever we're developing on a branch using a different patch version.

Then, we should keep the Selenium and Capybara integration in order to make it possible to run system tests.

Finally, instead of running `bundle`, it'd be great if we could run `bin/setup`, since it also installs Node packages, copies the `database.yml.example` and `secrets.yml.example` files, and sets up the database. If it isn't possible to set up the database at this point due to username/password mismatches (maybe we should change the `config/database.yml.example` file to avoid this :thinking:), we could add the commands to install Node packages and copy the `database.yml.example` and `secrets.yml.example` files to the `postCreateCommand`.

As a minor thing, changing the `name` attribute to something containing "Consul Democracy" would also make sense IMHO.

javierm:
I've just opened #5530, which will hopefully make it possible to run `bin/setup` without having to copy and edit the `database.yml` file. We might need to add `POSTGRES_HOST: postgres` in the compose file, and maybe also change the `database.yml.example` file so it allows setting the database with an environment variable :thinking:.

javierm:
> as far as I know, a minor change in ruby version should not stop the application from working.

Sorry, I didn't fully explain this one :pray:. For a reason not related to this pull request, we're considering adding `ruby file: ".ruby-version"` to the `Gemfile`. This way, developers using RVM don't accidentally use the wrong Ruby version when switching to a different branch (RVM doesn't automatically change the Ruby version when switching to a branch with a different `.ruby-version` file).

With this option, the application wouldn't run when using a different patch version, so I guess the codespaces integration would break if that's the case :thinking:.

javierm:
@diegozen On Monday we merged pull request #5530. With that, and the patch previously mentioned, maybe it's no longer necessary to manually edit the `database.yml` file? :thinking:

PR #5113, opened 2023-05-19
Show email digest only if proposals are enabled

decabeza:
## Objectives

Users can select the option "**Receive a summary of proposal notifications**" on the My Account `/account` page.

These notifications are created by proposal authors for users who have supported the proposal. 

To avoid confusion and improve the UX, this PR hides that checkbox if the proposals feature is disabled, as it doesn't make sense to show the option if those notifications cannot be created. 😌 

## Visual Changes

proposals_notifications

### DIFF
--- a/app/views/account/show.html.erb
+++ b/app/views/account/show.html.erb
@@ -54,9 +54,11 @@
               <%= f.check_box :newsletter %>
             </div>
 
-            <div>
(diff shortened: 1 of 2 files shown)


Senen (approved):
Nice catch! 👏🏼

PR #5115, opened 2023-05-23
Remove unused field and related code

taitus:
## References
Related PR: #1140 

## Objectives
Remove unused field and related code

## Notes
As we can see in the commit messages this field with related code was added in the PR #1140, but in the same PR all related code was removed. I think that we can remove this field, since it does not appear to be used in the application.


db/migrate/20230523090028_remove_email_on_proposal_notification_from_users.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveEmailOnProposalNotificationFromUsers < ActiveRecord::Migration[6.0]
+  def change
+    remove_column :users, :email_on_proposal_notification, :boolean
```
javierm, line 3:
```suggestion
    remove_column :users, :email_on_proposal_notification, :boolean, default: true
```

This way running this migration and then reverting it will result in the same properties being applied to this  field.

javierm (commented):
@taitus Nice catch! I've left a comment about a minor thing.

One more thing: could we squash these commits into one or change their order? Right now one test in `system/management/users_spec.rb:45` fails after the first commit, which might be an issue when running `git bisect` in the future.

javierm (approved):

PR #5117, opened 2023-05-24
Update spec

taitus:
## References
Related PR: #5113 
Related commit:  b9da024d8eaeb

## Objectives
In order to prevent "flakys" we configure the database before starting the browser with a call to `visit`


Senen (approved):
Nice catch! I did not realize that the test was calling the database after the visit  🙏🏼

taitus:
> Nice catch! I did not noticed that the test was calling the database after the visit 🙏🏼
@Senen I didn't notice it either

PR #5123, opened 2023-06-05
Fix cached tenant secrets when running tests

taitus:
## Objectives

When running multiple specs that have to overwrite the rails secrets, it can happen that the condition:
"@cached_rails_secrets != Rails.application.secrets" is not met and unexpected secrets are returned.

We have found this case while experimenting with tests related to the Tenant secrets. In one case, assigning cached rails secrets to nil resulted in a failure to detect when the 'rails.application.secrets' had changed.


javierm (commented):
Was going to approve it and accidentally pressed the "merge" button instead :relieved:.

PR #5130, opened 2023-06-13
Fix Failing Test: SDG management relations_spec

karim-semmoud:
## References
When cloning or forking the latest consul, installing it locally, there are two failing scenarios when running the tests. 
spec/system/sdg_management/relations_spec.rb
- "allows adding the goals, global targets and local targets and marks the resource as reviewed"
- "does not show the review notice when resource was already reviewed"
Screenshot from 2023-06-10 14-25-48

When looking at the screenshots of the tests we see that the data is not being updated
failures_r_spec_example_groups_sdg_relations_edit_allows_adding_the_goals__global_targets_and_local_targets_and_marks_the_resource_as_reviewed_71

**Steps to reproduce:**
Follow the steps to install Consul locally on the [Consul documentation](https://docs.consulproject.org/docs/english-documentation/introduction/local_installation) and then run the tests
```
bin/rake db:test:prepare
bin/rspec --format html --out rspec_results.html

```
**Expected Result:**
All tests should pass without any failures.

**Actual Result:**
Two scenarios related to SDG goal/target management are failing.

## Objectives

This pull request aims to address and resolve 2 failing test scenarios related to SDG goal/target management. 
The proposed changes in this pull request aim to rectify the issues causing the test failure and improve code reusability.

Modifications Made:

1. Created a reusable function, click_sdg_goal_or_target_with_autocomplete, to add an SDG goal or target with autocomplete functionality. This function accepts a code parameter and simulates the interaction of filling in the "Goals and Targets" field and selecting the desired SDG goal or target **using the already existing working code on the same test** 

```
def click_sdg_goal_or_target_with_autocomplete(code)
   fill_in "Goals and Targets", with: code
   within(".amsify-list") do
       find("[data-val='#{code}']").click
    end
 end
```
  
2. . Fixed the non-working find(:css, ".sdg-related-list-selector-input").set("1.2, 2.1,") line with the newly created click_sdg_goal_or_target_with_autocomplete function. This modification ensures that the SDG goals or targets with codes are correctly selected.

3. . Replaced repetitive code blocks with calls to the click_sdg_goal_or_target_with_autocomplete function for improved code readability and future use.
Screenshot from 2023-06-10 14-38-33

## Visual Changes

There a no visual changes, as this pull request aims only at fixing failing Rpec tests

## Notes

Please note that in the function click_sdg_goal_or_target_with_autocomplete(code), the when code contains more that one dot, it must be put between brackets
```
 click_sdg_goal_or_target_with_autocomplete(1)
 click_sdg_goal_or_target_with_autocomplete(1.1)
 click_sdg_goal_or_target_with_autocomplete("1.1.1") 
 and not  click_sdg_goal_or_target_with_autocomplete(1.1.1)
```

### DIFF
--- a/spec/support/common_actions.rb
+++ b/spec/support/common_actions.rb
@@ -79,6 +79,13 @@ module CommonActions
     end
   end
 
+  def click_sdg_goal_or_target_with_autocomplete(code)
+    fill_in "Goals and Targets", with: code
+    within(".amsify-list") do
+      find("[data-val='#{code}']").click
+    end
+  end
+
   def remove_sdg_goal_or_target_tag(code)
     within "span[data-val='#{code}']" do
       click_button "Remove"
--- a/spec/system/sdg_management/relations_spec.rb
+++ b/spec/system/sdg_management/relations_spec.rb
@@ -297,7 +297,9 @@ describe "SDG Relations" do
 
       visit sdg_management_edit_legislation_process_path(process)
 
-      find(:css, ".sdg-related-list-selector-input").set("1.2, 2, 1.1.1, ")
+      click_sdg_goal(2)
+      click_sdg_goal_or_target_with_autocomplete(1.2)
+      click_sdg_goal_or_target_with_autocomplete("1.1.1")
 
       click_button "Update Process"
 
@@ -339,7 +341,8 @@ describe "SDG Relations" do
       debate = create(:sdg_review, relatable: create(:debate, title: "SDG debate")).relatable
 
       visit sdg_management_edit_debate_path(debate, filter: "sdg_reviewed")
-      find(:css, ".sdg-related-list-selector-input").set("1.2, 2.1,")
+      click_sdg_goal_or_target_with_autocomplete(1.2)
+      click_sdg_goal_or_target_with_autocomplete(2.1)
       click_button "Update Debate"
 
       expect(page).not_to have_content "Debate updated successfully and marked as reviewed"
--- a/spec/system/sdg_management/relations_spec.rb
+++ b/spec/system/sdg_management/relations_spec.rb
@@ -360,18 +360,15 @@ describe "SDG Relations" do
       create(:sdg_local_target, code: "1.1.1")
       visit sdg_management_edit_legislation_process_path(process)
 
-      fill_in "Goals and Targets", with: "3"
-      within(".amsify-list") { find(:css, "[data-val='3']").click }
+      click_sdg_goal_or_target_with_autocomplete(3)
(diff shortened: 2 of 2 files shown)


javierm:
Hi, @karim-semmoud :smile:.

Thank you very much for this pull request! :pray:

I can't reproduce the failing tests on my machine :thinking:. Do these tests always fail on your machine, or only sometimes? If it's just sometimes, have you found a way to run these tests so they fail more or less consistently (and, if possible, without the need to run the whole test suite)? If they always fail, could you tell us which version of the Chrome/Chromium browser are you using?

Thanks!

karim-semmoud:
Hi @javierm 
Thanks for your time and review.
These 2 specific tests are failing 100% of the time on a new installation, I was never able to make them pass successfully.
I also tried to run the whole suite or just the files but still the same, only these 2 tests fails

I am on Ubuntu 22.04.2 LTS
I tried the following versions of Chrome but with the same failed result
Google Chrome 114.0.5735.133 
Google Chrome 112.0.5615.165

javierm:
Unfortunately the test passes on our machines with the same browser, so we need more help :pray:. But it's great that you can reproduce it 100% of the time :smile:.

Could you add `save_and_open_screenshot` after the line `find(:css, ".sdg-related-list-selector-input").set("1.2, 2.1,")` (or the line `find(:css, ".sdg-related-list-selector-input").set("1.2, 2, 1.1.1, ")` in the other test) and run the test? Then you can send us the screenshot you get. 

The thing is, those tests check that, when people fill in that field with a value ending in a comma, the right goal/target is added, so we'd like to keep testing that somehow if possible :thinking:.

Once again, thanks!

karim-semmoud:
@javierm 
Thank you so much for your help, it really helped me look toward the right direction.
I was able to make the test work :partying_face: 
My laptop has an AZERTY keyboard, so I only had French keybord setup
I added an English QWERTY keyboard and made it as default.
The original tests all pass now without any changes

image

### Notes
- The English Qwerty must be in first position, as default for the test to pass.
- With French AZERTY in first position, it looks like the comma sign is not interpreted correctly
capybara-202306240919547605761342
- When trying in real on chrome browser with only French Azerty keyboard setup, it works fine. pressing the comma sign does add the tag. It is only the test that fails

Once again again for your help , it saves me a lot of headaches :relieved:

javierm:
Wow, thanks a lot for the explanation!

Next time we use unusual methods in the tests (like `set` here), we'll take into account that some tests might not pass on machines using a different keyboard layout.

Issue #5135, opened 2023-06-20
Markdown not rendering correctly


coslajohn:
### Description

[comment]: # (Feature or Bug? eg. Type: Bug)
*Type*: Bug

### Current Behavior
[comment]: # (Describe what actually happened.)
When using the text editor in Collaborative Legislation the markdown for a table renders correctly in the preview part of the text editor.
![image](https://github.com/consul/consul/assets/102246590/4e9b7718-bc50-4672-a67b-5c05e52bd719)

When the process is published and viewed in the process the table doesn't render and the markdown commands are shown as text

![image](https://github.com/consul/consul/assets/102246590/8c2f356b-6826-4446-bb2b-49e80bea2045)

The opposite behaviour happens with the markdown code for header # It renders correctly in the process but doesn't render in the  text editor preview

![image](https://github.com/consul/consul/assets/102246590/fb63883f-81f2-4293-9816-00bbdc0e868f)
![image](https://github.com/consul/consul/assets/102246590/547d3b42-519d-40af-b3e2-bbf53e5cc691)

### Expected Behavior
[comment]: # (Describe what you expected to happen.)

Markdown elements to render consistently

karim-semmoud:
Hi @coslajohn 
I hope you are doing great.

It looks like the gem [Redcarpet](https://github.com/vmg/redcarpet)  has `tables: false` by default

Can you try this changes 
in `lib/admin_legislation_sanitizer.rb` add `tr th td table thead tbody`
```
class AdminLegislationSanitizer < WYSIWYGSanitizer
  def allowed_tags
    super + %w[img h1 h4 h5 h6 tr th td table thead tbody]
  end

  def allowed_attributes
    super + %w[alt src id]
  end
end
```
in `app/models/legislation/draft_version.rb` add `tables: true`

```
  def body_html
    renderer = Redcarpet::Render::HTML.new(with_toc_data: true)

    Redcarpet::Markdown.new(renderer, tables: true).render(body)
  end
```

This this changes you can now see the tables
![Screenshot from 2023-06-21 10-27-01](https://github.com/consul/consul/assets/114252883/19c158db-7a6c-4c3a-b87c-da479ad5e5ac)
![Screenshot from 2023-06-21 10-26-30](https://github.com/consul/consul/assets/114252883/717e7c5a-f58b-401c-966e-ce3c250a88cf)

coslajohn:
Thanks @karim-semmoud  that seems to work

PR #5136, opened 2023-06-22
Render markdown tables in legislation draft

karim-semmoud:
## Reference
Markdown not rendering correctly #5135 

## Objectives

This pull request aims at fixing the display of tables in the Legislation Draft from as a user, it also add test in order to make sure that the tables are rendered.

1- Allow table tags in Admin Legislation Sanitizer

2- Add Tables option to Redcarpet in Legislation draft
We used render_options and extensions variable in order to align with the existing code at app/helpers/application_helper.rb

3- Add custom CSS for Table inside Legislation draft using global table variables. we also make the table responsive with scrolling

4- Add Test to render markdown tables in Legislation drafts as a user and as an admin

5- Add Test for Admin Legislation Sanitizer. We also strengthen allowed and not allowed parameters 

## Visual Changes

Screenshot from 2023-06-26 18-11-49
Screenshot from 2023-06-26 18-11-03

## Notes
- See [Redcarpet](https://github.com/vmg/redcarpet) for more options

### DIFF
--- a/app/assets/stylesheets/legislation_process.scss
+++ b/app/assets/stylesheets/legislation_process.scss
@@ -459,6 +459,10 @@
         padding: rem-calc(32) rem-calc(32) rem-calc(32) rem-calc(48);
       }
 
+      table {
+        @include table-scroll;
+      }
+
       h2 {
         font-weight: 400;
         margin-bottom: rem-calc(32);
--- a/app/models/legislation/draft_version.rb
+++ b/app/models/legislation/draft_version.rb
@@ -22,15 +22,7 @@ class Legislation::DraftVersion < ApplicationRecord
   scope :published, -> { where(status: "published").order("id DESC") }
 
   def body_html
-    render_options = {
-      with_toc_data: true
-    }
-    renderer = Redcarpet::Render::HTML.new(render_options)
-    extensions = {
-      tables: true
-    }
-
-    Redcarpet::Markdown.new(renderer, extensions).render(body)
+    ApplicationController.helpers.markdown(body, with_toc_data: true)
   end
 
   def toc_html
--- a/app/models/legislation/draft_version.rb
+++ b/app/models/legislation/draft_version.rb
@@ -22,13 +22,11 @@ class Legislation::DraftVersion < ApplicationRecord
   scope :published, -> { where(status: "published").order("id DESC") }
 
   def body_html
-    ApplicationController.helpers.markdown(body, with_toc_data: true)
+    MarkdownConverter.new(body, with_toc_data: true).render
   end
 
   def toc_html
-    renderer = Redcarpet::Render::HTML_TOC.new(with_toc_data: true)
-
-    Redcarpet::Markdown.new(renderer).render(body)
+    MarkdownConverter.new(body).render_toc
   end
 
   def display_title
--- a/app/models/legislation/draft_version.rb
+++ b/app/models/legislation/draft_version.rb
@@ -22,9 +22,15 @@ class Legislation::DraftVersion < ApplicationRecord
   scope :published, -> { where(status: "published").order("id DESC") }
 
   def body_html
-    renderer = Redcarpet::Render::HTML.new(with_toc_data: true)
+    render_options = {
+      with_toc_data: true
+    }
+    renderer = Redcarpet::Render::HTML.new(render_options)
+    extensions = {
+      tables: true
+    }
 
-    Redcarpet::Markdown.new(renderer).render(body)
+    Redcarpet::Markdown.new(renderer, extensions).render(body)
   end
 
   def toc_html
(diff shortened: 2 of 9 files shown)


app/assets/stylesheets/legislation_process.scss
```diff
@@ -986,3 +987,38 @@
...
+      &:nth-child(odd) {
+        background: $table-background;
+      }
+
+      &:nth-child(even) {
+        background: $table-striped-background;
+      }
+    }
+  }
+}
```
taitus, line 1024: I see that you have used the variables available for the tables. But currently in consul for the tables that appear in the public part the styles from `/assets/stylesheets/layout.scss:1203` are being applied.

What do you think if we remove these styles that have been added and then it will render the same as the rest of the tables?

What might be a good idea is to add the scroll. I don't think it would be necessary to add the new section // 10. Legislation draft tables and we can add the following code on `legislation_process.scss:503` inside the `.draft-text` class:

```
table {
  @include table-scroll;
}
```

app/models/legislation/draft_version.rb
```diff
@@ -22,9 +22,16 @@ class Legislation::DraftVersion < ApplicationRecord
   scope :published, -> { where(status: "published").order("id DESC") }
 
   def body_html
-    renderer = Redcarpet::Render::HTML.new(with_toc_data: true)
-
-    Redcarpet::Markdown.new(renderer).render(body)
+    # See https://github.com/vmg/redcarpet for options
```
taitus, line 25: At Consul we try to avoid putting comments in the code, the commit message might be a good place to add this reference to the options offered by RedCarpet 🤔

spec/system/legislation/draft_versions_spec.rb
```diff
@@ -397,4 +397,30 @@
...
+      expect(page).to have_css("table")
+      expect(page).to have_content "Roberta"
+      expect(page).to have_content "25"
+    end
+
+    scenario "See table as an admin" do
+      login_as(administrator)
+
+      visit path
+      click_link class: "fullscreen-toggle"
```
taitus, line 419: How about using `click_link "Launch text editor"` instead of `click_link class: "fullscreen-toggle"` in this test?

taitus (commented):
Hi, @karim-semmoud 

Thanks a lot for this pull request! 🎉

karim-semmoud:
Hi @taitus 
Thank you for the review and comments.

1 -How about using click_link "Launch text editor" instead of click_link class: "fullscreen-toggle" in this test?
I agree, I will make the changes 

2 -At Consul we try to avoid putting comments in the code, the commit message might be a good place to add this reference to the options offered by RedCarpet thinking
My bad, I copied the code from app/helpers/application_helper.rb without removing the comment line.
I also removed the comment line from application_helper.rb if that is ok

3-Table CSS
I agree with you , it makes more sense and keep consistency in the code 

I will make the changes necessary and commit them

Extra Note:
Regarding the CSS for the Legislation draft, we may need in the future to use custom CSS for all the element inside Legislation Draft.
If an installation uses "fancy" colors for the text, table, headings ( for example pink on black ), I don't think it is a good idea to reuse the same CSS inside the Legislation Draft as it is supposed to look neutral and "official".
Ideally we would set neutral and "official document "looking font family, size, colors, headings, tables, links etc ... 
Currently we only set the background color to white, but use the main body color etc ... 
But I guess this is a discussion to be continued on a different pull request

javierm:
>  I copied the code from app/helpers/application_helper.rb

This comment made me wonder why we've got two different ways of rendering markdown, with only one of them enabling certain extensions. Maybe it made sense before we stopped storing the generated HTML in the database but, if so, I think we can unify the code :thinking:.

@karim-semmoud @taitus Do you think it'd make sense? Here are a couple of commits that make the `Legislation::DraftVersion` use the same code as the `markdown` helper. @karim-semmoud If you think this is a good idea, could you add them to this pull request (using `git am <file>`)?

[0001-Use-the-same-extensions-in-all-markdown-renderers.txt](https://github.com/consul/consul/files/11908009/0001-Use-the-same-extensions-in-all-markdown-renderers.txt)
[0002-Extract-markdown-helper-logic-to-a-class.txt](https://github.com/consul/consul/files/11908011/0002-Extract-markdown-helper-logic-to-a-class.txt)

javierm (commented):
@karim-semmoud I've left a small comment; other than that, could you squash the first four commits of this pull request (all the commits except the ones I've posted in another comment) into one, since there are three commits that basically fix the first one?

Thanks a lot!

app/assets/stylesheets/legislation_process.scss
```diff
@@ -986,3 +990,4 @@
     transform: translateX(-50%);
   }
 }
+
```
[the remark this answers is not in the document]
karim-semmoud: Yes sure, it has been removed

karim-semmoud:
@javierm 
I applied and pushed your 2 commits as you advised
I also squashed my previous commits into a single one
Let me know if there is anything else needed

javierm (approved):
@karim-semmoud Once again, thank you very much for your contributions!

PR #5145, opened 2023-06-29
Enable mousewheel when focusing on the map

javierm:
## References

* We disabled the mouswheel in pull request #5097

## Objectives

* Make it possible to use the zoom with the mousewheel while avoiding the "scroll trap" (on desktop computers) at the same time

### DIFF
--- a/app/assets/javascripts/map.js
+++ b/app/assets/javascripts/map.js
@@ -81,12 +81,20 @@
       App.Map.addGeozones(map);
     },
     leafletMap: function(element) {
-      var centerData, mapCenterLatLng;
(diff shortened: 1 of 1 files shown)


Senen:
In the commit message, you mention the Leaflet Gesture Handling Plugin. I tried it a time ago, and it worked mostly fine except in Firefox. Maybe it was a problem with my phone where I tested it. Just in case it helps, you can find the commit below:

* https://github.com/rockandror/consul/commit/77663f526215a4131148154e409c9798c95b014a

Senen (approved):
👍🏼

PR #5146, opened 2023-06-29
Restart DelayedJob workers after they crash

javierm:
## References

* Closes consul/installer#212

## Background

We're receiving reports from Consul installations saying that, once in a while, DelayedJob processes stop.

In some cases we've solved this issue by monitoring these processes with Systemd or Monit, but implementing support for these tools in a way that works on existing installations is difficult.

On the other hand, DelayedJob provides a simple way to monitor its processes that might not be as powerful as systemd but it's much better than doing nothing and it's easy to make it work on existing installations

## Objectives

* Make it easier to maintain Consul installations running on production

## Notes

:warning: In order to stop delayed job, we now need to pass the `-n` option: `RAILS_ENV=production bin/delayed_job -n 2 stop`

We might switch to systemd in the future, particularly if we upgrade Puma (see pull request #4922).

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -40,6 +40,7 @@ set :puma_conf, "#{release_path}/config/puma/#{fetch(:rails_env)}.rb"
 
 set :delayed_job_workers, 2
 set :delayed_job_roles, :background
+set :delayed_job_monitor, true
 
 set :whenever_roles, -> { :app }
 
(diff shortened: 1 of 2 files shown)


Senen (approved):
I tested it in the staging server by killing the delayed_job processes by hand, and the processes were restarted automatically. 

Also tested a server reboot and the Capistrano `delayed_job` tasks (delayed_job:start delayed_job:stop delayed_job:restart). 

Also, checked how a deployment will create monitoring processes for existing installations.

PR #5148, opened 2023-07-07
Order expired polls by ends date

decabeza:
## Objectives

When a user visits the polls that have already finished they appear sorted by date of creation.

This PR changes the order so that expired polls appear sorted by ends date. I think it's more interesting and better for the UX to see the most recent processes  especially for new users who want to explore the latest web activity. 😌

### DIFF
--- a/app/models/poll.rb
+++ b/app/models/poll.rb
@@ -57,8 +57,25 @@ class Poll < ApplicationRecord
 
   def self.sort_for_list(user = nil)
     all.sort do |poll, another_poll|
-      [poll.weight(user), poll.starts_at, poll.name] <=>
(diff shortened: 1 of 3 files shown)


taitus:
Hi @decabeza 

Thanks for the PR 👏 

Just a question regarding this change, do you think it would make sense to add the "weight" concept in this new sorting for expired polls? 🤔 

Before this change, `weight` was given more importance than `starts_at`, which makes me think that it would be reasonable to keep the same criteria as before. That is to say first to take into account the `weight`, then in this case the `ends_at` and finally the `name`. 

Greetings!

taitus:
It has been added to the commit of this PR to also take into account the weight and name when sorting expired polls.

Senen (approved):

PR #5158, opened 2023-07-22
Depend on a specific version of Node.js

javierm:
## References

* Depends on #5157

## Objectives

* Make sure all Consul Democracy installations compile the assets using the same version of Node.js, so we don't get inconsistent errors
* Make it easier to use Node packages in the future

## Documentation

We need to update the [technical documentation](https://docs.consuldemocracy.org/tech_docs/) in order to add instructions about how to install a Node Version Manager and use it to install Node.js.


config/deploy.rb
```diff
@@ -89,6 +102,40 @@ def main_deploy_server
...
+          execute fetch(:fnm_update_command)
+        end
+
+        execute fetch(:fnm_install_node_command)
+      end
+    end
+  end
+end
+
+task :map_node_bins do
```
Senen, line 125: Hi @javierm, I'm experiencing an error when deploying in a server prepared with the new version of the installer (one without the former nodejs version 10):

```
00:39 delayed_job:restart
      01 /home/deploy/consul/rvm1scripts/rvm-auto.sh . bundle exec bin/delayed_job -m -n 2 restart
      01 bundler: failed to load command: bin/delayed_job (bin/delayed_job)
      01
      01 /home/deploy/consul/shared/bundle/ruby/3.1.0/gems/execjs-2.8.1/lib/execjs/runtimes.rb:58:in `autodetect'
      01 : Could not find a JavaScript runtime. See https://github.com/rails/execjs for a list of available runtimes. (ExecJS::RuntimeUnavailable)
      01 	from /home/deploy/consul/shared/bundle/ruby/3.1.0/gems/execjs-2.8.1/lib/execjs.rb:5:in `<module:ExecJS>'
      01 	from /home/deploy/consul/shared/bundle/ruby/3.1.0/gems/execjs-2.8.1/lib/execjs.rb:4:in `<top (required)>'
      01 	from /home/deploy/consul/shared/bundle/ruby/3.1.0/gems/activesupport-6.1.7.4/lib/active_support/dependencies.rb:332:in `require'
      01 	from /home/deploy/consul/shared/bundle/ruby/3.1.0/gems/activesupport-6.1.7.4/lib/active_support/dependencies.rb:332:in `block in require'
      01 	from /home/deploy/consul/shared/bundle/ruby/3.1.0/gems/activesupport-6.1.7.4/lib/active_support/dependencies.rb:299:in `load_dependency'
      01 	from /home/deploy/consul/shared/bundle/ruby/3.1.0/gems/activesupport-6.1.7.4/lib/active_support/dependencies.rb:332:in `require'
      01 	from /home/deploy/consul/shared/bundle/ruby/3.1.0/gems/autoprefixer-rails-8.2.0/lib/autoprefixer-rails/processor.rb:2:in `<top (required)>'
      01 	from /home/deploy/consul/shared/bundle/ruby/3.1.0/gems/autoprefixer-rails-8.2.0/lib/autoprefixer-rails.rb:36:in `require_relative'
      01 	from /home/deploy/consul/shared/bundle/ruby/3.1.0/gems/autoprefixer-rails-8.2.0/lib/autoprefixer-rails.rb:36:in `<top (required)>'
      01 	from /home/deploy/.rvm/rubies/ruby-3.1.4/lib/ruby/3.1.0/bundler/runtime.rb:60:in `require'
      01 	from /home/deploy/.rvm/rubies/ruby-3.1.4/lib/ruby/3.1.0/bundler/runtime.rb:60:in `block (2 levels) in require'
      01 	from /home/deploy/.rvm/rubies/ruby-3.1.4/lib/ruby/3.1.0/bundler/runtime.rb:55:in `each'
      01 	from /home/deploy/.rvm/rubies/ruby-3.1.4/lib/ruby/3.1.0/bundler/runtime.rb:55:in `block in require'
      01 	from /home/deploy/.rvm/rubies/ruby-3.1.4/lib/ruby/3.1.0/bundler/runtime.rb:44:in `each'
      01 	from /home/deploy/.rvm/rubies/ruby-3.1.4/lib/ruby/3.1.0/bundler/runtime.rb:44:in `require'
      01 	from /home/deploy/.rvm/rubies/ruby-3.1.4/lib/ruby/3.1.0/bundler.rb:186:in `require'
      01 	from /home/deploy/consul/releases/20231010154502/config/application.rb:20:in `<top (required)>'
      01 	from /home/deploy/consul/releases/20231010154502/config/environment.rb:2:in `require_relative'
      01 	from /home/deploy/consul/releases/20231010154502/config/environment.rb:2:in `<top (required)>'
      01 	from bin/delayed_job:3:in `require'
      01 	from bin/delayed_job:3:in `<top (required)>'
      01 	from /home/deploy/.rvm/rubies/ruby-3.1.4/lib/ruby/3.1.0/bundler/cli/exec.rb:58:in `load'
      01 	from /home/deploy/.rvm/rubies/ruby-3.1.4/lib/ruby/3.1.0/bundler/cli/exec.rb:58:in `kernel_load'
      01 	from /home/deploy/.rvm/rubies/ruby-3.1.4/lib/ruby/3.1.0/bundler/cli/exec.rb:23:in `run'
      01 	from /home/deploy/.rvm/rubies/ruby-3.1.4/lib/ruby/3.1.0/bundler/cli.rb:486:in `exec'
      01 	from /home/deploy/.rvm/rubies/ruby-3.1.4/lib/ruby/3.1.0/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
      01 	from /home/deploy/.rvm/rubies/ruby-3.1.4/lib/ruby/3.1.0/bundler/vendor/thor/lib/thor/invocation.rb:127:in `invoke_command'
      01 	from /home/deploy/.rvm/rubies/ruby-3.1.4/lib/ruby/3.1.0/bundler/vendor/thor/lib/thor.rb:392:in `dispatch'
      01 	from /home/deploy/.rvm/rubies/ruby-3.1.4/lib/ruby/3.1.0/bundler/cli.rb:31:in `dispatch'
      01 	from /home/deploy/.rvm/rubies/ruby-3.1.4/lib/ruby/3.1.0/bundler/vendor/thor/lib/thor/base.rb:485:in `start'
      01 	from /home/deploy/.rvm/rubies/ruby-3.1.4/lib/ruby/3.1.0/bundler/cli.rb:25:in `start'
      01 	from /home/deploy/.rvm/rubies/ruby-3.1.4/lib/ruby/gems/3.1.0/gems/bundler-2.3.26/libexec/bundle:48:in `block in <top (required)>'
      01 	from /home/deploy/.rvm/rubies/ruby-3.1.4/lib/ruby/3.1.0/bundler/friendly_errors.rb:120:in `with_friendly_errors'
      01 	from /home/deploy/.rvm/rubies/ruby-3.1.4/lib/ruby/gems/3.1.0/gems/bundler-2.3.26/libexec/bundle:36:in `<top (required)>'
      01 	from /home/deploy/.rvm/rubies/ruby-3.1.4/bin/bundle:31:in `load'
      01 	from /home/deploy/.rvm/rubies/ruby-3.1.4/bin/bundle:31:in `<main>'
      01 	from /home/deploy/.rvm/rubies/ruby-3.1.4/bin/ruby_executable_hooks:22:in `eval'
      01 	from /home/deploy/.rvm/rubies/ruby-3.1.4/bin/ruby_executable_hooks:22:in `<main>'
```


It seems it fails because we are not initializing the nodejs version with fnm. I tried to add delayed_job to map_node_bins with no luck.

Also realized that we probably need to install nodejs for the other application roles in case they run in separated servers: `web app db importer cron background`. What do you think?

javierm: > Also realized that we probably need to install nodejs for the other application roles in case they run in separated servers: web app db importer cron background. What do you think?

Right now, we only install Ruby on the `:app` role, so if we change this, I guess we'd have to change that too :thinking:.

I don't know about this because I've never worked in an environment where a server didn't have the `app` role.

javierm: > I tried to add delayed_job to map_node_bins with no luck

Just to confirm, have you tried adding `bin/delayed_job`, `delayed_job`, or both?

javierm: > I tried to add delayed_job to map_node_bins with no luck

Just to confirm, have you tried adding `bin/delayed_job`, `delayed_job`, or both?

Maybe we could try setting `:delayed_job_bin_path` to `fnm exec bin` :thinking:, although it looks like a hack.

Senen: I tried both `delayed_job`, `bin/delayed_job`. Today I tried `bundle exec bin/delayed job` but it failed because the command contains multiple words.

Now I tried `bundle`, and it works fine. Two deployment commands were changed after this update:

```
00:38 whenever:update_crontab
      01 export PATH="$HOME/.fnm:$PATH" && cd /home/deploy/consul/releases/20231011070838 && eval "$(fnm env)" && fnm exec /home/deploy/consul/rvm1scripts/rvm-auto.sh . bundle exec whe…
      01 [write] crontab file updated
      01
    ✔ 01 deploy@127.0.0.1 1.106s
00:40 delayed_job:restart
      01 export PATH="$HOME/.fnm:$PATH" && cd /home/deploy/consul/releases/20231011070838 && eval "$(fnm env)" && fnm exec /home/deploy/consul/rvm1scripts/rvm-auto.sh . bundle exec bin…
      01 delayed_job.0: trying to stop process with pid 45804 sending TERM and waiting 20s ...
      01 delayed_job.0: process with pid 45804 successfully stopped.
      01
      01 delayed_job.0: process with pid 231965 started.
      01
      01 delayed_job.1: trying to stop process with pid 45822 sending TERM and waiting 20s ...
      01
      01 delayed_job.1: process with pid 45822 successfully stopped.
      01
      01 delayed_job.1: process with pid 231979 started.
    ✔ 01 deploy@127.0.0.1 12.065s
```

Senen: > Right now, we only install Ruby on the :app role, so if we change this, I guess we'd have to change that too 🤔.

> I don't know about this because I've never worked in an environment where a server didn't have the app role.

If we want to run a separate server for the background tasks (delayed_job) we can set the `app` and the `background` roles to that server and it should work. 👍🏼 

I'm also not an expert in executing consul in separate servers with different roles. I was worried that if we wanted to restart delayed_job with Capistrano in a server that does not have the nodejs version it would fail. But the solution is as easy as adding the role app to that server and that's it, capistrano will install Ruby, nodejs and all the dependencies.

Senen:
Hi @javierm, I've been testing deployments from this implementation (with `bundle` within `map_node_bins`) on servers built with the installer with Consul Democracy 2.0.1 and on servers with Consul Democracy with the new nodejs version:

It worked well in all cases, but I found that on the server with Consul democracy 2.0.1, the fnm entries were not added to the deploy user `.bashrc` file, as you can see in the following deployment log:

```
03:00 install_node
      01 export PATH="$HOME/.fnm:$PATH" && cd /home/deploy/consul/releases/20231011105903 && eval "$(fnm env)" && fnm use --install-if-missing
      01 bash: fnm: command not found
      01
      01 bash: fnm: command not found
      02 export PATH="$HOME/.fnm:$PATH" && cd /home/deploy/consul/releases/20231011105903 && eval "$(fnm env)"
      02 bash: fnm: command not found
      02
    ✔ 02 deploy@127.0.0.1 0.005s
      03 curl -fsSL https://fnm.vercel.app/install | bash -s -- --install-dir "$HOME/.fnm" --skip-shell
      03 Checking dependencies for the installation script...
      03 Checking availability of curl...
      03 OK!
      03 Checking availability of unzip...
      03 OK!
      03 Downloading https://github.com/Schniz/fnm/releases/latest/download/fnm-linux.zip...
      03 #=#=#
      03 ##O#- #
      03 ##O=#  #
      03 #=#=-#  #
      03 -#O#- #   #
                                                                           0.3%
##                                                                         3.6%
#######                                                                   10.1%
############                                                              17.4%
#################                                                         24.6%
#######################                                                   32.7%
#############################                                             40.5%
##################################                                        48.3%
#######################################                                   54.8%
###########################################                               60.0%
##################################################                        70.2%
#######################################################                   76.9%
#############################################################             85.8%
#####################################################################     97.0%
######################################################################## 100.0%
    ✔ 03 deploy@127.0.0.1 2.713s
      01 Installing Node v18.18.0 (x64)
      01 Using Node v18.18.0
      01
    ✔ 01 deploy@127.0.0.1 8.649s
```

javierm:
> I think adding those entries to the deploy user .bashrc file is important so interactive sessions load the proper nodejs environment through fnm automatically.

I agree! No idea what's going on here :thinking:. Maybe next week we can share a screen and try to find out.

config/deploy.rb
javierm: > Now I tried bundle, and it works fine. Two deployment commands were changed after this update:

Great catch!

While testing this changed, I noticed two additional issues we had:

* Running tasks like `cap staging deploy:restart` wouldn't work because the `map_node_bins` task wasn't executed in this case
* Running `cap staging puma:start` failed as well (although `puma:restart` worked)

So I've updated the code in order to make sure `map_node_bins` is always executed and `puma` is included in the `fnm_map_bins` list.

javierm: So now, instead of running `rake`, we run:

```bash
export PATH="$HOME/.fnm:$PATH" && cd /home/deploy/consul/current && eval "$(fnm env)" && fnm exec /home/deploy/consul/rvm1scripts/rvm-auto.sh . bundle exec rake
```

Sometimes the world is a horrible place :rofl:.

Senen: > Sometimes the world of programming is a horrible place 🤣.

Yes, I had the same feeling these days 😓 . Hopefully, at some point we might simplify it a bit 🤞🏼

Senen:
Hi @javierm, 

Debugging a few deployments, I found that when executing the `fnm_setup_command`, it fails when `fnm` is not installed as expected, but it does not throw an exception, so the `fnm_install_command` is never reached. This silent failure is because the command `fnm_setup_command` uses eval, and the failure happens within the bash eval function.

Adding any `fnm` command in the `fnm_setup_command` fails as expected and throws the exception. I tried the following, and it worked fine:

```
set :fnm_setup_command, -> do
                          "export PATH=\"#{fetch(:fnm_path)}:$PATH\" && " \
                            "cd #{release_path} && fnm env && eval \"$(fnm env)\""
                        end
```

config/deploy.rb
Senen: > So I've updated the code in order to make sure map_node_bins is always executed and puma is included in the fnm_map_bins list. [Changes done today](https://github.com/consuldemocracy/consuldemocracy/compare/588c43892ca..d6eb6ce6ad3353049d826600d8e60e6815c83f80).

It seems everything works like a charm now! 👍🏼

config/deploy.rb
```diff
@@ -37,6 +37,17 @@ def main_deploy_server
 
 set :local_user, ENV["USER"]
 
+set :fnm_path, "$HOME/.fnm"
+set :fnm_install_command, "curl -fsSL https://fnm.vercel.app/install | " \
+                          "bash -s -- --install-dir \"#{fetch(:fnm_path)}\""
+set :fnm_update_command, "#{fetch(:fnm_install_command)} --skip-shell"
+set :fnm_setup_command, -> do
```
Senen, line 44: > I agree! No idea what's going on here 🤔. Maybe next week we can share a screen and try to find out.

Hi @javierm, 

Debugging a few deployments, I found that when executing the `fnm_setup_command`, it fails when `fnm` is not installed as expected, but it does not throw an exception, so the `fnm_install_command` is never reached. This silent failure is because the command `fnm_setup_command` uses eval, and the failure happens within the bash eval function.

Adding any `fnm` command in the `fnm_setup_command` fails as expected and throws the exception. I tried the following, and it worked fine:

```
set :fnm_setup_command, -> do
                          "export PATH=\"#{fetch(:fnm_path)}:$PATH\" && " \
                            "cd #{release_path} && fnm env && eval \"$(fnm env)\""
                        end
```

javierm:
@Senen I've updated the code:

* Added `> /dev/null` to the `fnm env` command so we don't constantly get its output while deploying
* Run `map_node_bins` after `git:create_release` so we don't use the wrong folder when running the FNM setup command
* Install node before running `bundle`, since `bundle` now depends on Node being installed :relieved:

Tried every scenario I could think :crossed_fingers:.

javierm:
> Another thing we can do is add the required NodeJS version in the README file and mention the need to install the package.json dependencies to precompile assets.

I've updated the README to add the required Node.js version :+1:. Not sure about the package.json dependencies, but I guess that one belongs to pull request #5159 :thinking:.

javierm:
> I think now we should force the Execjs runtime environment so if developers do not have NodeJS installed, they get a very explicit error.

Good question :thinking:.

Dockerfile
```diff
@@ -40,6 +39,13 @@ RUN mkdir -p $RAILS_ROOT/tmp/pids
 # Set our working directory inside the image
 WORKDIR $RAILS_ROOT
 
+# Install Node
+COPY .node-version ./
+ENV PATH=/usr/local/node/bin:$PATH
```
Senen, line 44: Hi @javierm, I think we need to add the folder (`/usr/local/node/bin`), where we install NodeJS, to the `/etc/sudoers.d/secure_path` in line 30,  so the node binaries are available for the consul user. Without this change, I have the Javascript Runtime not found error when running any command that loads the application.

javierm: @Senen Awesome! :tada: I've just updated the pull request.

Senen (approved):
Hi @javierm,

As we talked about, if we finally decide to set the Execjs runtime to NodeJS, we'll do it in a separate PR.

javierm:
@Senen We finally made it! :tada:

PR #5159, opened 2023-07-22
Allow adding Node packages as dependencies

javierm:
## References

* Depends on #5158

## Objectives

* Properly organize our JavaScript dependencies
* Allow using JavaScript libraries that are only distributed as packages


config/deploy.rb
```diff
@@ -30,7 +30,7 @@ def main_deploy_server
 set :use_sudo, false
 
 set :linked_files, %w[config/database.yml config/secrets.yml]
-set :linked_dirs, %w[.bundle log tmp public/system public/assets
+set :linked_dirs, %w[.bundle log tmp node_modules public/system public/assets
```
Senen, line 33: Hi @javierm,

I'm not sure about this change as it seems it's behaving unexpectedly 🤔 .

I think our intention is to have a shared folder named `node_modules` and reuse its contents among different releases. The `npm install` command overwrites the symlink and creates a real `node_modules` folder within the new release instead of respecting the symlinked directory. Maybe [it's related to this old npm issue](https://github.com/npm/npm/issues/10013) 🤔 .

It's happening the same to me when running `npm install` from the Consul Democracy installer consuldemocracy/installer#223 changes.

Two solutions come to my mind:
1º Keep investigations on how to use the shared folder
2º Do not use a shared folder and install all the application nodejs dependencies within each release folder

Senen: I found [another old Capistrano thread](https://github.com/capistrano/npm/issues/7) talking about the usage of shared folders and node_modules.

Quoting the link above:
>One downside to having node_modules in shared/ is you'd have to manually (by default) run npm install again on a rollback. If you think about it, node_modules aren't really shared since they're tied to a specific release (their versions at least). It's wasteful, and a bit slower, but it might be more proper to not share them.

javierm: > If you think about it, node_modules aren't really shared since they're tied to a specific release (their versions at least)

I see :thinking:. I guess what we need (ideally) is something similar to the `bundle` folder: a place where all versions of all packages are downloaded and stored forever, so they don't need to be re-downloaded on each deploy if they've already been downloaded in the past.

I'll have a look.

javierm: Oh, apparently NPM automatically caches packages at `~/.npm`, so I guess we're fine here :ok_hand:.

I've updated the pull request so we don't share the `node_modules` folder across releases :+1:.

.dockerignore
```diff
@@ -24,6 +24,7 @@ public/tenants/*/sitemap.xml
 public/assets/
 public/machine_learning/data/
 public/tenants/*/machine_learning/data/
+node_modules/
```
Senen, line 27: Hi @javierm, 

Should we remove this line following the same conclusion we reached [here](https://github.com/consuldemocracy/consuldemocracy/pull/5159#discussion_r1350765100)?

Senen (approved):
👏🏼

PR #5163, opened 2023-07-28
ENS: Enable password complexity

taitus:
## Objectives

As it seems that adding complexity to the password is something that
might be wanted from the Consul applications, we added the necessary
changes to allow it.

In this version we simply:
- Uncomment the configuration variable "password_complexity"
- Set this variable without any restrictions
- Adapt the application so that everything still works normally.

### DIFF
--- a/app/assets/javascripts/managers.js
+++ b/app/assets/javascripts/managers.js
@@ -2,16 +2,33 @@
   "use strict";
   App.Managers = {
     generatePassword: function() {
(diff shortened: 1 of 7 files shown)


Senen (commented):
Hi @taitus, I just saw a duplicate comment! Feel free to ignore it. I'll approve it when you decide whether that comment leaves or stays with us. 😉

Senen (approved):

PR #5166, opened 2023-08-05
Reduce the size of JavaScript code on most pages

javierm:
:warning: Currently these changes seem to break the navigation with the browser's back button in some cases, probably due to an incompatibility with Turbolinks.

## Objectives

* Make the page load faster, particularly on devices with a slow internet connection

## Notes

With these changes, on production environments (where assets are compressed), the size of the JavaScript file in most of the pages is reduced from 461KB to 141KB. Overall, our [demo homepage](https://demo.consuldemocracy.org) is now 23% smaller (from 1.37MB to 1.05MB).

### DIFF
--- a/app/assets/config/manifest.js
+++ b/app/assets/config/manifest.js
@@ -6,6 +6,7 @@
 //= link application-rtl.css
 //= link application.js
 //= link html_editor_loader.js
+//= link legislation_annotatable_loader.js
 //= link map_loader.js
 //= link markdown_editor_loader.js
 //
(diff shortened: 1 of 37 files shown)


javierm:
Closing since I haven't managed to make these changes compatible with Turbolinks. I've tried both loading the scripts inside the `<head>` tag and loading them inside the `<body>` tag, and both options fail in different ways.

Hopefully we'll find a way to deal with this issue in the future, but for now, we'll keep working with Turbolinks like we've done so far, using just one huge JavaScript file that loads on every page.

Issue #5167, opened 2023-08-07
Error when trying to change the name of an image


microweb10:
### Description

When updating an image's name, the attachment should not be checked.

### Steps to reproduce in https://demo.consuldemocracy.org

1. Set the max size for images to 10Mb in [the admin section](https://demo.consuldemocracy.org/admin/settings#tab-images-and-documents).
2. Create any process that allows you to attach an image, for instance, [a new Proposal](https://demo.consuldemocracy.org/proposals/new), with an attached image of 10Mb size.
3. Set the max size for images to 1Mb in [the admin section](https://demo.consuldemocracy.org/admin/settings#tab-images-and-documents).
4. Try to edit the name of the image attached in [the edit a Proposal form](https://demo.consuldemocracy.org/proposals/7/edit).

You will get the following error message:
```
1 error prevented this Proposal from being saved. 
Please check the marked fields to know how to correct them:
```
But no field will be marked in red color.

![Screenshot](https://github.com/consuldemocracy/consuldemocracy/assets/942995/7efaf1ff-edcf-4c49-89ea-643a73000353)

Senen:
Solved by:
*  #5168

PR #5168, opened 2023-08-07
Do not validate the attachment when updating the name

microweb10:
## References

https://github.com/consuldemocracy/consuldemocracy/issues/5167

## Objective

Do not run file size and content type validations for already uploaded attachments.

This change solves the problem described in #5167. 

As the administrator may change the accepted mime types for documents or images through the application settings, applying the same to file content type validation makes sense.

## Notes

1. Feel free to refactor or add specs if you think it's needed. 😉 
2. I don't know if it makes sense to keep [this validation](https://github.com/consuldemocracy/consuldemocracy/blob/master/app/models/image.rb#L20). 🤔

### DIFF
--- a/app/models/concerns/attachable.rb
+++ b/app/models/concerns/attachable.rb
@@ -9,7 +9,7 @@ module Attachable
               presence: true,
               file_content_type: {
                 allow: ->(record) { record.accepted_content_types },
-                if: -> { association_class && attachment.attached? },
+                if: -> { association_class && attachment.attached? && attachment.new_record? },
                 message: ->(record, *) do
                   I18n.t("#{record.model_name.plural}.errors.messages.wrong_content_type",
                          content_type: record.attachment_content_type,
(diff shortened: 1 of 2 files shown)


Senen:
Hi @microweb10, sorry for the late response! 🙏🏼 

Thank you for the pull request! 🎉  I noticed it was happening the same with the file content type validation, so I added another commit and a few specs.

> I don't know if it makes sense to keep [this validation](https://github.com/consuldemocracy/consuldemocracy/blob/master/app/models/image.rb#L20). 🤔

I do not understand what you mean 🤔. The image dimensions validation does not have the same problem, as it only applies to new attachments. In my opinion, this validation makes sense as it allows administrators to control the uploaded image's minimum width and height and this can help users to choose images with better quality that will look better than smaller images.

microweb10:
> I do not understand what you mean 🤔. The image dimensions validation does not have the same problem, as it only applies to new attachments. In my opinion, this validation makes sense as it allows administrators to control the uploaded image's minimum width and height and this can help users to choose images with better quality that will look better than smaller images. 

Never mind 👌 I think I got confused with something else. 😅

PR #5173, opened 2023-08-10
Replace back link on budget results

decabeza:
## Objectives

To enhance the UX a bit, this PR modifies the "Go back" link on the results page of a Participatory Budget. 

Instead of redirecting users to `/budgets`, which could be confusing if there are multiple PBs, it now directs them to the corresponding PB. This change helps maintain context, especially when dealing with different PBs.

### DIFF
--- a/app/views/budgets/executions/show.html.erb
+++ b/app/views/budgets/executions/show.html.erb
@@ -15,7 +15,7 @@
   <div class="expanded no-margin-top padding header">
     <div class="row">
       <div class="small-12 column">
-        <%= back_link_to budgets_path %>
+        <%= back_link_to budget_path(@budget) %>
         <h2 class="margin-top">
           <%= t("budgets.executions.heading") %><br>
           <span><%= @budget.name %></span>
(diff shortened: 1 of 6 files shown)


decabeza:
Hi @taitus!

About the first comment I really don't know. 😅 I think it was a while ago, the commit you send is from 2018, and I think it was simply because the [budgets/index](https://github.com/consuldemocracy/consuldemocracy/blob/c6fead6a728efb3b07bd7ffa1a4473657cfa89d4/app/views/budgets/index.html.erb) and [budgets/show](https://github.com/consuldemocracy/consuldemocracy/blob/c6fead6a728efb3b07bd7ffa1a4473657cfa89d4/app/views/budgets/show.html.erb) views were very different, and for some reason it was decided to show the index instead of the show. 🤔 

Currently the [budgets/index](https://github.com/consuldemocracy/consuldemocracy/blob/master/app/views/budgets/index.html.erb) and [budgets/show](https://github.com/consuldemocracy/consuldemocracy/blob/master/app/views/budgets/show.html.erb) views barely have differences, but the idea of the change in this PR is that the user does not lose the context in the case that there are several budgets open simultaneously. 😌 

About the second comment, I have added a couple of commits to the PR to change the go back link also for stats and milestones. 😉

taitus (approved):

PR #5174, opened 2023-08-11
Order homepage cards by creation date

decabeza:
## Objectives

When an administrator creates cards for the homepage from `admin/widget/cards/new` they are displayed in order of **creation date** showing the _newly created one last_.

Since this section is usually used to show "what's new", this PR changes the order to show the cards by **descending creation date** so that the _last card created is shown first_.

### DIFF
--- a/app/models/widget/card.rb
+++ b/app/models/widget/card.rb
@@ -16,6 +16,6 @@ class Widget::Card < ApplicationRecord
   end
 
   def self.body
-    where(header: false, cardable_id: nil).order(:created_at)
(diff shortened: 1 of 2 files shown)


taitus:
Hi @decabeza 

Since we don't know how changing the order of the card display might affect existing installations, we've created this [PR](https://github.com/consuldemocracy/consuldemocracy/pull/5276) to allow for a customized order to be defined for each installation.

Thanks, without this PR, this need would not have been identified 👏

PR #5170, opened 2023-08-09
New Crowdin updates

ConsulBot:

PR #5176, opened 2023-08-17
Update translations from Crowdin

Senen:
## References

Replaces PR:
* #5170 

## Objectives

Update translations to the latest and replace old references.


config/locales/sl-SI/rails.yml
```diff
@@ -10,18 +10,18 @@ sl:
...
+      -
+      -
+      -
       - Maj
-      - Jun
-      - Jul
+      -
+      -
       - Avg
-      - Sep
```
javierm, line 21: Since now some months are present and some aren't, we're getting a strange behavior :thinking:.

```
I18n.with_locale(:sl) { l 1.month.ago }
=> "Pon, 17  2023 20:59:10 +0200"
I18n.with_locale(:sl) { l Time.current }
=> "Čet, 17 Avg 2023 20:59:21 +0200"
```

Note that in the first case the month isn't displayed at all, not even in English.

Since Slovenian isn't even enabled, I'm fine with it behaving this way, though :ok_hand:. So I'm going to approve this pull request as is, but if you feel we need to change it, feel free to do so.

javierm (approved):

config/locales/sl-SI/rails.yml
Senen: > Note that in the first case the month isn't displayed at all, not even in English.

Yes, I know. Translation arrays do not access the fallback when the array translation key is present in the current locale. So translation arrays may be fully translated or will behave this way.  IMO, we have three options here:

1. Leave it as it is. IMO, this version reflects better the current Slovenian translations at Crowdin. As you said, Slovenian is not enabled.
2. Remove Slovenian translations from Crowdin so the array translation key will be removed and fallbacks will work properly. This option handicap is that will remove contributions from translators.
3. Remove incomplete translation arrays from this PR. The handicap here is that Crowdin will recreate those changes in the next PR


I think the best is the first option.

Senen: Having an option in Crowdin to avoid sending incomplete arrays would be nice 🤔.

Issue #4995, opened 2022-09-23
Changes to simplify the use of Admin interface


Pomerange:
### Why is this Needed?

The following ideas are proposed to simplify the use of the Administration interface to make it easier to use and understand for new admins of Consul. The current interface is based on how the software is organised and has grown organically as consul was more developed. The new interface should help organise CONSUL from a administrator's point of view. 

### Description

The issue contains different ideas to reorganise the menus that exist within the administration interface. 

Left menu bar reorganization. The bar can be split in four different types of menus: 
a) Modules menus (Proposals (with submenu proposals dashboard), Debates, Comments, PB, Polls, Collaborative legislation) 
b) Offline participation menus (Voting booths, Signature sheets) 
c) General administration tasks (Site content, Messages to users, Moderated content, Profiles, Statistics) 
d) Advanced administration (Settings)

Settings/Advanced administration reorganization.
- Move “level 1-5 public official” settings to the corresponding “Profiles” menu. 
- Split the “Configuration settings” in multiple tabs. For example: a) the settings that have to do with social networks, website title, etc; b) the settings that have to do with modules settings as number of supports for a proposal, months to archive proposals, etc; 
- “Map configuration” settings should be merged with the “Manage Geozones” menu, and rename it “Map and Geozones”. 
- “Remote census configuration” should be merged with the “Manage local census” and rename it “Census/User verification”

Move the submenu Moderated content. This menu in a first look seems redundant with the moderation interface. It is difficult to understand that this is a “double check” menu of the moderators actions. It would be better to move it to the Moderation interface (it would be in any case only visible to admins, not to moderators), and rename it with a clear name “Administrators review of moderated content”.

Rename Milestones. The “milestones” elements (appearing in Proposals, Collaborative legislation, PB, etc) are wrongly named. They are not “Milestones”, but “Project progress updates”. It would be better to use this description (even when longer), to make it easy to understand without additional explanations. The same with the related tabs “Following” that could be also called “Project progress updates”

### Definition of Done
This is a large issue with different elements, which may be worked on separately or prioritised differently than described here.

javierm:
Regarding the menu, one thing I'd like to watch out for: we currently have a two-level menu (that is, some menus have submenus). Introducing a third level (submenus with submenus) might also bring its own set of usability issues. We'll need to check that.

[...]

PR #5251, opened 2023-09-17
Move customizations to more intuitive admin menu section

dnsos:
## References

#4995 (no complete resolution of this issue, but an improvement discussed in it)

## Objectives

There were already some menu items to customization pages under the "Site content" admin menu ("Custom pages", "Custom information texts", "Custom documents"). It therefore makes sense to move "Custom images" and "Custom content blocks" (which were previously located under "Settings") to "Site content" as well.

## Visual Changes

New site content menu
New Settings menu

## Notes

It's possible that some admins may have memorized "Custom images" and "Custom content blocks" to be located under "Settings". However, it is more intuitive to have them under "Site content", so the possible confusion at first will probably pay off in the long-term, when all admins have more intuitive access to them.

### DIFF
--- a/app/components/admin/menu_component.rb
+++ b/app/components/admin/menu_component.rb
@@ -55,8 +55,7 @@ class Admin::MenuComponent < ApplicationComponent
 
     def settings?
       controllers_names = ["settings", "tenants", "tags", "geozones", "local_census_records", "imports"]
-      controllers_names.include?(controller_name) &&
-        controller.class.module_parent != Admin::Poll::Questions::Answers
+      controllers_names.include?(controller_name)
     end
 
     def customization?
--- a/app/components/admin/menu_component.rb
+++ b/app/components/admin/menu_component.rb
@@ -54,15 +54,14 @@ class Admin::MenuComponent < ApplicationComponent
     end
 
     def settings?
-      controllers_names = ["settings", "tenants", "tags", "geozones", "images",
-                           "content_blocks", "local_census_records", "imports"]
+      controllers_names = ["settings", "tenants", "tags", "geozones", "local_census_records", "imports"]
       controllers_names.include?(controller_name) &&
         controller.class.module_parent != Admin::Poll::Questions::Answers
     end
 
     def customization?
-      ["pages", "banners", "information_texts", "documents"].include?(controller_name) ||
-        homepage? || pages?
+      controllers_names = ["pages", "banners", "information_texts", "documents", "images", "content_blocks"]
+      controllers_names.include?(controller_name) || homepage? || pages?
     end
 
     def homepage?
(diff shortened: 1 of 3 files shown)


taitus (approved):

app/components/admin/menu_component.rb
```diff
@@ -54,15 +54,14 @@ def profiles?
     end
 
     def settings?
-      controllers_names = ["settings", "tenants", "tags", "geozones", "images",
-                           "content_blocks", "local_census_records", "imports"]
+      controllers_names = ["settings", "tenants", "tags", "geozones", "local_census_records", "imports"]
       controllers_names.include?(controller_name) &&
         controller.class.module_parent != Admin::Poll::Questions::Answers
```
taitus, line 59: Hi @dnsos,

Reviewing this PR, I think this line was added to allow for the "images" value in the controllers_names.

Could it be that by moving this "images" to another method, this line can be removed?

javierm:
@dnsos Sorry for intruding! :pray: I updated the branch so the tests would run (didn't run last time for some reason). I've just undone the changes.

app/components/admin/menu_component.rb
dnsos: Hi @taitus, thanks for the hint, I wasn't aware of that special case.

After removing the line I have tested the app manually and there is indeed no problem. I've tried to explain this to the best of my knowledge in [809875e](https://github.com/consuldemocracy/consuldemocracy/pull/5251/commits/809875ed357b95f9edddf881dd3439b70f53e625).

 A spec to verify that the line is unnecessary could look something like that (note the used controller and the text to match):

```rb
# spec/components/admin/menu_component_spec.rb
require "rails_helper"

describe Admin::MenuComponent, :admin, controller: Admin::Poll::Questions::Answers::ImagesController do
  let(:component) { Admin::MenuComponent.new }

  it "highlights currently active list item" do
    render_inline component
    expect(page).to have_css "[aria-current='true'] a", exact_text: "Polls"
  end
end
```

It feels a little overspecific though, given that the menu component has a lot of other cases that wouldn't be covered by such a minimal test. Happy to receive directions on whether such a test would be valuable.

taitus: Hi @dnsos 

Thanks for adding the changes. The new commit added is just what was missing. 👏 

In my opinion the test you suggest is not necessary to add for now. Since we don't seem to have any test for that component and it's a bit weird that there is only one and it's at that level of detail. 🤔 

What you could do if you don't mind would be to rebase the `master` branch and remove the commit `Merge branch 'consuldemocracy:master' into intuitive-admin-menu-sections`. I think that this way it will be ready for the merge.

Thank you very much!

dnsos: Hey @taitus thanks for the reply and I agree on your opinion about the test.

The merge commit is now gone!

javierm:
@dnsos Thank you very much! :smile:

PR #5273, opened 2023-10-09
ENS: Enable lockable devise module

taitus:
# Objectives
Add lockable devise module in order to use:
- maximum_attempts
- unlock_in

# How
- Allow to enable/disable devise lockable through secrets (devise_lockable). This secret is at the application level. This means that it cannot be configured for each Tenant.

- Allow customize `maximum_attemps` and `unlock_in` for each Tenant with the desired value. Or, as we can do with multitenant, which defines some default values, use the same values for all tenants and only overwrite them in the tenants that want to modify them.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -425,6 +425,14 @@ class User < ApplicationRecord
     end
   end
 
+  def self.maximum_attempts
+    (Tenant.current_secrets.dig(:security, :lockable, :maximum_attempts) || 20).to_i
+  end
+
+  def self.unlock_in
+    (Tenant.current_secrets.dig(:security, :lockable, :unlock_in) || 1).to_f.hours
+  end
+
   private
 
     def clean_document_number
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -3,7 +3,7 @@ class User < ApplicationRecord
   attribute :registering_from_web, default: false
 
   devise :database_authenticatable, :registerable, :confirmable, :recoverable, :rememberable,
-         :trackable, :validatable, :omniauthable, :password_expirable, :secure_validatable,
+         :trackable, :validatable, :omniauthable, :password_expirable, :secure_validatable, :lockable,
          authentication_keys: [:login]
 
   acts_as_voter
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -3,8 +3,9 @@ class User < ApplicationRecord
   attribute :registering_from_web, default: false
 
   devise :database_authenticatable, :registerable, :confirmable, :recoverable, :rememberable,
-         :trackable, :validatable, :omniauthable, :password_expirable, :secure_validatable, :lockable,
+         :trackable, :validatable, :omniauthable, :password_expirable, :secure_validatable,
          authentication_keys: [:login]
+  devise :lockable if Rails.application.config.devise_lockable
 
   acts_as_voter
   acts_as_paranoid column: :hidden_at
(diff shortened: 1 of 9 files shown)


app/models/custom/user.rb
```diff
@@ -0,0 +1,5 @@
+require_dependency Rails.root.join("app", "models", "user").to_s
+
+class User
+  devise :lockable if Rails.application.config.devise_lockable
```
Senen, line 4: We can add this line to the `app/models/user.rb` file directly. It seems to work fine!

```
  devise :database_authenticatable, :registerable, :confirmable, :recoverable, :rememberable,
         :trackable, :validatable, :omniauthable, :password_expirable, :secure_validatable,
         authentication_keys: [:login]
  devise :lockable if Rails.application.config.devise_lockable
```

config/application_custom.rb
```diff
@@ -1,4 +1,5 @@
 module Consul
   class Application < Rails::Application
+    config.devise_lockable = Rails.application.secrets.devise_lockable
```
Senen, line 3: We can move this line to the `config/application.rb` file.

Senen (commented):
Hi @taitus, I found two potential changes. Let me know what you think.

Checking the original MR!28 from Cabildo de Tenerife - Isla Abierta repository I think we can add the specs added to the [spec/model/custom/user_spec.rb](https://git.nubarron.teide.int/shs-consul/cabildoabierto-consul/-/merge_requests/28/diffs#diff-content-0fadc12704ff3803d554ab2564ae756ba752fd61) to the `spec/models/user_spec.rb`. What do you think?

app/models/user.rb
```diff
@@ -416,6 +416,14 @@ def add_subscriptions_token
     update!(subscriptions_token: SecureRandom.base58(32)) if subscriptions_token.blank?
   end
 
+  def self.maximum_attempts
+    (Tenant.current_secrets.security[:lockable][:maximum_attempts] || 20).to_i
```
Senen, line 420: This version is giving me problems with other specs. Why not the [Cabildo Teneride implementation](https://git.nubarron.teide.int/shs-consul/cabildoabierto-consul/-/merge_requests/28/diffs#0c788cab1662321bd1e65ad8c84fbe4b6008c822_17_16)?

```
  1) Users::SessionsController access logs when feature is enabled when a sign in process fails it calls the authentication logger
     Failure/Error: (Tenant.current_secrets.security[:lockable][:maximum_attempts] || 20).to_i
     
     NoMethodError:
       undefined method `[]' for nil:NilClass
     
           (Tenant.current_secrets.security[:lockable][:maximum_attempts] || 20).to_i
                                           ^^^^^^^^^^^
     # ./app/models/user.rb:420:in `maximum_attempts'
```

taitus: Great that you spotted this!

In the Cabildo a change was made that does not have this PR. I have just added it.

The difference is to use "dig" so that in case of not having the secret defined it does not fail.

```Tenant.current_secrets.dig(:security, :lockable, :maximum_attempts)```

**Review by Senen — APPROVED** [review]
👏🏼

File: config/application.rb [hunk 4]
```diff
@@ -140,6 +140,8 @@ class Application < Rails::Application
 
     # Set to true to enable managing different tenants using the same application
     config.multitenancy = Rails.application.secrets.multitenancy
+
```
Senen, line 143: Should we add a comment to the `config.devise_lockable` like we do for `config.multitenancy`? Should we use alphabetical order in the custom application config options, we are adding?

Senen: cc @taitus

javierm (approved):

PR #5276, opened 2023-10-10
Allow sorting homepage cards

taitus:
## References
#5174 

## Objectives
Allows for the customization of card order. This way, each installation can display them in the desired sequence without being constrained by a fixed order.

### DIFF
--- a/app/components/admin/widget/cards/row_component.html.erb
+++ b/app/components/admin/widget/cards/row_component.html.erb
@@ -9,6 +9,12 @@
     <%= card.link_url %>
   </td>
 
+  <% unless header_section? %>
+    <td>
+      <%= card.order %>
+    </td>
+  <% end %>
+
   <!-- remove conditional once specs have image validations -->
   <td>
     <% if card.image.present? %>
--- a/app/components/admin/widget/cards/row_component.rb
+++ b/app/components/admin/widget/cards/row_component.rb
@@ -5,4 +5,10 @@ class Admin::Widget::Cards::RowComponent < ApplicationComponent
     @card = card
     @options = options
   end
+
+  private
+
+    def header_section?
+      card.header_or_sdg_header?
+    end
 end
--- a/app/models/widget/card.rb
+++ b/app/models/widget/card.rb
@@ -12,12 +12,14 @@ class Widget::Card < ApplicationRecord
   validates :link_url, presence: true, if: -> { !header? || link_text.present? }
   validates :order, numericality: { greater_than_or_equal_to: 1 }
 
+  scope :sort_by_order, -> { order(:order, :created_at) }
+
   def self.header
     where(header: true)
   end
 
   def self.body
-    where(header: false, cardable_id: nil).order(:created_at)
+    where(header: false, cardable_id: nil).sort_by_order
   end
 
   def header_or_sdg_header?
--- a/app/models/widget/card.rb
+++ b/app/models/widget/card.rb
@@ -10,6 +10,7 @@ class Widget::Card < ApplicationRecord
 
   validates_translation :title, presence: true
   validates :link_url, presence: true, if: -> { !header? || link_text.present? }
+  validates :order, numericality: { greater_than_or_equal_to: 1 }
 
   def self.header
     where(header: true)
--- a/app/models/widget/card.rb
+++ b/app/models/widget/card.rb
@@ -18,4 +18,12 @@ class Widget::Card < ApplicationRecord
   def self.body
     where(header: false, cardable_id: nil).order(:created_at)
   end
+
+  def header_or_sdg_header?
+    header? || sdg_header?
+  end
+
+  def sdg_header?
+    cardable == WebSection.find_by!(name: "sdg")
+  end
 end
--- a/app/views/admin/widget/cards/_form.html.erb
+++ b/app/views/admin/widget/cards/_form.html.erb
@@ -30,7 +30,7 @@
   </div>
 
   <div class="row">
-    <% unless card.header? %>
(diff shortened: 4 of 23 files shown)


spec/system/home_spec.rb
```diff
@@ -143,6 +143,34 @@
     expect(page).not_to have_css(".title", text: "Featured")
   end
 
+  describe "Cards are ordered" do
+    scenario "by order field" do
+      create(:widget_card, title: "Card one", link_text: "Link one", order: 2)
+      create(:widget_card, title: "Card two", link_text: "Link two", order: 1)
+      create(:widget_card, title: "Card three", link_text: "Link three", order: 0)
```
javierm, line 150: Since we aren't usingthe `link_text` field, maybe we can simplify the test. Also note the test would pass if cards were displayed in creation order, with the last card appearing first, so maybe we can change the orders to something like:

```suggestion
      create(:widget_card, title: "Card one", order: 3)
      create(:widget_card, title: "Card two", order: 1)
      create(:widget_card, title: "Card three", order: 2)
```

javierm (commented):
I certainly prefer this solution rather than having an inaccessible sortable table with drag and drop like we do in poll question answers :+1:.

I've left a few comments. Let me know what you think!

app/views/admin/widget/cards/_form.html.erb
```diff
@@ -31,11 +31,13 @@
 
   <div class="row">
     <% unless card.header? %>
-      <div class="column">
-        <%= f.label :columns %>
-        <p class="help-text"><%= t("admin.site_customization.pages.cards.columns_help") %></p>
-        <div class="small-12 medium-4 large-2">
-          <%= f.select :columns, (1..12), label: false %>
+      <div class="clear">
```
javierm, line 34: I think this `<div>` tag isn't necessary :thinking:.

config/locales/en/admin.yml
```diff
@@ -1648,6 +1648,7 @@ en:
           create_card: Create card
           no_cards: There are no cards.
           columns_help: "Width of the card in number of columns. On mobile screens it's always a width of 100%."
+          order_help: "You can enter the position where this card will be shown to the user in the homepage"
```
javierm, line 1651: I think we can remove "to the user" :thinking:.

On the other hand, I wonder whether it makes sense to mention "homepage" when sometimes we mean a custom page and sometimes we mean the SDG index :thinking:.

app/views/admin/widget/cards/_form.html.erb
```diff
@@ -31,12 +31,18 @@
...
-        <p class="help-text"><%= t("admin.site_customization.pages.cards.columns_help") %></p>
-        <div class="small-12 medium-4 large-2">
+      <div class="clear">
+        <div class="small-12 medium-6 column">
+          <%= f.label :columns %>
+          <p class="help-text"><%= t("admin.site_customization.pages.cards.columns_help") %></p>
           <%= f.select :columns, (1..12), label: false %>
         </div>
+
+        <div class="small-12 medium-6 column">
```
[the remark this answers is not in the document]
javierm: I forgot we are also using cards in custom pages, so we could add the order to the `PagesController` as well :thinking:.

spec/system/admin/widgets/cards_spec.rb
```diff
@@ -25,6 +26,7 @@
         expect(page).to have_content "Card description"
         expect(page).to have_content "Link text"
         expect(page).to have_content "consul.dev"
+        expect(page).to have_css("td", exact_text: "12")
```
javierm, line 29: Since none of the methods above or below this one use parentheses, do you think it makes sense to remove them here too? :thinking:

Same thing in a few other changes in this pull request where we could `have_css "` instead of `have_css("`.

```suggestion
        expect(page).to have_css "td", exact_text: "12"
```

app/models/widget/card.rb
```diff
@@ -10,12 +10,13 @@ class Widget::Card < ApplicationRecord
...
   validates :link_url, presence: true, if: -> { !header? || link_text.present? }
+  validates :order, numericality: { greater_than_or_equal_to: 1 }
 
   def self.header
     where(header: true)
   end
 
   def self.body
-    where(header: false, cardable_id: nil).order(:created_at)
+    where(header: false, cardable_id: nil).order(:order, :created_at)
```
javierm, line 20: Since we're using `.order(:order, :created_at)` three times, do you think it makes sense to extract this logic to a scope?

```suggestion
    where(header: false, cardable_id: nil).by_order # or by_position, or similar
```

app/components/admin/widget/cards/row_component.rb
```diff
@@ -1,8 +1,10 @@
 class Admin::Widget::Cards::RowComponent < ApplicationComponent
-  attr_reader :card, :options
+  attr_reader :card, :header_section, :options
+  alias_method :header_section?, :header_section
 
-  def initialize(card, **options)
+  def initialize(card, header_section, **options)
```
javierm, line 5: We're using an anonymous parameter here, but in the table component we're using a named parameter :thinking:. Is it on purpose? :thinking:

app/components/admin/widget/cards/row_component.html.erb
```diff
@@ -9,6 +9,12 @@
     <%= card.link_url %>
   </td>
 
+  <% unless header_section? %>
```
javierm, line 12: We could use the `header?(card)` helper method here, right? We could also move that method to the model :thinking:.

javierm (commented):
@taitus I've left some comments. Let me know what you think!

javierm (approved):

PR #5303, opened 2023-10-24
Map node binaries whenever `rvm1:hook` is called and evaluate the `fnm_setup_command` dynamically

Senen:
# References

* #5158 

The previous version worked fine when using Capistrano to deploy a new release, but it did not work for other commands, like `cap npm install` or `cap puma:start`, as they do not create a release. Therefore, the `git:create_release` was never called, and the `map_node_bins` was not invoked.

## Objectives

Since we need node binaries available to the deploy user when executing any process that runs the application, like puma, delayed_job, and rake tasks, among others, it makes sense to load FNM to use the correct Node version in the same places where Capistrano loads RVM to load the correct Ruby version.

With dynamic loading of the `fnm_setup_command`, we get the correct path when deploying a new release and also when running any other capistrano command that does not deploy a new release.

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -60,7 +60,7 @@ namespace :deploy do
   Rake::Task["delayed_job:default"].clear_actions
   Rake::Task["puma:smart_restart"].clear_actions
 
-  after "git:create_release", "map_node_bins"
+  after "rvm1:hook", "map_node_bins"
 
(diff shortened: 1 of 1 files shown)


javierm (approved):
Superb catch! :clap: So great we've detected this case before releasing a new version :smile:.

PR #5320, opened 2023-11-17
Filter sensitive information from airbrake parameters

Senen:
## References
We detected sensitive information was sent and registered as plain text through the airbrake integration.

## Objectives

Filter sensitive data when using airbrake to track exceptions. 
Use the same rules we use in Consul Democracy for logging.

We needed to rename the initializer file name `filter_parameter_logging.rb` so it's loaded before the airbrake initializer.


javierm (commented):
@Senen Nice catch! :tada:

I've left a comment; let me know what you think!

javierm (approved):

Issue #5160, opened 2023-07-24
InvalidAuthenticityToken when voting a comment


javierm:
## Steps to reproduce

1. Without signing in, go to the main Participatory Budgeting page
2. Click on a heading
3. Click on an investment which has at least one comment
4. Click on the "I agree" button to agree with the comment

## Expected result

You're redirected to the sign in page

## Actual result

You get a 422 ActionController::InvalidAuthenticityToken error.

## Notes

* The same issue might be happening in other places, since it seems to be related to Turbolinks and the CSRF token
* Pull request #5118 deals with votes, so we need to check whether this behavior persists with the changes in that pull request

Thanks @decabeza and @microweb10 for reporting! :wink:

microweb10:
No problem! 😉 
We have temporarily applied a patch by using the method `$.rails.refreshCSRFTokens()`
In case anyone is interested, the commit can be found [here](https://github.com/democrateam/consul/pull/137/commits/19af86728865987053fe27f5f5b840fbcf8962b3).

javierm:
@microweb10 @decabeza I'm not able to reproduce this issue on the demo anymore :thinking:. Could you check whether you're able to reproduce it? Maybe I'm doing something wrong...

microweb10:
I was just able to reproduce it. 😅 

![Captura de Pantalla 2023-08-08 a las 11 35 03](https://github.com/consuldemocracy/consuldemocracy/assets/942995/66d2a2c2-c586-4643-a529-936b514727ad)

javierm:
Today I can reproduce it as well :sweat_smile:.

Senen:
>Pull request https://github.com/consuldemocracy/consuldemocracy/pull/5118 deals with votes, so we need to check whether this behavior persists with the changes in that pull request

Hi @javierm, the error still happens with the #5118 changes. I've just opened a PR to include the Democrateam fix.

PR #5329, opened 2023-11-30
Refresh CSRF Token when using Turbolinks

Senen:
## References

This PR fixes:
* #5160

## Objectives
Refresh the authenticity token when browsing the application with Turbolinks.

## Notes
We cannot reproduce the error in the test or development environment. So, we cannot write a test that demonstrates it. We verified manually that this commit fixes the problem described in #5160 when running the application in staging or production environments.


javierm (approved):
Awesome! :tada:

PR #5330, opened 2023-12-01
Make it easier to find permanent links to documents

### LINKED ISSUE #5112: Use URL blobs for documents
### User Story

Administrators can upload documents from [Admin > Site content > Documents](https://demo.consulproject.org/admin/site_customization/documents).

Usually they want to link to it within the web site, so they open the document and copy the URL.

The problem is that `ActiveDirectory` creates different links for different computers so when a different user tries to open the file it doesn't work.

### Description

When the document is opened, the URL generated is:
`https://demo.consulproject.org/rails/active_storage/disk/.../dummy.pdf`

The solution is for the administrator to use _**right click > copy link**_ to get the absolute URL: `https://demo.consulproject.org/rails/active_storage/blobs/../dummy.pdf` but this is not very intuitive. 😅 

### Definition of Done
Maybe it's a good idea to replace the document URL link to File name and add the "view document" and "copy link" actions.

### Before
before

### After
after

Senen:
## References

This PR tries to address the issues described in:
*  #5112

## Objectives

1. Use the Rails recommended way to render document download links
2. Provide an obvious place from where to copy the permanent link to documents so administrators can paste it anywhere they want.

## Visual Changes

**Before**

Captura de pantalla 2023-12-01 a las 13 20 00

**After**

Link "Show link" added. It opens an alert box 
Captura de pantalla 2023-12-01 a las 13 18 55

Modal dialog with the permanent URL
Captura de pantalla 2024-01-23 a las 13 59 49

Alert box with the permanent URL (Previous implementation)

Captura de pantalla 2023-12-01 a las 13 19 06

### DIFF
--- a/app/components/admin/poll/questions/answers/documents/table_actions_component.html.erb
+++ b/app/components/admin/poll/questions/answers/documents/table_actions_component.html.erb
@@ -2,6 +2,6 @@
                                                    destroy_path: document_path(document)) do |actions| %>
   <%= actions.action(:download,
                      text: t("documents.buttons.download_document"),
-                     path: document.attachment,
+                     path: rails_blob_path(document.attachment, disposition: "attachment"),
                      rel: "nofollow") %>
 <% end %>
--- a/app/components/admin/site_customization/documents/copy_link_dialog_component.html.erb
+++ b/app/components/admin/site_customization/documents/copy_link_dialog_component.html.erb
@@ -1,0 +1,8 @@
+<section id="<%= dialog_id %>" class="small reveal" data-reveal aria-labelledby="<%= dialog_label_id %>">
+  <h2 id="<%= dialog_label_id %>"><%= t("admin.site_customization.documents.copy_link_dialog.title") %></h2>
+  <button class="close-button" aria-label="<%= t("admin.shared.close_modal") %>" type="button" data-close>
+    <span aria-hidden="true">&times;</span>
+  </button>
+  <p><%= t("admin.site_customization.documents.copy_link_dialog.description") %></p>
+  <code><%= polymorphic_url(document.attachment) %></code>
+</section>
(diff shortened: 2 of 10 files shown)


javierm:
@Senen I see I started a review a month ago but didn't finish it; the only comment so far is: "In some browsers (for example, Chromium 118) it doesn't seem to be possible to select the text displayed inside an alert, so it wouldn't be possible to effectively copy the link". I'd rather wait until we release version 2.1, but up to you! :wink:

Senen:
> @Senen I see I started a review a month ago but didn't finish it; the only comment so far is: "In some browsers (for example, Chromium 118) it doesn't seem to be possible to select the text displayed inside an alert, so it wouldn't be possible to effectively copy the link". I'd rather wait until we release version 2.1, but up to you! 😉

Hi @javierm, I did not know about the text selection issue on some browsers.  We can leave it for the next release. What do you think about using a modal dialog instead of an alert?

javierm:
@Senen Haven't looked into it enough to decide. I think I prefer the "copy link" idea that Alberto mentioned, but this is just my initial impression so apologies in advance if I change my mind :pray:.

Senen:
> @Senen Haven't looked into it enough to have a clear preference. I think I prefer the "copy link" idea that Alberto mentioned, but this is just my initial impression so apologies in advance if I change my mind 🙏.

No problem, I have no hurry 😌 . I leave here some thoughts of mine about the solutions I found for this issue:

1. Install a third-party library to allow copy content to the clipboard by clicking the link. I discarded this solution to avoid installing more dependencies and its potential browser compatibility problems.
2. Use a copy link, as Alberto suggests. The problem is that administrators following the link and copying from the URL would have the same error Alberto describes. We should clarify in the interface that the user may not follow the link but use the link context menu and choose the "Copy link" option. This solution seems tricky to me and error-prone.
3. Use ~an alert~ a modal dialog to avoid following the link and showing the original link to the user before its changed by the Active Storage Redirects controller.


Maybe I'm missing some other better solution.

Senen:
Closing as in the latest version this does not happen anymore.

PR #5369, opened 2024-01-30
Make comments specs faster

### LINKED ISSUE #2637: Make comments tests faster
## Context

Many models can be commented (debates, proposals, investments, polls).

Comments for all these models work pretty much the same way, however we have a large duplication of these specs and these tests take a long time to run.

## What

Create a single `spec/system/comments_spec.rb` file and add tests for every scenario involving comments while also covering every model that accepts comments.

A similar task was done in pull request #3713.

taitus:
## References

* Closes #2637 

## Objectives

* Remove duplicated specs for comment section
* Make our test suite about 7% faster


spec/system/comments_spec.rb
```diff
@@ -285,7 +370,7 @@
         expect(page).not_to have_content "Wrong comment"
       end
 
-      visit debate_path(debate)
+      visit polymorphic_path(resource)
```
javierm, line 373:
```suggestion
      refresh
```

spec/system/comments_spec.rb
```diff
@@ -317,20 +402,20 @@
   end
 
   scenario "Reply to reply" do
-    create(:comment, commentable: debate, body: "Any estimates?")
+    create(:comment, commentable: resource, body: "Any estimates?")
 
-    login_as(create(:user))
-    visit debate_path(debate)
+    login_as(create(:user, :level_two))
```
javierm, line 407: Would it be possible to add the reason why we add `:level_two` to the commit message?

spec/system/comments_spec.rb
```diff
@@ -398,52 +476,60 @@
...
+    visit polymorphic_path(resource)
 
-    fill_in "Leave your comment", with: "Testing submit button!"
-    click_button "Publish comment"
+    fill_in fill_text, with: "Testing submit button!"
+    click_button button_text
 
     expect(page).to have_button "Publish comment", disabled: true
     expect(page).to have_content "Testing submit button!"
     expect(page).to have_button "Publish comment", disabled: false
```
javierm, line 520:
```suggestion
    expect(page).to have_button button_text, disabled: true
    expect(page).to have_content "Testing submit button!"
    expect(page).to have_button button_text, disabled: false
```

spec/system/comments/budget_investments_spec.rb
```diff
@@ -1,415 +1,12 @@
 require "rails_helper"
 
 describe "Commenting Budget::Investments" do
-  let(:user) { create(:user) }
   let(:investment) { create(:budget_investment) }
 
   it_behaves_like "flaggable", :budget_investment_comment
```
javierm, line 6: Is it OK to remove this line? :thinking:

spec/system/comments_spec.rb
```diff
@@ -480,64 +564,64 @@
...
-      click_button "Publish comment"
-
-      within "#comments" do
-        expect(page).to have_content "I am your Admin!"
-        expect(page).to have_content "Administrator ##{admin.id}"
-        expect(page).to have_css "div.is-admin"
-        expect(page).to have_css "img.admin-avatar"
+    context "comment as administrator" do
+      scenario "can create comment" do
+        admin = create(:administrator, description: "valuator user")
```
javierm, line 577: When first reading it, I thought we would be in the valuation section :thinking:. How about:

```suggestion
        admin = create(:administrator, description: "admin user")
```

javierm (commented):
This is amazing! :tada:

I've left some comments. Let me know what you think!

spec/system/comments_spec.rb
```diff
@@ -295,17 +380,17 @@
...
-    comment = create(:comment, commentable: debate, user: citizen)
+    comment = create(:comment, commentable: resource)
 
-    login_as(manuela)
-    visit debate_path(debate)
+    login_as(user)
+    visit polymorphic_path(resource)
 
-    click_link "Reply"
+    within "#comment_#{comment.id}" do
```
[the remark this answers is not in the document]
taitus: This is due to the case of `legislation_annotation` before in the original test no comment was created as it simply took the one created by default when creating a `legislation_annotation`.

```    
    annotation = create(:legislation_annotation, author: citizen)
    comment = annotation.comments.first
``` 

Now to try to unify this test, we always create a `comment`, and in this case as we also created the `legislation_annotation` we have 2 comments, so it is necessary to add the `click_link` inside the `within`.

I'm going to add an explanation to the commit message.

spec/system/comments_spec.rb
```diff
@@ -342,54 +427,47 @@
...
   scenario "Errors on reply" do
-    comment = create(:comment, commentable: debate, user: user)
+    comment = create(:comment, commentable: resource, user: user)
 
     login_as(user)
-    visit debate_path(debate)
+    visit polymorphic_path(resource)
 
-    click_link "Reply"
+    within "#comment_#{comment.id}" do
```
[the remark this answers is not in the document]
taitus: Same as previous comment

spec/system/comments_spec.rb
```diff
@@ -454,19 +540,17 @@
...
-      moderator = create(:moderator, user: manuela)
-      comment = create(:comment, commentable: debate, user: citizen)
+      comment = create(:comment, commentable: resource)
 
-      login_as(manuela)
-      visit debate_path(debate)
+      visit polymorphic_path(resource)
 
-      click_link "Reply"
+      within "#comment_#{comment.id}" do
```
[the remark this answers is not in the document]
taitus: Same as previous comment

spec/system/comments_spec.rb
```diff
@@ -480,64 +564,64 @@
...
-      fill_in "Leave your comment", with: "I am your Admin!"
-      check "comment-as-administrator-debate_#{debate.id}"
-      click_button "Publish comment"
-
-      within "#comments" do
-        expect(page).to have_content "I am your Admin!"
-        expect(page).to have_content "Administrator ##{admin.id}"
-        expect(page).to have_css "div.is-admin"
-        expect(page).to have_css "img.admin-avatar"
+    context "comment as administrator" do
```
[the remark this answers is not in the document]
taitus: The reason was that this context existed in the original file (`spec/system/comments/budget_investments_spec.rb`). 

But seeing that it doesn't make much sense, I've just removed it by using the commit: **Simplify administrators comments tests**

spec/system/comments_spec.rb
```diff
@@ -480,64 +564,64 @@
...
+        admin   = create(:administrator)
+        comment = create(:comment, commentable: resource)
 
-      login_as(manuela)
-      visit debate_path(debate)
+        login_as(admin.user)
+        visit polymorphic_path(resource)
 
-      click_link "Reply"
+        within "#comment_#{comment.id}" do
```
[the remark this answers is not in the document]
taitus: Same as previous comments

spec/system/comments_spec.rb
taitus: We are solved this question adding:

```
  let(:user) do
    if factory == :legislation_question
      create(:user, :level_two)
    else
      create(:user)
    end
  end
  ```
  
  and use `login_as(user)` in the test. This behaviour was to allow skip `require_verified_resident_for_commentable?` condition in comments/form_component.html.erb.

File: app/components/comments/form_component.rb [hunk 11]
```diff
@@ -0,0 +1,32 @@
...
+    def cache_key
+      [
+        locale_and_user_status,
+        parent_id,
+        commentable_cache_key(commentable),
+        valuation
+      ]
+    end
+
+    def comments_closed_text(commentable)
```
javierm, line 25:
```suggestion
    def comments_closed_text
```

Since this isn't a helper method anymore, we don't need to pass the commentable.

Note that we'll also have to change the place where this method is called.

javierm (commented):
Amazing!!!

(That's three exclamation marks, for no reason :rofl:).

I've left a minor comment. Let me know what you think.

javierm (approved):

PR #5370, opened 2024-01-30
Make legislation summary configurable

coslajohn:
## User Story
The Legislation Summary contains useful information for Administrators.  One of the biggest complaints I receive from my users is about being able to view/download/summarise comments as there is often a statutory obligation on the city to review and publish comments in other places.

When carrying out a legislation process there is often a long time period between the phases which means at the moment that the Admin doesn't get to see the summary until the process is complete which can be many months after a phase is carried out.

Because the information is useful to Admins, there is no reason why an Admin shouldn't see it all of the time, therefore I propose to make the Summary visible to Admins all the time in the same way as all of the other phases (debate, proposals etc) are visible all of the time.

It is also possible that this information may be useful to citizens as part of the legislation process as it provides an overview of the previous phases, so can give a sense of what has already happened without having to read all of the previous comments/proposals etc

I am suggesting that this should be an option for the Admin to switch on/off/publish as this makes it consistent with the way other phases behave in Legislation

## Objectives

The Legislation Summary is a useful overview for admins. This change makes the Legislation Summary tab visible to admins at all times.
At the moment the summary is only published when the process is finished.
A configuration item is added to allow the admin to publish the summary on a chosen date

This pull request also tidies up the layout of the Legislation configuration page by enclosing the configuration items in a row and adding labels and descriptions to some options

## original layout
image
## revised layout
image

## new alignment and options
image
 
## Summary displayed in advance of final publication
image

## Notes

This adds a db migration to add fields for summary_publication_date and summary_publication_enabled

### DIFF
--- a/app/controllers/admin/legislation/processes_controller.rb
+++ b/app/controllers/admin/legislation/processes_controller.rb
@@ -72,6 +72,8 @@ class Admin::Legislation::ProcessesController < Admin::Legislation::BaseControll
         :background_color,
         :font_color,
         :related_sdg_list,
+        :summary_publication_date,
+        :summary_publication_enabled,
         translation_params(::Legislation::Process),
         documents_attributes: document_attributes,
         image_attributes: image_attributes
--- a/app/models/abilities/administrator.rb
+++ b/app/models/abilities/administrator.rb
@@ -18,7 +18,8 @@ module Abilities
       can :create, Legislation::Proposal
       can :show, Legislation::Proposal
       can :proposals, ::Legislation::Process
-
+      can :summary, ::Legislation::Process
+      
       can :restore, Legislation::Proposal
       cannot :restore, Legislation::Proposal, hidden_at: nil
 
--- a/app/models/abilities/everyone.rb
+++ b/app/models/abilities/everyone.rb
@@ -20,8 +20,13 @@ module Abilities
       can :read_executions, Budget, phase: "finished"
       can [:read, :debate, :draft_publication, :allegations, :result_publication,
            :proposals, :milestones], Legislation::Process, published: true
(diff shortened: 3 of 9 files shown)


javierm:
Hi, @coslajohn :smile:.

Could you add some background information regarding why it's useful to publish the summary on a different date?

coslajohn:
> Hi, @coslajohn 😄.
> 
> Could you add some background information regarding why it's useful to publish the summary on a different date?

@javierm  I have updated the PR with more information on the background for this change

javierm:
@coslajohn It looks like this pull request was opened right when GitHub Actions was making dozens of tests fail. Would it be possible to update it against the latest master branch, so the tests run again? :pray:

```
git rebase master
git push -f <might_need_to_indicate_your_repo>
```

I've run some tests on my machine and it looks like some fail because we're changing the current behaviour. Most of the failures are related to the same question: what should we do about existing installations where the summary can be accessed by everyone right now? Should we add a migration task for existing processes so the `summary_publication_enabled` fielt is set depending on whether it could be accessed using the logic we had until now? :thinking:

This leads to the question: would it make sense to make the legislation summary accessible if *either* the conditions with current logic or the new one are fulfilled? That is, the summary would be accessible if the process is finished and result publication is enabled *or* if the process is not finished but the summary publication is enabled. Of course, this could also be confusing for administrators :thinking:.

Sorry to think out loud here. Any thoughts about all this? Might need to ask the Foundation as well :thinking:.

lucialuzuriaga:
Hi @coslajohn! Since this PR is in a draft state has been open for a long time without updates, I am going to close it. If you or anyone else would like to pick this back up in the future, please feel free to comment or open a new PR. Thanks!

PR #5390, opened 2024-02-09
Geozone Admin maps

coslajohn:
## Objectives

This pull request renders a map and geojson geozones on the Geozones admin page making it easier to see if Geojson has been imported correctly

## Visual Changes

The map is appended to the bottom of the Geozone admin page.  Clicking on the location link opens the geozone edit page

image

### DIFF
--- a/app/components/admin/geozones/index_component.html.erb
+++ b/app/components/admin/geozones/index_component.html.erb
@@ -29,5 +29,8 @@
     <% end %>
   </tbody>
 </table>
-
+<div class="small-12 column">
+  <h2><%= t("admin.geozones.index.geojson_map") %></h2>
+  <p class="help-text">  <%= t("admin.geozones.index.geojson_map_help") %></p>
+</div>
 <%= render Shared::MapLocationComponent.new(nil, geozones_data: geozones_data) %>
--- a/app/components/admin/geozones/index_component.html.erb
+++ b/app/components/admin/geozones/index_component.html.erb
@@ -29,3 +29,6 @@
     <% end %>
   </tbody>
 </table>
+<div class="budgets-map">
+  <%= render Shared::MapLocationComponent.new(nil, investments_coordinates: nil, geozones_data: nil) %>
+</div>
--- a/app/components/admin/geozones/index_component.html.erb
+++ b/app/components/admin/geozones/index_component.html.erb
@@ -29,8 +29,10 @@
     <% end %>
   </tbody>
 </table>
-<div class="small-12 column">
-  <h2><%= t("admin.geozones.index.geojson_map") %></h2>
-  <p class="help-text">  <%= t("admin.geozones.index.geojson_map_help") %></p>
-</div>
-<%= render Shared::MapLocationComponent.new(nil, geozones_data: geozones_data) %>
+
+<section>
+  <h3><%= t("admin.geozones.index.geojson_map") %></h3>
+  <p class="help-text"><%= t("admin.geozones.index.geojson_map_help") %></p>
+
+  <%= render Shared::MapLocationComponent.new(nil, geozones_data: geozones_data) %>
+</section>
--- a/app/components/admin/geozones/index_component.html.erb
+++ b/app/components/admin/geozones/index_component.html.erb
@@ -30,5 +30,5 @@
   </tbody>
 </table>
 <div class="budgets-map">
-  <%= render Shared::MapLocationComponent.new(nil, investments_coordinates: nil, geozones_data: nil) %>
+  <%= render Shared::MapLocationComponent.new(nil, investments_coordinates: nil, geozones_data: geozones_data) %>
 </div>
--- a/app/components/admin/geozones/index_component.html.erb
+++ b/app/components/admin/geozones/index_component.html.erb
@@ -29,6 +29,5 @@
     <% end %>
   </tbody>
 </table>
-<div class="budgets-map">
-  <%= render Shared::MapLocationComponent.new(nil, investments_coordinates: nil, geozones_data: geozones_data) %>
-</div>
+
+<%= render Shared::MapLocationComponent.new(nil, geozones_data: geozones_data) %>
--- a/app/components/admin/geozones/index_component.rb
+++ b/app/components/admin/geozones/index_component.rb
@@ -22,8 +22,12 @@ class Admin::Geozones::IndexComponent < ApplicationComponent
     def geozones_data
        geozones.map do |geozone|
         {
+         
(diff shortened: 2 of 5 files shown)


javierm:
@coslajohn Thank you very much; this is very useful.

Here's a patch that simplifies the code a bit. Let me know what you think!

```diff
diff --git a/app/components/admin/geozones/index_component.html.erb b/app/components/admin/geozones/index_component.html.erb
index 953873ffc0..5c4062879a 100644
--- a/app/components/admin/geozones/index_component.html.erb
+++ b/app/components/admin/geozones/index_component.html.erb
@@ -29,6 +29,5 @@
     <% end %>
   </tbody>
 </table>
-<div class="budgets-map">
-  <%= render Shared::MapLocationComponent.new(nil, investments_coordinates: nil, geozones_data: geozones_data) %>
-</div>
\ No newline at end of file
+
+<%= render Shared::MapLocationComponent.new(nil, geozones_data: geozones_data) %>
diff --git a/app/components/admin/geozones/index_component.rb b/app/components/admin/geozones/index_component.rb
index d6d9ee6ec9..bdade404e8 100644
--- a/app/components/admin/geozones/index_component.rb
+++ b/app/components/admin/geozones/index_component.rb
@@ -19,17 +19,14 @@ class Admin::Geozones::IndexComponent < ApplicationComponent
         t("shared.no")
       end
     end
+
     def geozones_data
-       geozones.map do |geozone|
+      geozones.map do |geozone|
         {
-         
           outline_points: geozone.outline_points,
           color: geozone.color,
-          headings: geozones.where(id: geozone).map do | geozone|
-             link_to(geozone.name, edit_admin_geozone_path(geozone.id))
-          end
+          headings: [link_to(geozone.name, edit_admin_geozone_path(geozone))]
         }
-       end
+      end
     end
-
 end
diff --git a/spec/system/admin/geozones_spec.rb b/spec/system/admin/geozones_spec.rb
index 6f24c36666..95c862bafc 100644
--- a/spec/system/admin/geozones_spec.rb
+++ b/spec/system/admin/geozones_spec.rb
@@ -143,31 +143,17 @@ describe "Admin geozones", :admin do
     end
   end
 
-
   scenario "Show polygons on geozone admin view" do
     Setting["feature.map"] = true
-
     geojson = '{ "geometry": { "type": "Polygon", "coordinates": [[-0.1,51.5],[-0.2,51.4],[-0.3,51.6]] } }'
-    geozone = create(:geozone, name: "Polygon me!")
-
-    visit edit_admin_geozone_path(geozone)
-    fill_in "GeoJSON data (optional)", with: geojson
-    fill_in "Color (optional)", with: "#f5c211"
-    click_button "Save changes"
-
-    expect(page).to have_content "Geozone updated successfully"
+    geozone = create(:geozone, name: "Polygon me!", geojson: geojson)
 
     visit admin_geozones_path
 
-    expect(page).to have_css ".map-polygon[fill='#f5c211']"
-    within(".map-location") { expect(page).not_to have_link "Area 51" }
-
     find(".map-polygon").click
 
     within ".map-location" do
       expect(page).to have_link "Polygon me!", href: edit_admin_geozone_path(geozone)
     end
   end
-
-
 end
```

app/components/admin/geozones/index_component.html.erb
```diff
@@ -29,3 +29,6 @@
     <% end %>
   </tbody>
 </table>
+<div class="budgets-map">
+  <%= render Shared::MapLocationComponent.new(nil, investments_coordinates: nil, geozones_data: geozones_data) %>
```
javierm, line 33: Do you think it makes sense to only render the map when at least one of the geozones contains GeoJSON data? :thinking:

I'm wondering because there might be people who don't use polygons, and they might see an empty map and wonder what's going on and might expect the "HTML coordinates" to be displayed there :thinking:.

coslajohn: @javierm what I have done for now is add a heading with a help text which explains the purpose of the map.  I think it is possibly useful to have the map because if the mapping is set up correctly in Settings, it should be a visual representation of the area the geozones relate to.

Something to think about for a future enhancement would be whether or not it is also useful to display the html image map on this page.

If the HTML map was also available on the Geozones page then maybe it would make sense to have an action button which toggled the display of each type of map on and off

There is possibly also a need for a setting which allows the admin to decide whether to display GeoJSON maps, HTML maps (or unlikely both). That way the Geojson maps could be dropped into the sidebar of Proposals etc where the HTML image map is currently displayed  or on /proposals/map

javierm:
@coslajohn Could you add the Spanish translations as well? Here's a patch.

```diff
diff --git a/app/components/admin/geozones/index_component.html.erb b/app/components/admin/geozones/index_component.html.erb
index a81b157724..c423a6f87a 100644
--- a/app/components/admin/geozones/index_component.html.erb
+++ b/app/components/admin/geozones/index_component.html.erb
@@ -29,8 +29,10 @@
     <% end %>
   </tbody>
 </table>
-<div class="small-12 column">
-  <h2><%= t("admin.geozones.index.geojson_map") %></h2>
-  <p class="help-text">  <%= t("admin.geozones.index.geojson_map_help") %></p>
-</div>
-<%= render Shared::MapLocationComponent.new(nil, geozones_data: geozones_data) %>
+
+<section>
+  <h3><%= t("admin.geozones.index.geojson_map") %></h3>
+  <p class="help-text"><%= t("admin.geozones.index.geojson_map_help") %></p>
+
+  <%= render Shared::MapLocationComponent.new(nil, geozones_data: geozones_data) %>
+</section>
diff --git a/config/locales/es/admin.yml b/config/locales/es/admin.yml
index 4bdd562f5e..ecde6bcaf1 100644
--- a/config/locales/es/admin.yml
+++ b/config/locales/es/admin.yml
@@ -1400,6 +1400,8 @@ es:
       index:
         title: Zonas
         create: Crear una zona
+        geojson_map: Mapa GeoJSON
+        geojson_map_help: "Si las coordenadas GeoJSON están disponibles, aparecerán en el siguiente mapa"
       geozone:
         name: Nombre
         external_code: Código externo
```

Other than adding these texts, this patch removes the `small-12 column` classes so we don't get extra padding and replaces the `<h2>` tag with `<h3>` since, in the admin section, the page title is displayed using `<h2>` tags instead of `<h1>` tags :sweat_smile: (we might change it in the future).

Let me know what you think! :wink:

coslajohn:
@javierm that is the patch applied. All good

javierm (commented):
@coslajohn Thank you very much! :pray:

PR #5439, opened 2024-03-22
Simplify signing in to our demo

javierm:
## Objectives

* Make it easier to sign in in our demo, without writing the username and password and even if missing our text on top of the page where the username and password are displayed
* Celebrate that we've just released a new version with a change in our demo :tada:

## Notes for reviewers

Another option would be to only fill in the email and autofocusing on the password field :thinking:. The advantage here would be that nobody would get scared thinking that Consul Democracy installations by default fill in the admin password for anyone :sweat_smile:.

## Release Notes

:warning: Do not include this pull request in the CHANGELOG when publishing a new release. This text is **not** meant to be included in the release notes :joy:, but a warning that the features related to the demo aren't included there.


javierm:
Closing for now, since it might not be clear that the demo behaves differently than a real installation due to being a demo :thinking:.

PR #5442, opened 2024-03-22
Update code of conduct contact address

javierm:
## Objectives

* Use the current address where people can report abuse


javierm:
@taitus Merging it without approval because the current address is already featuring on github discussions. Please review it anyway! :pray:

Issue #1676, opened 2017-06-21
Legislation SCSS refactor


decabeza:
## Background

Recently we added to consul the [sccs-lint gem](https://github.com/brigade/scss-lint) to improve the SCSS code.

Now if you run on console `scss-lint` only appears warnings on `legislation_process.scss` file.

Use `scss-lint --format=Stats` command to show a summary.

```
27  SelectorDepth        (across 1 files)
22  NestingDepth         (across 1 files)
--  -------------        ----------------
49  total                (across 1 files)
```

<details>
  <summary>Click to view all warnings</summary>
<pre>
app/assets/stylesheets/legislation_process.scss:212:9 [W] NestingDepth: Nesting should be no greater than 4, but was 5
app/assets/stylesheets/legislation_process.scss:217:9 [W] NestingDepth: Nesting should be no greater than 4, but was 5
app/assets/stylesheets/legislation_process.scss:331:9 [W] NestingDepth: Nesting should be no greater than 4, but was 5
app/assets/stylesheets/legislation_process.scss:335:9 [W] NestingDepth: Nesting should be no greater than 4, but was 5
app/assets/stylesheets/legislation_process.scss:537:9 [W] NestingDepth: Nesting should be no greater than 4, but was 5
app/assets/stylesheets/legislation_process.scss:586:11 [W] NestingDepth: Nesting should be no greater than 4, but was 5
app/assets/stylesheets/legislation_process.scss:636:9 [W] NestingDepth: Nesting should be no greater than 4, but was 5
app/assets/stylesheets/legislation_process.scss:640:9 [W] NestingDepth: Nesting should be no greater than 4, but was 5
app/assets/stylesheets/legislation_process.scss:650:9 [W] NestingDepth: Nesting should be no greater than 4, but was 5
app/assets/stylesheets/legislation_process.scss:696:9 [W] NestingDepth: Nesting should be no greater than 4, but was 5
app/assets/stylesheets/legislation_process.scss:760:11 [W] NestingDepth: Nesting should be no greater than 4, but was 5
app/assets/stylesheets/legislation_process.scss:778:9 [W] NestingDepth: Nesting should be no greater than 4, but was 5
app/assets/stylesheets/legislation_process.scss:811:9 [W] NestingDepth: Nesting should be no greater than 4, but was 5
app/assets/stylesheets/legislation_process.scss:816:9 [W] NestingDepth: Nesting should be no greater than 4, but was 5
[...]

PR #5448, opened 2024-03-26
Replace SCSS Lint with Stylelint

javierm:
## References

* [SCSS Lint recommeds using Stylelint instead since 2016](https://github.com/sds/scss-lint/commit/4ac76f957)
* This change will make it easier to move to Dart Sass, necessary for pull request #5445

## Objectives

* Make it easier to move to Dart Sass
* Make it easier to add more styling rules for CSS/SCSS files in the future
* Remove a broken rule in the styles for the "Skip to main content" link

## Notes

Check the commit message for an explanation regarding which Stylelint rules are equivalent to the SCSS Lint rules we were using.

Most of the syntax problems reported by Stylelint are the same that SCSS Lint used to report; see issue #1676.


javierm:
@decabeza Hope you like this one! :smile:

(Because, if you don't, I'm afraid there isn't much we can do about it :sweat_smile:).

taitus (approved):

PR #5454, opened 2024-03-28
Change the order in which PB Ballot votes are displayed 

coslajohn:
This pull request explicitly sets a display order for the votes selected by a Voter. The display order is set as the order in which they were selected

The reason for this change is to provide a first step into providing voting types which use ranking. In this scenario the order in which votes are displayed is considered to be the priority selected by the voter. The voter can change priority by selecting proposals in a different order

It may be possible to extend this further in future by attaching a rank position to a vote and/or displaying the rank position on the vote

In the screenshots, the voter selected the proposals in order of price and they are now correctly displayed in that order

image
image

### DIFF
--- a/app/components/budgets/ballot/ballot_component.html.erb
+++ b/app/components/budgets/ballot/ballot_component.html.erb
@@ -36,9 +36,7 @@
         <% end %>
 
         <ul class="ballot-list">
-          <%= render Budgets::Ballot::InvestmentComponent.with_collection(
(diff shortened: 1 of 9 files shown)


javierm:
@coslajohn This is great, thanks! :pray:

I've slightly refactored the code and added some tests. Those changes are pending a review by @taitus.

taitus (approved):

javierm:
@coslajohn Once again, thank you very much! :tada:

PR #5455, opened 2024-03-28
Change the English translation of check ballot to avoid confusion

### LINKED ISSUE #5322: Behaviour of Submit Ballot in PB Voting
### User Story

### Why is this Needed?

The behaviour of the the Voting is confusing for users. The Submit Ballot implies that a vote has not been made until Submit Balot is pressed.

### Current Behavior
When you Vote it takes you to a screen which suggests that you need to Submit Ballot. This is not clear to the user whether they have voted or not. It has caused a lot of confusion for users

When I log in (and I don't know if this is because of caching -  there doesn't appear to be any Vote Now button or link to voting?
image

 I need to click into a project and the click back before the correct view appears
image

### Expected Behavior
When you press the Vote button a message is flashed saying you vote has been submitted and the user can then either Vote by selecting another proposal or be taken to the Voting summary

It might be best if the voting summary is displayed permanently when viewing a budget which you are allowed to vote in.

This view appears as expected AFTER you vote

image

A secondary request my users have asked if it is possible to have the Vote button in list view so that you can vote for multiple proposals without having to navigate in and out of proposals.

coslajohn:
This fixes #5322 

The current english translation of check_ballot is Submit my ballot. This is incorrect and confusing for users
The Spanish translation looks to be correct

Before
image

After
image

### DIFF
--- a/config/locales/en/budgets.yml
+++ b/config/locales/en/budgets.yml
@@ -106,7 +106,7 @@ en:
           change_vote_link: "change your vote"
           different_heading_assigned: "You have active votes in another heading: %{heading_link}"
           change_ballot: "If you change your mind you can remove your votes in %{check_ballot} and start again."
-          check_ballot_link: "submit my ballot"
+          check_ballot_link: "check my votes"
           zero: You have not voted any investment project in this group.
           verified_only: "To create a new budget investment %{verify}."
           create: "Create a budget investment"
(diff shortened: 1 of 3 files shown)


javierm (approved):
@coslajohn Thank you very much! :pray:

I've updated the pull request because we didn't only have the same text three times in the YAML file :sweat_smile:, but we also had it a few times in tests checking this link is present.

(We've thought about removing this kind of duplication from the tests in the past, but we think that the resulting complexity in the tests wouldn't make it worth it.)

By the way, since we aren't native English speakers and basically choose which preposition to use by rolling the dice, contributions fixing typos or changing texts that sound strange are more than welcome :smile:.

PR #5479, opened 2024-04-05
Remove legacy code from Puma config.

### LINKED ISSUE #5478: Legacy code in puma config
According to this https://stackoverflow.com/questions/62458471/is-establish-connection-on-worker-boot-still-required-on-rails-6-and-puma

this snippet is no longer required in config/puma/defaults.rb

```
before_fork do
  ActiveRecord::Base.connection_pool.disconnect!
end

on_worker_boot do
  ActiveSupport.on_load(:active_record) do
    ActiveRecord::Base.establish_connection
  end
end

```
These are no longer included in the puma config generated by rails new ( as they were in v5.1)

janimo:
Fixes #5478.

### DIFF
--- a/config/puma/defaults.rb
+++ b/config/puma/defaults.rb
@@ -23,13 +23,3 @@ on_restart do
   puts "Refreshing Gemfile"
   ENV["BUNDLE_GEMFILE"] = ""
 end
-
-before_fork do
-  ActiveRecord::Base.connection_pool.disconnect!
-end
-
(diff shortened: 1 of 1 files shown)


javierm (approved):
@janimo Thank you very much for noticing the issue, and even bigger thanks for fixing it! :pray:

I've checked possible side-effects with Apartment; there seems to be none :smile:.

PR #5484, opened 2024-04-08
Reduce the size of JavaScript code on most pages

javierm:
⚠️ Right now, this pull request requires removing Turbolinks, so we aren't merging it until we decide whether the overall experience improves thanks to this change.

## Objectives

* Make the page load faster, particularly on devices with a slow internet connection

## Notes

With these changes, on production environments (where assets are compressed), the size of the JavaScript file in most of the pages is reduced from 454KB to 136KB. Overall, our [demo homepage](https://demo.consuldemocracy.org) is now 23% smaller TODO: update info.

I haven't managed to make these changes compatible with Turbolinks. I've tried both loading the scripts inside the <head> tag and loading them inside the <body> tag, and both options fail in different ways. Using `data-turbolinks-track="refresh"` results in strange behavior where a page is sometimes loaded twice and breaks the request referer as well as the Turbolinks integration with the browser's back button.

### DIFF
--- a/app/assets/config/manifest.js
+++ b/app/assets/config/manifest.js
@@ -6,6 +6,7 @@
 //= link application-rtl.css
 //= link application.js
 //= link html_editor_loader.js
+//= link legislation_annotatable_loader.js
 //= link map_loader.js
 //= link markdown_editor_loader.js
 //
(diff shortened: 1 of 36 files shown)


javierm:
I only created this pull request because I couldn't reopen #5166. This is an updated version of #5166 but, just like that one, we're closing it until we decide what to do regarding Turbolinks. I've only opened this one in order to have a reference when dealing with #5485.

Issue #5615, opened 2024-07-19
maps not rendering mutipolygons correctly


coslajohn:
When maps are rendered if they contain overlapping polygons it creates a hole

![image](https://github.com/user-attachments/assets/3d78efd7-b99d-4046-9f12-6aa172e1608c)

It should render as

![image](https://github.com/user-attachments/assets/b2bed8aa-8405-4090-82d4-3eba981a142b)

This is caused by the way consul flattens all of the polygons into a single set of outline points before rendering in map.js

I have a fix and shall raise a PR

PR #5616, opened 2024-07-19
Render Geozone Maps Collections and Multipolygons

coslajohn:
## References

This fixes issue #5615 

## Objectives

This change adds the ability to correctly interpret modern geoJson in a standard feature collection format
eg

```
{
"type": "FeatureCollection",
"name": "Central_North_Perth",
"crs": { "type": "name", "properties": { "name": "urn:ogc:def:crs:OGC:1.3:CRS84" } },
"features": [
{ "type": "Feature", "properties": { "INTERZONE": "S02002230", "NAME": "MONCRIEFFE AND FRIARTON", "TOTPOP2011": 4366, "RESPOP2011": 3806, "HHCNT2011": 1707, "STDAREAHA": 186.058189, "STDAREAKM2": 1.860581 }, "geometry": { "type": "MultiPolygon", "coordinates": [ [ [ [ -3.418866445343083, 56.377887663872443 ], [ -3.419601634341667, 56.377844202989245 ],
```

It will also correctly interpret Feature and geometry formats

## Visual Changes

Previous behaviour
Multipolygons render with holes
image

New behavior
Mutipolygons render correctly
image

New Functionality
Feature collection with multiple polygons
image

## Notes

By passing valid geoJson directly to leaflet, there is no longer any need to normalize the data and create outline points in the geozone model.

The previous behaviour used to accept invalid geojson fragments which started with eg
```
{
  "geometry": {
    "type": "Polygon",
    "coordinates": [
      [
        [-3.93798565212626, 56.9471386917935],
  
}
```

The Geojson needs to be vaild so must start with a type eg
```
{
  "type": "Feature",
  "geometry": {
    "type": "Polygon",
    "coordinates": [
      [
        [-3.93798565212626, 56.9471386917935],
```

UPDATE The code will now use the normalize method to convert any partial geojson fragments into standard feature collections for processing


coslajohn:
Guidance on what is valid GeoJSON


# GeoJSON Format for `Geozones`

## Overview

The `Geozone` model in our application expects the `geojson` attribute to be a valid GeoJSON string.  This documentation provides guidelines on the correct format for entering GeoJSON data and describes the valid fragments that are accepted.

## Correct Format for GeoJSON

ConsulDemocracy mapping using Leaflet. Leaflet requires coordinates to be in WGS84 (Longitude, Latitude) format.

The `geojson` attribute must conform to the following GeoJSON formats:

### 1. **FeatureCollection**

A `FeatureCollection` is a collection of features. The GeoJSON string must be an object with the following structure:

```json
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "geometry": {
        "type": "GeometryType",
        "coordinates": [...]
      },
      "properties": {}
    }
  ],
  "properties": {}  // Optional
}
```

- **`type`**: Should be `"FeatureCollection"`.
- **`features`**: An array of `Feature` objects.
- **`properties`**: An optional object that can hold additional metadata about the feature collection.

### 2. **Feature**

A `Feature` represents a single spatial object and must have the following structure:

```json
{
  "type": "Feature",
  "geometry": {
    "type": "GeometryType",
    "coordinates": [...]
  },
  "properties": {}
}
```

- **`type`**: Should be `"Feature"`.
- **`geometry`**: An object describing the type of geometry and its coordinates.
- **`properties`**: An object that can hold additional metadata about the feature.

### 3. **Geometry Object**

A `Geometry` object can be of the following types:

- **`Point`**
- **`LineString`**
- **`Polygon`**
- **`MultiPoint`**
- **`MultiLineString`**
- **`MultiPolygon`**
- **`GeometryCollection`**

Each type has a specific structure:

#### **Point**

```json
{
  "type": "Point",
  "coordinates": [longitude, latitude]
}
```

#### **LineString**

```json
{
  "type": "LineString",
  "coordinates": [[longitude1, latitude1], [longitude2, latitude2], ...]
}
```

#### **Polygon**

```json
{
  "type": "Polygon",
  "coordinates": [
    [
      [longitude1, latitude1], 
      [longitude2, latitude2], 
      ...
      [longitude1, latitude1]  // Closing the polygon
    ]
  ]
}
```

#### **MultiPolygon**

```json
{
  "type": "MultiPolygon",
  "coordinates": [
    [
      [
        [longitude1, latitude1], 
        [longitude2, latitude2], 
        ...
        [longitude1, latitude1]  // Closing the polygon
      ],
      ...
    ],
    ...
  ]
}
```

#### **GeometryCollection**

```json
{
  "type": "GeometryCollection",
  "geometries": [
    {
      "type": "GeometryType",
      "coordinates": [...]
    },
    ...
  ]
}
```

### 4. **Valid Fragments**

The following GeoJSON fragments are acceptable and will be normalized by the system:

- **Geometry Objects**: Directly include objects like `Point`, `LineString`, `Polygon`, `MultiPoint`, `MultiLineString`, `MultiPolygon`, or `GeometryCollection`.

  ```json
  {
    "type": "Point",
    "coordinates": [-73.981149, 40.747028]
  }
  ```

  ```json
  {
    "type": "Polygon",
    "coordinates": [
      [
        [-73.981149, 40.747028],
        [-73.983141, 40.734168],
        [-74.005390, 40.734315],
        [-74.004792, 40.750279],
        [-73.981149, 40.747028]  // Closing the polygon
      ]
    ]
  }
  ```

- **Feature**: Contains a `geometry` object and optional `properties`.

  ```json
  {
    "type": "Feature",
    "geometry": {
      "type": "Polygon",
      "coordinates": [[
        [-73.981149, 40.747028],
        [-73.983141, 40.734168],
        [-74.005390, 40.734315],
        [-74.004792, 40.750279],
        [-73.981149, 40.747028]
      ]]
    },
    "properties": {}
  }
  ```

- **FeatureCollection**: Contains an array of `Feature` objects.

  ```json
  {
    "type": "FeatureCollection",
    "features": [
      {
        "type": "Feature",
        "geometry": {
          "type": "Point",
          "coordinates": [-73.981149, 40.747028]
        },
        "properties": {}
      },
      {
        "type": "Feature",
        "geometry": {
          "type": "Polygon",
          "coordinates": [[
            [-73.981149, 40.747028],
            [-73.983141, 40.734168],
            [-74.005390, 40.734315],
            [-74.004792, 40.750279],
            [-73.981149, 40.747028]
          ]]
        },
        "properties": {}
      }
    ],
    "properties": {}  // Optional
  }
  ```

## Notes

- **`properties` Attribute**: In the `FeatureCollection`, `Feature`, and `Geometry` objects, if the `properties` attribute is missing, it will be added automatically with an empty object.
- **Closing Polygons**: Ensure that polygons and multi-polygons properly close their boundaries by repeating the first coordinate at the end.

javierm:
Hi, @coslajohn :smile:. Thank you for the report and the fix! :tada:

Before reviewing, could you have a look at the [messages reported by our linters](https://github.com/consuldemocracy/consuldemocracy/actions/runs/10041099109/job/27748458110?pr=5616)? :pray:

coslajohn:
@javierm I have fixed the linters so ready for review

app/models/concerns/geojson_format_validator.rb
```diff
@@ -3,21 +3,127 @@ def validate_each(record, attribute, value)
...
+        end
+      elsif geojson["type"] == "Feature"
+        unless valid_coordinates?(geojson["geometry"])
+          record.errors.add(
+            attribute,
+            :invalid_coordinates,
+            message: I18n.t("errors.geozone.attributes.geojson.invalid_coordinates")
+          )
+        end
       end
```
javierm, line 34: I'm writing some tests for this validation and it seems like we aren't validating the coordinates when we have a top-level geometry object :thinking:. Is it intentional?

javierm (commented):
@coslajohn Really appreciate this pull request :pray:; we couldn't have done it by ourselves.

I've left a few questions about points where I'm not sure about the intended behaviour. Could you help me with those?

Thanks!

app/models/concerns/geojson_format_validator.rb
coslajohn: I can't really remember it is a while since I looked at this. The validation actually needs another look as it was originally done to validate the geojson when it was done using points.  If we are being correct then a geojson fragment which begins with geometry should be invalid as it isn't valid geojson as per the spec. I think I was trying to include some degree of backward compatibility to avoid breaking existing geozones which contain invalid geojson but I don't think it is totally correct

app/assets/javascripts/map.js
```diff
@@ -220,17 +222,29 @@
...
+      var geojsonData = JSON.parse(geozone.outline_points);
 
-      if (geozone.headings !== undefined) {
-        polygon.bindPopup(geozone.headings.join("<br>"));
-      }
+      // Create a GeoJSON layer
+      var geoJsonLayer = L.geoJSON(geojsonData, {
+        style: function(feature) {
+          return {
+            color: geozone.color || feature.properties.color || "blue",
```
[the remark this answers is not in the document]
coslajohn: Seems sensible to prioritise feature.properties. Again seems sensible to move the default blue to the geozone model

app/models/geozone.rb
```diff
@@ -21,26 +21,66 @@ def safe_to_destroy?
...
+        # Handle raw geometry (coordinates) which should be a Feature
+        elsif parsed_geojson["type"] && parsed_geojson["coordinates"]
+
+          wrap_in_feature_collection(wrap_in_feature(parsed_geojson))
+        # Handle a valid geometry with type and coordinates
+        elsif parsed_geojson["geometry"] &&
+              parsed_geojson["geometry"]["type"] &&
+              parsed_geojson["geometry"]["coordinates"]
+          wrapped_feature = wrap_in_feature(parsed_geojson["geometry"])
+          wrapped_feature["properties"] ||= {}
```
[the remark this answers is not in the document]
coslajohn: you are right the line is probably redundant

coslajohn:
I have extended the code now to add layers to the map so that geozones can be toggled on and off. This is useful when geozones are overlapping

image

app/models/concerns/geojson_format_validator.rb
javierm: > I can't really remember it is a while since I looked at this. 

Once again, sorry for the late review! :pray: (I wonder whether I should make that sentence part of my nickname :sweat_smile:)

> If we are being correct then a geojson fragment which begins with geometry should be invalid as it isn't valid geojson as per the spec. 

Sorry, my "top-level geometry object" comment was ambiguous. What I meant was something like:

```
{
  "type": "Point",
  "coordinates": [-73.981149, 40.747028]
}
```

If I understand the specification correctly, this is valid GeoJSON, right? So, should we check the coordinates in this case?

javierm: By the way, just to clarify, if we need to check the coordinates in this case, I'll change the code accordingly (have already done it in a development branch in case it's needed). It's been a while since you worked on this code, so asking you to do it wouldn't be fair.

javierm:
Hi, @coslajohn :smile:.

> I have extended the code now to add layers to the map so that geozones can be toggled on and off. This is useful when geozones are overlapping

Cool, thanks! :tada:

It looks like, after these changes, some tests are failing with the message:

```
      ActionController::RoutingError:
        No route matches [GET] "/assets/images/layers.png"
```

Any ideas?

By the way, apparently the linters are reporting an error because of the ES6 syntax added here. We still use ES5 in order to support old browsers because on these browsers using the ES6 syntax generates an error that breaks features that only work with JavaScript enabled. Support for using the website without JavaScript is something we keep improving, but we aren't done yet.

coslajohn:
Hi @javierm the error is caused by a missing icon in the Leaflet installation.

There is a reference to it here https://stackoverflow.com/questions/71674735/rails-7-importmap-leaflet-css-images-path-fix

javierm:
Oh, I see, thanks @coslajohn! I've updated the pull request to include the latest changes in the master branch and changed the last commit in order to load the layers image. Vendor assets in Rails are certainly tricky, particularly because they load in a different way on a production environment.

app/assets/javascripts/map.js
```diff
@@ -84,8 +83,15 @@
       }
 
       App.Map.addInvestmentsMarkers(investmentsMarkers, createMarker);
-      App.Map.addGeozones(map);
+      // TODO: I wonder if it makes sense in the small map when there are only markers
```
javierm, line 86: Layers are certainly useful :+1:. Thanks for adding the feature! :tada:

I'm wondering if it makes sense to disable this feature in some cases, since it's currently shown even when there are no geozones (so only the "markers" layer is there, and so disabling it isn't so useful) and the map shown on the investments is usually pretty small and the icon takes some space :thinking:.

Any thoughts on this?

javierm:
Hi, @coslajohn :smile:.

I've added some tests for the GeoJSON validation and done some refactorings and small fixes in order to make the tests pass (in particular, we're now supporting GeoJSON containing Feature objects that contain GeometryCollection objects). I've also added a couple of commits with additional validations for Polygon and LineString objects.

There's one change I'd like to ask you about: I've removed (at least for now; adding it back would be easy) the code that wrapped the normalized GeoJSON in a `FeatureCollection` object (but I've kept the code that wraps it in a `Feature` object when necessary). Is this OK, or is there's something I'm missing that might cause an issue?

Please let me know what you think and whether I'm introducing an issue without realizing it! :pray:

Other than that, I'm going to add a small review to the latest changes (the ones regarding the layers to toggle geozones).

app/assets/javascripts/map.js
coslajohn: I don't have a strong opinion about this. There is a case where it would still be useful and that is in a multi heading budget, even if geozones are not in use being able to toggle on/off layers for each heading is still quite helpful

coslajohn:
> Hi, @coslajohn 😄.
> 
> There's one change I'd like to ask you about: I've removed (at least for now; adding it back would be easy) the code that wrapped the normalized GeoJSON in a `FeatureCollection` object (but I've kept the code that wraps it in a `Feature` object when necessary). Is this OK, or is there's something I'm missing that might cause an issue?
> 
> Please let me know what you think and whether I'm introducing an issue without realizing it! 🙏
> 

Hmmm....  taking out the FeatureCollection wrapping probably doesn't make any functional difference as any geojson which is being normalized is only ever likely to be a single feature. However, I think my preference would be to keep the FeatureCollection wrapping. The reason for this is that most of the large GIS systems eg ArcGIS, QGIS  export geojson as a complete FeatureCollection and so for completeness and for possible futureproofing in case we ever add a different mapping system I would tend to leave it in.

Having said that I am happy to go with whatever you think best

javierm:
> Hmmm.... taking out the FeatureCollection wrapping probably doesn't make any functional difference as any geojson which is being normalized is only ever likely to be a single feature. However, I think my preference would be to keep the FeatureCollection wrapping. The reason for this is that most of the large GIS systems eg ArcGIS, QGIS export geojson as a complete FeatureCollection and so for completeness and for possible futureproofing in case we ever add a different mapping system I would tend to leave it in.
>
> Having said that I am happy to go with whatever you think best

Sounds reasonable :+1:. I've re-added the code and added this explanation to the commit message.

app/assets/javascripts/map.js
javierm: > even if geozones are not in use being able to toggle on/off layers for each heading is still quite helpful

I'm not sure I'm following :thinking:. If geozones aren't used, what's supposed to happen when you toggle on/off a heading? If I understand correctly, right now turning on/off a heading only changes the visibility of the associated geozone polygon/feature, right? Sorry if I'm missing something! :pray:

javierm: By the way, I've uploaded a couple of changes.

Since the layers were links to headings or geozones, it was very easy to accidentally click on a link when trying to toggle the visibility of the layer, particularly for people with motor disabilities. So now we're using plain text for the name of a layer.

To avoid the internationalization of the "Markers: " text for now we're not including this layer in the layer control. This also simplifies the logic about the markers layer appearing on the "Create new proposal" form (it no longer appears there).

Finally, in the layers control we're only including geozones with GeoJSON available and we're only adding the layers control if there's more than one geozone with GeoJSON available.

What do you think? :thinking:

coslajohn: that all sounds good to me.  I think what I was thinking with the heading layers was that at some point all markers associated with that heading could be toggled on/off enabling you see only the marker locations for a heading even if the heading doesn't have a geozone associated with it

javierm: > that heading could be toggled on/off enabling you see only the marker locations for a heading even if the heading doesn't have a geozone associated with it

That would be interesting indeed. Future pull requests for this would be welcome :smile:.

javierm (approved):
@taitus I'm approving these changes, but I've also modified parts of the code (the tests, in particular). Could you have a look?

taitus (approved):

PR #5617, opened 2024-07-20
Replace legislation processes form markdown

decabeza:
## Objectives

Replace legislation processes form markdown to CKEditor fields.

### Before
before

### After
after

### DIFF
--- a/app/views/legislation/processes/_process.html.erb
+++ b/app/views/legislation/processes/_process.html.erb
@@ -13,7 +13,7 @@
     </div>
 
     <div class="small-12 medium-11 column end">
-      <%= markdown(process.summary.presence || first_paragraph(process.description)) %>
+      <%= auto_link_already_sanitized_html AdminWYSIWYGSanitizer.new.sanitize((process.summary.presence || first_paragraph(process.description))) %>
     </div>
 
     <div class="small-12 medium-11 column end">
--- a/app/assets/stylesheets/layout.scss
+++ b/app/assets/stylesheets/layout.scss
@@ -916,11 +916,11 @@ form {
   }
 
   .html-area:not(.form-error) {
-    height: 272px;
+    height: $line-height * 12;
     margin-bottom: $line-height;
 
     &.admin {
-      height: 572px;
+      height: $line-height * 20;
     }
   }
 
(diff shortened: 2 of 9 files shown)


javierm:
@decabeza Interesting :thinking:. Do these changes work with existing data which might have been created using markdown (like `##` for `<h2>` tags, `*` for `<strong>`, ...)?

app/views/legislation/processes/_process.html.erb
```diff
@@ -13,7 +13,7 @@
     </div>
 
     <div class="small-12 medium-11 column end">
-      <%= markdown(process.summary.presence || first_paragraph(process.description)) %>
+      <%= auto_link_already_sanitized_html AdminWYSIWYGSanitizer.new.sanitize((process.summary.presence || first_paragraph(process.description)) %>
```
javierm, line 16: Tests are reporting a syntax error here; it looks like there's an extra parenthesis after `sanitize` :thinking:.

decabeza: It seems that a `)` was missing at the end. 😌

decabeza:
> Hi, @decabeza 😄.
> 
> This is interesting 🤔. Do these changes work with existing data which might have been created using markdown (like `##` for `<h2>` tags, `*` for `\<strong\>`, ...)?

Ouch! It seem doesn't work. 😬 For example something like:

```markdown
# Title 
## Subtitle
**Bold**
```
Should be:
```html
<h1>Title</h1>
<h2>Subtitle</h2>
<p><strong>Bold</strong></p>
```

But after this changes, it renders as:
```html
<p># Title</p>
<p>## Subtitle</p>
<p>**Bold**</p>

```
So, I imagine it will be necessary to create a Rake so that if those fields with content already exist they should be "translated" from Markdown to HTML. 😅  I think we could use `redcarpet` as well to do the opposite of what it does [app/lib/markdown_converter.rb](https://github.com/consuldemocracy/consuldemocracy/blob/master/app/lib/markdown_converter.rb). 🤔

lucialuzuriaga (commented):
Hi @decabeza! Since this PR is in a draft state has been open for a long time without updates, I am going to close it. If you or anyone else would like to pick this back up in the future, please feel free to comment or open a new PR. Thanks!

PR #5636, opened 2024-07-22
Restrict access to Admin functions by IP Address

coslajohn:
## Objectives

For security purposes conditionally restrict access to Admin functions to authorised IP addresses

This pull request introduces a RestrictIP module to implement conditional IP restrictions based on the configuration in secrets.yml. 

By default, access is unrestricted.  This will satisfy a security need of some municipalities to restrict access to admin functions to only their network

## Visual Changes

image

## Notes

Access is only restricted if valid IP addresses or ranges are specified in secrets.yml.

Next enhancement will be to read from a Setting rather than secrets.yml

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -781,6 +781,9 @@ Style/RaiseArgs:
 Style/RedundantArgument:
   Enabled: true
 
+Style/RedundantBegin:
+  Enabled: true
+
 Style/RedundantCondition:
   Enabled: true
 
--- a/app/controllers/admin/base_controller.rb
+++ b/app/controllers/admin/base_controller.rb
@@ -1,4 +1,5 @@
 class Admin::BaseController < ApplicationController
+  include IpDeniedHandler
   layout "admin"
   before_action :authenticate_user!
 
--- a/app/controllers/concerns/ip_denied_handler.rb
+++ b/app/controllers/concerns/ip_denied_handler.rb
@@ -1,0 +1,17 @@
+module IpDeniedHandler
+  extend ActiveSupport::Concern
+
+  included do
+    before_action :restrict_ip, unless: :allowed_ip?
+  end
+
+private
+
+  def restrict_ip
+    redirect_to root_path, alert: t("ip_denied_handler.unauthorized")
+  end
+
+  def allowed_ip?
+    RestrictAdminIps.new(request.remote_ip).allowed?
+  end
+end
--- a/app/controllers/graphql_controller.rb
+++ b/app/controllers/graphql_controller.rb
@@ -9,25 +9,23 @@ class GraphqlController < ApplicationController
   class QueryStringError < StandardError; end
 
   def execute
-    begin
-      raise GraphqlController::QueryStringError if query_string.nil?
+    raise GraphqlController::QueryStringError if query_string.nil?
 
-      result = ConsulSchema.execute(
-        query_string,
-        variables: prepare_variables,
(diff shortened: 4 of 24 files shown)


javierm:
Hi, @coslajohn :smile:. Thanks for the pull request!

I know this is still a work in progress, but here are some preliminary comments.

> Next enhancement will be to read from a Setting rather than secrets.yml

For now, we can keep it in the `secrets.yml` file. Maybe under the `security` section of that file?

Since this is a value that might be different for different tenants, instead of using `Rails.application.secrets`, do you think it makes sense to use `Tenant.current_secrets` instead?

Also, it looks like [our linters check is reporting some issues](https://github.com/consuldemocracy/consuldemocracy/actions/runs/10044544657/job/27759826792?pr=5636). Could you have a look? You can also run a similar check locally (although this is only about 95% reliable) using `bundle exec pronto run`.

coslajohn:
@javierm I have accepted the recommendation about reading secrets from Tenants and fixed the linter errors as far as I can see.  I need a little help with the spec file for this as no idea where to start!

javierm:
@coslajohn Thanks a lot! We really appreciate it.

We'll take it from here and have a look at the tests. It'll take a while, though (August is a slow month here, and it's not like we're usually the fastest reviewers ever :relieved:).

coslajohn:
@taitus @javierm just noticed that when I try this in production under release 2 I am getting an error 
 uninitialized constant Admin::BaseController::RestrictIP (NameError)
 
  zeitwerk error "expected app/controllers/concerns/restrict_ip.rb to define constant RestrictIp  but didn't (Zeitwerk::NameError)"
  
  Can you suggest how I fix this?

javierm:
Hi, @coslajohn :smile:.

Luckily, this is an easy one! The problem is that Zeitwerk expects the file `restrict_ip` to define a class named `RestrictIp` instead of `RestrictIP`.

There are at least two options here.

The safe one is renaming `RestrictIP` to `RestrictIp`.

The one which is probably be safe but might have side-effects is adding `inflect.acronym "IP"` inside the `ActiveSupport::Inflector.inflections(:en) do |inflect|` block in `config/initializers/inflections.rb`.

Hope this helps!

coslajohn:
brilliant thanks
________________________________

javierm:
Hi, @coslajohn :smile:.

I see that here we're using generic names like `RestrictIp` and `allowed_ips` and a concern that might be included in other controllers. Does it mean that you intend to use the same list of IPs to restrict access to other sections? I was wondering whether people might want to use different addresses to restrict different sections, in which case maybe it'd make sense to use `RestrictAdminIps` and `allowed_admin_ips` or something :thinking:.

coslajohn:
to be honest I can't think of any other section that you would want to restrict access to.

The purpose of this is to allow access to the admin functions to be restricted as that is a requirement of our security assessments.

I don't have any strong opinion on naming convention.

While it works, the use case is really best suited to municipalities where the admins are on a corporate network. I really need to think about implementing 2FA for Admins but I don't have time for that just now
________________________________

javierm (approved):
Thank you so much! And sorry once again for the late review :pray:.

I've added some tests. To simplify the tests, I've extracted the main logic into a different class. Also, since we already had an `AccessDeniedHandler`, I've renamed the concern to `IpDeniedHandler`.

@taitus Could you have a look at these changes?

taitus (approved):

PR #5692, opened 2024-09-16
Fix environment variables and version warnings in Docker

javierm:
## References

* [LegacyKeyValueFormat documentation](https://docs.docker.com/reference/build-checks/legacy-key-value-format/)
* [The version top-level element is obsolete](https://docs.docker.com/reference/compose-file/version-and-name/#version-top-level-element-obsolete)

## Background

We were getting warnings in our CI:

> Legacy key/value format with whitespace separator should not be used
> LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy
> "ENV key value" format

We were also getting a warning in Docker Compose:

```
docker-compose.yml: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion
```

## Objectives

* Use the right format when declaring environment variables in Dockerfile
* Remove an obsolete attribute in docker-compose that results in a warning

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 FROM ruby:3.2.5-bullseye
 
-ENV DEBIAN_FRONTEND noninteractive
+ENV DEBIAN_FRONTEND=noninteractive
 
 # Install essential Linux packages
 RUN apt-get update -qq \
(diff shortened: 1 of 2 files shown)


taitus (approved):

javierm:
The test failure is not related to this pull request :relieved:.

PR #5762, opened 2024-10-28
Add unique index to poll voters table

javierm:
## References

* We avoided the creation of duplicate voters and added a task to delete existing ones in pull request #5532

## Objectives

* Simplify the code that makes sure no duplicate voters can be created
* Make sure that duplicate voters can't even be created using custom code or the rails console

## Release notes

:warning: We've added a unique index to the poll voters table. If you executed the release tasks when upgrading to Consul Democracy 2.2 or you've never used Consul Democracy 2.1, this shouldn't affect you. However, if you run into issues when upgrading due to duplicate voters in your database, run `bin/rake polls:remove_duplicate_voters RAILS_ENV=production`.

**Note for ourselves**: don't remove the task to delete duplicate voters in version 2.4 so it's easier to find its code while upgrading, just in case.

### DIFF
--- a/app/controllers/officing/voters_controller.rb
+++ b/app/controllers/officing/voters_controller.rb
@@ -14,18 +14,14 @@ class Officing::VotersController < Officing::BaseController
     @poll = Poll.find(voter_params[:poll_id])
     @user = User.find(voter_params[:user_id])
 
-    @user.with_lock do
-      @voter = Poll::Voter.new(document_type: @user.document_type,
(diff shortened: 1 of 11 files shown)


taitus:
Hi @javierm 

I've made some changes on this branch and removed a few others, then pushed them to my [fork](https://github.com/taitus/consuldemocracy/tree/unique_index_on_poll_voters_wip).
If you agree with the suggestion, we can add it here.

PR #5766, opened 2024-10-30
Reduce code duplication in Graphql

taitus:
## References
Bump graphql from 1.13.22 to 2.0.31 #5748 

## Objectives

Apply refactors to reduce code duplication:
  - Introduce a helper method to handle connection_type fields more efficiently.
  - Simplify the definition of id-based fields using a new id_field helper, reducing redundancy in the QueryType and other types.

### DIFF
--- a/app/graphql/types/query_type.rb
+++ b/app/graphql/types/query_type.rb
@@ -1,61 +1,41 @@
 module Types
   class QueryType < Types::BaseObject
     field :budgets, Types::BudgetType.connection_type, "Returns all budgets", null: false
-    field :budget, Types::BudgetType, "Returns budget for ID", null: false do
-      argument :id, ID, required: true, default_value: false
-    end
+    object_by_id_field :budget, Types::BudgetType, "Returns budget for ID", null: false
 
     field :comments, Types::CommentType.connection_type, "Returns all comments", null: false
-    field :comment, Types::CommentType, "Returns comment for ID", null: false do
-      argument :id, ID, required: true, default_value: false
-    end
+    object_by_id_field :comment, Types::CommentType, "Returns comment for ID", null: false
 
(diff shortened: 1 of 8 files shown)


app/graphql/types/query_type.rb
```diff
@@ -1,61 +1,20 @@
...
-    field :users, Types::UserType.connection_type, "Returns all users", null: false
-    field :user, Types::UserType, "Returns user for ID", null: false do
-      argument :id, ID, required: true, default_value: false
-    end
-
-    field :votes, Types::VoteType.connection_type, "Returns all votes", null: false
-    field :vote, Types::VoteType, "Returns vote for ID", null: false do
-      argument :id, ID, required: true, default_value: false
-    end
+    def self.connection_and_id_fields(name, type)
```
javierm, line 3: If we combine the ideas about the individual methods:

```suggestion
    def self.collection_and_object_by_id_fields(name, type)
```

app/graphql/types/query_type.rb
```diff
@@ -1,61 +1,20 @@
...
+      connection_field name.to_s.pluralize.to_sym, type, "Returns all #{name.to_s.pluralize}", null: false
+      id_field name, type, "Returns #{name.to_s.capitalize}", null: false
```
javierm, line 5: I guess we're missing the `for ID` we used to have :thinking:. By the way, what's the reason to introduce `capitalize` here? We're in the middle of a sentence, and we don't use `capitalize` in the `connection_field` method :thinking:.

```suggestion
      id_field name, type, "Returns #{name} for ID", null: false
```

javierm (commented):
@taitus Cool! I've left some comments, mostly about naming. Let me know what you think! :wink:

javierm (approved):

PR #5794, opened 2024-11-20
include physical ballots in total number of voters

coslajohn:
## Objectives

In Participatory Budgeting
When uploading ballot sheets the ballots are correctly counted

In stats the number of votes is correctly displayed

The number of participants is not updated to reflect offline or physical voters.

locality stats

This change assumes each line in a ballot sheet represents a participant and includes physical voters in the final total

## Notes

This only partially improves the PB stats.
The participants per district table is not updated because there is no way to associate the Polling Booth Location with the district.

The solution to this will be to make the Polling Booth Location selectable from a drop down of either Budget Heading Scope of Operation or from the list of geozones. This would then allow the ballot sheets to be counted by location allowing the update of participants by district

There may be some redundant code here still as it refers to poll_ballot_voters and as far as I can see that isn't used anywhere?

### DIFF
--- a/app/models/budget/stats.rb
+++ b/app/models/budget/stats.rb
@@ -50,7 +50,11 @@ class Budget::Stats
   end
 
   def total_participants_vote_phase
-    (balloters + poll_ballot_voters).uniq.count
+    (balloters + poll_ballot_voters).uniq.count + total_physical_balloters
+  end
+  
+  def total_physical_balloters
+    budget.ballots.where(physical: true).count
   end
 
   def total_budget_investments


javierm:
Hi, @coslajohn :smile:.

Once again, sorry for the late reply :pray:; took a while to check this :relieved:.

> There may be some redundant code here still as it refers to poll_ballot_voters and as far as I can see that isn't used anywhere?

If I remember correctly, when a citizen goes to a booth in order to vote, the assigned polling officer will use the "Validate document" feature in the "Polling officers" area to register the fact that the citizen has voted. Then, after the voting has finished, the polling officers enter the data in the physical ballots into the application.

So, if I understand correctly, if we count both the citizens that voted (which is what the `poll_ballot_voters` method returns) and the physical ballots, we would be counting voters twice. In other words, we don't add participants when uploading ballot sheets because they were already added when they voted.

Does it make sense? Please correct me if I'm wrong! I was never involved in a participatory process with physical ballots, so I could be mistaken here.

coslajohn:
Hi @javierm. I had actually forgotten about this one.....

Thanks for clarifying the behaviour. I didn't know about the the document validation updating the voter count.

We don't use the document validation when doing manual voting.  The GDPR and practical issues with managing voter documents are too large unless you have a team the size of the original Madrid team! Most manual votes are collected by show of hands or paper voting in town hall events, rather than in a formal polling station environment which is how it appears to have been originally designed

I wonder if this could be solved with a feature setting "Skip Document Verification". What do you think?

The manual voting is a really useful feature but it is one which needs a bit of work to make it more flexible. It would be really good to be able to upload totals for each budget investment in the same way as the manual voting in polls works.

javierm:
@coslajohn I see... So, just to be sure, when people vote in town hall events, there's no way to check whether a person has already voted using the web and therefore they've voting twice, right?

Regarding the setting, I'm wondering whether it'd make sense to compare the amount of people who have voted in a booth with the amount of physical ballots, and from there deduce whether we're using document verification :thinking:. In the case you mention, for example, the amount of people who have voted in a booth should be zero, right?

If it doesn't make sense, then a setting would probably be the way to go :thinking:.

And, yes, the offline voting part is certainly very Madrid-specific. Improving it has been on our TODO list for years; really hope we'll get there eventually.

PR #5797, opened 2024-11-25
Make nested imageable specs faster

taitus:
## Objectives
As we did in PR #5369 , we reduce the number of tests we are going to run. In this case we remove the shared example and create a new file in system folder to run the same tests but with a different factory in each execution for each test. 

As they are the same tests and based on the same functionality, we use this related PR technique (random factory) to make sure that everything still works.


spec/system/nested_imageable_spec.rb
```diff
@@ -0,0 +1,264 @@
...
+        "Proposal created successfully"
+      end
+    end
+  end
+
+  context "New and edit path" do
+    before do
+      create(:administrator, user: user) if admin_section?(path) || management_section?(path)
+      imageable.update!(author: user) if edit_path?(path)
+      do_login_for(user, management: management_section?(path))
```
javierm, line 63: Maybe we don't need to pass the `path` parameter to these methods, since it's the result of a `let` block :thinking:.

```suggestion
      create(:administrator, user: user) if admin_section? || management_section?
      imageable.update!(author: user) if edit_path?
      do_login_for(user, management: management_section?)
```

spec/system/tags/budget_investments_spec.rb
```diff
@@ -181,7 +181,7 @@
 
     click_button "Create Investment"
 
-    expect(page).to have_content "Investment created successfully."
+    expect(page).to have_content "Budget Investment created successfully."
```
javierm, line 184: Since we're change it here, shall we change this text in other spec files?

spec/system/nested_imageable_spec.rb
```diff
@@ -0,0 +1,264 @@
...
+  def admin_section?(path)
+    path.starts_with?("/admin/")
+  end
+
+  def management_section?(path)
+    path.starts_with?("/management/")
+  end
+
+  def edit_path?(path)
+    path.match?("/edit")
```
javierm, line 262:
```suggestion
    path.match?("/edit/")
```

So something like `/editors` doesn't match.

spec/system/nested_imageable_spec.rb
```diff
@@ -0,0 +1,264 @@
...
+    elsif factory == :budget_investment
+      [
+        new_budget_investment_path(budget_id: imageable.budget_id),
+        new_management_budget_investment_path(budget_id: imageable.budget_id)
+      ].sample
+    elsif factory == :future_poll_question_option
+      new_admin_option_image_path(option_id: imageable.id)
+    elsif factory == :proposal
+      [new_proposal_path, edit_proposal_path(imageable)].sample
+    end
```
javierm, line 26: I'm wondering whether it makes sense to use a `case` here :thinking:.

```suggestion
    case factory
    when :budget
      new_admin_budgets_wizard_budget_path
    when :budget_investment
      [
        new_budget_investment_path(budget_id: imageable.budget_id),
        new_management_budget_investment_path(budget_id: imageable.budget_id)
      ].sample
    when :future_poll_question_option
      new_admin_option_image_path(option_id: imageable.id)
    when :proposal
      [new_proposal_path, edit_proposal_path(imageable)].sample
    end
```

Which one do you prefer?

javierm (commented):
Great! I've left a few comments; let me know what you think!

spec/system/nested_imageable_spec.rb
javierm: Maybe we could add the `Style/CaseLikeIf` rubocop rule :thinking:.

spec/system/nested_imageable_spec.rb
javierm: Sorry, I meant:

```suggestion
    path.end_with?("/edit")
```

javierm (approved):

PR #5857, opened 2025-01-28
Budgets: allow admins to change winners & make columns adaptive to current phase

DominikPeters:
## Objectives

Consul currently determines the winners in a participatory budget by an algorithm triggered by clicking the "calculate winning investments" button. It does not allow admins to change which projects win. But in many context, being able to change the winners would be useful:
* Some cities have rules like "projects must receive at least 100 votes to win"
* Some projects may be incompatible with each other, so if both are selected as winner by the algorithm, this needs to be overridden
* Maybe some project almost fit in the budget but overshot it a bit. The city may have the flexibility to increase the available budget slightly, and make the project win after all.
* The city may want to use an entirely different winner determination method such as the Method of Equal Shares.

To allow this, this PR adds a toggle for each project in the budget investments table, allowing admins to change which projects win. This is the same style of toggle already used for whether a project is selected, and whether it is shown to valuators.

**Changes to columns**. In addition, this PR adds a column "Votes" showing the number of votes each project received during the balloting phase. Adding two more columns (votes and winner toggle) makes the table a bit overwhelming, so I added logic that decides which columns to show based on the phase of the budget. For example, once we reach balloting, columns related to valuation are not shown anymore. And the two new columns are not shown until we start balloting. The columns shown can still be customized, and the state is saved in a cookie as before, except that the cookie is not specific to each phase.

## References

No directly relevant issue, but the possibility of adding a winner toggle is mentioned at the bottom of https://github.com/consuldemocracy/consuldemocracy/issues/5501

## Visual Changes

Investment projects view during "Reviewing ballots" phase: (note no winner toggle for projects that are not selected)
image

During the "valuating projects" phase, showing different default columns:
image

During the "finished" phase:
image

## Future

I'm working on the [Method of Equal Shares](https://equalshares.net/) method, which is a voting method for participatory budgets that provides proportional representation. Some cities using Consul have expressed interest in using this method, so I'm working on features that make this possible. Next steps would be:
* Implement an export feature that exports (anonymized) raw votes so the voting outcome can be computed by an external tool
* Perhaps: Implement the Method of Equal Shares directly in Consul, with admins able to select their desired winner determination method in the settings.

### DIFF
--- a/app/assets/javascripts/columns_selector.js
+++ b/app/assets/javascripts/columns_selector.js
@@ -2,15 +2,12 @@
   "use strict";
   App.ColumnsSelector = {
     initColumns: function() {
-      var c_value, columns;
+      var columns;
       App.ColumnsSelector.hideAll();
-      c_value = App.ColumnsSelector.currentValue();
-      if (c_value.length === 0) {
-        c_value = $("#js-columns-selector").data("default");
-        App.Cookies.saveCookie($("#js-columns-selector").data("cookie"), c_value, 30);
-      }
-      columns = c_value.split(",");
-      columns.forEach(function(column) {
+      columns =
+        App.Cookies.getCookie($("#js-columns-selector").data("cookie")) ||
+        $("#js-columns-selector").data("default");
+      columns.split(",").forEach(function(column) {
         $("[data-field=" + column + "]").removeClass("hidden");
         $("#column_selector_" + column).prop("checked", true);
       });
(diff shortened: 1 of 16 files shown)


DominikPeters:
Hi! Please let me know if I can clarify anything about this, and whether adding features like this fits or doesn't fit with your plans. Many thanks!

DominikPeters:
Thanks for your answer 😃 No rush, I just wanted to check that we might get to it sometime down the road. Good luck with hiring :)

PR #5859, opened 2025-01-28
Add chromium-driver to Dockerfile

### LINKED ISSUE #5854: Docker: running system tests fails due to absence of chromedriver
In the [documentation about using docker](https://github.com/consuldemocracy/consuldemocracy/blob/master/docs/en/installation/docker.md), it says:

> System tests also work out of the box, although they might fail the first time while the tool running the tests downloads the right version of Chromedriver (which is needed to run them), and only "headless" mode (with a browser running in the background) is supported, which is the mode you'd probably use more than 95% of the time anyway. For example, to run the tests for the homepage:
>
> `POSTGRES_PASSWORD=password docker-compose run app bundle exec rspec spec/system/welcome_spec.rb`

For me, as predicted, the tests fail the first time, but they continue to fail after. The errors are of form:
```
  26.2) Failure/Error: example.run
        
        Selenium::WebDriver::Error::WebDriverError:
          unable to connect to /home/consul/.cache/selenium/chromedriver/linux64/132.0.6834.110/chromedriver 127.0.0.1:9515
          # /usr/local/bundle/gems/selenium-webdriver-4.25.0/lib/selenium/webdriver/common/service_manager.rb:132:in `connect_until_stable'
          # ... omitted ...
          # ./spec/spec_helper.rb:41:in `block (3 levels) in <top (required)>'
          # /usr/local/bundle/gems/i18n-1.14.6/lib/i18n.rb:353:in `with_locale'
          # ./spec/spec_helper.rb:40:in `block (2 levels) in <top (required)>'
```
Do you have ideas for how to fix it, i.e. to install/start chromedriver? Thanks for your help.

DominikPeters:
## References

* Closes #5854

## Objectives

I followed the instructions in [`docker.md`](https://github.com/consuldemocracy/consuldemocracy/blob/master/docs/en/installation/docker.md) to install Consul on my Mac. It all worked, except that systems tests did not work due to a problem with chromedriver.

This PR adds `chromium-driver` to the Dockerfile. This fixed the problem for me. I also tested this version of the Dockerfile on a fresh machine (GitHub Codespaces) and it didn't cause problems there. However, on GitHub Codespaces the systems tests also work without installing chromium-driver in the Dockerfile, so the problem I have on my machine remains a little mysterious.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,7 @@ RUN apt-get update -qq \
     unzip
 
 # Install Chromium for E2E integration tests
-RUN apt-get update -qq && apt-get install -y chromium
+RUN apt-get update -qq && apt-get install -y chromium chromium-driver
 
 # Files created inside the container repect the ownership
 RUN adduser --shell /bin/bash --disabled-password --gecos "" consul \


javierm (approved):
Hi, @DominikPeters :smile:.

Thanks a lot for this pull request! :pray: To be honest, I've no idea why this change fixes your issue :sweat_smile: but, at the same time, I can't think of a scenario where this change makes matter worse either :thinking:. I've done a few tests and everything seems fine.

So let's merge this change for now and we'll keep an eye on it in case there are issues in the future.

Once again, thanks a lot for reporting the issue and, particularly, for opening a pull request to fix it! :tada:

javierm:
Force-pushing since I couldn't find the test report from last time :relieved:. I've also updated the commit message in order to include the info mentioned on the issue.

PR #5927, opened 2025-03-20
Replace link with button in ballot investment component

cyrillefr:
## References

Contributes to #5461 (first item ie first TODO)
## Objectives

Replacing `link_to` with `button_to` elements to improve Accessibility.
The first of many changes. This one is a bit of a try. Next ones should be addressed in other PRs. 

## Visual Changes

There should be none. I only had to explicitly use a cursor pointer in the css to get the same ux than with the link.

## Notes

> Mention rake tasks or actions to be done when deploying this changes to a server (if any).
> Explain any caveats, or important things to notice like deprecations (if any).

Issue #5461, opened 2024-03-31
Use buttons instead of links in the public area


javierm:
### User Story

As mentioned in commits 5311daadf and bb958daf0, using links combined with JavaScript to generate POST (or, in this case, DELETE) requests to the server has a few issues. Quoting from the first of these commits:

> First, screen readers will announce them as "links". People using screen readers usually associate links with "things that get you somewhere" and buttons with "things that perform an action". So when something like "Delete, link" is announced, they'll probably think this is a link which will take them to another page where they can delete a record.
>
> Furthermore, the URL of the link for the "destroy" action might be the same as the URL for the "show" action (only one is accessed with a DELETE request and the other one with a GET request). That means screen readers could announce the link like "Delete, visited link", which is very confusing.
>
> These links won't work when opening links in a new tab, since opening links in a new tab always results in a GET request to the URL the link points to.
>
> Finally, submit buttons work without JavaScript enabled, so they'll work even if the JavaScript in the page hasn't loaded (for whatever reason).

## Implementation details

See pull request #5459 for an example, since this pull request does the exact same thing for links in the admin section.

It'll probaby be a good idea to add a `Shared::ActionComponent` class, and make `Admin::ActionComponent` class call `Shared::ActionComponent` class. Most of the code in `Admin::ActionComponent` would be moved to `Shared::ActionComponent`, keeping only the things that are only relevant in the admin area (like the `default_path` method).

Here's a patch adding a `# TODO` element to the places where we detected this is needed. Note the patch was based on version 2.1.1, so it's possible that it doesn't apply cleanly if you're using a more recent version of Consul Democracy.

```diff
diff --git a/app/components/budgets/ballot/investment_component.html.erb b/app/components/budgets/ballot/investment_component.html.erb
index 58e62f4ea9..8e0d5af93c 100644
--- a/app/components/budgets/ballot/investment_component.html.erb
+++ b/app/components/budgets/ballot/investment_component.html.erb
@@ -3,6 +3,7 @@
   <%= investment_price %>
 
   <% if budget.balloting? %>
+    <%# TODO %>
     <%= link_to delete_path,
                 title: t("budgets.ballots.show.remove"),
                 class: "remove-budget-investment",
diff --git a/app/components/documents/document_component.html.erb b/app/components/documents/document_component.html.erb
index 1de9100811..1a73c49212 100644
--- a/app/components/documents/document_component.html.erb
+++ b/app/components/documents/document_component.html.erb
[...]

PR #5955, opened 2025-04-04
Replace link with button in various components part i

cyrillefr:
## References

Contributes to #5461 

## Objectives

Replacing `link_to` with `button_to` elements to improve Accessibility.
Follow-up of #5927 
Three files here (and 3 commits)

## Visual Changes

There should be none. 
I  had to modify css in case there were a difference in styling, so that buttons are styled like links.

### DIFF
--- a/app/assets/stylesheets/layout.scss
+++ b/app/assets/stylesheets/layout.scss
@@ -1154,6 +1154,10 @@ form {
       line-height: rem-calc(120);
       vertical-align: top;
     }
+
+    button {
+      @include link;
+    }
   }
 }
 
--- a/app/assets/stylesheets/layout.scss
+++ b/app/assets/stylesheets/layout.scss
@@ -1811,8 +1811,11 @@ table {
   color: $delete;
 }
 
-.dropdown-pane a {
-  color: $text-medium;
+.dropdown-pane {
+  a,
+  button {
+    color: $text-medium;
+  }
 }
 
 // 17. Activity
--- a/app/assets/stylesheets/layout.scss
+++ b/app/assets/stylesheets/layout.scss
@@ -821,6 +821,16 @@ footer {
 
 .auth-page {
 
+  .cancel-button {
+    cursor: pointer;
+
+    &:hover,
+    &:active,
+    &:focus {
+      text-decoration: underline;
+    }
+  }
+
   .footer {
     margin-top: 0;
   }
--- a/app/assets/stylesheets/layout.scss
+++ b/app/assets/stylesheets/layout.scss
@@ -2134,7 +2134,7 @@ table {
     text-align: center;
   }
 
-  a {
+  button {
     font-weight: bold;
     margin-right: $line-height;
     text-decoration: none;
@@ -2152,6 +2152,14 @@ table {
     &::before {
       margin-right: $font-icon-margin;
     }
+
+    &:hover {
+      cursor: pointer;
+    }
+  }
+
+  form {
+    display: inline;
   }
 }
 
--- a/app/assets/stylesheets/layout.scss
+++ b/app/assets/stylesheets/layout.scss
@@ -1190,9 +1190,17 @@ form {
   @include breakpoint(medium) {
 
     .mark-notification {
+      @include anchor-color;
+      cursor: pointer;
       position: absolute;
       right: 12px;
       top: 12px;
+
+      &:hover,
+      &:active,
+      &:focus {
+        @include anchor-color-hover;
+      }
     }
   }
 
--- a/app/views/devise/shared/_links.html.erb
+++ b/app/views/devise/shared/_links.html.erb
@@ -11,10 +11,4 @@
     <%= link_to t("devise_views.shared.links.new_unlock"), new_unlock_path(resource_name) %><br>
   <% end -%>
 
-  <%- if devise_mapping.omniauthable? && devise_mapping.name == "user" %>
-    <%- resource_class.omniauth_providers.each do |provider| %>
-      <%= link_to t("devise_views.shared.links.signin_with_provider", provider: provider.to_s.titleize), omniauth_authorize_path(resource_name, provider), method: :post %><br>
-    <% end -%>
-  <% end -%>
-
 </div>
--- a/app/views/follows/_follow_button.html.erb
+++ b/app/views/follows/_follow_button.html.erb
@@ -1,19 +1,19 @@
 <div class="js-follow">
   <span class="followable-content">
     <% if follow.followable.followed_by?(current_user) %>
-      <%= link_to t("shared.following"),
-                  follow_path(follow),
-                  method: :delete, remote: true,
-                  title: unfollow_text(follow.followable),
-                  class: "button expanded" %>
+      <%= button_to t("shared.following"),
+                    follow_path(follow),
+                    method: :delete, remote: true,
+                    title: unfollow_text(follow.followable),
+                    class: "button expanded" %>
 
     <% else %>
-      <%= link_to follow_text(follow.followable),
-                  follows_path(followable_id: follow.followable.id,
-                               followable_type: follow.followable.class.name),
-                  method: :post, remote: true,
-                  title: follow_text(follow.followable),
-                  class: "button hollow expanded" %>
+      <%= button_to follow_text(follow.followable),
+                    follows_path(followable_id: follow.followable.id,
+                                 followable_type: follow.followable.class.name),
+                    remote: true,
+                    title: follow_text(follow.followable),
+                    class: "button hollow expanded" %>
     <% end %>
   </span>
 </div>
--- a/app/views/relationable/_score.html.erb
+++ b/app/views/relationable/_score.html.erb
@@ -1,15 +1,15 @@
 <small><%= t("related_content.is_related") %></small>
 
-<span class="relate-content-score">
-  <%= link_to t("related_content.score_positive"),
-              score_positive_related_content_path(related),
-              method: :put,
-              remote: true,
-              class: "score-positive" %>
+<div class="relate-content-score">
+  <%= button_to t("related_content.score_positive"),
+                score_positive_related_content_path(related),
+                method: :put,
+                remote: true,
+                class: "score-positive" %>
 
-  <%= link_to t("related_content.score_negative"),
-              score_negative_related_content_path(related),
-              method: :put,
-              remote: true,
-              class: "score-negative" %>
-</span>
+  <%= button_to t("related_content.score_negative"),
+                score_negative_related_content_path(related),
+                method: :put,
+                remote: true,
+                class: "score-negative" %>
+</div>
--- a/app/views/shared/_flag_actions.html.erb
+++ b/app/views/shared/_flag_actions.html.erb
@@ -8,7 +8,8 @@
       <span class="icon-flag flag-disable"></span>
     </button>
     <span class="dropdown-pane" id="flag-drop-<%= dom_id(flaggable) %>" data-dropdown data-auto-focus="true">
-      <%= link_to t("shared.flag"), polymorphic_path(flaggable, action: :flag), method: :put, remote: true %>
+      <%= button_to t("shared.flag"), polymorphic_path(flaggable, action: :flag),
+                    method: :put, remote: true %>
     </span>
   <% end %>
 
@@ -21,7 +22,8 @@
       <span class="icon-flag flag-active"></span>
     </button>
     <span class="dropdown-pane" id="unflag-drop-<%= dom_id(flaggable) %>" data-dropdown data-auto-focus="true">
-      <%= link_to t("shared.unflag"), polymorphic_path(flaggable, action: :unflag), method: :put, remote: true %>
+      <%= button_to t("shared.unflag"), polymorphic_path(flaggable, action: :unflag),
+                    method: :put, remote: true %>
     </span>
   <% end %>
 </span>
--- a/app/views/users/registrations/finish_signup.html.erb
+++ b/app/views/users/registrations/finish_signup.html.erb
@@ -20,7 +20,7 @@
   <% end %>
 
   <%= f.submit t("devise_views.users.registrations.new.submit"), class: "button expanded" %>
-  <div class="text-center">
-    <%= link_to t("devise_views.users.registrations.new.cancel"), destroy_user_session_path, class: "delete", method: :delete %>
-  </div>
 <% end %>
+<div class="text-center">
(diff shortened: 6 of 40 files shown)


javierm:
Hi, @cyrillefr :smile:.

Once again, thank you very much for taking care of this issue! :tada: And thank you for rebasing it so the tests are in the commit where its associated code changes :pray:.

Do you mind if we wait until you've finished replacing all the links mentioned on issue #5461 before reviewing the pull request? That way it'd probably be easier to catch hypothetical inconsistencies :thinking:.

spec/components/devise/omniauth_form_component_spec.rb
```diff
@@ -22,35 +22,31 @@
 
       render_inline component
 
-      expect(page).to have_link "Twitter"
-      expect(page).to have_link count: 1
```
javierm, line 26: Instead of removing this line, we could replace it with:

```ruby
      expect(page).to have_button count: 1
```

This way we still test that the rest of the buttons aren't rendered in this case.

Same in the other tests of this file :wink:.

cyrillefr: Oki, I just added that.

cyrillefr:
Hello @javierm ,

No problem.
I will add all links replacements (one per commit) in this PR.

cyrillefr:
Hello @javierm , 

I have proceeded on every TODO listed on #5461.

The reason I have modified the css sometimes is to keep the new button styling the same as the previous link.
Sometimes, the new button appeared just like the previous link and so there was nothing to do.

spec/system/dashboard/dashboard_spec.rb
```diff
@@ -10,15 +10,15 @@
...
     expect(page).to have_link("E-mail")
     expect(page).to have_link("Poster")
   end
 
   scenario "Publish link dissapears after proposal's publication" do
     visit proposal_dashboard_path(proposal)
-    click_link "Publish proposal"
+    click_button "Publish proposal"
 
     expect(page).not_to have_link("Publish proposal")
```
taitus: In the commit “Replace link with button in Dashboard show”, maybe we should update this line:
```suggestion
    expect(page).not_to have_button("Publish proposal")
```

app/views/devise/shared/_links.html.erb
```diff
@@ -13,7 +13,7 @@
 
   <%- if devise_mapping.omniauthable? && devise_mapping.name == "user" %>
     <%- resource_class.omniauth_providers.each do |provider| %>
-      <%= link_to t("devise_views.shared.links.signin_with_provider", provider: provider.to_s.titleize), omniauth_authorize_path(resource_name, provider), method: :post %><br>
+      <%= button_to t("devise_views.shared.links.signin_with_provider", provider: provider.to_s.titleize), omniauth_authorize_path(resource_name, provider), class: "button-as-link", form_class: "button_to inline" %><br>
     <% end -%>
   <% end -%>
```
taitus: In the commit “Replace link with button in devise shared links”, the following block was updated:

```  
  <%- if devise_mapping.omniauthable? && devise_mapping.name == "user" %>
    <%- resource_class.omniauth_providers.each do |provider| %>
      <%= button_to t("devise_views.shared.links.signin_with_provider", provider: provider.to_s.titleize), omniauth_authorize_path(resource_name, provider), class: "button-as-link", form_class: "button_to inline" %><br>
    <% end -%>
  <% end -%>
  ```  
  
 However, this condition will never be true because devise_mapping.name returns a symbol (:user), not a string ("user").
 
Since these links already appear at the very top of the page and the condition is always false, maybe we could consider removing the entire block, along with the CSS changes added in the same commit? 🤔

File: app/views/follows/_follow_button.html.erb [hunk 4]
```diff
@@ -8,12 +8,12 @@
...
-                               followable_type: follow.followable.class.name),
-                  method: :post, remote: true,
-                  title: follow_text(follow.followable),
-                  class: "button hollow expanded" %>
+      <%= button_to follow_text(follow.followable),
+                    follows_path(followable_id: follow.followable.id,
+                                 followable_type: follow.followable.class.name),
+                    remote: true,
+                    title: follow_text(follow.followable),
+                    class: "button hollow expanded" %>
```
taitus, line 16: In the commit “Replace link with buttons in follow button”, I was wondering if would it make sense to apply the same change to the “unfollow” link as well? 🤔 

It’s in the same file:

```
<%= link_to t("shared.following"),
            follow_path(follow),
            method: :delete, remote: true,
            title: unfollow_text(follow.followable),
            class: "button expanded" %>
```

spec/system/notifications_spec.rb
```diff
@@ -88,8 +88,9 @@
     click_link "Read"
 
     expect(page).to have_css(".notification", count: 1)
+    # binding.irb
```
taitus: In the commit “Replace link with button in notifications”, I noticed there’s a leftover `# binding.irb` comment.

If it’s no longer needed, it’d be great to remove it 🙏

app/assets/stylesheets/layout.scss
```diff
@@ -1190,9 +1207,17 @@ form {
...
+      cursor: pointer;
       position: absolute;
       right: 12px;
       top: 12px;
+
+      &:hover,
+      &:active,
+      &:focus {
+        @include anchor-color-hover;
+      }
```
taitus, line 1217: In the commit “Replace link with button in notifications”, it looks like the following CSS was added. Would it make sense to replace this with just: `@include link;`? 

```suggestion
    .mark-notification {
      @include link;
      position: absolute;
      right: 12px;
      top: 12px;
```

app/views/relationable/_score.html.erb
```diff
@@ -1,15 +1,17 @@
...
-              score_positive_related_content_path(related),
-              method: :put,
-              remote: true,
-              class: "score-positive" %>
+  <%= button_to t("related_content.score_positive"),
+                score_positive_related_content_path(related),
+                method: :put,
+                remote: true,
+                class: "score-positive",
+                form_class: "button_to inline" %>
```
taitus: In the commit “Replace link with button in relationable scores”, the `button_to` was added inside a `<span>`:

```
<span class="relate-content-score">
  <%= button_to ... %>
</span>
```

Since `button_to` generates a `<form>`, this results in invalid HTML (`<form>` inside `<span>`).

What do you think about replacing the `<span>` with a `<div class="relate-content-score">`? 🤔 

Also, I noticed this was added to the `button_to`:

`form_class: "button_to inline"`

Since we don’t have a `.button_to` class defined, and we already have a scoped `.relate-content-score` class in `layout.scss`, maybe we could move the styling there?

For example:

```
.relate-content-score {
  ...
  ...
  form {
    display: inline;
  }
}
```
and remove the line: `form_class: "button_to inline"` 🤔

app/assets/stylesheets/layout.scss
```diff
@@ -1803,7 +1828,8 @@ table {
   color: $delete;
 }
 
-.dropdown-pane a {
+.dropdown-pane a,
+button {
```
taitus: In the commit “Replace link with button in flag unflag actions”, the following CSS was added:
```
.dropdown-pane a,
button {
```

This would affect all buttons across the application, not just the ones inside .dropdown-pane. Would it make sense to scope it like this instead? 🤔 
```suggestion
.dropdown-pane {
  a,
  button {
```

app/views/shared/_flag_actions.html.erb
```diff
@@ -4,11 +4,12 @@
       <span class="divider">&nbsp;|&nbsp;</span>
     <% end %>
 
-    <button type="button" data-toggle="flag-drop-<%= dom_id(flaggable) %>" title="<%= t("shared.flag") %>">
+    <button type="button" id="upper-flag-button" data-toggle="flag-drop-<%= dom_id(flaggable) %>" title="<%= t("shared.flag") %>">
```
taitus: In the commit “Replace link with button in flag unflag actions”, some `ids` were introduced:

```
id="upper-flag-button"
id: "lower-flag-button"
id="upper-unflag-button"
id: "lower-unflag-button"
```

It seems these were added to simplify the tests and avoid relying on repeated button labels like:
```
click_button "Flag as inappropriate"
click_button "Flag as inappropriate"
```

However, what do you think about not adding the ids in this commit, and instead updating the tests like this?
```
scenario "Flagging as inappropriate" do
  login_as(user)
  visit path

  within "##{dom_id(flaggable)} .flag-content" do
    click_button "Flag as inappropriate"
    within("form") { click_button "Flag as inappropriate" }

    expect(page).to have_button "Unflag"
    expect(page).to have_button "Unflag", visible: :hidden
    expect(page).not_to have_button "Flag as inappropriate", visible: :all
  end

  refresh

  within "##{dom_id(flaggable)} .flag-content" do
    expect(page).to have_button "Unflag", visible: :hidden
    expect(page).not_to have_button "Flag as inappropriate", visible: :all
  end
end
```

Let me know what you think!
Or if you prefer, I can go ahead and update the remaining tests in this commit myself 🙌

app/views/users/registrations/finish_signup.html.erb
```diff
@@ -20,7 +20,7 @@
   <% end %>
 
   <%= f.submit t("devise_views.users.registrations.new.submit"), class: "button expanded" %>
-  <div class="text-center">
-    <%= link_to t("devise_views.users.registrations.new.cancel"), destroy_user_session_path, class: "delete", method: :delete %>
-  </div>
 <% end %>
+<div class="text-center">
+  <%= button_to t("devise_views.users.registrations.new.cancel"), destroy_user_session_path, class: "delete", method: :delete %>
```
taitus: In the commit “Replace link with button in finish signup in registrations”, it looks like the delete class was modified to style the cancel button.

What do you think about avoiding changes to the existing `delete` class, and instead introducing a more specific class like cancel-button?
```suggestion
  <%= button_to t("devise_views.users.registrations.new.cancel"), destroy_user_session_path, class: "delete cancel-button", method: :delete %>
```

That way, we could scope its styles directly under `.auth-page`, like this:

```
// 05. Auth pages
// --------------

.auth-page {

  .footer {
    margin-top: 0;
  }

  .cancel-button {
    cursor: pointer;

    &:hover,
    &:active,
    &:focus {
      text-decoration: underline;
    }
  }
}
```

app/views/verification/letter/new.html.erb
```diff
@@ -42,7 +42,7 @@
           <%= t("verification.letter.new.office") %>
         </li>
         <li>
-          <%= link_to t("verification.letter.new.send_letter"), letter_path, method: :post %>
+          <%= button_to t("verification.letter.new.send_letter"), letter_path, method: :post %>
```
taitus: In the commit “Replace link with button in letter verification”, I think we could remove the method: :post from the button without affecting behaviour 🤔

Since button_to uses POST by default, this could simplify the code a bit.
```suggestion
          <%= button_to t("verification.letter.new.send_letter"), letter_path %>
```

taitus:
Thanks @cyrillefr,
 
Yes, exactly. I think rewriting the history via rebase is totally fine in this case.

I really appreciate you taking the time to apply the suggested changes 😊.

Let me know if you run into any issues 🙌

spec/system/dashboard/dashboard_spec.rb
cyrillefr: Yes, obviously, testing that the 'publish' -link- after having pressed what it is now a 'publish' -button- is a bad (outdated since changes) test.
To be updated with suggested change.

app/views/devise/shared/_links.html.erb
cyrillefr: If I remember well for this one, I did not find any tests and I modified a bit that part to get it to be displayed and tested it on dev. I saw afterwars on followinf PR there were Twitter, FB etc. auth methods.
So, I trust you in that it is not used and deleted the block (also restored the CSS file to its previous state).

app/views/follows/_follow_button.html.erb
cyrillefr: I simply forgot the 2nd part in this file :( So I amended the commit with the 'Following' button

spec/system/notifications_spec.rb
cyrillefr: Argh. I deleted it.

app/assets/stylesheets/layout.scss
cyrillefr: I have stuck to the original style in all the commits.

In the original, a link_to + style did not draw an underline when hovered (I don't know how though).
But since there are now buttons and no more links, it makes sense not to have underline displayed.

If `@include link` here, an underline will be displayed (with a button) and it will not stick to the original style.

What do you think ?

taitus: You're absolutely right, good catch! 👌  I hadn't noticed the underline was added with @include link. That said, I'd still lean toward keeping the mixin, even if it introduces an underline that wasn't in the original 🤔. Mainly because it simplifies the SCSS and, visually, the button still behaves like a link, so the underline doesn't feel out of place. What do you think?

app/views/relationable/_score.html.erb
cyrillefr: I have changed this part in this way.

app/assets/stylesheets/layout.scss
cyrillefr: Yes, I forgot the curly brace ... fixed

app/views/shared/_flag_actions.html.erb
cyrillefr: I have updated this part following your advice

app/views/users/registrations/finish_signup.html.erb
cyrillefr: Yes, it is better, sorry I did not see that. I modified this way.

app/views/verification/letter/new.html.erb
cyrillefr: Oups, corrected.

cyrillefr:
Hello @taitus ,
I just pushed all modifications.
There is an error related to correction asked for 'Replace link with button in devise shared links'
 cf. https://github.com/consuldemocracy/consuldemocracy/pull/5955#discussion_r2158715957
Deleting the part the key `signin_with_provider` raise an error in `i18n_spec.rb` spec related to an unused key.
So, I deleted it just in the english locale (to get the specs pass).
But, if this is correct and this key is not used, then every key should be removed (in another PR ?).

cyrillefr:
Hello @taitus ,

Just red your last comment and updated the mentioned commit accordingly regarding the `es` translation.

Note that the tool that checks for labels usage raises an error only with the `en` key. I don't know if one can add the `es` key, but it would be great for some similar next use case.

PR #5966, opened 2025-04-14
Vendor foundation_rails_helper/form_builder.rb to drop gem dependency

taitus:
## References

Related discussion: #5494 

## Objectives

The foundation_rails_helper gem is no longer maintained and not compatible with Rails 7.1. We've vendored the FormBuilder class to avoid blocking the upgrade.

## Notes

This copy is under the MIT license and retains original attribution. Safe to remove once legacy Foundation CSS is deprecated.


vendor/foundation_rails_helper/form_builder.rb
```diff
@@ -0,0 +1,245 @@
+# frozen_string_literal: true
+
+# Copyright (c) 2015 Sébastien Gruhier (http://xilinus.com/)
+# MIT License
+# Vendored copy from https://github.com/sgruhier/foundation_rails_helper/blob/master/lib/foundation_rails_helper/form_builder.rb
+#
+# This file was vendored to work around the Rails 7.1 incompatibility,
+# since the `foundation_rails_helper` gem is no longer maintained.
+#
+# Safe to remove once we drop legacy Foundation CSS support.
```
javierm, line 10: I think this isn't really true :thinking:. The `foundation_rails_helper` name is misleading in our case because the funcionality we use is not related to Foundation.

```suggestion
```

vendor/foundation_rails_helper/form_builder.rb
```diff
@@ -0,0 +1,245 @@
+# frozen_string_literal: true
+
+# Copyright (c) 2015 Sébastien Gruhier (http://xilinus.com/)
+# MIT License
```
javierm, line 4: While these lines are OK, the license says:

> The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.

So what do you think about copying the LICENSE file from the foundation_rails_helper project to `vendor/foundation_rails_helper/`?

javierm (approved):

PR #5967, opened 2025-04-15
Update readme

lucialuzuriaga:
## Objectives

The installation guide varies depending on the user's operating system. While there are specific guidelines, there is no reference to them in the README nor in the local_installation.md file itself, making it easy to overlook the appropriate guide. The purpose of this change is to provide additional references in the main installation guides for a smoother installation process.

### DIFF
--- a/README_ES.md
+++ b/README_ES.md
@@ -32,7 +32,8 @@ Puedes acceder a la página principal del proyecto en [http://consuldemocracy.or
 
 ## Configuración para desarrollo y tests
 
-**NOTA**: para unas instrucciones más detalladas, consulta la [documentación de instalación local](docs/es/installation/local_installation.md).
+**NOTA**:
+El proceso de instalación varía según el sistema operativo. Por favor, consulta la [documentación de instalación local](docs/es/installation/local_installation.md) apropiada para tu SO.
 
 Prerrequisitos: tener instalado git, Ruby 3.3.8, CMake, pkg-config, Node.js 20.19.2, ImageMagick y PostgreSQL (9.5 o superior).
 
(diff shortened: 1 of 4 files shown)


README_ES.md
```diff
@@ -32,7 +32,9 @@ Puedes acceder a la página principal del proyecto en [http://consuldemocracy.or
 
 ## Configuración para desarrollo y tests
 
-**NOTA**: para unas instrucciones más detalladas, consulta la [documentación de instalación local](docs/es/installation/local_installation.md).
+**NOTA**:
+
+El proceso de instalación varía según el sistema operativo. Por favor, consulta la [documentación de instalación local](docs/es/installation/local_installation.md) apropiada para tu SO.
```
taitus: Right now, the English version has the note text on the same line as **NOTE**:, while the Spanish one adds a line break after **NOTA**:.

Could we use the same formatting in both cases?

docs/en/installation/local_installation.md
```diff
@@ -2,6 +2,13 @@
 
 Before installing Consul Democracy and having it up and running make sure you have all [prerequisites](prerequisites.md) installed.
 
+The installation process will vary depending on your operative system:
+
+- [Ubuntu Linux](docs/en/installation/ubuntu.md)
+- [Debian Linux](docs/en/installation/debian.md)
+- [macOS](docs/en/installation/macos.md)
+- [Windows](docs/en/installation/windows.md)
```
taitus: The links in this section point to `docs/en/installation/ubuntu.md` and similar paths, but since we’re already inside `docs/en/installation/local_installation.md`, GitHub tries to resolve them as nested paths like `docs/en/installation/docs/en/installation/ubuntu.md`, which don’t exist.

I think we should update the links to be relative to the current file, like this:

```suggestion
- [Ubuntu Linux](ubuntu.md)
- [Debian Linux](debian.md)
- [macOS](macos.md)
- [Windows](windows.md)
```

Same in the Spanish version.

PR #5989, opened 2025-05-14
Make nested documentable specs faster

taitus:
## References

## Objectives

As we did in PR #5369 and #5797, we reduce the number of tests we are going to run. In this case we remove two shared example files and create a new file in system folder to run the same tests but with a different factory in each execution for each test.

As they are the same tests and based on the same functionality, we use this related PR technique (random factory) to make sure that everything still works.

Shared examples removed:
- spec/shared/system/documentable.rb
- spec/shared/system/nested_documentable.rb


spec/system/nested_documentable_spec.rb
```diff
@@ -0,0 +1,270 @@
...
+
+  let(:factory) { factories.sample }
+  let!(:documentable) { create(factory) }
+  let!(:user) { create(:user, :level_two) }
+  let(:path) { attachable_path_for(factory, documentable) }
+  let(:submit_button_text) { submit_button_text_for(factory, path) }
+  let(:notice_text) { notice_text_for(factory, path) }
+
+  context "New and edit path" do
+    describe "When allow attached documents setting is enabled" do
```
javierm: Right now, we've got "When allow attached documents setting is enabled", "Only for edit path" and "When allow attached documents setting is disabled" blocks :thinking:. I find the approach in `shared/system/nested_documentable` a bit easier to follow.

My suggestion would be to move the "Only for edit path" block outside "New and edit path", and remove the "When allow attached documents setting is enabled" line. That means tests in the "New and edit path" block wouldn't have any additional contexts/descriptions except the ones disabling attached documents. This way we could also remove the duplication in the `before` blocks :thinking:.

spec/system/nested_documentable_spec.rb
```diff
@@ -0,0 +1,270 @@
...
+      before do
+        Setting["uploads.documents.max_amount"] = 1
+        create(:administrator, user: user) if admin_section?(path)
+        documentable.update!(author: user) if edit_path?(path)
+        do_login_for(user, management: management_section?(path))
+        visit path
```
javierm: Having a `visit` (or calling `login_as_manager`) inside a `before` block makes it hard to find cases where  we might create data after a `visit` :thinking:. I know it's the same as `nested_imageable`, but we did that one before having to spend dozens of hours dealing with flaky tests :sweat_smile:.

What do you think about keeping the `visit` inside each test, as it was in the `shared/system_nested_documentable` file? And do you think it makes sense to change the `nested_imageable` tests so they don't have a `visit` in a `before` block?

spec/system/nested_documentable_spec.rb
```diff
@@ -0,0 +1,270 @@
...
+    end
+
+    describe "When allow attached documents setting is disabled" do
+      before do
+        Setting["feature.allow_attached_documents"] = false
+        create(:administrator, user: user) if admin_section?(path)
+        create(:document, documentable: documentable)
+        documentable.update!(author: user) if edit_path?(path)
+        do_login_for(user, management: management_section?(path))
+        visit path
```
javierm: I know it only affects one test, but, for the same reasons mentioned earlier, what do you think about moving these lines inside the `scenario` block?

spec/support/common_actions/attachables.rb
```diff
@@ -0,0 +1,112 @@
+module Attachables
+  def imageable_attach_new_file(path, success = true)
```
javierm: Since we're changing this method, shall we use a named parameter here? At some point, I'd love to add the [Style/OptionalBooleanParameter](https://docs.rubocop.org/rubocop/cops_style.html#styleoptionalbooleanparameter) rubocop rule, which we follow most of the time :thinking:.

We'd only have to change the one place that calls this method with `false` as an argument.

```suggestion
  def imageable_attach_new_file(path, success: true)
  ```
```

spec/support/common_actions/attachables.rb
```diff
@@ -0,0 +1,112 @@
...
+    attach_new_file("Add image", "nested-image", "image", "Choose image", path, success)
+  end
+
+  def documentable_attach_new_file(path, success = true)
```
javierm: Since we're changing this method, shall we use a named parameter here? At some point, I'd love to add the [Style/OptionalBooleanParameter](https://docs.rubocop.org/rubocop/cops_style.html#styleoptionalbooleanparameter) rubocop rule, which we follow most of the time :thinking:.

We'd only have to change the one place that calls this method with `false` as an argument.

```suggestion
  def documentable_attach_new_file(path, success: true)
```

spec/system/nested_documentable_spec.rb
```diff
@@ -0,0 +1,270 @@
...
+  let!(:documentable) { create(factory) }
+  let!(:user) { create(:user, :level_two) }
+  let(:path) { attachable_path_for(factory, documentable) }
+  let(:submit_button_text) { submit_button_text_for(factory, path) }
+  let(:notice_text) { notice_text_for(factory, path) }
+
+  context "New and edit path" do
+    describe "When allow attached documents setting is enabled" do
+      before do
+        Setting["uploads.documents.max_amount"] = 1
```
javierm: Does this change mean we're no longer testing the cases where multiple documents are attached? :thinking: If so, I'd vote to revert the commit introducing it.

spec/system/nested_documentable_spec.rb
```diff
@@ -0,0 +1,270 @@
...
+
+        expect(page).to have_field("Title", with: "empty.pdf")
+      end
+
+      scenario "Should not update nested document file title with
+                file name after choosing a file when title already defined" do
+        click_link "Add new document"
+        within "#nested-documents" do
+          fill_in "Title", with: "My Title"
+          attach_file "Choose document", file_fixture("empty.pdf")
```
javierm: We're removed an expectation here because "These expectations are already covered by attach_new_file", but here we aren't using `attach_new_file` :thinking:. Should we bring the expectation back?

spec/system/nested_documentable_spec.rb
```diff
@@ -0,0 +1,270 @@
...
+      scenario "Should show successful notice when resource filled correctly and after valid file uploads" do
+        fill_in_required_fields(factory, path)
+
+        documentable_attach_new_file(file_fixture("empty.pdf"))
+        click_button submit_button_text
+
+        if factory != :dashboard_action
+          within "#documents" do
+            expect(page).to have_content "Documents (1)"
+            expect(page).to have_css "a[rel=nofollow]", text: "empty.pdf"
```
javierm: We used to check that the link pointed to a URL ending with `.pdf`, but we're no longer doing so :thinking:. Should we add the expectation back?

spec/system/nested_documentable_spec.rb
```diff
@@ -0,0 +1,270 @@
...
+        expect(page).to have_content notice_text
+      end
+
+      scenario "Should show successful notice when resource filled correctly and after valid file uploads" do
+        fill_in_required_fields(factory, path)
+
+        documentable_attach_new_file(file_fixture("empty.pdf"))
+        click_button submit_button_text
+
+        if factory != :dashboard_action
```
javierm: This would actually mean that the code inside this condition might not be executed during a test run :thinking:. I'm wondering whether we should keep the `let(:factory) { (factories - [:dashboard_action]).sample }` for this case :thinking:.

spec/system/nested_documentable_spec.rb
```diff
@@ -0,0 +1,270 @@
...
+  context "Show path" do
+    let(:factory) { (factories - [:dashboard_action]).sample }
+    let(:path) do
+      case factory
+      when :budget_investment
+        budget_investment_path(budget_id: documentable.budget_id, id: documentable.id)
+      when :proposal
+        proposal_path(id: documentable.id)
+      end
+    end
```
javierm:
```suggestion
    let(:path) { polymorphic_path(documentable) }
```

spec/system/nested_documentable_spec.rb
```diff
@@ -0,0 +1,270 @@
...
+        scenario "Should not be able when no user logged in" do
+          visit path
+
+          expect(page).not_to have_button "Delete document"
+        end
+
+        scenario "Should be able when documentable author is logged in" do
+          create(:document, documentable: documentable)
+          documentable.update!(author: user)
+          do_login_for(user, management: management_section?(path))
```
javierm: In this test, we're never in the management section, right? :thinking:

spec/system/nested_documentable_spec.rb
```diff
@@ -0,0 +1,270 @@
...
+        end
+
+        scenario "Administrators cannot destroy documentables they have not authored", :admin do
+          visit path
+
+          expect(page).not_to have_button "Delete document"
+        end
+
+        scenario "Users cannot destroy documentables they have not authored" do
+          do_login_for(user, management: management_section?(path))
```
javierm: In this test, we're never in the management section, right? :thinking:

spec/system/nested_documentable_spec.rb
```diff
@@ -0,0 +1,270 @@
...
+          expect(page).not_to have_button "Delete document"
+        end
+      end
+    end
+
+    context "Destroy" do
+      scenario "Should show success notice after successful document upload" do
+        create(:document, documentable: documentable)
+        documentable.update!(author: user)
+        do_login_for(user, management: management_section?(path))
```
javierm: In this test, we're never in the management section, right? :thinking:

spec/system/nested_imageable_spec.rb
```diff
@@ -1,60 +1,20 @@
 require "rails_helper"
 
 describe "Nested imageable" do
-  factories = [
-    :budget,
-    :budget_investment,
-    :future_poll_question_option,
-    :proposal
-  ]
+  factories = [:budget, :budget_investment, :future_poll_question_option, :proposal]
```
javierm: Any specific reason to use one line instead of multiple ones? :thinking:

spec/system/nested_documentable_spec.rb
taitus: Hi @javierm Thanks for the suggestion! I tried to group the before blocks like you said, but I had a problem.

In the "When allow attached documents setting is disabled" part, we need to set a different flag (Setting["feature.allow_attached_documents"] = false) and then run visit path after that.

So I kept "When allow attached documents setting is disabled" as a separate block, just to avoid repeating visit path in this block.

Maybe we could move the "Only for edit path" block outside like you suggested, but keep "When allow attached documents setting is enabled" grouped? 🤔 

Let me know if you think we could do it in a better way!

taitus: I just saw your [next comment](https://github.com/consuldemocracy/consuldemocracy/pull/5989#discussion_r2126412280), and I think it helps with this problem.

If we put visit path inside each test (like in shared/system/nested_documentable), then the issue I said before is not a problem anymore. So we can apply the change you suggested 👍  🙏

spec/system/nested_documentable_spec.rb
taitus: You're right, this test doesn't use `attach_new_file`, but it does replicate the same steps manually (click, attach_file).

The reason why I removed the expectation you mentioned is because I thought the only difference between using attach_new_file and this test is that here we fill in the title and I thought that filling in the title wasn't a good enough reason to keep the expectation.

Let me know if you'd prefer to keep it for extra clarity though!

spec/system/nested_documentable_spec.rb
taitus: If we use the factory you suggest, we'd stop testing the rest of the expectations for `:dashboard_action` 🤔

spec/system/nested_imageable_spec.rb
taitus: Good question.  Though I'm not sure if that's a question for me, or for the person who committed this over 6 months ago (my past self), or maybe for the one who reviewed and approved the PR? 😄

javierm: I've no idea what you're talking about :thinking:. Isn't this change part of the "Unify lets from documentable and imageable to attachables methods" commit, included in this pull request? :thinking:

spec/system/nested_documentable_spec.rb
javierm: The thing is that attaching the file generates an AJAX request, and now we're no longer checking whether that request has finished, which could result in another test handling the request :thinking:.

spec/system/nested_documentable_spec.rb
javierm: Great point! Hadn't realized that. Let's talk about it.

spec/support/common_actions/attachables.rb
```diff
@@ -0,0 +1,112 @@
...
+    fill_in_new_investment_title with: "Budget investment title"
+    fill_in_ckeditor "Description", with: "Budget investment description"
+    check :budget_investment_terms_of_service
+  end
+
+  def fill_proposal
+    fill_in_new_proposal_title with: "Proposal title"
+    fill_in "Proposal summary", with: "Proposal summary"
+    check :proposal_terms_of_service
+  end
```
javierm: We've got almost identical methods in `spec/support/common_actions/maps.rb` :thinking:. Shall we move them to `common_actions` or similar? :thinking: Also, right now we've got `fill_proposal` (here), `fill_in_proposal` (in `common_actions`) and `fill_in_proposal_form` (in maps) :joy:.

spec/system/nested_imageable_spec.rb
taitus: Sorry!!! My bad, I misunderstood the comment 🙏  and my blame totally failed 🤣

spec/system/nested_documentable_spec.rb
javierm: You're right! There are similar conditions in `nested_imageable`, so let's keep it like this :+1:.

spec/system/nested_documentable_spec.rb
```diff
@@ -0,0 +1,314 @@
...
+
+    scenario "Shows or hides new document link depending on max documents limit" do
+      do_login_for(user, management: management_section?(path))
+      visit path
+
+      expect(page).to have_link id: "new_document_link"
+
+      documentable_attach_new_file(file_fixture("empty.pdf"))
+
+      expect(page).not_to have_css "#new_document_link"
```
javierm: This test is now different :thinking:. Previously, we tested that the `#new_document_link` wasn't present after clicking on "Add new document" and before attaching the file. Now we test it isn't present after attaching the file. Is this intentional?

spec/system/nested_documentable_spec.rb
```diff
@@ -0,0 +1,314 @@
...
+
+    scenario "Should update nested document file title with
+              file name after choosing a file when no title defined" do
+      do_login_for(user, management: management_section?(path))
+      visit path
+
+      documentable_attach_new_file(file_fixture("empty.pdf"))
+
+      expect(page).to have_field("Title", with: "empty.pdf")
+    end
```
javierm: Shall we group these tests together? The only thing that changes is the expectation :thinking:.

spec/system/nested_documentable_spec.rb
```diff
@@ -0,0 +1,314 @@
...
+
+    scenario "Should show successful notice when resource filled correctly and after valid file uploads" do
+      Setting["uploads.documents.max_amount"] = 3
+      do_login_for(user, management: management_section?(path))
+      visit path
+
+      fill_in_required_fields(factory, path)
+
+      %w[clippy empty logo].take(documentable.class.max_documents_allowed).each do |filename|
```
javierm: Seems a bit strange to take 3 elements from an array with 3 elements :thinking:. I know it was in the original test, but the original test didn't have the `Setting["uploads.documents.max_amount"] = 3` line :thinking:.

```suggestion
      %w[clippy empty logo].each do |filename|
```

spec/system/nested_documentable_spec.rb
```diff
@@ -0,0 +1,314 @@
...
+        click_link "Add new document"
+
+        attach_file "Choose document", file_fixture("#{filename}.pdf")
+
+        within all(".document-fields").last do
+          expect(page).to have_css ".loading-bar.complete"
+        end
+      end
```
javierm, line 156: Maybe we could add an extra expectation here to check that there's no longer a button to add a document :thinking:. Since this is the only test with multiple uploads...

spec/system/nested_documentable_spec.rb
javierm: I see now there's a line saying: `expect(page).to have_link text: "empty.pdf"` :thinking:. This checks the text of the link, but previously we tested the URL :thinking:.

spec/system/nested_documentable_spec.rb
```diff
@@ -0,0 +1,314 @@
...
+    scenario "Should show persisted documents and remove nested_field" do
+      create(:document, documentable: proposal)
+      login_as user
+      visit edit_proposal_path(proposal)
+
+      expect(page).to have_css ".document-fields", count: 1
+
+      click_link "Remove document"
+
+      expect(page).not_to have_css ".document-fields"
```
javierm: What do you think about adding the `#new_document_link` expectations here, as part of the «Remove redundant tests from "Only for edit path" describe» commit? That means we'd need to add `Setting["uploads.documents.max_amount"] = 1` to this test or move it to a new `before` block at the beginning of the file.

```suggestion
      visit edit_proposal_path(proposal)

      expect(page).to have_css ".document-fields", count: 1
      expect(page).not_to have_link "Add new document"

      click_link "Remove document"

      expect(page).not_to have_css ".document-fields"
      expect(page).to have_link "Add new document"
```

spec/system/nested_documentable_spec.rb
```diff
@@ -0,0 +1,314 @@
...
+      Setting["uploads.documents.max_amount"] = 1
+      create(:administrator, user: user) if admin_section?(path)
+      documentable.update!(author: user) if edit_path?(path)
+    end
+
+    scenario "Shows or hides new document link depending on max documents limit" do
+      do_login_for(user, management: management_section?(path))
+      visit path
+
+      expect(page).to have_link id: "new_document_link"
```
javierm: Since we're at it:

```suggestion
      expect(page).to have_link "Add new document"
```

Same for the `not_to have_css` expectation below.

spec/system/nested_documentable_spec.rb
```diff
@@ -0,0 +1,314 @@
...
+        expect(reader.info[:Title]).not_to eq "logo_with_metadata.pdf"
+        expect(reader.info[:Producer]).not_to eq "Test Producer"
+        expect(reader.info).to eq({})
+      end
+    end
+
+    describe "When allow attached documents setting is disabled" do
+      before do
+        Setting["feature.allow_attached_documents"] = false
+        create(:document, documentable: documentable)
```
javierm: Note that, with this line, the "Add new document" link wouldn't be present even if the setting was enabled :thinking:. Maybe we should keep separate tests for the  "Add new document" link and the "#documents" element :thinking:. What do you think?

spec/system/nested_documentable_spec.rb
```diff
@@ -0,0 +1,314 @@
...
+          expect(page).to have_link text: "empty.pdf"
+          expect(page).to have_css "a[rel=nofollow]", text: "empty.pdf"
+          expect(page).to have_link text: "clippy.pdf"
+          expect(page).to have_css "a[rel=nofollow]", text: "clippy.pdf"
+          expect(page).to have_link text: "logo.pdf"
+          expect(page).to have_css "a[rel=nofollow]", text: "logo.pdf"
+          expect(page).not_to have_css "a[target=_blank]"
+        end
+      end
+      expect(page).to have_content notice_text
```
javierm: Shall we move this expectation before `if factory != :dashboard_action`? That way we make it clear that the expectations inside the `if` block are checked after the request to submit the form has finished.

spec/system/nested_documentable_spec.rb
```diff
@@ -0,0 +1,314 @@
...
+          expect(page).not_to have_button "Delete document"
+        end
+
+        scenario "Users cannot destroy documentables they have not authored" do
+          login_as(user)
+          visit path
+
+          expect(page).not_to have_button "Delete document"
+        end
+      end
```
javierm: No user interaction in these tests :thinking:. What do you think about moving them to component tests for the `Documents::DocumentComponent` class?

PR #5990, opened 2025-05-20
Regroup all tests related with imageable

taitus:
## References

Related PR: #5989 

## Objectives

Regroup in one unique file all tests related with imageable.


spec/system/nested_imageable_spec.rb
```diff
@@ -170,4 +170,17 @@
       expect(page).to have_css ".image-fields", count: 1, visible: :all
     end
   end
+
+  context "Show path" do
+    let(:factory) { (factories - [:proposal, :budget, :future_poll_question_option]).sample }
```
javierm, line 175: There's a test in budget investments checking "Index should show investment descriptive image only when is defined" and another one in proposals checking "Index should show proposal descriptive image only when is defined". Would it make sense to remove those tests and change this one accordingly? :thinking: Maybe we could also include budgets, since budgets can have associated images as well since a few years ago.

javierm:
```suggestion
    let(:factory) { :budget_investment }
```

Would a bit easier to follow IMHO :thinking:.

spec/system/nested_imageable_spec.rb
```diff
@@ -170,4 +170,17 @@
...
+    let(:path) { polymorphic_path(imageable) }
+    let!(:image) { create(:image, imageable: imageable) }
+
+    scenario "Show descriptive image and image title when exists" do
```
javierm: I think "when exists" is ambiguous :thinking:. What is the thing that exist? :thinking: The image? The title? Both?

javierm (approved):

PR #5991, opened 2025-05-21
Fix PDF poster generated in dashboard

taitus:
## Objectives
The proposal code text rendered correctly in the browser preview, but when generating the final PDF, it was left-aligned and hidden behind an image.

This PR fixes that issue by:
- Adding specific styles to center the text in the generated PDF.

## Visual Changes
Before
- Image showing the text hidden behind the image
Image showing the text hidden behind the image

After
- Image showing the centered and readable text 
Image showing the centered and readable text

### DIFF
--- a/app/assets/stylesheets/dashboard.scss
+++ b/app/assets/stylesheets/dashboard.scss
@@ -545,6 +545,10 @@
     }
   }
 
+  .proposal-code {
+    text-align: center;
+  }
+
   .poster-content {
     margin: 0 auto;
     max-width: 90%;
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/dashboard.scss
```diff
@@ -683,7 +683,8 @@
       }
 
       &.proposal-code {
-        font-size: 1.7cm;
+        font-size: 2cm;
```
javierm: Whether this style is the same as in the preview or not depends on your screen resolution and font size :thinking:. So, when in doubt, I'd vote to leave the original style as it was.

```suggestion
        font-size: 1.7cm;
```

javierm (approved):

PR #5994, opened 2025-05-30
Remove Ahoy ensure_uuid logic and uuidtools dependency

taitus:
## References
Related PR's: #4635 and #4971
Closes #5936 
  
## Objectives

Remove legacy Ahoy ensure_uuid logic from the Ahoy initializer, as it is no longer needed.
This allow remove the uuidtools gem dependency.

We originally added the `ensure_uuid` method and the `uuidtools gem` to handle
the conversion of Ahoy tokens (visit_token and visitor_token) into UUIDs,
following the [Ahoy 2.x upgrade guide](https://github.com/ankane/ahoy/blob/v2.0.0/docs/Ahoy-2-Upgrade.md).

However, since Ahoy v3.3.0, the gem generates valid UUIDs for tokens by default as we can see in the README under the [Token Generation](https://github.com/ankane/ahoy/blob/v3.3.0/README.md#token-generation) section.


javierm (commented):
@taitus Great catch! :tada:

Only a couple of comments regarding the commits:

1. Would it make sense to have just one commit instead of three? The second and third commit are only removing code that isn't used after the changes in the first one :thinking:. Maybe we could add something to the first commit message instead, like: "since we aren't manually generating UUIDs anymore, we no longer need the `uuidtools` dependency" :thinking:.
2. It looks like the token generation change is present in Ahoy since version 2.0.0, as seen in the [Ahoy 2.0.0 README](https://github.com/ankane/ahoy/blob/v2.0.0/README.md#token-generation), but the commit message mentions Ahoy 3.3.

PR #6040, opened 2025-07-09
Remove redundant tests

taitus:
## References

Pending

## Objectives

Pending

## Notes

Pending

### DIFF
--- a/spec/system/admin/budgets_wizard/budgets_spec.rb
+++ b/spec/system/admin/budgets_wizard/budgets_spec.rb
@@ -2,7 +2,7 @@ require "rails_helper"
 
 describe "Budgets wizard, first step", :admin do
   describe "New" do
-    scenario "Create budget - Knapsack voting (default)" do
+    scenario "Create budget is always create with Knapsack voting and in draft mode (default)" do
       visit admin_budgets_path
       click_button "Create new budget"
       click_link "Create multiple headings budget"
@@ -15,10 +15,14 @@ describe "Budgets wizard, first step", :admin do
 
       expect(page).to have_content "New participatory budget created successfully!"
 
-      click_link "Go back to edit budget"
+      within("#side_menu") { click_link "Participatory budgets" }
+      within("tr", text: "M30 - Summer campaign") { click_link "Edit" }
 
       expect(page).to have_field "Name", with: "M30 - Summer campaign"
       expect(page).to have_select "Final voting style", selected: "Knapsack"
+      expect(page).to have_content "This participatory budget is in draft mode"
+      expect(page).to have_link "Preview"
+      expect(page).to have_button "Publish budget"
     end
 
     scenario "Create budget - Approval voting" do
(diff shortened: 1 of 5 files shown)


spec/system/admin/budgets_wizard/budgets_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 describe "Budgets wizard, first step", :admin do
   describe "New" do
-    scenario "Create budget - Knapsack voting (default)" do
+    scenario "Create budget is always create with Knapsack voting and in draft mode (default)" do
```
javierm: There seems to be a typo here :thinking:. Maybe we could write something like:

```suggestion
    scenario "Create budget with Knapsack voting and in draft mode (default)" do
```

spec/system/admin/budgets_wizard/budgets_spec.rb
```diff
@@ -19,6 +19,13 @@
 
       expect(page).to have_field "Name", with: "M30 - Summer campaign"
       expect(page).to have_select "Final voting style", selected: "Knapsack"
+
+      within("#side_menu") { click_link "Participatory budgets" }
+      within("tr", text: "M30 - Summer campaign") { click_link "Edit" }
+
+      expect(page).to have_content "This participatory budget is in draft mode"
+      expect(page).to have_link "Preview"
+      expect(page).to have_button "Publish budget"
```
javierm: Judging from the commit message, it looks like the `click_link "Edit"` part is supposed to replace `click_link "Go back to edit budget"`, which would make sense IMHO. So then it would be something like:

```diff
diff --git a/spec/system/admin/budgets_wizard/budgets_spec.rb b/spec/system/admin/budgets_wizard/budgets_spec.rb
index 08131e74b6..a7c7230917 100644
--- a/spec/system/admin/budgets_wizard/budgets_spec.rb
+++ b/spec/system/admin/budgets_wizard/budgets_spec.rb
@@ -15,14 +15,11 @@ describe "Budgets wizard, first step", :admin do
 
       expect(page).to have_content "New participatory budget created successfully!"
 
-      click_link "Go back to edit budget"
-
-      expect(page).to have_field "Name", with: "M30 - Summer campaign"
-      expect(page).to have_select "Final voting style", selected: "Knapsack"
-
       within("#side_menu") { click_link "Participatory budgets" }
       within("tr", text: "M30 - Summer campaign") { click_link "Edit" }
 
+      expect(page).to have_field "Name", with: "M30 - Summer campaign"
+      expect(page).to have_select "Final voting style", selected: "Knapsack"
       expect(page).to have_content "This participatory budget is in draft mode"
       expect(page).to have_link "Preview"
       expect(page).to have_button "Publish budget"
```

javierm (commented):
@taitus I've left a couple of comment; let me know what you think! :wink:

spec/system/admin/budgets_wizard/budgets_spec.rb
javierm: Looks like my suggestion was completely wrong :rofl:. Thanks for noticing!

PR #6046, opened 2025-07-24
Add support for OIDC authentication

Anamika1608:
## Objectives

To integrate the OIDC Authentication Support

## Visual Representation

image

## Implementation video

https://github.com/user-attachments/assets/e9d63dbc-9629-49f4-98c8-c0c76d2bf950


javierm:
@Anamika1608 Looks like the linters are failing :thinking:. Could you have a look?

bin/setup
```diff
@@ -17,21 +17,13 @@ FileUtils.chdir APP_ROOT do
...
+  # unless File.exist?("config/database.yml")
+  #   FileUtils.cp "config/database.yml.sample", "config/database.yml"
+  # end
 
   puts "\n== Preparing database =="
   system! "bin/rails db:prepare"
-  system! "RAILS_ENV=test bin/rails db:prepare"
 
   puts "\n== Removing old logs and tempfiles =="
   system! "bin/rails log:clear tmp:clear"
```
javierm: Are these changes intentional? :thinking:

Anamika1608:
@javierm, I've fixed it. Can you have a look now?

app/assets/stylesheets/layout.scss
```diff
@@ -1453,6 +1454,16 @@ table {
   }
 }
 
+.button.button-oidc {
+  @include has-fa-icon(openid, brands);
+  background: #f2e8fc;
+  border-left: 3px solid #954ae0;
+
+  &::before {
+    color: #954ae0;
```
javierm: Looks like we can't use multiple colors in font awesome icons, but maybe we could use one of the official colors of the openID icon :thinking:. I've found some [open ID logo guidelines](https://openid.net/wp-content/uploads/2023/06/openid-logo-guidelines.pdf). For now, maybe we can use:

```suggestion
    color: #f7931e;
```

The problem of this color is that it doesn't offer the minimum contrast required for accessibility against a light background (see the [WebAIM contrast checker](https://webaim.org/resources/contrastchecker/)). Now, logos do not need to follow contrast requirements as far as I know, but we should still try to provide as much contrast as possible. So, for the background color (4 lines above this one), how about:

```css
  background: #fafafa;
```

config/initializers/devise.rb
```diff
@@ -296,6 +296,14 @@
   end
   config.omniauth :saml, saml_settings.merge(setup: ->(env) { OmniauthTenantSetup.saml(env) })
 
+  config.omniauth :openid_connect, name: :oidc, scope: [:openid, :email, :profile], response_type: :code,
+                                   issuer: Rails.application.secrets.oidc_issuer, discovery: true, client_auth_method: :basic,
```
javierm: In the commit that added these lines, could you explain why we're using these options? :pray:

app/lib/omniauth_tenant_setup.rb
```diff
@@ -21,6 +21,11 @@ def saml(env)
                 secrets.saml_idp_metadata_url, secrets.saml_idp_sso_service_url)
     end
 
+    def oidc(env)
```
javierm, line 24: We should add this method in the same commit we add the code calling it. Or the other way around: introduce the code calling this method in the same commit we add this method (like we did in the SAML pull request). Sometimes it's also OK to add the method in a commit before adding the code calling it.

The problem is that, if we add the method after a commit containing code calling it, the commit which introduced code calling it will be broken. Having broken commits can make debugging with `git bisect` harder.

spec/components/devise/omniauth_form_component_spec.rb
```diff
@@ -62,5 +63,14 @@
...
     end
+
+    it "renders the OIDC link when the feature is enabled" do
+      Setting["feature.oidc_login"] = true
+
+      render_inline component
+
+      expect(page).to have_button "OIDC"
+      expect(page).to have_button count: 1
+    end
```
javierm, line 74: Would it be possible to add system tests as well, similar to the ones for SAML?

bin/setup
Anamika1608: nope, these are not intentional, never mind, I've edited that commit code to have the prev changes.

app/assets/stylesheets/layout.scss
Anamika1608: I think, your suggestion is fine, I've made the background white and border and logo color to orange.

config/initializers/devise.rb
Anamika1608: sure,

* **`name: :oidc`** → Name of the login provider
* **`scope: [:openid, :email, :profile]`** → Ask the provider for user identity, email, and profile info.
* **`response_type: :code`** → Use secure **Authorization Code Flow** for token exchange.
* **`issuer: Rails.application.secrets.oidc_issuer`** → Base URL of OIDC provider (here I used auth0)
* **`discovery: true`** → Auto-fetch provider’s config instead of hardcoding endpoints.
* **`client_auth_method: :basic`** → Send client credentials via HTTP Basic Auth when getting tokens.

app/lib/omniauth_tenant_setup.rb
Anamika1608: yeah, i got it. I've moved into one commit now :)

config/initializers/devise.rb
```diff
@@ -296,6 +296,14 @@
...
+                                   client_options: {
+                                     identifier: Rails.application.secrets.oidc_client_id,
+                                     secret: Rails.application.secrets.oidc_client_secret,
+                                     redirect_uri: Rails.application.secrets.oidc_redirect_uri
+                                   }, setup: ->(env) { OmniauthTenantSetup.oidc(env) }
+
```
javierm: I see you changed these lines after the linter reported an issue. I think your original code was more readable because it used one line per option. So here's an alternative approach, which uses your original code but a different indentation:

```suggestion
  config.omniauth :openid_connect,
                  name: :oidc,
                  scope: [:openid, :email, :profile],
                  response_type: :code,
                  issuer: Rails.application.secrets.oidc_issuer,
                  discovery: true,
                  client_auth_method: :basic,
                  client_options: {
                    identifier: Rails.application.secrets.oidc_client_id,
                    secret: Rails.application.secrets.oidc_client_secret,
                    redirect_uri: Rails.application.secrets.oidc_redirect_uri
                  },
                  setup: ->(env) { OmniauthTenantSetup.oidc(env) }
```

config/locales/en/settings.yml
```diff
@@ -91,6 +91,8 @@ en:
       google_login_description: "Allow users to sign up with their Google Account"
       wordpress_login: "Wordpress login"
       wordpress_login_description: "Allow users to sign up with their Wordpress Account"
+      oidc_login: "OIDC login"
+      oidc_login_description: "Allow users to sign up with OIDC"
```
javierm: Since administrators might not know what OIDC stands for, maybe we could include the full name:

```suggestion
      oidc_login: " OpenID Connect login"
      oidc_login_description: "Allow users to sign up with OpenID Connect (OIDC)"
```

I'm wondering what to do about the button to sign in with OIDC :thinking:. I guess we can leave "OIDC" for now and change it later if necessary.

docs/en/features/oauth.md
```diff
@@ -21,6 +21,8 @@ user_wordpress_oauth2_omniauth_authorize GET|POST /users/auth/wordpress_oauth2(.
 user_wordpress_oauth2_omniauth_callback GET|POST /users/auth/wordpress_oauth2/callback(.:format) users/omniauth_callbacks#wordpress_oauth2
 user_saml_omniauth_authorize GET|POST /users/auth/saml(.:format) users/omniauth_callbacks#passthru
 user_saml_omniauth_callback GET|POST /users/auth/saml/callback(.:format) users/omniauth_callbacks#saml
+user_openid_connect_omniauth_authorize GET|POST /users/auth/openid_connect(.:format) users/omniauth_callbacks#passthru
+user_openid_connect_omniauth_callback GET|POST /users/auth/openid_connect/callback(.:format) users/omniauth_callbacks#openid_connect
```
javierm: Is there a typo here? :thinking: Here's what I get when running `rails routes`:

```suggestion
user_oidc_omniauth_authorize GET|POST /users/auth/oidc(.:format) users/omniauth_callbacks#passthru
user_oidc_omniauth_callback GET|POST /users/auth/oidc/callback(.:format) users/omniauth_callbacks#oidc
```

spec/system/users_auth_spec.rb
```diff
@@ -645,6 +645,204 @@
...
+          }
+        )
+
+        OmniAuth.config.add_mock(:oidc, expired_oidc_hash)
+
+        visit new_user_session_path
+        click_button "Sign in with OIDC"
+
+        expect(page).to have_content "Successfully identified as Oidc"
+        expect_to_be_signed_in
```
javierm, line 824: Could you tell me a bit about what the implications of an expired token are? I didn't expect someone to be able to sign in with an expired token :thinking:. Or maybe I'm not understanding this test :sweat_smile:. Instead of "handle gracefully", what do you think about changing the title of the test to describe the expected behavior?

javierm (commented):
@Anamika1608 Here's another review. Let me know what you think! :wink:

app/assets/stylesheets/layout.scss
```diff
@@ -1453,6 +1454,16 @@ table {
   }
 }
 
+.button.button-oidc {
+  @include has-fa-icon(openid, brands);
+  background: #fafafa;
```
javierm: Sorry, I know I'm the one who suggested this color :pray:, but I've just noticed that for all the other buttons we try to use similar colors (only lighter) for text and background. So that would mean:

```suggestion
  background: #fdf9f1
```

config/initializers/devise.rb
Anamika1608: thank you, I also liked this one, but since linter was failing so had to come up with that :/

but now after your suggestion, it's good

docs/en/features/oauth.md
Anamika1608: yeah it was typo :| but I've fixed it

app/assets/stylesheets/layout.scss
Anamika1608: yeah this looks good now :)

spec/system/users_auth_spec.rb
Anamika1608: An expired OIDC token means you can’t use it to call APIs anymore, but it doesn’t invalidate the fact that the user has been authenticated. This test is checking that app still lets the user sign in even if the access token is expired.

And also we can pick up any of the below title, let me know which one  

- signs user in despite expired OIDC token
- allows user authentication when OIDC token has expired

Anamika1608: I've made it 2nd one for now, let me know if that's fine.

javierm: Understood, thanks! I prefer "allows user authentication when OIDC token has expired". No strong preference, though.

Anamika1608: perfect, done

app/assets/stylesheets/layout.scss
```diff
@@ -1453,6 +1454,16 @@ table {
...
+  border-left: 3px solid #fdf9f1;
```
javierm: I think there's a typo here, right?

```suggestion
  background: #fdf9f1;
  border-left: 3px solid #f7931e;
```

Anamika1608: yeah, it was typo, sorry for that, i've fixed it now :)

PR #6061, opened 2025-08-06
WIP: Poll text answers

taitus:
# Before creating a pull request

Please make sure to read and follow the coding conventions shown in [CONTRIBUTING](https://github.com/consuldemocracy/consuldemocracy/blob/master/CONTRIBUTING.md)
For more information please view [Coding Conventions](https://github.com/consuldemocracy/consuldemocracy/blob/master/docs/en/open_source/coding_conventions.md)

**Here's a quick summary of some points to check**

* Include specs to test any changes you've made
* Our CI will check whether the rest of the application is still working properly; check its build and make sure all tests are passing
* Run `bundle exec pronto run` and fix any issues it reports; these issues will also be automatically reported on the pull request
* Follow [the seven rules of a great commit message](https://chris.beams.io/posts/git-commit/)
* Add or modify i18n translations only in the base languages, English (en) and Spanish (es); we manage all other languages through the [Crowdin integration](https://translate.consuldemocracy.org/).

## References

> Reference any related issues/Pull Requests/errors/etc... to your work.

## Objectives

> What are the objectives of these changes?

## Visual Changes

* Any visual changes? please attach screenshots (or gifs) showing them.
* If modified views are public (not the admin panel), try them in mobile display (with your browser's developer console) and add screenshots.

## Notes

* Mention rake tasks or actions to be done when deploying this changes to a server (if any).
* Explain any caveats, or important things to notice like deprecations (if any).

PR #6065, opened 2025-08-18
Allow voting for essay option in unique and multiple question

taitus:
## References

Based on and partially extracted from #5827 by [werdenktwas-gmbh](https://github.com/werdenktwas-gmbh)
The original PR implemented three features in one branch:
	1.	New "essay" question type (open text answers).
	2.	Open-text answer option for unique/multiple-choice questions.
	3.	Preliminary results view for admins.

This PR only addresses (2)

## Objectives
Allow voting for essay option in public section

## Visual Changes
- Option text option for unique and multiple questions
Option text option for unique and multiple questions

## Notes
Builds on  #6061


config/locales/es/admin.yml
```diff
@@ -1170,6 +1170,10 @@ es:
           documents_list: Lista de documentos
           document_title: Título
           document_actions: Acciones
+          open_text:
+            field: Texto libre
+            true: "Si"
```
javierm: Typo:

```suggestion
            true: "Sí"
```

app/components/polls/questions/question_component.rb
```diff
@@ -48,28 +48,41 @@ def multiple_choice_help_text
       )
     end
 
-    def multiple_choice_field(option)
-      choice_field(option) do
+    def choice_field(option)
+      label_tag("web_vote_option_#{option.id}") do
+        html = input_tag(option) + option.title
+        html += open_text_tag(option) if option.open_text?
+        html
```
javierm: According to the W3C validator:

> The label element may contain at most one button, input, meter, output, progress, select, or textarea descendant.

So it looks like we need to move the `open_text_tag` bit outside the `<label>` tag :thinking:.

app/components/polls/questions/question_component.rb
```diff
@@ -48,28 +48,41 @@ def multiple_choice_help_text
...
+        existing_text_for(option),
+        id: "web_vote_option_#{option.id}_text",
+        disabled: disabled?,
+        class: "open-text",
+        data: { selects: "web_vote_option_#{option.id}" }
+      )
+    end
+
+    def existing_text_for(option)
+      question.answers.find_by(author: current_user, option: option)&.answer.to_s
```
javierm: Right now the answer is lost when submitting an invalid form, so we probably have to use `form.object.answers` as we do in `existing_answer` :thinking:.

By the way, is there a difference between `existing_text_for` and `existing_answer`?

app/views/admin/poll/questions/options/_form.html.erb
```diff
@@ -18,6 +18,12 @@
     <% end %>
   </div>
 
+  <div class="row">
+    <div class="column">
+      <%= f.check_box :open_text, checked: @option.open_text? %>
```
javierm: What do you think about adding an extra checkbox when creating a question "allow open answers as options", and then only rendering this checkbox when that one was checked?

The idea is that, when creating a question, right now it isn't exactly clear what you need to do in order to add a question that is both open-ended and contains predefined options :thinking:.

javierm (commented):
This is just a very high-level review. I'm wondering whether it's better to release version 2.4.0 before finishing this pull request, and including this one in version 2.5.0 :thinking:.

app/views/admin/poll/questions/options/_form.html.erb
```diff
@@ -18,6 +18,15 @@
...
```
javierm: Do we need the `checked` attribute? 🤔


```suggestion
      <%= f.check_box :open_text %>
```

app/assets/stylesheets/admin/polls/questions/options/table.scss
```diff
@@ -17,6 +18,12 @@
         width: calc(100% * 7 / 12);
       }
     }
+
+    &.open-text {
+      @include breakpoint(medium) {
+        width: calc(100% * 2 / 12);
```
javierm, line 23: What's the purpose of this line? 🤔

app/assets/stylesheets/admin/polls/questions/options/table.scss
```diff
@@ -6,7 +6,8 @@
       vertical-align: top;
     }
 
-    &.description {
+    &.description,
+    &.open-text {
```
javierm: `td.open-text` is going to be either "Yes" or "No". Do we need the `overflow-wrap` rule in this case? 🤔

app/components/admin/poll/questions/options/table_component.html.erb
```diff
@@ -16,6 +17,7 @@
       <tr id="<%= dom_id(option) %>" class="poll_question_option" data-option-id="<%= option.id %>">
         <td><%= option.title %></td>
         <td class="description"><%= wysiwyg(option.description) %></td>
+        <td class="open-text"><%= t("admin.questions.show.answers.open_text.#{option.open_text}") %></td>
```
javierm: Would it make sense to use `shared.yes` and `shared.no`? 🤔

app/views/admin/poll/questions/options/_form.html.erb
```diff
@@ -18,6 +18,15 @@
...
+      <span class="help-text">
```
javierm: Shall we add an `ID` to this element, and an `aria-describedby` attribute to the checkbox?

config/locales/en/admin.yml
```diff
@@ -1171,6 +1171,11 @@ en:
           documents_list: Documents list
           document_title: Title
           document_actions: Actions
+          open_text:
+            field: Open text
+            true: "Yes"
+            false: "No"
+            help: "Allows users to write their own answer when selecting this option. Useful for options like 'Other (please specify)'."
```
javierm: I think double quotes are more common than single quotes in the public area 🤔.

```suggestion
            help: "Allows users to write their own answer when selecting this option. Useful for options like \"Other (please specify)\"."
```

Same in the Spanish text.

spec/system/admin/poll/questions/options/options_spec.rb
```diff
@@ -46,6 +46,27 @@
...
       expect("First").to appear_before("Second")
     end
+
+    scenario "Create option with open text enabled" do
+      question = create(:poll_question, poll: future_poll)
+
+      visit admin_question_path(question)
+      click_link "Add answer"
+
+      expect(page).to have_unchecked_field "Allow entering text as answer"
```
javierm: Not a strong prefence here, but I think we usually use `have_field (...) checked: false` instead of `have_unchecked_field` 🤔.

spec/system/admin/poll/questions/options/options_spec.rb
```diff
@@ -46,6 +46,27 @@
...
+
+      fill_in "Answer", with: "Other (please specify)"
+      check "Allow entering text as answer"
+
+      expect(page).to have_checked_field "Allow entering text as answer"
```
javierm: I think we can remove this line 🤔. We've just used `check`, so the checkbox is obviously checked 🤔.

app/components/polls/questions/question_component.rb
```diff
@@ -48,31 +52,51 @@ def multiple_choice_help_text
...
     end
 
-    def multiple_choice_field(option)
-      choice_field(option) do
+    def choice_field(option)
+      html = label_tag("web_vote_option_#{option.id}") do
+        input_tag(option) + option.title
+      end
+      html += open_text_tag(option) if option.allows_custom_text?
+      html
```
javierm: What do you think about something like this?

```suggestion
      label_with_input(option) + open_text_tag(option)
```

And extract methods accordingly.

app/assets/stylesheets/polls/form.scss
```diff
@@ -23,6 +23,7 @@
 
     label {
       @include radio-or-checkbox-and-label-alignment;
+      flex-wrap: wrap;
```
javierm: What's this line for? 🤔 This would allow the label text to appear below the radio button or checkbox, which usually makes it harder to understand that they're related.

app/assets/stylesheets/polls/form.scss
```diff
@@ -31,6 +32,11 @@
         white-space: pre;
       }
     }
+
+    .open-text {
+      margin-left: $line-height * 0.8;
```
javierm: Where does the 0.8 part come from? 🤔.

Other than that:

```suggestion
      margin-#{$global-left}: $line-height * 0.8;
```

app/components/polls/questions/question_component.rb
```diff
@@ -48,31 +52,51 @@ def multiple_choice_help_text
...
-    def choice_field(option, &block)
-      label_tag("web_vote_option_#{option.id}") do
-        block.call + option.title
-      end
+    def open_text_tag(option)
+      text_area_tag(
+        "web_vote[#{question.id}][answer][#{option.id}]",
+        existing_text_for(option),
+        id: "web_vote_option_#{option.id}_text",
+        disabled: disabled?,
```
javierm: Do we need this attribute? 🤔 We're already disabling the fieldset when `disabled?` is true, right?

app/models/poll/answer.rb
```diff
@@ -7,10 +7,19 @@ class Poll::Answer < ApplicationRecord
 
   validates :question, presence: true
   validates :author, presence: true
-  validates :answer, presence: true
+  validates :answer, presence: true, unless: ->(poll_answer) { poll_answer.option&.allows_custom_text? }
```
javierm: Not sure I understand this one 🤔. I'd kind of expect the opposite: optional if the option doesn't allow custom text 🤔.

app/models/poll/answer.rb
```diff
@@ -7,10 +7,19 @@ class Poll::Answer < ApplicationRecord
...
+  validates :answer, length: { maximum: ->(*) { Poll::Answer.answer_max_length }}
```
javierm, line 12: This also applies to generic open answers, right? If so, what do you think about adding this rule in a different commit? 🤔

app/models/poll/question/option.rb
```diff
@@ -51,6 +51,10 @@ def with_read_more?
     description.present? || images.any? || documents.any? || videos.any?
   end
 
+  def allows_custom_text?
+    open_text?
+  end
+
```
javierm: Any reasons to have two different names for the same thing? 🤔

config/locales/en/general.yml
```diff
@@ -646,6 +646,7 @@ en:
     description:
       multiple: "You can select a maximum of %{maximum} answers."
     read_more_about: "Read more about:"
+    open_text_aria_label: "Specify answer for %{option}"
```
javierm: Usually label texts don't start with a verb 🤔.

```suggestion
    open_text_aria_label: "Answer for %{option}"
```

Same in Spanish.

javierm (commented):
I've left some comments. Let me know what you think!

app/models/poll/answer.rb
taitus: This is intentional. `allows_custom_text?` means the additional text is allowed, not required. For open-text options we intentionally allow a blank custom answer, while regular options and generic open questions still require answer to be present. There's already a spec covering this behaviour (stores the selected open-text option even when custom text is blank).

javierm: Then maybe we shouldn't require an answer in regular options 🤔.

app/assets/stylesheets/admin/polls/questions/options/table.scss
taitus: It reserves some width for the new column on medium and larger screens. Without it, the "Open text" header wraps onto two lines. The 2/12 value follows the same proportional layout already used by the description column. If you have a better approach in mind, I'd be happy to change it.

app/components/polls/questions/question_component.rb
taitus: Do you mean something like this?

```
def choice_field(option)
  label_with_input(option) + open_text_tag(option)
end

def label_with_input(option)
  label_tag("web_vote_option_#{option.id}") do
    input_tag(option) + option.title
  end
end

def open_text_tag(option)
  return "".html_safe unless option.allows_custom_text?
  text_area_tag(...)
end
```

The only thing I'm not entirely convinced about is that, when reading `choice_field`, it looks as though `open_text_tag` is always rendered, while the conditional is actually hidden inside `open_text_tag` 🤔 .

Another option could be:

```
def choice_field(option)
  safe_join([
    option_label(option),
    open_text_tag(option) if option.allows_custom_text?
  ].compact)
end

def option_label(option)
  label_tag("web_vote_option_#{option.id}") do
    input_tag(option) + option.title
  end
end
```

Do you have a preference? I'm not entirely sure which approach would be better here.

app/assets/stylesheets/polls/form.scss
taitus: It was chosen to align the textarea with the option text rather than with the checkbox. Using the full `$line-height` moves it slightly too far to the right, while `0.8 * $line-height` matches the current checkbox and label spacing more closely.

Can you think of a better way to express this without using 0.8?

Although this does not remove the magic number, perhaps giving the value a descriptive name would make the intention clearer 🤔 :
```
$option-text-indent: $line-height * 0.8;

.open-text {
  display: block;
  margin-#{$global-left}: $option-text-indent;
  width: calc(100% - #{$option-text-indent});
}
```

What do you think?

app/models/poll/question/option.rb
taitus: My intention was to make the conditions easier to read. `allows_custom_text?` felt a bit more expressive of the intended behavior than `open_text?`.

But if you don't think it adds any clarity, I don't have any problem simplifying it and using `open_text?` everywhere.

spec/models/poll/answer_spec.rb
```diff
@@ -25,18 +25,33 @@
       expect(answer).not_to be_valid
     end
 
-    it "is not valid without an answer" do
+    it "is not valid without an answer when question is open-ended" do
+      answer.question = create(:poll_question_open)
+      answer.option = nil
```
javierm, line 30: Do we need this line? 🤔

spec/models/poll/answer_spec.rb
```diff
@@ -25,18 +25,33 @@
...
+
       expect(answer).not_to be_valid
     end
 
-    it "is not valid without an answer when question is open-ended" do
-      answer.question = create(:poll_question_open)
-      answer.answer = nil
+    it "is valid without answer text when an option is present" do
+      expect(answer).to be_valid
+      expect(answer.answer).to be nil
```
javierm: I know this works, but maybe we could write something similar to what the test description says 🤔. That way we won't have to check what the factory does in order to understand the test.

```suggestion
      answer.option = answer.question.options.sample
      answer.answer = nil
      
      expect(answer).to be_valid
```

spec/models/poll/partial_result_spec.rb
```diff
@@ -2,18 +2,35 @@
...
-      option = create(:poll_question_option, title_en: "One", title_es: "Uno", question: question)
+    let(:partial_result) { build(:poll_partial_result) }
+
+    it "is valid" do
+      expect(partial_result).to be_valid
+    end
+
+    it "is valid without answer text when an option is present" do
+      expect(partial_result).to be_valid
+      expect(partial_result.answer).to be nil
```
javierm: I know this works, but maybe we could write something similar to what the test description says 🤔. That way we won't have to check what the factory does in order to understand the test.

app/views/admin/poll/questions/options/_form.html.erb
```diff
@@ -18,6 +18,16 @@
...
+  <div class="row">
+    <div class="column">
+      <% allows_custom_text_help_id = f.field_id(:allows_custom_text, :help) %>
+      <%= f.check_box :allows_custom_text, aria: { describedby: allows_custom_text_help_id } %>
+      <span class="help-text" id="<%= allows_custom_text_help_id %>">
+        <%= t("admin.questions.show.answers.allows_custom_text.help") %>
+      </span>
+    </div>
+  </div>
+
```
javierm: Shall we take this chance to move this view to a component? Then we can extract `allows_custom_text_help_id` to a method and move the `help_id` part of the system test to a component.

app/components/admin/poll/questions/options/table_component.html.erb
```diff
@@ -4,6 +4,7 @@
     <tr>
       <th><%= t("admin.questions.show.answers.title") %></th>
       <th scope="col" class="description"><%= t("admin.questions.show.answers.description") %></th>
+      <th scope="col" class="open-text"><%= t("admin.questions.show.answers.allows_custom_text.field") %></th>
```
javierm: To avoid duplication in the translation files:

```suggestion
      <th scope="col" class="open-text"><%= Poll::Question::Option.human_attribute_name(:allows_custom_text)
      %></th>
```

Then we can remove the translations for `admin.questions.show.answers.allows_custom_text.field`.

spec/models/poll/question/option_spec.rb
```diff
@@ -139,4 +144,18 @@
       end
     end
   end
+
+  describe "#allows_custom_text" do
+    it "defaults to false" do
+      option = create(:poll_question_option)
```
javierm: Here we might be testing the model or the factory 🤔. To make sure we're testing the model:

```suggestion
      option = Poll::Question::Option.new
```

javierm (commented):
@taitus More cowbell, I mean, comments!

spec/models/poll/answer_spec.rb
```diff
@@ -75,18 +90,25 @@
...
+      option = create(:poll_question_option, title: "One", question: question)
+
+      expect(build(:poll_answer, option: option, answer: "legacy snapshot")).to be_valid
+    end
 
-      create(:poll_question_option, title: "Two", question: question)
-      create(:poll_question_option, title: "Three", question: create(:poll_question, poll: create(:poll)))
+    it "is not valid when answer exceeds maximum length" do
+      question = create(:poll_question_open)
+      answer = build(:poll_answer, question: question, answer: "a" * (Poll::Answer.answer_max_length + 1))
```
javierm: In similar tests, we use the number instead of the method 🤔.

```suggestion
      answer = build(:poll_answer, question: question, answer: "a" * 1001)
```

Same in the other test, only with `1000` and `1001`.

db/schema.rb
```diff
@@ -1017,7 +1017,6 @@
     t.index ["author_id"], name: "index_poll_answers_on_author_id"
     t.index ["option_id", "author_id"], name: "index_poll_answers_on_option_id_and_author_id", unique: true
     t.index ["option_id"], name: "index_poll_answers_on_option_id"
-    t.index ["question_id", "answer"], name: "index_poll_answers_on_question_id_and_answer"
```
javierm, line 1020: Would it make sense to replace this one with an index by both `question_id` and `option_id`? 🤔 Not sure; just wondering.

spec/models/poll/answer_spec.rb
```diff
@@ -25,18 +25,33 @@
...
 
-    it "is not valid without an answer when question is open-ended" do
-      answer.question = create(:poll_question_open)
-      answer.answer = nil
+    it "is valid without answer text when an option is present" do
+      expect(answer).to be_valid
+      expect(answer.answer).to be nil
+    end
+
+    it "is not valid without an option when question accepts options" do
```
javierm, line 43: For completeness sake, shall we also add a test checking that it's valid when one of the options allows custom text?

app/components/polls/questions/question_component.rb
```diff
@@ -56,32 +60,56 @@ def multiple_choice_help_text
...
-    def multiple_choice_field(option)
-      choice_field(option) do
+    def choice_field(option)
+      safe_join([
+        option_label(option),
+        (open_text_tag(option) if option.allows_custom_text?)
+      ])
+    end
+
+    def option_label(option)
```
javierm: The name `option_label` might be a bit misleading here, since it displays both an input and a label 🤔. Maybe `option_field`? Although then `choice_field` and `option_field` would be too similar...

app/assets/javascripts/polls/form.js
```diff
@@ -1,6 +1,14 @@
 (function() {
   "use strict";
   App.PollsForm = {
+    syncOpenText: function(input) {
+      var $input = $(input);
+      var $textarea = $(".open-text[data-selects='" + $input.attr("id") + "']");
+
+      if ($textarea.length) {
+        $textarea.prop("disabled", !$input.prop("checked") || $input.prop("disabled"));
```
javierm: Do we need the second part of the condition? 🤔

app/assets/javascripts/polls/form.js
```diff
@@ -11,16 +19,32 @@
...
+    },
+
     initialize: function() {
       $(".poll-form .multiple-choice").each(function() {
         App.PollsForm.updateMultipleChoiceStatus(this);
       });
 
-      $(".poll-form .multiple-choice :checkbox").on("change", function() {
-        var fieldset = $(this).closest("fieldset");
+      $(".poll-form .multiple-choice").off("change.pollsform").on("change.pollsform", ":checkbox", function() {
```
javierm: We've never used the `.off` + `.on` pattern. Why do we need it here and not anywhere else? 🤔

By the way, why aren't we using the `change` event? 🤔

app/components/polls/questions/question_component.rb
```diff
@@ -56,32 +60,56 @@ def multiple_choice_help_text
...
-    def choice_field(option, &block)
-      label_tag("web_vote_option_#{option.id}") do
-        block.call + option.title
-      end
+    def open_text_tag(option)
+      text_area_tag(
+        "web_vote[#{question.id}][answer][#{option.id}]",
+        existing_text_for(option),
+        id: "web_vote_option_#{option.id}_text",
+        class: "open-text",
```
javierm: The text area now relates to database columns called `allows_custom_text` and `answer`. So I guess the `open-text` class is a bit confusing 🤔. How about `custom-answer`?

Same in other places where we use the `open-text` class.

app/components/polls/questions/question_component.rb
```diff
@@ -56,32 +60,56 @@ def multiple_choice_help_text
...
 
-    def choice_field(option, &block)
-      label_tag("web_vote_option_#{option.id}") do
-        block.call + option.title
-      end
+    def open_text_tag(option)
+      text_area_tag(
+        "web_vote[#{question.id}][answer][#{option.id}]",
+        existing_text_for(option),
+        id: "web_vote_option_#{option.id}_text",
```
javierm:
```suggestion
        id: "web_vote_option_#{option.id}_answer",
```

app/assets/javascripts/polls/form.js
```diff
@@ -1,6 +1,14 @@
 (function() {
   "use strict";
   App.PollsForm = {
+    syncOpenText: function(input) {
+      var $input = $(input);
+      var $textarea = $(".open-text[data-selects='" + $input.attr("id") + "']");
+
+      if ($textarea.length) {
```
javierm: Is this condition necessary? 🤔

app/models/poll/question.rb
```diff
@@ -80,7 +84,11 @@ def find_or_initialize_user_answer(user, option_id: nil, answer_text: nil)
...
       answer.option = option
-      answer.answer = option.title
+      if option.allows_custom_text?
+        answer.answer = answer_text.presence
+      else
+        answer.answer = nil
+      end
     else
       answer.answer = answer_text
     end
```
javierm, line 90: Would this make sense?

```ruby
    answer.option = question_options.find(option_id) if accepts_options?
    answer.answer = answer_text.presence

    answer
```

app/models/poll/web_vote.rb
```diff
@@ -64,16 +64,18 @@ def given_answers(params)
...
+        answer_text = answer_param.to_s.strip
         if answer_text.present?
           [question.find_or_initialize_user_answer(user, answer_text: answer_text)]
         else
           []
         end
       else
         Array(question_params[:option_id]).map do |option_id|
-          question.find_or_initialize_user_answer(user, option_id: option_id)
+          answer_text = answer_param&.dig(option_id)&.to_s&.strip
```
javierm: `nil.to_s` exists, so we could write:

```suggestion
          answer_text = answer_param&.dig(option_id).to_s.strip
```

Related rubocop rule: Lint/RedundantSafeNavigation

app/components/officing/results/form_component.html.erb
```diff
@@ -11,10 +11,10 @@
   <% poll.questions.for_physical_votes.each do |question| %>
     <fieldset class="row">
       <legend class="column"><%= question.title %></legend>
-      <% question.question_options.each_with_index do |option, i| %>
+      <% question.question_options.each do |option| %>
         <div class="small-12 medium-6 large-3 column end">
-          <%= label_tag "questions_#{question.id}_#{i}", option.title %>
-          <%= number_field_tag "questions[#{question.id}][#{i}]", answer_result_value(question.id, i), min: 0 %>
+          <%= label_tag "questions_#{question.id}_#{option.id}", option.title %>
+          <%= number_field_tag "questions[#{question.id}][#{option.id}]", answer_result_value(question.id, option.id), min: 0 %>
```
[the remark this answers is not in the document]
javierm: False positive. We aren't enabling the autocomplete linter.

app/assets/javascripts/polls/form.js
```diff
@@ -11,16 +17,32 @@
...
     },
+
     initialize: function() {
       $(".poll-form .multiple-choice").each(function() {
         App.PollsForm.updateMultipleChoiceStatus(this);
       });
 
       $(".poll-form .multiple-choice :checkbox").on("change", function() {
-        var fieldset = $(this).closest("fieldset");
+        App.PollsForm.updateMultipleChoiceStatus($(this).closest("fieldset")[0]);
```
javierm: Is the `[0]` necessary? 🤔


```suggestion
        App.PollsForm.updateMultipleChoiceStatus($(this).closest("fieldset"));
```

Same for the radio buttons.

app/assets/javascripts/polls/form.js
```diff
@@ -1,6 +1,12 @@
 (function() {
   "use strict";
   App.PollsForm = {
+    syncOpenText: function(input) {
+      var $input = $(input);
+      var $textarea = $(".custom-answer[data-selects='" + $input.attr("id") + "']");
+
+      $textarea.prop("disabled", !$input.prop("checked"));
```
javierm: Maybe we can simplify this a little bit 🤔.

```suggestion
      var $textarea = $("[data-selects='" + input.id + "']");

      $textarea.prop("disabled", !input.checked);
```

db/migrate/20250818091122_add_allows_custom_text_to_poll_question_answers.rb
```diff
@@ -0,0 +1,5 @@
+class AddAllowsCustomTextToPollQuestionAnswers < ActiveRecord::Migration[7.1]
+  def change
+    add_column :poll_question_answers, :allows_custom_text, :boolean, default: false
```
javierm: For consistency with the `allow_custom_content` attributes in budget headings:

```suggestion
    add_column :poll_question_answers, :allow_custom_text, :boolean, default: false
```

app/assets/stylesheets/polls/form.scss
```diff
@@ -31,6 +31,12 @@
         white-space: pre;
       }
     }
+
+    .custom-answer {
+      display: block;
```
javierm: Is this line necessary? 🤔

app/components/admin/poll/questions/options/form_component.html.erb
```diff
@@ -0,0 +1,36 @@
+<%= render "shared/globalize_locales", resource: option %>
+
+<%= translatable_form_for(option, url: url) do |f| %>
+
+  <%= render "shared/errors", resource: option %>
+
+  <%= f.hidden_field :given_order,
+                     value: option.persisted? ? option.given_order : option.class.last_position(option.question_id) + 1 %>
```
javierm, line 8: Could you add to the commit message the reason why we can remove the `|| @question.id` part of this line? :pray:

(Or maybe change this line in a previous commit.)

app/components/officing/results/form_component.html.erb
javierm, line 17: There's only one test covering this change, and it's a bit hard to follow 🤔. What do you think of this patch?

```diff
diff --git a/spec/system/officing/results_spec.rb b/spec/system/officing/results_spec.rb
index 17fe8ca865..a82f8ad0b3 100644
--- a/spec/system/officing/results_spec.rb
+++ b/spec/system/officing/results_spec.rb
@@ -132,18 +132,17 @@ describe "Officing Results", :with_frozen_time do
     expect(page).to have_content(I18n.l(Date.current.to_date, format: :long))
     expect(page).to have_content(booth_name)
 
-    within("#question_#{question_1.id}_0_result") do
-      expect(page).to have_content("5555")
-      expect(page).not_to have_content("7777")
-    end
+    expect(page).to have_table rows: [
+      { "Answer" => "Yes", "Votes" => "5555" },
+      { "Answer" => "No", "Votes" => "200" }
+    ]
+    expect(page).not_to have_content("7777")
 
     expect(partial_result.reload.answer).to be nil
 
     within("#white_results") { expect(page).to have_content("6") }
     within("#null_results")  { expect(page).to have_content("7") }
     within("#total_results") { expect(page).to have_content("8") }
-    within("#question_#{question_1.id}_0_result") { expect(page).to have_content("5555") }
-    within("#question_#{question_1.id}_1_result") { expect(page).to have_content("200") }
     expect(page).not_to have_content "What do you want?"
   end
 end
```

app/components/polls/questions/question_component.rb
```diff
@@ -56,32 +60,56 @@ def multiple_choice_help_text
...
+      answer_for(option)&.answer.to_s
     end
 
     def checked?(option)
-      form.object.answers[question.id].find { |answer| answer.option_id == option.id }
+      answer_for(option).present?
+    end
+
+    def answer_for(option)
+      answers_for_question.find { |a| a.option_id == option.id }
```
javierm: Why have we renamed the `answer` block parameter to `a`? We usually use more descriptive names 🤔.

```suggestion
      answers_for_question.find { |answer| answer.option_id == option.id }
```

Or, alternatively (haven't tested it):

```suggestion
      answers_for_question.find_by(option_id: option.id)
```

app/components/polls/questions/question_component.rb
```diff
@@ -40,8 +40,12 @@ def options_read_more_links
       end, ", ")
     end
 
+    def answers_for_question
+      form.object.answers[question.id] || []
```
javierm, line 44: In order to return similar objects in both cases:

```suggestion
      form.object.answers[question.id] || question.answers.none
```

app/components/polls/questions/question_component.rb
```diff
@@ -56,32 +60,56 @@ def multiple_choice_help_text
...
-      end
+    def open_text_tag(option)
+      text_area_tag(
+        "web_vote[#{question.id}][answer][#{option.id}]",
+        existing_text_for(option),
+        id: "web_vote_option_#{option.id}_answer",
+        class: "custom-answer",
+        maxlength: Poll::Answer.answer_max_length,
+        rows: 1,
+        "aria-label": t("poll_questions.open_text_aria_label", option: option.title),
```
javierm: Since we've renamed `open-text` to `custom-answer`, should we also rename the I18n key? 🤔

app/controllers/officing/results_controller.rb
```diff
@@ -51,7 +51,7 @@ def build_results
             option_id: option.id
           )
           partial_result.officer_assignment_id = @officer_assignment.id
-          partial_result.answer = option.title
+          partial_result.answer = nil
```
javierm: Do we need this line? 🤔 I know we've talked about it, but I don't remember our conclusion 😌.

app/models/poll/partial_result.rb
```diff
@@ -9,11 +9,9 @@ class Poll::PartialResult < ApplicationRecord
 
   validates :question, presence: true
   validates :author, presence: true
-  validates :answer, presence: true
-  validates :answer, inclusion: { in: ->(partial_result) { partial_result.option.possible_answers }},
```
javierm, line 13: Does this mean we can remove the `possible_answers` method in `Poll::Question::Option` as well?

javierm (commented):
@taitus Here are a few more comments. Let me know what you think!

app/assets/javascripts/polls/form.js
```diff
@@ -11,16 +16,32 @@
...
+      $(fieldset).find(":checkbox").each(function() {
+        App.PollsForm.syncOpenText(this);
+      });
+    },
+    updateRadioChoiceStatus: function(fieldset) {
+      $(fieldset).find(":radio").each(function() {
+        App.PollsForm.syncOpenText(this);
+      });
     },
+
```
javierm: Shall we remove this line? We usually don't add blank lines between object properties 🤔.

spec/components/admin/poll/questions/options/form_component_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+describe Admin::Poll::Questions::Options::FormComponent do
+  let(:question) { create(:poll_question) }
+  let(:option) { build(:poll_question_option, question: question) }
+  let(:component) { Admin::Poll::Questions::Options::FormComponent.new(option, url: "/") }
+
+  it "renders the form" do
+    render_inline component
+
+    expect(page).to have_field "Answer"
+    expect(page).to have_css "label", text: "Description"
```
javierm: In order to check that the label is correctly associated to its input:

```suggestion
    expect(page).to have_field "Description (optional)"
```

spec/components/admin/poll/questions/options/form_component_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+    expect(page).to have_button "Save"
+  end
+
+  describe "allows custom text checkbox" do
+    it "includes accessibility attributes" do
+      render_inline component
+
+      checkbox = page.find_field("Allow custom text")
```
javierm:
```suggestion
      checkbox = page.find_field("Allow custom text", type: :checkbox)
```

app/components/polls/questions/question_component.rb
taitus: After a failed submit, `WebVote#update` replaces `@answers` with Arrays from `given_answers`.

spec/components/polls/questions/question_component_spec.rb
```diff
@@ -80,11 +84,13 @@
     end
 
     it "selects the option when users have already voted" do
-      create(:poll_answer, author: user, question: question, option: option_yes)
+      option_yes.update!(allow_custom_text: true)
+      create(:poll_answer, author: user, question: question, option: option_yes, answer: "because yes")
```
javierm: "Because yes" sounds a bit strange in English 🤔.

```suggestion
      create(:poll_answer, author: user, question: question, option: option_yes, answer: "because")
```

spec/components/polls/questions/question_component_spec.rb
```diff
@@ -80,11 +84,13 @@
...
 
       render_inline Polls::Questions::QuestionComponent.new(question, form: form)
 
       expect(page).to have_field "Yes", type: :radio, checked: true
+      expect(page).to have_field "Answer for Yes", type: :textarea, with: "because yes"
```
javierm:
```suggestion
      expect(page).to have_field "Answer for Yes", type: :textarea, with: "because"
```

spec/factories/polls.rb
```diff
@@ -66,6 +66,13 @@
       end
     end
 
+    trait :yes_no_custom_text do
+      after(:create) do |question|
+        create(:poll_question_option, question: question, title: "Yes", allow_custom_text: true)
+        create(:poll_question_option, question: question, title: "No", allow_custom_text: true)
+      end
+    end
```
javierm, line 75: Do we need to create the `votation_type_unique` here, like we do in the `:yes_no` trait? 🤔 No idea, and maybe it wasn't necessary in the `:yes_no` trait. It's just that the difference feels a bit strange.

spec/models/poll/partial_result_spec.rb
```diff
@@ -2,18 +2,37 @@
...
+      question = create(:poll_question, :yes_no)
 
-      create(:poll_question_option, title: "Two", question: question)
-      create(:poll_question_option, title: "Three", question: create(:poll_question, poll: create(:poll)))
+      expect(build(:poll_partial_result, question: question, option: nil)).not_to be_valid
+    end
+
+    it "is valid with an option for the question" do
+      question = create(:poll_question, :yes_no)
+      option = question.question_options.first
```
javierm: Is this test redundant? Seems to be similar to the "is valid without answer text when an option is present" test 🤔.

spec/system/polls/polls_spec.rb
```diff
@@ -200,7 +200,8 @@
       question = create(:poll_question, :yes_no, poll: poll, title: "Do you agree?")
 
       poll.update!(geozone_restricted_to: [geozone])
-      create(:poll_answer, author: user, question: question, answer: "Yes")
+      option_yes = question.option_for("Yes")
+      create(:poll_answer, author: user, question: question, option: option_yes)
```
javierm: To make more obvious what's changing:

```suggestion
      create(:poll_answer, author: user, question: question, option: question.option_for("Yes"))
```

spec/system/polls/polls_spec.rb
```diff
@@ -231,7 +232,8 @@
       user = create(:user, :level_two, geozone: geozone)
       question = create(:poll_question_multiple, :abc, poll: poll, title: "Which ones are better?")
 
-      create(:poll_answer, author: user, question: question, answer: "Answer A")
+      option_a = question.option_for("Answer A")
+      create(:poll_answer, author: user, question: question, option: option_a)
```
javierm:
```suggestion
      create(:poll_answer, author: user, question: question, option: question.option_for("Answer A"))
```

spec/system/polls/votation_types_spec.rb
```diff
@@ -51,9 +51,78 @@
...
+    within_fieldset("Is it that bad?") do
+      expect(page).to have_field "Answer for Yes", type: :textarea, disabled: true
+      expect(page).to have_field "Answer for No", type: :textarea, disabled: true
+
+      choose "Yes"
+
+      expect(page).to have_field "Answer for Yes", type: :textarea, disabled: false
+      expect(page).to have_field "Answer for No", type: :textarea, disabled: true
+
+      fill_in "Answer for Yes", with: "because yes"
```
javierm:
```suggestion
      fill_in "Answer for Yes", with: "because"
```

spec/system/polls/votation_types_spec.rb
```diff
@@ -51,9 +51,78 @@
...
+    end
+
+    expect(page).to have_button "Vote"
+  end
+
   scenario "Maximum votes has been reached" do
     question = create(:poll_question_multiple, :abc, poll: poll, max_votes: 2)
-    create(:poll_answer, author: author, question: question, answer: "Answer A")
+    option_a = question.option_for("Answer A")
+    create(:poll_answer, author: author, question: question, option: option_a)
```
javierm:
```suggestion
    create(:poll_answer, author: author, question: question, option: question.option_for("Answer A"))
```

javierm (approved):

PR #6068, opened 2025-08-21
AI translations

stirkac:
# Before creating a pull request

Please make sure to read and follow the coding conventions shown in [CONTRIBUTING](https://github.com/consuldemocracy/consuldemocracy/blob/master/CONTRIBUTING.md)
For more information please view [Coding Conventions](https://github.com/consuldemocracy/consuldemocracy/blob/master/docs/en/open_source/coding_conventions.md)

**Here's a quick summary of some points to check**

* Include specs to test any changes you've made
* Our CI will check whether the rest of the application is still working properly; check its build and make sure all tests are passing
* Run `bundle exec pronto run` and fix any issues it reports; these issues will also be automatically reported on the pull request
* Follow [the seven rules of a great commit message](https://chris.beams.io/posts/git-commit/)
* Add or modify i18n translations only in the base languages, English (en) and Spanish (es); we manage all other languages through the [Crowdin integration](https://translate.consuldemocracy.org/).

## References

> Reference any related issues/Pull Requests/errors/etc... to your work.

## Objectives

## Visual Changes

* Any visual changes? please attach screenshots (or gifs) showing them.
* If modified views are public (not the admin panel), try them in mobile display (with your browser's developer console) and add screenshots.

## Notes

* Mention rake tasks or actions to be done when deploying this changes to a server (if any).
* Explain any caveats, or important things to notice like deprecations (if any).


app/components/admin/settings/llm_configuration_tab_component.html.erb
```diff
@@ -0,0 +1,42 @@
...
+            onchange: "this.form.submit();"
+          ) %>
+      </td>
+    </tr>
+  <% end %>
+
+  <% if Setting["llm.provider"].present? && Setting["llm.model"].present? %>
+    <%= render Admin::Settings::RowComponent.new("llm.use_llm_for_translations", type: :feature, tab: tab) %>
+  <% end %>
+
```
taitus: This new `llm_configuration_tab_component.html.erb` view keeps the same structure as `Admin::Settings::TableComponent`, which is great for consistency.

That said, we could consider using the `Admin::Settings::RowComponent`, just like in the other tabs.

Using `RowComponent` would mean slightly extending it so it can render dropdowns (like your current `Admin::Settings::DropdownSettingComponent`) by passing a list of options. Doing that would help us standardize how setting selectors should be rendered when adding them to settings tabs in the future.

We could also add a `disabled:` option to enable or disable any row, which would help us avoid using `onchange` events in the components.

What do you think of this suggestion? 🤔 

```suggestion

  <%= render Admin::Settings::RowComponent.new("llm.provider", type: :dropdown, tab: tab, options: provider_options) %>
  <%= render Admin::Settings::RowComponent.new("llm.model", type: :dropdown, tab: tab, options: model_options, disabled: model_disabled?) %>
  <%= render Admin::Settings::RowComponent.new("llm.use_llm_for_translations", type: :feature, tab: tab, disabled: feature_disabled?) %>

```

If that approach makes sense to you, [here’s a patch](https://github.com/user-attachments/files/22779753/admin_settings_component.patch) you can apply directly using `git apply`.

This patch adapts the rest of the settings components to support this change and align with the proposed use of `RowComponent`.

In short, it includes the following updates for `DropdownFormComponent`:
	•	Renames `DropdownSettingComponent` to `DropdownFormComponent` for consistency with existing form components.
	•	Removes the `onchange` handler and relies on the disabled flag propagated from the `RowComponent`.
	•	Adds aria-labelledby, aria-describedby, and a submit button to improve accessibility and keyboard navigation.
	•	With the use of the `RowComponent`, we no longer need the hint parameter since we can now use the setting’s description directly.
	•	Unifies the `blank_option` logic inside the component view instead of passing it as a parameter.

And other changes like:
	•	Extends `RowComponent` to support the new `:dropdown` type and a `disabled:` flag.
	•	Updates `FeaturedSettingsFormComponent` to accept and propagate the `disabled:` argument.
  •   Refactors the LLM configuration tab to use `RowComponent` rows instead of custom <tr> markup for better consistency and maintainability.

app/assets/stylesheets/admin/settings/dropdown_settings.scss
```diff
@@ -0,0 +1,19 @@
...
+  .help-text {
+    display: block;
+  }
+
+  [type="submit"] {
+    @include regular-button;
+    display: block;
+    margin-top: $line-height;
+  }
+}
```
taitus: If we apply the patch suggested in `app/components/admin/settings/llm_configuration_tab_component.html.erb`, we should review which of the styles in `app/assets/stylesheets/admin/settings/dropdown_settings.scss` are still needed. At minimum, we should rename the file to `dropdown_form.scss` to match the new component name.

One specific concern: `@include regular-button;` on the submit button is overriding the `"button hollow expanded"` classes used by `Admin::Settings::DropdownFormComponent`. That means the hollow style won’t take effect.

app/lib/remote_translations/caller.rb
```diff
@@ -1,6 +1,27 @@
 class RemoteTranslations::Caller
   attr_reader :remote_translation
 
+  def self.available_locales
+    translation_provider ? translation_provider::AvailableLocales.locales : []
```
taitus: Since translation_provider always returns a class (RemoteTranslations::Llm or RemoteTranslations::Microsoft), the fallback to [] is never actually used.

```suggestion
    translation_provider::AvailableLocales.locales
```

app/lib/remote_translations/caller.rb
```diff
@@ -1,6 +1,27 @@
...
+  end
+
+  def self.configured?
+    return true if llm?
+
+    Setting["feature.remote_translations"].present? && Tenant.current_secrets.microsoft_api_key.present?
```
taitus: Since the `feature.remote_translations` flag is already checked in `remote_translation_enabled?`, there’s no need to verify it again here.

```suggestion
    llm? || Tenant.current_secrets.microsoft_api_key.present?
```

app/lib/remote_translations/llm/client.rb
```diff
@@ -0,0 +1,39 @@
...
+
+      TOKENS_PER_WORD = 3 # approximation
+
+      class LLMTranslationError < StandardError; end
+
+      def initialize
+        @context = Config.context
+        @chat = @context.chat(provider: Setting["llm.provider"].downcase.to_sym, model: Setting["llm.model"])
+        @prompt = YAML.load_file("config/llm_prompts.yml", aliases: true)["remote_translation_prompt"]
+      end
```
taitus, line 10: Maybe we could keep the initialize method focused only on assigning instance variables, and move the dynamic parts (like loading the prompt or building the chat) into small private methods instead.

```suggestion
def initialize
  @context = build_context
  @chat = build_chat
  @prompt = load_prompt
end

private

def build_context
  Config.context
end

def build_chat
  build_context.chat(provider: Setting["llm.provider"].downcase.to_sym, model: Setting["llm.model"])
end

def load_prompt
  YAML.load_file("config/llm_prompts.yml", aliases: true)["remote_translation_prompt"]
end
```

app/lib/remote_translations/llm/client.rb
```diff
@@ -0,0 +1,39 @@
...
+            output_locale: locale
+          }
+          text_prompt = prompt % text_placeholders
+
+          if text_prompt.size * TOKENS_PER_WORD < RubyLLM.models.find(Setting["llm.model"]).context_window
+            chat.ask(text_prompt).content
+          else
+            raise LLMTranslationError, "Text to translate is too long for the model context window"
+          end
+        end
```
taitus, line 35: Since translatable fields are limited to 6,000 characters, the context-window check doesn’t seem necessary. If that’s the case, we could simplify request_translation to a single call.

```suggestion
        def request_translation(text, locale)
          text_prompt = prompt % { input_text: text, output_locale: locale }
          chat.ask(text_prompt).content
        end
```

app/lib/remote_translations/llm/config.rb
```diff
@@ -0,0 +1,22 @@
+module RemoteTranslations
+  module Llm
+    class Config
+      class << self
+        def context
+          # set credentials in the secrets.yml file, referencing https://rubyllm.com/configuration/#provider-configuration
```
taitus: We usually don’t leave inline comments; instead, we could include that context in the commit message or in the documentation 🤔

app/lib/remote_translations/llm/config.rb
```diff
@@ -0,0 +1,22 @@
...
+          @context ||= RubyLLM.context do |config|
```
taitus:
```suggestion
          @context = RubyLLM.context do |config|
```
I’m wondering if memoizing the context could cause the first tenant’s configuration to be propagated and reused by others. If that’s the case, rebuilding the context on each call would ensure that the correct settings are applied for the active tenant.

config/locales/en/admin.yml
```diff
@@ -1347,6 +1347,13 @@ en:
           title: Cookies consent
         cookies_consent_banner:
           title: Cookies consent banner
+        llm:
+          title: LLM Settings
```
taitus, line 1353: “LLM Settings” might sound a bit too technical for administrators who aren’t familiar with the term Large Language Model 🤔

config/locales/en/admin.yml
```diff
@@ -1347,6 +1347,13 @@ en:
...
+          provider: "LLM Provider"
+          default_help_text: "Providers will be <strong>disabled</strong> until credentials are configured in the secrets.yml."
+          provider_blank_option: "None"
+          model_blank_option: "None"
+          model: "Model"
```
taitus: These translations don’t seem necessary, since the component is already using the ones defined under a different path in config/locales/en/settings.yml.

spec/components/layout/remote_translations_button_component_spec.rb
```diff
@@ -25,6 +25,7 @@
     it "displays the text in English with a locale needing parsing" do
       I18n.with_locale(:"zh-CN") { render_inline component }
 
+      expect(component.render?).to be true
```
taitus: We usually use:
```suggestion
      expect(page).to be_rendered
```

Same for line 50 in this file.

spec/lib/remote_translations/caller_selection_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+
+    it "falls back to microsoft settings when llm? is false" do
+      allow(RemoteTranslations::Caller).to receive(:llm?).and_return(false)
+      Setting["feature.remote_translations"] = true
+      allow(Tenant).to receive_message_chain(:current_secrets, :microsoft_api_key).and_return("key")
+
+      expect(RemoteTranslations::Caller.configured?).to be true
+    end
+  end
+end
```
taitus: Maybe we could move all these tests into spec/lib/remote_translations/caller_spec.rb to keep everything related to the caller grouped in one place.

Regarding the existing specs in this file we have two options:

1) Add a second context for the LLM client:
`let(:client) { RemoteTranslations::Llm::Client }`

2) Decouple from client logic and stub at the caller level:

```
expect_any_instance_of(RemoteTranslations::Caller).to receive(:translations).exactly(2).times.and_return(responses[0], responses[1])
```

spec/system/admin/llm/llm_settings_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+      # Wait for the AJAX request to complete
+      expect(page).to have_content "Value updated"
+
+      # Verify settings persisted
+      expect(Setting.find_by!(key: "llm.provider").reload.value).to eq "openai"
+      expect(Setting.find_by!(key: "llm.model").reload.value).to eq "gpt-4o-mini"
+      expect(Setting.find_by!(key: "llm.use_llm_for_translations").reload.value).to eq "active"
+    end
+  end
+end
```
taitus: Maybe we could move these specs to spec/system/admin/settings_spec.rb.

config/locales/en/admin.yml
muki: If anyone has a suggestion with what to replace it I'll take it, but I feel "LLM Settings" is exactly what this is. I think "AI settings", while arguably more colloquial is (imo) wrong. We should be as specific as possible. 

Personally, I don't really share the concern that it might be too technical, but I could be wrong. Even if it is too technical I'd argue it's still "correct" and therefore "right" and maybe a nice small barrier to entry (don't touch knobs you don't understand).

All of this is to say that I have no better idea but am very open to anyone else's.

app/components/admin/settings/llm_configuration_tab_component.html.erb
stirkac: @taitus  I was coming from a standpoint that as little of existing code should be fettled with, that's why I separated out all the components, code and tests. Your suggestions are appreciated and make the feature feel more at home.
I accepted the changes to be in line with suggestions. Let me know if the corrected implementation makes sense.

spec/components/layout/remote_translations_button_component_spec.rb
```diff
@@ -5,8 +5,8 @@
   let(:component) { Layout::RemoteTranslationsButtonComponent.new(translations) }
 
   before do
-    allow(RemoteTranslations::Microsoft::AvailableLocales).to receive(:remote_available_locales)
-                                                          .and_return(%w[de en es fr pt zh-Hans])
+    allow(RemoteTranslations::Caller).to receive(:available_locales)
+      .and_return(%w[de en es fr pt zh-Hans zh-CN pt-BR])
```
taitus: Good catch adding "zh-CN" and "pt-BR" so the tests pass 👏

However, since we're now stubbing `RemoteTranslations::Caller.available_locales` instead of `RemoteTranslations::Microsoft::AvailableLocales.remote_available_locales`, the locales "pt" and "zh-Hans" are no longer needed in the return list for the tests to pass.

```suggestion
    allow(RemoteTranslations::Caller).to receive(:available_locales)
      .and_return(%w[de en es fr zh-CN pt-BR])
```

spec/lib/remote_translations/caller_selection_spec.rb
taitus: Since these tests were moved to `spec/lib/remote_translations/caller_spec.rb,` this file can be removed.

spec/lib/remote_translations/caller_spec.rb
```diff
@@ -226,4 +226,67 @@
       caller.call
     end
   end
+
+  describe ".configured?" do
+    it "is true if llm? is true regardless of microsoft key" do
+      allow(RemoteTranslations::Caller).to receive(:llm?).and_return(true)
+      allow(Tenant).to receive_message_chain(:current_secrets, :microsoft_api_key).and_return(nil)
```
taitus:
```suggestion
      stub_secrets(microsoft_api_key: nil)
```

We have a helper method called `stub_secrets` to stub application secrets. We can use it here. Same for line 241.

spec/lib/remote_translations/caller_spec.rb
```diff
@@ -226,4 +226,67 @@
...
+
+      expect(RemoteTranslations::Caller.configured?).to be true
+    end
+
+    it "falls back to microsoft settings when llm? is false" do
+      allow(RemoteTranslations::Caller).to receive(:llm?).and_return(false)
+      Setting["feature.remote_translations"] = true
```
taitus: `Setting["feature.remote_translations"] = true` The test seems to pass without it, so we could remove it if it's not needed 🤔

spec/system/admin/settings_spec.rb
```diff
@@ -370,4 +370,72 @@
...
+
+  describe "LLM settings" do
+    context "Nothing is configured" do
+      scenario "Provider and model are disabled, but Settings render without errors" do
+        visit admin_settings_path
+        click_link "LLM Settings"
+
+        expect(page).to have_content "LLM Provider"
+      end
+    end
```
taitus: Lately, since we introduced components into the project, we try to test the component itself whenever possible instead of adding new system specs.

Since the next test already covers LLM tab flow, we could remove this one and test it directly in the component instead.

app/components/admin/settings/dropdown_form_component.html.erb
```diff
@@ -0,0 +1,21 @@
...
+      <%= f.select :value,
+                   options,
+                   {
+                     include_blank: t("admin.settings.index.blank_option"),
+                     label: false
+                   },
+                   {
+                     "aria-labelledby": dom_id(setting, :title),
+                     "aria-describedby": dom_id(setting, :description)
+                   } %>
```
taitus, line 15: I know I suggested this structure in the previous review, but it would be better to add a dynamic id here to avoid duplicate IDs if this component is rendered multiple times on the same page.

```suggestion
                   {
                     id: dom_id(setting, :value),
                     "aria-labelledby": dom_id(setting, :title),
                     "aria-describedby": dom_id(setting, :description)
                   } %>
```

spec/system/admin/settings_spec.rb
```diff
@@ -370,4 +370,72 @@
...
+
+        # Wait for the AJAX request to complete
+        expect(page).to have_content "Value updated"
+
+        # Verify settings persisted
+        expect(Setting.find_by!(key: "llm.provider").reload.value).to eq "OpenAI"
+        expect(Setting.find_by!(key: "llm.model").reload.value).to eq "gpt-4o-mini"
+        expect(Setting.find_by!(key: "llm.use_llm_for_translations").reload.value).to eq "active"
+      end
+    end
```
taitus, line 406: I think having just a happy path here is enough. We can cover the disabled states and other variations in the component specs.

We've had some flaky tests before when hitting the DB after a visit, so we try to avoid that now.
In this case, just checking that the field stays selected after the update should be good enough.


```suggestion
    context "Required LLM setup is configured" do
      before do
        allow(RemoteTranslations::Llm::Config)
          .to receive(:providers).and_return({ OpenAI: { enabled: true } })

        ruby_llm_models = [double("Model", name: "GPT-4.1 mini", id: "gpt-4o-mini")]
        allow(RubyLLM.models).to receive(:by_provider).with(:openai).and_return(ruby_llm_models)
      end

      scenario "Configure provider, model and enable usage" do
        visit admin_settings_path
        click_link "LLM Settings"

        within "tr", text: "LLM Provider" do
          expect(page).to have_select selected: "None"

          select "OpenAI"
          click_button "Update"

          expect(page).to have_select selected: "OpenAI"
        end

        expect(page).to have_content "Value updated"

        within "tr", text: "Model" do
          expect(page).to have_select selected: "None"

          select "GPT-4.1 mini"
          click_button "Update"

          expect(page).to have_select selected: "GPT-4.1 mini"
        end

        expect(page).to have_content "Value updated"

        within "tr", text: "Content Translation" do
          expect(page).to have_button "No"

          click_button "No"

          expect(page).to have_button "Yes"
        end

        expect(page).to have_content "Value updated"
      end
    end
```

spec/lib/remote_translations/llm/config_spec.rb
```diff
@@ -0,0 +1,46 @@
+require "rails_helper"
+
+describe RemoteTranslations::Llm::Config do
+  # Ensure memoized context from RemoteTranslations::Llm::Config doesn't leak doubles
+  after do
+    if RemoteTranslations::Llm::Config.instance_variable_defined?(:@context)
+      RemoteTranslations::Llm::Config.remove_instance_variable(:@context)
+    end
+  end
```
taitus: Since we're no longer memoizing @context, these lines could be removed?

spec/lib/remote_translations/llm/config_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+      allow(Tenant).to receive(:current_secrets).and_return(double(llm: secrets))
+
+      config_double = double("RubyLLM::Configuration", openai_access_token: nil)
+      allow(config_double).to receive(:openai_access_token=)
+      context_double = double(config: config_double)
+      allow(RubyLLM).to receive(:context).and_yield(config_double).and_return(context_double)
+
+      expect { RemoteTranslations::Llm::Config.context }.not_to raise_error
+    end
+  end
```
taitus: What do you think about using the stub_secrets method here?

```suggestion
  describe ".context" do
    before { stub_secrets(llm: { openai_api_key: "1234" }) }

    it "creates a context with tenant secrets without errors" do
      config = instance_double("RubyLLM::Configuration")
      expect(config).to receive(:openai_api_key=).with("1234")

      context = double("RubyLLM::Context", config: config)
      expect(RubyLLM).to receive(:context).and_yield(config).and_return(context)

      expect { RemoteTranslations::Llm::Config.context }.not_to raise_error
    end
  end
```

spec/lib/remote_translations/llm/config_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+      end
+
+      stub_const("RubyLLM::Providers::OpenAI", stub_provider)
+      allow(RubyLLM::Providers).to receive(:constants).and_return([:OpenAI])
+
+      providers = RemoteTranslations::Llm::Config.providers
+      expect(providers).to include(:OpenAI)
+      expect(providers[:OpenAI]).to include(:enabled)
+      expect(providers[:OpenAI][:enabled]).to be true
+    end
```
taitus:
```suggestion
    it "returns a hash of providers with their enabled status" do
      context = double("RubyLLM::Context", config: double)
      allow(RemoteTranslations::Llm::Config).to receive(:context).and_return(context)

      stub_provider = Class.new do
        def self.configured?(_config)
          true
        end
      end

      stub_const("RubyLLM::Providers::OpenAI", stub_provider)
      allow(RubyLLM::Providers).to receive(:constants).and_return([:OpenAI])

      providers = RemoteTranslations::Llm::Config.providers

      expect(providers).to eq({ OpenAI: { enabled: true } })
    end
```

spec/lib/remote_translations/llm/config_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
```
taitus: We can stub the provider in the before block. Same way we usually do in other specs:

```suggestion
  describe ".providers" do
    before do
      dummy_provider = Class.new do
        def self.configured?(_config)
          true
        end
      end

      stub_const("RubyLLM::Providers::OpenAI", dummy_provider)
    end

    it "maps provider enabled status using RubyLLM providers" do
      context = double("RubyLLM::Context", config: instance_double("RubyLLM::Configuration"))
      allow(RemoteTranslations::Llm::Config).to receive(:context).and_return(context)
      allow(RubyLLM::Providers).to receive(:constants).and_return([:OpenAI])

      providers = RemoteTranslations::Llm::Config.providers

      expect(providers).to eq({ OpenAI: { enabled: true } })
    end
  end
```

app/lib/remote_translations/llm/client.rb
```diff
@@ -0,0 +1,42 @@
+module RemoteTranslations
+  module Llm
+    class Client
+      attr_reader :context, :chat, :prompt
+
+      TOKENS_PER_WORD = 3 # approximation
```
taitus: Since we're no longer referencing `TOKENS_PER_WORD`, this line can be removed.

spec/lib/remote_translations/llm/client_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+      RemoteTranslations::Llm::Config.remove_instance_variable(:@context)
+    end
+  end
+
+  it "calls chat.ask for each field and returns contents" do
+    client = RemoteTranslations::Llm::Client.new
+    result = client.call(["Hello", "World"], "es")
+
+    expect(result).to eq(["translated", "translated"])
+  end
```
taitus, line 32: I think we can remove: 
- `Setting["llm.use_llm_for_translations"]`, since it isn't referenced in `RemoteTranslations::Llm::Client`.
- The `YAML` stub and `RubyLLM.models.find` chain, as we don't rely on `context_window` or prompt content in this test.
- The `after` cleanup and `allow_any_instance_of(RemoteTranslations::Llm::Config)` lines, since `Config.context` no longer memoizes.

So we could simplify the test setup to stub only what's actually used in the client flow:

```suggestion
describe RemoteTranslations::Llm::Client do
  let(:chat) { instance_double("RubyLLM::Chat", ask: double(content: "translated")) }
  let(:config)  { instance_double("RubyLLM::Configuration") }
  let(:context) { instance_double("RubyLLM::Context", chat: chat, config: config) }

  before do
    Setting["llm.provider"] = "OpenAI"
    allow(RemoteTranslations::Llm::Config).to receive(:context).and_return(context)
  end

  it "calls chat.ask for each field and returns contents" do
    client = RemoteTranslations::Llm::Client.new
    result = client.call(["Hello", "World"], "es")

    expect(result).to eq(["translated", "translated"])
  end
end
```

Not sure if it would be worth adding a separate example to check the prompt interpolation (or adding those expectations here).
Since the prompt is fictitious and may change, it might not add much value. Unless we just want to verify that both %{output_locale} and %{input_text} are interpolated correctly. 🤔

```
it "interpolates the prompt with the expected text" do
  allow(YAML).to receive(:load_file).and_return(
    { "remote_translation_prompt" => "Please translate to %{output_locale}: %{input_text}" }
  )

  expect(chat).to receive(:ask).with("Please translate to es: Hello")
    .and_return(double(content: "translated"))
  expect(chat).to receive(:ask).with("Please translate to es: World")
    .and_return(double(content: "translated"))

  client = RemoteTranslations::Llm::Client.new
  result = client.call(["Hello", "World"], "es")

  expect(result).to eq(%w[translated translated])
end
```

docs/img/translations/remote_translations/display-llm-translations-en.png
taitus, line 1: This image is outdated.

docs/en/features/user_content_translations.md
taitus, line 1: I can take care of updating the `docs/es/features/user_content_translations.md` file.
I'm leaving this comment here so we don't forget.

config/locales/en/settings.yml
```diff
@@ -240,3 +237,13 @@ en:
       more_info_link_description: Link shown in the consent banner to the cookie policy page.
       version_name: Version name
       version_name_description: Change the version name to force all users to go through the cookie consent process again. This feature can be handy when introducing drastic cookie policy changes requiring a new user acknowledgement.
+    llm:
+      title: LLM Settings
+      provider: "LLM Provider"
+      default_help_text: "Providers will be <strong>disabled</strong> until credentials are configured in the secrets.yml."
```
taitus: This key "default_help_text" is no longer used in the component, but the text still makes sense as the setting description.
If we rename it to "provider_description", the row component will automatically display it as the setting description.
Since we usually avoid adding HTML tags in setting descriptions, we could probably remove the `<strong>`  as well. 🤔

```suggestion
      provider_description: "Providers will be disabled until credentials are configured in the secrets.yml."
```

Same in `config/locales/es/settings.yml`

app/components/admin/settings/dropdown_form_component.rb
taitus, line 1: We could add a new spec file at
`spec/components/admin/settings/dropdown_form_component_spec.rb`
to cover this component with component specs (similar to the existing `FeaturedSettingsFormComponent` tests).

app/components/admin/settings/featured_settings_form_component.rb
taitus, line 1: We could add a few more specs in the FeaturedSettingsFormComponent tests to cover the recent changes in this component.

app/components/admin/settings/llm_configuration_tab_component.rb
taitus, line 1: We could add a new spec file at
`spec/components/admin/settings/llm_configuration_tab_component_spec.rb`
to cover this component with component specs. The most interesting part to test here would probably be the disabled states depending on the component's configuration.

app/lib/remote_translations/llm/client.rb
```diff
@@ -0,0 +1,40 @@
+module RemoteTranslations
+  module Llm
+    class Client
+      attr_reader :context, :chat, :prompt
+
+      class LLMTranslationError < StandardError; end
```
taitus: Could it be that this line is no longer needed?

config/locales/en/settings.yml
```diff
@@ -240,3 +240,13 @@ en:
       more_info_link_description: Link shown in the consent banner to the cookie policy page.
       version_name: Version name
       version_name_description: Change the version name to force all users to go through the cookie consent process again. This feature can be handy when introducing drastic cookie policy changes requiring a new user acknowledgement.
+    llm:
+      title: LLM Settings
+      provider: "LLM Provider"
+      provider_description: "Providers will be disabled until credentials are configured in the secrets.yml."
+      provider_blank_option: "None"
+      model_blank_option: "None"
```
taitus: These two translations could probably be removed, as they're no longer referenced in the code.

app/components/admin/settings/llm_configuration_tab_component.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def models
+    provider = Setting["llm.provider"]
+    return {} if provider.blank?
+
+    RubyLLM.models.by_provider(provider.downcase.to_sym).each_with_object({}) do |model, hash|
+      hash[model.name] = {
+        id: model.id,
+        enabled: true
```
taitus: I think the `enabled: true` key might be unused 🤔

taitus (commented):
Hi @stirkac 

Thanks again for applying all the changes to this PR!

In addition to the three small comments you'll see in the review, I'm sharing a couple of patches that you can apply if they make sense to you:

[docs-and-es-translations.patch](https://github.com/user-attachments/files/23499645/docs-and-es-translations.patch)

- Includes the Spanish documentation update and a small structural tweak in the English version (no text changes).
- Since we now have a Spanish version, both guides reference separate images (`display-llm-translations-en.png` and `display-llm-translations-es.png`).
You’ll just need to rename the original image from `display-llm-translations.png` to `display-llm-translations-en.png` and add the [Spanish one](https://github.com/user-attachments/assets/bc953ef8-9316-4c04-86f6-a4abd4fc357a).
- I'd also suggest adding a "Pricing" subsection under Using LLM Translation (same level as Configuration), with a short generic explanation about how costs vary by provider/model 🤔

[specs-changes.patch](https://github.com/user-attachments/files/23502422/specs-changes-v2.patch)

- Contains the adjustments we discussed for the tests related to `spec/shared/system/remotely_translatable.rb`, along with a few simplifications and some alignment with our component conventions.
For the remotely translatable changes, I opted to allow a new provider and make one call for `:microsoft` and another for `:llm`. Since the Microsoft option will likely be deprecated in upcoming releases, we can revisit and optimize those tests later on.

They're all small tweaks, that's why I preferred to share them as patches rather than add tiny inline comments in the review.

If anything doesn't quite fit or you'd rather approach it differently, just let me know and we'll figure it out together.

Thanks again! 🙌

PR #6070, opened 2025-09-04
Fix total votes in multilingual polls

### LINKED ISSUE #4744: Voting counts are locale dependant
### User Story

Hi, I'm part of Enreda team, we make a lot of CONSUL implementations and yesterday we found a really ugly bug ⚠️ 

### Description

The problem is that `poll_answers` are saving the title (as string) of the corresponding translation of `poll_question_answer` instead of the foreign key to the `poll_question_answer` id.

To illustrate this I paste the table schema:
```ruby
  create_table "poll_answers", id: :serial, force: :cascade do |t|
    t.integer "question_id"
    t.integer "author_id"
    # this line 👇 
    t.string "answer"
    t.datetime "created_at"
    t.datetime "updated_at"
    t.index ["author_id"], name: "index_poll_answers_on_author_id"
    t.index ["question_id", "answer"], name: "index_poll_answers_on_question_id_and_answer"
    t.index ["question_id"], name: "index_poll_answers_on_question_id"
  end

```

The consequences are that the result of the poll turns out to be different depending on the language each user had set when voting AND the language set for the user viewing the results.

### Quick fix

I made the following quick fix for the implementation that was facing the issue:

In `models/polls/question/answer.rb` change `total_votes` function
```ruby
  def total_votes
    result = 0
    for locale in I18n.available_locales do
      I18n.with_locale(locale) do
        result += Poll::Answer.where(question_id: question, answer: title).count + ::Poll::PartialResult.where(question: question).where(answer: title).sum(:amount)
      end
    end
    result
  end
```

The problem with this is that if an admin user changes the title of one of the answers for a question, the previous votes for that answer, that have the old title, get lost.

### Fix

I think the right way to fix this would be to add a new field `question_answer_id` in the `poll_answer` model to be independent from the translation. The bug seems to be affecting only `total_votes` functionality, but maybe I'm missing something.

taitus:
## References

* Closes #4744

## Objectives

- Make Option#total_votes independent of locale by counting Poll::Answer via option_id
- Sum PartialResult using all translated titles

### DIFF
--- a/app/models/poll/question/option.rb
+++ b/app/models/poll/question/option.rb
@@ -40,8 +40,7 @@ class Poll::Question::Option < ApplicationRecord
   end
 
   def total_votes
-    Poll::Answer.where(question_id: question, answer: title).count +
-      ::Poll::PartialResult.where(question: question).where(answer: title).sum(:amount)
+    answers.count + partial_results.sum(:amount)
   end
 
   def total_votes_percentage
--- a/spec/models/poll/question/option_spec.rb
+++ b/spec/models/poll/question/option_spec.rb
@@ -60,4 +60,83 @@ describe Poll::Question::Option do
       expect(option.with_read_more?).to be_truthy
     end
   end
+
+  describe "#total_votes" do
+    let!(:question) { create(:poll_question) }
+
+    context "with translated options" do
+      let(:option_yes) { create(:poll_question_option, question: question, title_en: "Yes", title_es: "Sí") }
+
+      it "group votes from different locales for the same option" do
+        create(:poll_answer, question: question, option: option_yes, answer: "Sí")
+        create(:poll_answer, question: question, option: option_yes, answer: "Yes")
+
+        expect(option_yes.total_votes).to eq 2
+      end
+    end
+
+    context "with options whose titles collide across locales" do
+      let!(:option_a) do
+        create(:poll_question_option, question: question, title_en: "Lead", title_es: "Plomo")
+      end
+      let!(:option_b) do
+        create(:poll_question_option, question: question, title_en: "Plomo", title_es: "Lead")
+      end
+
+      it "keeps votes isolated to the correct option" do
+        create(:poll_answer, question: question, option: option_a, answer: "Plomo")
+        create(:poll_answer, question: question, option: option_a, answer: "Lead")
+
+        expect(option_b.total_votes).to eq 0
+      end
+    end
+
+    context "with partial results" do
+      it "sums amounts by option_id" do
+        option = create(:poll_question_option, question: question, title_en: "Yes", title_es: "Sí")
+        booth_assignment = create(:poll_booth_assignment, poll: question.poll)
+
+        create(:poll_partial_result,
+               booth_assignment: booth_assignment,
+               question: question,
+               option: option,
+               answer: "Yes",
+               amount: 2)
+
+        create(:poll_partial_result,
+               booth_assignment: create(:poll_booth_assignment, poll: question.poll),
+               question: question,
+               option: option,
+               answer: "Sí",
+               amount: 3)
+
+        expect(option.total_votes).to eq 5
+      end
+    end
+
+    context "with both answers and partial results" do
+      it "sums poll answers and partial results amounts" do
+        option = create(:poll_question_option, question: question, title_en: "Yes", title_es: "Sí")
+
+        create(:poll_answer, question: question, option: option, answer: "Yes")
+        create(:poll_answer, question: question, option: option, answer: "Sí")
+
+        create(:poll_partial_result,
+               booth_assignment: create(:poll_booth_assignment, poll: question.poll),
+               question: question,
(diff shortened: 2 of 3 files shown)


app/models/poll/question/option.rb
```diff
@@ -40,8 +40,9 @@ def self.last_position(question_id)
   end
 
   def total_votes
-    Poll::Answer.where(question_id: question, answer: title).count +
-      ::Poll::PartialResult.where(question: question).where(answer: title).sum(:amount)
+    web = Poll::Answer.where(option_id: id).count
+    partials = ::Poll::PartialResult.where(option_id: id).sum(:amount)
+    web + partials
```
javierm: A couple of options here. First:

```suggestion
    answers.count + partial_results.sum(:amount)
```

Second option would be extracting methods for these ones:

```suggestion
    web_votes + booth_votes
    
    # (...)
    
    def web_votes
      # (...)
    end
      
    def booth_votes
      # (...)
    end
```

As a positive side-effect, it'll make it easier to customize these methods.

spec/models/poll/question/option_spec.rb
```diff
@@ -60,4 +60,50 @@
...
+        option = create(:poll_question_option, question: question, title_en: "Yes", title_es: "Sí")
+        booth_assignment = create(:poll_booth_assignment, poll: question.poll)
+
+        create(:poll_partial_result,
+               booth_assignment: booth_assignment,
+               question: question,
+               option: option,
+               amount: 2)
+
+        expect(option.total_votes).to eq 2
```
javierm: This test passes with the old code too :thinking:. What do you think about passing `answer:` too, so it fails with the old code?

```suggestion
        create(:poll_partial_result,
               booth_assignment: booth_assignment,
               question: question,
               option: option,
               answer: "Yes",
               amount: 2)

        create(:poll_partial_result,
               booth_assignment: create(:poll_booth_assignment, poll: question.poll),
               question: question,
               option: option,
               answer: "Sí",
               amount: 3)

        expect(option.total_votes).to eq 5
```

Should we also

spec/models/poll/question/option_spec.rb
```diff
@@ -60,4 +60,50 @@
       expect(option.with_read_more?).to be_truthy
     end
   end
+
+  describe "#total_votes" do
```
javierm, line 64: What do you think about adding a test for the scenario where there are both poll answers and partial results? :thinking:

(I know we're testing it in `question_spec.rb`.)

spec/models/poll/question_spec.rb
```diff
@@ -46,4 +46,30 @@
...
+    let!(:option_yes) { create(:poll_question_option, question: question, title_en: "Yes", title_es: "Sí") }
+    let!(:option_no) { create(:poll_question_option, question: question, title_en: "No", title_es: "No") }
+
+    before do
+      create(:poll_answer, question: question, option: option_yes, answer: "Sí")
+      create(:poll_answer, question: question, option: option_yes, answer: "Yes")
+      create(:poll_answer, question: question, option: option_no, answer: "No")
+    end
+
+    it "has the same options_total_votes in :es and :en" do
```
javierm:
```suggestion
    it "includes answers in every language" do
```

spec/models/poll/question_spec.rb
```diff
@@ -46,4 +46,30 @@
       end
     end
   end
+
+  describe "#options_total_votes" do
```
javierm, line 50: Since we're at it, shall we add a test that checks that options for other questions aren't included in the count, even if they've got the same `answer:` value?

javierm:
Sorry, I approved by accident!

spec/models/poll/question_spec.rb
```diff
@@ -46,4 +46,43 @@
...
+             booth_assignment: booth_assignment,
+             question: question,
+             option: option_yes,
+             amount: 4)
+
+      expect(question.options_total_votes).to eq 7
+    end
+
+    it "does not include votes from other questions even with same answer text" do
+      other_question = create(:poll_question)
```
javierm: Maybe the test is a bit more strict if using a question for the same poll :thinking:. What do you think?

```suggestion
      other_question = create(:poll_question, poll: question.poll)
```

spec/models/poll/question_spec.rb
```diff
@@ -46,4 +46,43 @@
...
+      other_option_yes = create(:poll_question_option,
+                                question: other_question,
+                                title_en: "Yes",
+                                title_es: "Sí")
+
+      create(:poll_answer, question: other_question, option: other_option_yes, answer: "Yes")
```
javierm, line 84: Shall we also add a partial result for the other question, so we check it isn't taken into account either?

javierm (approved):
@taitus I'm approving the pull request (this time on purpose :joy:) but I've left a couple of ideas I'm not sure about.

javierm (approved):

Issue #6047, opened 2025-07-24
The collaborative legislation text box does not fit the size of the text.


FranMerBoa:
the titles of collaborative legislation do not look good when they are very large. The text box must be adapted to the size of the text.

<img width="2802" height="676" alt="Image" src="https://github.com/user-attachments/assets/24567def-5c5e-4f76-a3ee-1d2a301091c5" />

PR #6071, opened 2025-09-06
Adjust legislation title height to prevent a layout break

KhanhVanCong:
The long question legislation process title will overflow the text box contain, block the description and make the layout so bad.

The question title box should fit the size of the text, no matter how long it is.

Increase the min-height of the quiz title container in legislation_process.scss to ensure long titles remain visually contained and do not overflow the header area. This change improves UI consistency for legislation process pages.

Issue: #6047

### DIFF
--- a/app/assets/stylesheets/legislation_process.scss
+++ b/app/assets/stylesheets/legislation_process.scss
@@ -173,13 +173,25 @@
   .quiz-header {
     margin-bottom: $line-height;
 
+    @include breakpoint(medium) {
+      display: flex;
+
+      > * {
+        @include grid-column-gutter;
+      }
+
+      .quiz-title {
+        flex-basis: 75%;
+      }
+
+      .quiz-next {
+        flex-basis: 25%;
+      }
+    }
+
     .quiz-title,
     .quiz-next {
       padding: $line-height;
-
-      @include breakpoint(medium) {
-        height: $line-height * 4;
-      }
     }
 
     .quiz-title {
@@ -200,16 +212,15 @@
     }
 
     .quiz-next-link {
-      display: block;
+      @include anchor-color;
 
       &:hover,
       &:active {
-        text-decoration: none;
+        @include anchor-color-hover;
       }
     }
 
     .quiz-next {
-      @include brand-color;
       background: #ccdbe5;
       font-size: $small-font-size;
       font-weight: bold;
@@ -219,12 +230,6 @@
 
       .icon-angle-right {
         vertical-align: middle;
-      }
-
-      &:hover,
-      &:active {
-        @extend %brand-background;
-        text-decoration: none;
       }
     }
   }
@@ -247,7 +252,7 @@
       display: inline-block;
       margin-bottom: calc($line-height / 2);
       margin-right: $line-height;
-      padding: calc($line-height / 2)  $line-height * 2;
+      padding: calc($line-height / 2) $line-height * 2;
       position: relative;
     }
 
--- a/app/views/legislation/questions/show.html.erb
+++ b/app/views/legislation/questions/show.html.erb
@@ -2,26 +2,20 @@
 
 <section class="debate-questions">
   <div class="quiz-header row small-collapse">
-    <div class="small-12 medium-9 column">
-      <div class="quiz-title">
-        <p class="quiz-header-title"><%= t("legislation.questions.show.title") %></p>
-        <strong><%= link_to @process.title, @process %></strong>
-      </div>
+    <div class="quiz-title">
+      <p class="quiz-header-title"><%= t("legislation.questions.show.title") %></p>
+      <strong><%= link_to @process.title, @process %></strong>
     </div>
-    <div class="small-12 medium-3 column">
+    <div class="quiz-next">
       <% if @question.next_question_id %>
         <%= link_to legislation_process_question_path(@process, @question.next_question_id), class: "quiz-next-link" do %>
-          <div class="quiz-next">
-            <%= t("legislation.questions.show.next_question") %>
-            <span class="icon-angle-right" aria-hidden="true"></span>
-          </div>
+          <%= t("legislation.questions.show.next_question") %>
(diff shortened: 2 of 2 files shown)


javierm:
@KhanhVanCong Thanks for the pull request! :tada:

I thought that, with this change, `.quiz-title` and `.quiz-next` would have the same height when the titles are short, but that doesn't seem to be the case :thinking:.

Maybe we could use a flex layout here, then :thinking:.

In this case, we could do some refactoring first. Does that sound appealing to you? If not, you can stop reading this message :smile:.

Here's something you can do to get familiar with how we structure the code. As you know, the code in `app/views/` is mostly code written before 2020; since then, we use `app/components/`. When we need to change a view, we usually create a component for part or all of its code.

So take this code in `app/views/legislation/questions/show.html.erb`:

```erb
  <div class="quiz-header row small-collapse">
    <div class="small-12 medium-9 column">
      <div class="quiz-title">
        <p class="quiz-header-title"><%= t("legislation.questions.show.title") %></p>
        <strong><%= link_to @process.title, @process %></strong>
      </div>
    </div>
    <div class="small-12 medium-3 column">
      <% if @question.next_question_id %>
        <%= link_to legislation_process_question_path(@process, @question.next_question_id), class: "quiz-next-link" do %>
          <div class="quiz-next">
            <%= t("legislation.questions.show.next_question") %>
            <span class="icon-angle-right" aria-hidden="true"></span>
          </div>
        <% end %>
      <% elsif @question.first_question_id %>
        <%= link_to legislation_process_question_path(@process, @question.first_question_id), class: "quiz-next-link" do %>
          <div class="quiz-next">
            <%= t("legislation.questions.show.first_question") %>
            <span class="icon-angle-right" aria-hidden="true"></span>
          </div>
        <% end %>
      <% end %>
    </div>
  </div>
```

And move it to a component. The component should be at `app/components/legislation/questions/header_component.html.erb` and `app/components/legislation/questions/header_component.rb`. I guess the initializer in the `Legislation::Questions::HeaderComponent` class will receive the `@question` object :thinking:, but if you find it easier you can pass it the `@process` object as well.

The second part is CSS. In the past, we only had a few CSS files in the project, each with more than a thousand lines of code. Now we use one CSS file per component. So take the following code from `app/assets/stylesheets/legislation_process.scss` and move it to `app/assets/stylesheets/legislation/questions/header.scss`:


```scss
  .quiz-header {
    margin-bottom: $line-height;

    .quiz-title,
    .quiz-next {
      padding: $line-height;

      @include breakpoint(medium) {
        min-height: $line-height * 4;
      }
    }

    .quiz-title {
      background: $highlight;

      .quiz-header-title {
        font-size: $small-font-size;
        font-weight: 700;
        margin-bottom: 0;
        text-transform: uppercase;
      }

      a {
        @include brand-color;
        @include header-font-size(h4);
        text-decoration: none;
      }
    }

    .quiz-next-link {
      display: block;

      &:hover,
      &:active {
        text-decoration: none;
      }
    }

    .quiz-next {
      @include brand-color;
      background: #ccdbe5;
      font-size: $small-font-size;
      font-weight: bold;
      text-align: right;
      text-transform: uppercase;
      transition: background 0.25s ease-out, background 0.25s ease-out;

      .icon-angle-right {
        vertical-align: middle;
      }

      &:hover,
      &:active {
        @extend %brand-background;
        text-decoration: none;
      }
    }
  }
```

Note you'll need to change `app/assets/stylesheets/application.scss` as well. Between these two lines:

```scss
@import "layout/**/*";
@import "machine_learning/**/*";
```

Add a line that says:

```scss
@import "legislation/**/*";
```

After that, we can change the code in the CSS and ERB files in order to use a flex layout. Are you familiar with flex layouts in CSS? If not, we can lend you a hand on this one :wink:.

javierm:
@KhanhVanCong Looks like the linters are reporting a failure; could you have a look? :pray:

app/assets/stylesheets/legislation_process.scss
```diff
@@ -200,11 +212,12 @@
     }
 
     .quiz-next-link {
-      display: block;
+      color: inherit;
 
       &:hover,
       &:active {
         text-decoration: none;
+        color: #03a8e1;
```
javierm: I like the idea of changing the behavior on hover so we no longer change the whole background :+1:. One thing to take into account: the [WebAIM contrast checker](https://webaim.org/resources/contrastchecker/) reports that this color doesn't have enough contrast with the background. So, for accessiblity and consistency, what do you think about:

```suggestion
        @include anchor-color-hover;
```

Since we no longer change the background, we can remove the `text-decoration: none;` and so it'll be more obvious that it's a link when hovering on it.

app/assets/stylesheets/legislation_process.scss
```diff
@@ -200,11 +212,12 @@
     }
 
     .quiz-next-link {
-      display: block;
+      color: inherit;
```
javierm: Here's a possible simplification:

```diff
diff --git a/app/assets/stylesheets/legislation_process.scss b/app/assets/stylesheets/legislation_process.scss
index df9f81e9f8..676b96d2a6 100644
--- a/app/assets/stylesheets/legislation_process.scss
+++ b/app/assets/stylesheets/legislation_process.scss
@@ -212,7 +212,7 @@
     }
 
     .quiz-next-link {
-      color: inherit;
+      @include anchor-color;
 
       &:hover,
       &:active {
@@ -222,7 +222,6 @@
     }
 
     .quiz-next {
-      @include brand-color;
       background: #ccdbe5;
       font-size: $small-font-size;
       font-weight: bold;
```

app/assets/stylesheets/legislation_process.scss
KhanhVanCong: I think it's perfect so far. Thank you so much ^-^

app/assets/stylesheets/legislation_process.scss
KhanhVanCong: I'll apply your suggestion

PR #6072, opened 2025-09-10
Avoid duplicate records in partial results

taitus:
## References
Related PR's:  Avoid duplicate records in poll answers #5539

## Objectives
- Avoid having the same partial result twice to the same question
- Make it possible to take internationalization into account when counting votes in the future (issue https://github.com/consuldemocracy/consuldemocracy/issues/4744)

## Release notes

:warning: We've added a new `option_id` field to the `poll_partial_results` table; a task that populates this field is part of the release tasks. If you've added custom code that affects poll partial result (which are used when voting offline in a booth), you might need to modify this task. Also, this task doesn't handle an edge case where the options for a question were renamed after the poll started (which isn't possible since pull request #4904) and an extremely rare case that only applies to multi-language polls and that we're not even sure is possible in practice. For either of these cases, you'll get a warning when running the task, so you can choose what to do about it. The task doesn't change records which already have an `option_id`, so it can be run as many times as you'd like to. Also note that, if you've got custom code to save records in the `poll_partial_results` table, you'll have to make sure that, from now on, that code populates both the `answer` column and the `option_id` column (except for open-ended questions, when `option_id` isn't necessary).

### DIFF
--- a/lib/tasks/polls.rake
+++ b/lib/tasks/polls.rake
@@ -116,33 +116,54 @@ namespace :polls do
     logger.info "Removing duplicate partial results in polls"
 
     Tenant.run_on_each do
-      duplicate_ids = Poll::PartialResult.where(option_id: nil)
-                                         .select(:question_id, :booth_assignment_id, :date, :answer)
-                                         .group(:question_id, :booth_assignment_id, :date, :answer)
-                                         .having("count(*) > 1")
-                                         .pluck(:question_id, :booth_assignment_id, :date, :answer)
+      Poll::Question.find_each do |question|
+        manageable_titles = PollPartialResultOptionFinder.new(question).manageable_choices.keys
 
-      duplicate_ids.each do |question_id, booth_assignment_id, date, answer|
-        partial_results = Poll::PartialResult.where(
-          question_id: question_id,
-          booth_assignment_id: booth_assignment_id,
-          date: date,
-          answer: answer,
-          option_id: nil
-        )
+        question.question_options.each do |option|
+          titles = option.translations.where(title: manageable_titles).select(:title).distinct
 
-        partial_results.excluding(partial_results.first).each do |partial_result|
-          partial_result.delete
+          groups = question.partial_results.where(option_id: nil, answer: titles)
+                           .select(:booth_assignment_id, :date)
+                           .group(:booth_assignment_id, :date)
+                           .having("count(*) > 1")
+                           .pluck(:booth_assignment_id, :date)
 
-          tenant_info = " on tenant #{Tenant.current_schema}" unless Tenant.default?
-          log_message = "Deleted duplicate record with ID #{partial_result.id} " \
-                        "from the #{Poll::PartialResult.table_name} table " \
-                        "with question_id #{question_id}, " \
-                        "booth_assignment_id #{booth_assignment_id}, " \
-                        "date #{date} " \
-                        "and answer #{answer}" + tenant_info.to_s
-          logger.info(log_message)
-          duplicate_records_logger.info(log_message)
+          groups.each do |booth_assignment_id, date|
+            partial_results = question.partial_results.where(
+              option_id: nil,
+              booth_assignment_id: booth_assignment_id,
+              date: date,
+              answer: titles
+            )
+
+            tenant_info = " on tenant #{Tenant.current_schema}" unless Tenant.default?
+
+            amounts_by_id = partial_results.pluck(:id, :amount).to_h
+            if amounts_by_id.values.uniq.size > 1
+              log_message = "Found duplicate partial results with different amounts " \
+                            "for question_id #{question.id}, " \
+                            "booth_assignment_id #{booth_assignment_id} " \
+                            "and date #{date}. " \
+                            "Keeping ID #{partial_results.first.id} " \
+                            "with amount #{partial_results.first.amount}. " \
+                            "Deleting partial results with these IDs and amounts: " \
+                            "#{amounts_by_id.except(partial_results.first.id)}" + tenant_info.to_s
+              logger.info(log_message)
+              duplicate_records_logger.info(log_message)
+            end
+
+            partial_results.excluding(partial_results.first).each do |partial_result|
+              partial_result.delete
+
+              log_message = "Deleted duplicate record with ID #{partial_result.id} " \
+                            "from the #{Poll::PartialResult.table_name} table " \
+                            "with question_id #{question.id}, " \
+                            "booth_assignment_id #{booth_assignment_id} " \
+                            "and date #{date}" + tenant_info.to_s
+              logger.info(log_message)
+              duplicate_records_logger.info(log_message)
+            end
+          end
         end
       end
     end
--- a/spec/lib/tasks/polls_spec.rb
+++ b/spec/lib/tasks/polls_spec.rb
@@ -258,6 +258,60 @@ describe "polls tasks" do
       expect(Poll::PartialResult.all).to match_array [result, other_result, other_booth_result]
     end
 
+    it "removes duplicate partial results in different languages" do
+      question = create(:poll_question_multiple, max_votes: 2)
+
+      create(:poll_question_option, question: question, title_en: "Yes", title_de: "Ja")
+      create(:poll_question_option, question: question, title_en: "No",  title_de: "Nein")
+      create(:poll_question_option, question: question, title_en: "Maybe", title_de: "Vielleicht")
+
+      create(:poll_partial_result,
+             question: question,
+             booth_assignment_id: booth_assignment.id,
+             date: Date.current,
+             answer: "Yes",
+             option: nil)
+      create(:poll_partial_result,
+             question: question,
+             booth_assignment_id: booth_assignment.id,
+             date: Date.current,
+             answer: "Ja",
+             option: nil)
+
+      expect(Poll::PartialResult.count).to eq 2
+
+      Rake.application.invoke_task("polls:remove_duplicate_partial_results")
+
+      expect(Poll::PartialResult.count).to eq 1
+    end
+
+    it "does not remove duplicate partial results when many options are possible" do
+      question = create(:poll_question, title: "How do you pronounce it?")
+
+      create(:poll_question_option, question: question, title_en: "A", title_es: "EI")
+      create(:poll_question_option, question: question, title_en: "E", title_es: "I")
+      create(:poll_question_option, question: question, title_en: "I", title_es: "AI")
+
+      create(:poll_partial_result,
+             question: question,
+             booth_assignment_id: booth_assignment.id,
+             date: Date.current,
+             answer: "I",
+             option: nil)
+      create(:poll_partial_result,
+             question: question,
+             booth_assignment_id: booth_assignment.id,
+             date: Date.current,
+             answer: "AI",
+             option: nil)
+
+      expect(Poll::PartialResult.count).to eq 2
+
+      Rake.application.invoke_task("polls:remove_duplicate_partial_results")
+
+      expect(Poll::PartialResult.count).to eq 2
+    end
+
     it "does not remove partial results with the same text and different options" do
       question = create(:poll_question_multiple, :abcde, max_votes: 4)
       option_a = question.question_options.find_by(title: "Answer A")
@@ -304,6 +358,34 @@ describe "polls tasks" do
       Tenant.switch("partial_results") do
         expect(Poll::PartialResult.count).to eq 1
       end
+    end
+
+    it "removes duplicates in different languages even when amounts differ" do
+      question = create(:poll_question_multiple, max_votes: 2)
+
+      create(:poll_question_option, question: question, title_en: "Yes", title_de: "Ja")
+
+      create(:poll_partial_result,
+             question: question,
+             booth_assignment_id: booth_assignment.id,
+             date: Date.current,
+             answer: "Yes",
+             option: nil,
+             amount: 3)
+
+      create(:poll_partial_result,
+             question: question,
+             booth_assignment_id: booth_assignment.id,
+             date: Date.current,
+             answer: "Ja",
+             option: nil,
+             amount: 5)
+
+      expect(Poll::PartialResult.count).to eq 2
+
+      Rake.application.invoke_task("polls:remove_duplicate_partial_results")
+
+      expect(Poll::PartialResult.count).to eq 1
     end
   end
 
@@ -381,6 +463,11 @@ describe "polls tasks" do
     it "removes duplicate partial results before populating the option_id column" do
       question = create(:poll_question_multiple, :abc)
 
+      localized_question = create(:poll_question_multiple)
+      create(:poll_question_option, question: localized_question, title_en: "Yes", title_de: "Ja")
+      create(:poll_question_option, question: localized_question, title_en: "No",  title_de: "Nein")
+      create(:poll_question_option, question: localized_question, title_en: "Maybe", title_de: "Vielleicht")
+
       result_attributes = {
         question_id: question.id,
         booth_assignment_id: booth_assignment.id,
@@ -391,14 +478,28 @@ describe "polls tasks" do
       result = create(:poll_partial_result, result_attributes)
       insert(:poll_partial_result, result_attributes)
 
+      localized_result_attributes = {
+        question: localized_question,
+        booth_assignment_id: booth_assignment.id,
+        date: Date.current,
+        option: nil
+      }
+      localized_result = create(:poll_partial_result, localized_result_attributes.merge(answer: "Yes"))
+      create(:poll_partial_result, localized_result_attributes.merge(answer: "Ja"))
+
       result.reload
+      localized_result.reload
+
       expect(result.option_id).to be nil
+      expect(localized_result.option_id).to be nil
 
       Rake.application.invoke_task("polls:populate_partial_results_option_id")
       result.reload
+      localized_result.reload
 
-      expect(Poll::PartialResult.count).to eq 1
+      expect(Poll::PartialResult.count).to eq 2
       expect(result.option_id).to eq question.question_options.find_by(title: "Answer A").id
+      expect(localized_result.option_id).to eq localized_question.question_options.find_by(title: "Yes").id
     end
 
     it "populates the option_id column on tenants" do


app/controllers/officing/results_controller.rb
```diff
@@ -82,7 +81,7 @@ def build_recounts
     def go_back_to_new(alert = nil)
```
javierm: Since now we're always passing the parameter, maybe we can write:

```suggestion
    def go_back_to_new(alert)
```

spec/models/poll/partial_result_spec.rb
```diff
@@ -21,6 +21,75 @@
       expect(build(:poll_partial_result, origin: "custom")).to be_valid
       expect(build(:poll_partial_result, origin: "web")).not_to be_valid
     end
+
+    describe "check uniqueness option" do
```
javierm: Here "option" made me think about an option parameter in a method :thinking:. Maybe we can also omit "check", since we're in a test :thinking:.

```suggestion
    describe "option_id uniqueness" do
```

What do you think?

lib/tasks/polls.rake
```diff
@@ -107,4 +107,82 @@ namespace :polls do
...
+            partial_results = Poll::PartialResult.where(
+              option_id: nil,
+              question_id: question.id,
+              booth_assignment_id: booth_assignment_id,
+              date: date,
+              answer: titles
+            )
+
+            partial_results.excluding(partial_results.first).each do |partial_result|
+              partial_result.delete
```
javierm, line 156: What should we do when there are two partial results with different amounts (unlikely, I know)? Back when we deleted duplicate poll answers, we kept the first one because there was no way to know which answer was the right one, but here public workers might have a way to find which one is the right one :thinking:.

My suggestion would be to keep deleting everything except the first one but, if there was a different amount, add a warning about it in the log message. That way public workers could double-check which amount is the right one.

app/models/poll/partial_result.rb
```diff
@@ -5,13 +5,15 @@ class Poll::PartialResult < ApplicationRecord
   belongs_to :author, ->   { with_hidden }, class_name: "User", inverse_of: :poll_partial_results
   belongs_to :booth_assignment
   belongs_to :officer_assignment
+  belongs_to :option, class_name: "Poll::Question::Option"
```
javierm, line 8: Wasn't sure where to add this comment :relieved:.

Should we also change `app/views/admin/poll/results/_result.html.erb` and `app/views/officing/results/index.html.erb`? There's code there that says:

```erb
<% by_question = @partial_results.group_by(&:question_id) %>

<%# (...) %>

        <% by_answer = by_question[question.id].present? ? by_question[question.id].group_by(&:answer) : {} %>

<%# (...) %>

          <td class="text-center"><%= by_answer[option.title].present? ? by_answer[option.title].sum(&:amount) : 0 %></td>
```

lib/tasks/polls.rake
```diff
@@ -107,4 +107,82 @@ namespace :polls do
...
+    logger.info "Removing duplicate partial results in polls"
+
+    Tenant.run_on_each do
+      Poll::Question.find_each do |question|
+        manageable_titles = PollPartialResultOptionFinder.new(question).manageable_choices.keys
+
+        question.question_options.each do |option|
+          titles = option.translations.where(title: manageable_titles).select(:title).distinct
+
+          groups = Poll::PartialResult.where(option_id: nil, question_id: question.id, answer: titles)
```
javierm: Should we use the association here? Or is there a reason not to do so? :thinking:

```suggestion
          groups = question.partial_results.where(option_id: nil, answer: titles)
```

lib/tasks/polls.rake
```diff
@@ -107,4 +107,82 @@ namespace :polls do
...
+          tenant_info = " on tenant #{Tenant.current_schema}" unless Tenant.default?
+
+          if ids.count == 0
+            logger.warn "Skipping poll_partial_results with the text \"#{choice}\" for the poll_question " \
+                        "with ID #{question.id}. This question has no poll_question_options " \
+                        "containing the text \"#{choice}\"" + tenant_info.to_s
+          else
+            logger.warn "Skipping poll_partial_results with the text \"#{choice}\" for the poll_question " \
+                        "with ID #{question.id}. The text \"#{choice}\" could refer to any of these " \
+                        "IDs in the poll_question_options table: #{ids.join(", ")}" + tenant_info.to_s
```
javierm: The name of the table is still `poll_questions_answers` :sweat_smile:.

```suggestion
                        "IDs in the poll_question_answers table: #{ids.join(", ")}" + tenant_info.to_s
```

lib/tasks/polls.rake
```diff
@@ -107,4 +107,82 @@ namespace :polls do
...
+          Poll::PartialResult.where(question_id: question.id, option_id: nil, answer: choice)
+                             .update_all(option_id: ids.first)
+        end
+
+        option_finder.unmanageable_choices.each do |choice, ids|
+          tenant_info = " on tenant #{Tenant.current_schema}" unless Tenant.default?
+
+          if ids.count == 0
+            logger.warn "Skipping poll_partial_results with the text \"#{choice}\" for the poll_question " \
+                        "with ID #{question.id}. This question has no poll_question_options " \
```
javierm: Since the name of the table is still `poll_questions_answers` and we're probably going to use this name in the `else` statement (see my comment there):

```suggestion
                        "with ID #{question.id}. This question has no poll_question_answers " \
```

app/lib/poll_partial_result_option_finder.rb
```diff
@@ -0,0 +1,10 @@
+class PollPartialResultOptionFinder < PollOptionFinder
+  private
+
+    def existing_choices
+      Poll::PartialResult
+        .where(question_id: question.id, option_id: nil)
+        .distinct
+        .pluck(:answer)
```
javierm: As mentioned in another comment, shall we use the association here?

```suggestion
      question.partial_results.where(option_id: nil).distinct.pluck(:answer)
```

lib/tasks/polls.rake
```diff
@@ -107,4 +107,96 @@ namespace :polls do
...
+
+            amounts_by_id = partial_results.pluck(:id, :amount).to_h
+            if amounts_by_id.values.uniq.size > 1
+              log_message = "Found duplicate partial results with different amounts " \
+                            "for question_id #{question.id}, " \
+                            "booth_assignment_id #{booth_assignment_id} " \
+                            "and date #{date}. " \
+                            "Keeping ID #{partial_results.first.id} " \
+                            "with amount #{partial_results.first.amount}. " \
+                            "Others: #{amounts_by_id.except(partial_results.first.id)}" + tenant_info.to_s
```
javierm: Maybe, to contrast with "keeping", we can use "deleting" :thinking:. And then clarify what the values in the hash are about.

```suggestion
                            "Deleting partial results with these IDs and amounts: #{amounts_by_id.except(partial_results.first.id)}" + tenant_info.to_s
```

I know the line I wrote is too long; I'll leave fixing that to you :wink:.

spec/system/officing/results_spec.rb
```diff
@@ -90,6 +90,7 @@
       booth_assignment: poll_officer.officer_assignments.first.booth_assignment,
       date: Date.current,
       question: question_1,
+      option: question_1.question_options.first,
```
javierm: What do you think about changing the `poll_partial_result` factory so we can pass either `option` or `answer` but don't need to pass both of them? Maybe we could do something similar to what we did to the `poll_answer` factory?

spec/system/officing/results_spec.rb
```diff
@@ -167,13 +161,25 @@
...
-    question_1.question_options.each_with_index do |answer, i|
-      within("#question_#{question_1.id}_#{i}_result") { expect(page).to have_content(answer.title) }
+    page.find("tr#question_#{question_1.id}_0_result") do |yes_result|
+      expect(yes_result).to have_css "td", text: "Yes"
+      expect(yes_result).to have_css "td", text: "33"
+    end
+
+    page.find("tr#question_#{question_1.id}_1_result") do |no_result|
+      expect(no_result).to have_css "td", text: "No"
+      expect(no_result).to have_css "td", text: "0"
```
javierm: Would it make sense to move these lines to the `Admin::Poll::Results::QuestionComponent` component tests? The "Edit result" test already covers user interaction in this section :thinking:.

app/components/admin/poll/results/question_component.rb
```diff
@@ -0,0 +1,23 @@
+class Admin::Poll::Results::QuestionComponent < ApplicationComponent
+  attr_reader :question, :booth_assignment, :date
+
+  def initialize(question, booth_assignment: nil, date: nil)
```
javierm: I think I liked it better when we passed `partial_results` to this component instead of passing `booth_assignment` and `date` :thinking:. If the `group_by(&:question_id)` was the problem, we could write:

```ruby
def by_option
  @by_option ||= partial_results.where(question: question).group_by(&:option_id)
end
```

I haven't tested it, though, so I might be missing something :thinking:.

app/components/admin/poll/results/question_component.rb
```diff
@@ -0,0 +1,23 @@
...
+    @question = question
+    @booth_assignment = booth_assignment
+    @date = date
+  end
+
+  def votes_for(option_id)
```
javierm: The commit for this change says:

> Add votes_for(option_title) helper

It's not a helper, though :thinking:.

app/components/admin/poll/results/question_component.html.erb
```diff
@@ -0,0 +1,17 @@
...
+    <tr>
+      <th><%= t("admin.results.result.table_answer") %></th>
+      <th class="text-center"><%= t("admin.results.result.table_votes") %></th>
+    </tr>
+  </thead>
+  <tbody>
+    <% question.question_options.each_with_index do |option, i| %>
+      <tr id="question_<%= question.id %>_<%= i %>_result">
+        <td><%= option.title %></td>
+        <td class="text-center"><%= votes_for(option.id) %></td>
```
javierm:
```suggestion
        <td class="text-center"><%= votes_for(option) %></td>
```

Then, inside the method we could use `id`, `title`, ... Making it easier for other institutions to customize the method.

app/components/admin/poll/results/question_component.rb
```diff
@@ -0,0 +1,23 @@
...
+    grouped = by_option[option_id] || []
+    grouped.sum(&:amount)
```
javierm: Would this make sense?

```suggestion
  partial_results.where(option: option).sum(&:amount)
```

app/components/officing/results/index_component.html.erb
```diff
@@ -0,0 +1,41 @@
...
+            <td id="null_results"><%= recounts.sum(:null_amount) %></td>
+            <td id="total_results"><%= recounts.sum(:total_amount) %></td>
+          </tr>
+        </tbody>
+      </table>
+
+      <% @poll.questions.each do |question| %>
+        <%= render Admin::Poll::Results::QuestionComponent.new(question,
+                                                               booth_assignment: @booth_assignment,
+                                                               date: @date) %>
```
javierm: Where is `@date` set? :thinking:

app/components/officing/results/index_component.html.erb
```diff
@@ -0,0 +1,41 @@
+<%= back_link_to new_officing_poll_result_path(poll) %>
```
javierm, line 3: To follow the convention for "main page components":

```suggestion
<% provide :main_class, "officing-results-index" %>
<%= back_link_to new_officing_poll_result_path(poll) %>
```

app/components/admin/poll/results/question_component.html.erb
```diff
@@ -0,0 +1,17 @@
+<h3><%= question.title %></h3>
```
javierm: Shall we add a `<div>` tag for the whole component, as we usually do in components? With a class like `admin-poll-results-question` :thinking:.

spec/components/admin/poll/results/question_component_spec.rb
```diff
@@ -0,0 +1,154 @@
+require "rails_helper"
+
+describe Admin::Poll::Results::QuestionComponent, :admin do
```
javierm: This component is also called from the officing section, so is the `:admin` tag necessary? :thinking:

spec/components/admin/poll/results/question_component_spec.rb
```diff
@@ -0,0 +1,163 @@
...
+  let(:poll) { create(:poll) }
+  let(:question) { create(:poll_question, poll: poll, title: "What do you want?") }
+
+  before do
+    create(:poll_question_option, question: question, title: "Yes")
+    create(:poll_question_option, question: question, title: "No")
+  end
+
+  it "renders question title and headers" do
+    render_inline Admin::Poll::Results::QuestionComponent.new(question, Poll::PartialResult.none)
```
javierm: In order to show the way this component is expected to be used, what do you think about:

```suggestion
    render_inline Admin::Poll::Results::QuestionComponent.new(question, poll.partial_results)
```

spec/components/admin/poll/results/question_component_spec.rb
```diff
@@ -0,0 +1,163 @@
...
+    end
+
+    page.find("tr#question_#{question.id}_1_result") do |no_result|
+      expect(no_result).to have_css "td", text: "No"
+      expect(no_result).to have_css "td", text: "5"
+    end
+  end
+
+  it "shows 0 when an option has no partial results" do
+    render_inline Admin::Poll::Results::QuestionComponent.new(question, Poll::PartialResult.none)
```
javierm: This test would pass even if the option had partial results, because we're passing `Poll::PartialResult.none` :thinking:.

```suggestion
    render_inline Admin::Poll::Results::QuestionComponent.new(question, poll.partial_results)
```

spec/components/admin/poll/results/question_component_spec.rb
```diff
@@ -0,0 +1,163 @@
...
+    expect(page).to have_css "td", text: "Yes"
+    expect(page).to have_css "td", text: "No"
+  end
+
+  it "sums votes by answer title" do
+    create(:poll_partial_result, question: question, answer: "Yes", amount: 2)
+    create(:poll_partial_result, question: question, answer: "No", amount: 5)
+
+    partials = Poll::PartialResult.where(question_id: question.id)
+    render_inline Admin::Poll::Results::QuestionComponent.new(question, partials)
```
javierm:
```suggestion
    render_inline Admin::Poll::Results::QuestionComponent.new(question, question.partial_results)
```

spec/components/admin/poll/results/question_component_spec.rb
```diff
@@ -0,0 +1,163 @@
...
+  end
+
+  it "ignores partial results from other questions" do
+    other_question = create(:poll_question, poll: poll)
+    create(:poll_question_option, question: other_question, title: "Yes")
+    create(:poll_question_option, question: other_question, title: "Irrelevant")
+    create(:poll_partial_result, question: other_question, answer: "Yes", amount: 9)
+    create(:poll_partial_result, question: other_question, answer: "Irrelevant", amount: 9)
+
+    render_inline Admin::Poll::Results::QuestionComponent.new(question, Poll::PartialResult.all)
```
javierm:
```suggestion
    render_inline Admin::Poll::Results::QuestionComponent.new(question, poll.partial_results)
```

spec/components/admin/poll/results/question_component_spec.rb
```diff
@@ -0,0 +1,163 @@
...
+      end
+
+      page.find("tr#question_#{question.id}_1_result") do |no_result|
+        expect(no_result).to have_css "td", text: "No"
+        expect(no_result).to have_css "td", text: "0"
+      end
+    end
+  end
+
+  context "with date" do
```
javierm: Are these tests still necessary, now that we pass partial results to the initializer instead of a booth and a date :thinking:.

spec/components/admin/poll/results/question_component_spec.rb
```diff
@@ -0,0 +1,163 @@
...
+    create(:poll_question_option, question: other_question, title: "Irrelevant")
+    create(:poll_partial_result, question: other_question, answer: "Yes", amount: 9)
+    create(:poll_partial_result, question: other_question, answer: "Irrelevant", amount: 9)
+
+    render_inline Admin::Poll::Results::QuestionComponent.new(question, Poll::PartialResult.all)
+
+    expect(page).to have_css "td", text: "0", count: 2
+  end
+
+  context "with booth_assignment" do
```
javierm: Are these tests still necessary, now that we pass partial results to the initializer instead of a booth and a date :thinking:.

spec/components/admin/poll/results/question_component_spec.rb
```diff
@@ -0,0 +1,163 @@
...
+
+    expect(page).to have_css "tbody tr", count: 2
+    expect(page).to have_css "td", text: "Yes"
+    expect(page).to have_css "td", text: "No"
+  end
+
+  it "sums votes by answer title" do
+    create(:poll_partial_result, question: question, answer: "Yes", amount: 2)
+    create(:poll_partial_result, question: question, answer: "No", amount: 5)
+
```
javierm: Since the title of the test says, "sums votes", shall we add a second partial result for at least one of the options? Then we could test that the sum is correctly displayed :thinking:. We could use a different date or booth if the unique index is the problem :thinking:.

app/models/poll/partial_result.rb
```diff
@@ -5,13 +5,15 @@ class Poll::PartialResult < ApplicationRecord
...
 
   validates :question, presence: true
   validates :author, presence: true
   validates :answer, presence: true
   validates :answer, inclusion: { in: ->(a) { a.question.possible_answers }},
                      unless: ->(a) { a.question.blank? }
```
javierm: What do you think about changing this validation rule so it uses `option.possible_answers` instead? We could do the same in `Poll::Answer` and then move the `Poll::Question#possible_answers` method to `Poll::Question::Option` :thinking:.

Also, we could change the `a` variable name. `a` might be a bad variable name for a poll answer, but it's definitely a terrible name for a poll partial result :joy:.

spec/components/admin/poll/results/question_component_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+    expect(page).to have_css "tbody tr", count: 2
+    expect(page).to have_css "td", text: "Yes"
+    expect(page).to have_css "td", text: "No"
+  end
+
+  it "sums votes by answer title" do
+    create(:poll_partial_result, question: question, answer: "Yes", amount: 2, date: Date.current)
+    create(:poll_partial_result, question: question, answer: "Yes", amount: 1, date: Date.yesterday)
+    create(:poll_partial_result, question: question, answer: "No", amount: 5)
```
javierm: We no longer sum votes by answer title but by option :thinking:. I know this works because the factories automatically add the `option` field :relieved:. But, still, what do you think about changing this test in the commit where we introduce the `option_id` column so it uses `option:` instead of `answer:`?

PR #6076, opened 2025-09-18
 Add setting to require consent for notifications 

johanndev1208:
Screenshot 2025-09-18 at 21 39 10

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,7 +1,7 @@
 class User < ApplicationRecord
   include Verification
   attribute :registering_from_web, default: false
-  %i[newsletter email_digest email_on_direct_message recommended_debates
+  %i[newsletter email_digest email_on_direct_message public_activity recommended_debates
      recommended_proposals].each do |field|
     attribute field, :boolean, default: -> { !Setting["feature.gdpr.require_consent_for_notifications"] }
   end
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -146,6 +146,20 @@ describe User do
       end
     end
 
+    describe "#public_activity" do
+      it "is true by default when the consent for notifications setting is disabled" do
+        Setting["feature.gdpr.require_consent_for_notifications"] = false
+
+        expect(build(:user).public_activity).to be true
+      end
+
+      it "is false by default when the consent for notifications setting is enabled" do
+        Setting["feature.gdpr.require_consent_for_notifications"] = true
+
+        expect(build(:user).public_activity).to be false
+      end
+    end
+
     describe "#recommended_debates" do
       it "is true by default when the consent for notifications setting is disabled" do
         Setting["feature.gdpr.require_consent_for_notifications"] = false
--- a/db/migrate/20251009085528_remove_default_value_in_user_public_activity.rb
+++ b/db/migrate/20251009085528_remove_default_value_in_user_public_activity.rb
@@ -1,0 +1,7 @@
+class RemoveDefaultValueInUserPublicActivity < ActiveRecord::Migration[7.1]
+  def change
+    change_table :users do |t|
+      t.change_default :public_activity, from: true, to: nil
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.1].define(version: 2025_10_09_085327) do
+ActiveRecord::Schema[7.1].define(version: 2025_10_09_085528) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "pg_trgm"
   enable_extension "plpgsql"
@@ -1608,7 +1608,7 @@ ActiveRecord::Schema[7.1].define(version: 2025_10_09_085327) do
     t.datetime "level_two_verified_at", precision: nil
     t.string "erase_reason"
     t.datetime "erased_at", precision: nil
-    t.boolean "public_activity", default: true
+    t.boolean "public_activity"
     t.boolean "newsletter"
     t.integer "notifications_count", default: 0
     t.boolean "registering_with_oauth", default: false
--- a/spec/factories/users.rb
+++ b/spec/factories/users.rb
@@ -7,7 +7,6 @@ FactoryBot.define do
     terms_of_service    { "1" }
     confirmed_at        { Time.current }
     date_of_birth       { 20.years.ago }
-    public_activity     { true }
 
     trait :incomplete_verification do
       after :create do |user|


app/models/user.rb
```diff
@@ -443,6 +462,12 @@ def slug
 
   private
 
+    def with_notification_setting
+      return false if Setting["feature.notifications"]
```
javierm: Returning `false` when checking whether we should send notifications because a feature called `notifications` is enabled might be a bit confusing :thinking:. I think we've got at least two options:

1. Rename the feature to something like `feature.disable_notifications`.
2. Keep using `feature.notifications` but change its meaning, so `true` means that notifications are enabled. This would require changing the description of the feature as well :relieved:.

What do you think? Any preferences? Are there any other options that we could use?

app/models/setting.rb
```diff
@@ -65,6 +65,7 @@ def mime_types
 
     def defaults
       {
+        "feature.notifications": true,
```
javierm: This value will be used in all existing installations when upgrading, when the task `settings:add_new_settings` runs as part of the upgrade process. This means that this would change the behavior in existing installations, which might not be what they expect when upgrading :thinking:.

For compatibility reasons, we could set it to `false` and, then, after releasing a new version, change it to `true`. For example, we did something like that for SDG (see commit 797f61c0ca0).

config/locales/en/settings.yml
```diff
@@ -94,6 +94,8 @@ en:
       oidc_login: "OpenID Connect login"
       oidc_login_description: "Allow users to sign up with OpenID Connect (OIDC)"
       saml_login: "SAML login"
+      notifications: "GDPR-compliant default settings"
+      notifications_description: Activate the GDPR-compliant default settings to ensure “privacy by design”.
```
javierm: Could you add to the commit message a link to the section of the GDPR that makes this setting necessary? It'd be great if you could add a quote of the relevant part of the GDPR too :pray:. In particular, it'd be great to have an explanation about why making the `email_digest` and `newsletter` columns `false` by default wouldn't be enough to comply with the GDPR :thinking:.

app/views/account/show.html.erb
```diff
@@ -45,25 +45,25 @@
             <h2><%= t("account.show.notifications") %></h2>
 
             <div>
-              <%= f.check_box :email_on_comment %>
+              <%= f.check_box :email_on_comment, disabled: Setting["feature.notifications"] %>
```
javierm: I'd vote for not showing these fields at all instead of disabling them when notifications are disabled :thinking:. If we disable them, citizens might wonder why those fields are there and whether they can enable them.

In any case, would it be possible to add a test checking what happens to these fields when this feature is enabled?

app/models/user.rb
```diff
@@ -98,12 +98,12 @@ class User < ApplicationRecord
   scope :officials,      -> { where(official_level: 1..) }
   scope :male,           -> { where(gender: "male") }
   scope :female,         -> { where(gender: "female") }
-  scope :newsletter,     -> { where(newsletter: true) }
+  scope :newsletter,     -> { Setting["feature.notifications"] ? where(newsletter: true) : none }
```
javierm: When this setting is enabled, should we also hide the features in the admin section that send newsletters and email digests? :thinking:

app/models/user.rb
```diff
@@ -98,12 +98,12 @@ class User < ApplicationRecord
...
   scope :for_render,     -> { includes(:organization) }
   scope :by_document,    ->(document_type, document_number) do
     where(document_type: document_type, document_number: document_number)
   end
-  scope :email_digest,   -> { where(email_digest: true) }
+  scope :email_digest,   -> { Setting["feature.notifications"] ? where(email_digest: true) : none }
```
javierm: What do you think about adding a test for the scopes, checking what happens when this setting is enabled?

javierm (commented):
@johanndev1208 Thanks once again for the pull request! I've left a few comments; let me know what you think :wink:.

app/models/user.rb
itw-smart: Thanks Javi, I agree that the current name is confusing. I will adjust the naming to something clearer, e.g. feature.disable_notifications, so the flag’s behavior is self-explanatory.

config/locales/en/settings.yml
itw-smart: Absolutely. We will update the commit message with a GDPR Article 25 reference and add a short quote. We will also clarify why simply setting email_digest and newsletter to false is not sufficient for compliance.

Ensure GDPR compliance by default (Article 25 GDPR – privacy by design and by default).
Notifications and optional features are disabled unless explicitly enabled by the user, so that only data necessary for the intended purpose is processed.

app/views/account/show.html.erb
itw-smart: We agree, hiding the fields is a better UX choice. We will remove them when notifications are disabled and add tests to cover both enabled and disabled cases.

app/models/user.rb
itw-smart: I dont exactly get what you mean here

app/models/user.rb
itw-smart: Good idea. We will add scope tests to ensure the queries behave correctly depending on the flag.

app/models/setting.rb
itw-smart: Good point. To avoid unexpected behavior during upgrades, we will set the default to false (notifications disabled). We can switch to true in a later release once it’s stable and clearly documented or the operator of the installations should change it manually.

app/models/user.rb
javierm: In the admin section, if you go to "Messages to users" -> "Newsletters", you can create and send newsletters to users subscribed to the newsletter. Similarly, if you go to "Messages to users" -> "System emails", there's a button to send the "Proposal notification digest" to the users returned by `User.email_digest`.

But when this option is enabled, there'll be no users to send the newsletter and email digest to :thinking:. I'm wondering whether it makes sense to hide the option to send the newsletter and emai digest in this case :thinking:.

johanndev1208: But before `Setting["feature.disable_notifications"]` changes `User.email_digest` could be empty.

johanndev1208:
@javierm Could you please run linters and tests?

javierm:
Hi, @johanndev1208 :smile:.

We've sent you an invitation to be a collaborator of the repo. Once you accept it, the tests and linters will run automatically when you push changes.

itw-smart:
Hi @javierm, the 5 failing tests are also failing on master branch without any changes from us. What can we do about them because they fail not because of our developments - or do we miss anything?

app/models/user.rb
javierm: @johanndev1208 Do you mean we should also disable the "Proposal notification digest" when `User.email_digest` is empty? If so, that's an interesting idea :thinking:.

To further discuss this topic I think first we'd need to know why making the `email_digest` and `newsletter` columns false by default wouldn't be enough to comply with the GDPR 🤔. As mentioned in another comment, could you add it to the commit message introducing this feature?

johanndev1208: @javierm I do not want to change behaviour of the "Proposal notification digest" as new featrure does not affect it.

itw-smart:
> Hi, @johanndev1208 😄.
> 
> Thanks for adding the GDPR reference in the commit! Having said that:
> 
> > Ensure GDPR compliance by default (Article 25 GDPR – privacy by design and by default)
> > Notifications and optional features are disabled unless explicitly enabled by the user, so that only data necessary for the intended purpose is processed.
> 
> I might be missing something here 🤔. If we set the the `email_digest` and `newsletter` columns to `false` by default, notifications will be disabled unless explicitly enabled by the user, right?
> 
> So it looks like that would be enough, right? For example, the [GDPR Newsletter Consent Examples](https://gdprinfo.eu/best-gdpr-newsletter-consent-examples-a-complete-guide-to-compliant-email-marketing) show checkboxes that aren't checked by default. Administrators having the option to disable newsletters completely doesn't seem to be necessary 🤔.
> 
> Maybe I'm wrong here 🤔. If so, could you send me a link to the article that quote comes from? I can't find it in the [GDPR article 25](https://gdpr-info.eu/art-25-gdpr) or the [GDPR article 25 explanation](https://gdpr-text.com/read/article-25/).

Hi Javi,

Yes, you’re right. If the GDPR compliance setting is enabled, all notification checkboxes for new users should be unchecked by default and must be actively selected by the user. The intention is not to disable the entire feature, but to ensure that consent is given proactively.

app/models/setting.rb
```diff
@@ -65,6 +65,7 @@ def mime_types
 
     def defaults
       {
+        "feature.disable_notifications": false,
```
javierm: Judging from the "Warnings for links out" pull request and what @itw-smart mentioned, there are going to be a few GDPR-related settings, so maybe we could add `gdpr` here :thinking:. Also, since with the latest changes this no longer disables notifications, what do you think about renaming the setting? Maybe something like:

```suggestion
        "feature.gdpr.require_consent_for_notifications": false,
```

app/models/user.rb
```diff
@@ -443,6 +444,16 @@ def slug
 
   private
 
+    def set_defaults
+      if Setting["feature.disable_notifications"]
+        self.newsletter = false
+        self.email_digest = false
+        self.email_on_direct_message = false
+        self.recommended_debates = false
+        self.recommended_proposals = false
```
javierm: These ones aren't really notifications, are they? :thinking: If changing them is necessary, could you mention the reason in the commit message?

app/models/user.rb
```diff
@@ -122,6 +122,7 @@ class User < ApplicationRecord
     where(date_of_birth: start_date.beginning_of_day..end_date.end_of_day)
   end
 
+  after_initialize :set_defaults, if: :new_record?
```
javierm: Would it make sense to use the `attribute` method here?

```suggestion
  %i[newsletter email_digest email_on_direct_message].each do |notifications_attribute|
    attribute notifications_attribute, default: -> { !Setting["feature.disable_notifications"] }
  end
```

config/locales/en/settings.yml
```diff
@@ -94,6 +94,8 @@ en:
       oidc_login: "OpenID Connect login"
       oidc_login_description: "Allow users to sign up with OpenID Connect (OIDC)"
       saml_login: "SAML login"
+      disable_notifications: "GDPR-compliant default settings"
+      disable_notifications_description: Activate the GDPR-compliant default settings to ensure “privacy by design”.
```
javierm: Since there are going to be a few GDPR settings and this is only one of them, maybe we could change these ones a little bit :thinking:. Maybe something like:

```suggestion
      disable_notifications: "Explicit newsletters consent"
      disable_notifications_description: Require explicit consent to send newsletters and notifications emails to ensure "privacy by design" as required by the GDPR.
```

@itw-smart This is just a draft; feel free to suggest a better title and description.

javierm (commented):
@johanndev1208 Looking better! :pray: I've left a couple of comments; let me know what you think!

app/models/user.rb
johanndev1208: It was in requirements. @itw-smart could you have a look, please.

app/models/user.rb
johanndev1208: Good point!
In case we do not have complex logic anymore, `attribute` is it better.

db/dev_seeds/settings.rb
```diff
@@ -2,6 +2,7 @@
   Setting.reset_defaults
 
   {
+    "feature.gdpr.require_consent_for_notifications": false,
```
javierm: Right now this isn't necessary because it uses the same value as the production settings. We can either remove it or make it true by default for development. Either option is fine with me; up to you! :wink:

spec/models/user_spec.rb
```diff
@@ -93,32 +93,204 @@
...
+      context "when GDPR-compliant default settings are on" do
+        let(:new_user) { build(:user) }
+
+        before do
+          Setting["feature.gdpr.require_consent_for_notifications"] = true
+        end
+
+        it "is false by default" do
+          expect(new_user.email_on_comment).to be false
+        end
```
javierm: Since the setting doesn't affect `email_on_comment`, maybe we could remove these changes and leave the `it "is false by default" do` test as it was :thinking:. What do you think?

spec/models/user_spec.rb
```diff
@@ -93,32 +93,204 @@
...
+      context "when GDPR-compliant default settings are on" do
+        let(:new_user) { build(:user) }
+
+        before do
+          Setting["feature.gdpr.require_consent_for_notifications"] = true
+        end
+
+        it "is false by default" do
+          expect(new_user.email_on_comment_reply).to be false
+        end
```
javierm: Same here: maybe we can leave the test as it was because the setting doesn't affect `email_on_comment_reply` :thinking:.

spec/models/user_spec.rb
```diff
@@ -93,32 +93,204 @@
...
-      it "is true by default" do
-        expect(subject.newsletter).to be true
+    describe "newsletter" do
+      context "when scope" do
+        let!(:user_1) { create(:user, email_digest: true) }
+        let!(:user_2) { create(:user, email_digest: false) }
+
+        it "has correct users" do
+          expect(User.email_digest.to_a).to include(user_1)
+          expect(User.email_digest.to_a).not_to include(user_2)
```
javierm: Is there a typo here? We're checking the `email_digest` behavior inside the `describe "newsletter" do` block :thinking:.

javierm (commented):
Sorry I missed these ones earlier :pray:. Nice job renaming the setting, by the way :+1:; I completely forgot that you needed to change the structure in `settings.yml`

javierm:
Understood! If it isn't clear whether recommendations are notifications, we can include them to be safe :ok_hand:.

Later today or tomorrow I'll add the Spanish texts and merge this pull request. Thanks a lot both of you! :pray:

PR #6084, opened 2025-09-30
Verification links with expiration time

johanndev1208:
**Here's a quick summary of some points to check**
Added expiry for email verification links (valid for 72 hours only)

## References

According [doc](https://github.com/heartcombo/devise/blob/cf93de390a29654620fdf7ac07b4794eb95171d0/lib/devise/models/recoverable.rb#L90C16-L90C19) currently token is used in invalid way. 

## Objectives

- Added expiry for email verification links (valid for 72 hours only)
- Aligns with GDPR principles of data minimization (Art. 5(1)(c))
  and storage limitation (Art. 5(1)(e))
- Improves overall account security (Art. 32 GDPR)

### DIFF
--- a/app/controllers/users/confirmations_controller.rb
+++ b/app/controllers/users/confirmations_controller.rb
@@ -37,32 +37,14 @@ class Users::ConfirmationsController < Devise::ConfirmationsController
 
   # GET /resource/confirmation?confirmation_token=abcdef
   def show
-    # In the default implementation, this already confirms the resource:
-    # self.resource = self.resource = resource_class.confirm_by_token(params[:confirmation_token])
     self.resource = resource_class.find_by!(confirmation_token: params[:confirmation_token])
 
-    yield resource if block_given?
-
-    # New condition added to if: when no password was given, display the "show" view
-    # (which uses "update" above)
-    if resource.encrypted_password.blank?
-      respond_with_navigational(resource) { render :show }
-    elsif resource.errors.empty?
-      set_official_position if resource.has_official_email?
-
-      if resource.confirm
-        set_flash_message(:notice, :confirmed) if is_flashing_format?
-
-        respond_with_navigational(resource) do
-          redirect_to after_confirmation_path_for(resource_name, resource)
-        end
-      else
-        respond_with_navigational(resource.errors, status: :unprocessable_entity) do
-          render :new, status: :unprocessable_entity
-        end
-      end
+    if resource.encrypted_password.present?
+      super
     else
-      respond_with_navigational(resource.errors, status: :unprocessable_entity) { render :new }
+      yield resource if block_given?
+
+      respond_with_navigational(resource) { render :show }
     end
   end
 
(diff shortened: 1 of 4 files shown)


app/models/verification/email.rb
```diff
@@ -23,7 +23,7 @@ def user
   end
 
   def generate_token
-    @plain_token, @encrypted_token = Devise.token_generator.generate(User, :email_verification_token)
+    @encrypted_token, @plain_token = Devise.token_generator.generate(User, :email_verification_token)
```
johanndev1208, line 26: https://github.com/heartcombo/devise/blob/cf93de390a29654620fdf7ac07b4794eb95171d0/lib/devise/models/recoverable.rb#L90C16-L90C19
in this resourse we can find what should we store to db and use in payload.

app/models/verification/management/email.rb
```diff
@@ -21,7 +21,7 @@ def user?
   def save
     return false unless valid?
 
-    plain_token, encrypted_token = Devise.token_generator.generate(User, :email_verification_token)
+    encrypted_token, plain_token = Devise.token_generator.generate(User, :email_verification_token)
```
javierm: Same here; instead of changing the variable names, we could use `email_verification_token: encrypted_token` in the `user.update!` call.

javierm (commented):
@johanndev1208 Thanks for the pull request! :pray:

For now, I'd vote to simply uncomment the `config.confirm_within = 3.days` and write the test for that :ok_hand:.

We can remove the rest of the changes for now, but if you think they're necessary, you can open a different pull request with an explanation about how they would affect existing Consul Democracy installations :thinking:.

app/controllers/users/confirmations_controller.rb
```diff
@@ -81,4 +67,8 @@ def allowed_params
     def set_official_position
       resource.add_official_position!(Setting["official_level_1_name"], 1)
     end
+
+    def confirmation_period_expired?
+      resource.present? && resource.send(:confirmation_period_expired?)
```
javierm: Do we need `resource.present?` here? 🤔 In the `show` action, we use `find_by!`, so this shouldn't be needed. In the `update` action we do use `find_by`, but this is probably something we forgot to change in commit 78c6395e5f4 🤔. The `update` action currently generates a 500 error when the resource isn't found.

So what do you think about using `find_by!` in the `update` action and removing the `resource.present? &&` part of this line?

spec/system/management/users_spec.rb
```diff
@@ -44,6 +44,24 @@
...
+                  confirmation_sent_at: Time.current)
+
+    travel_to(4.days.from_now) do
+      visit user_confirmation_path(confirmation_token: "plain-token")
+
+      expect(page).to have_content "Re-send confirmation instructions"
+      expect(page).to have_content "You need to be verified within 3 days"
+      expect(page).not_to have_field "New access password"
+      expect(user.reload.encrypted_password).to be_blank
+      expect(user.reload).not_to be_confirmed
```
javierm: Here we're checking the database after starting the browser. Shall we remove these lines? We're already checking this in the controller tests after all 🤔.

javierm (approved):

taitus (approved):

PR #6125, opened 2025-10-24
Add missing investments filter on admin activity page

taitus:
## Objectives

* Fix a bug where filters for debates and proposals appeared but the filter for investments did not.
* Remove unused or obsolete ActiveRecord scopes across the code.
* Each removed scope was confirmed to have no remaining references in the application since their respective commits.

### DIFF
--- a/app/models/community.rb
+++ b/app/models/community.rb
@@ -49,8 +49,8 @@ class Community < ApplicationRecord
     end
 
     def users_who_topics_author
-      author_ids = topics.pluck(:author_id)
-      User.by_authors(author_ids)
+      ids = topics.pluck(:author_id)
+      User.with_ids(ids)
     end
 
     def author_from_community
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -111,7 +111,7 @@ class User < ApplicationRecord
   scope :erased,         -> { where.not(erased_at: nil) }
   scope :active,         -> { excluding(erased) }
   scope :public_for_api, -> { all }
-  scope :by_authors,     ->(author_ids) { where(id: author_ids) }
+  scope :with_ids,       ->(ids) { where(id: ids) }
   scope :by_comments,    ->(commentables) do
     joins(:comments).where("comments.commentable": commentables).distinct
   end


app/models/poll/answer.rb
```diff
@@ -12,6 +12,5 @@ class Poll::Answer < ApplicationRecord
   validates :answer, inclusion: { in: ->(poll_answer) { poll_answer.option.possible_answers }},
                      if: ->(poll_answer) { poll_answer.option.present? }
 
-  scope :by_author, ->(author_id) { where(author_id: author_id) }
```
javierm, line 15: Since we're removing all these `by_author` scopes, what do you think about removing or renaming the `User.by_authors` scope as well? The naming doesn't sense at all, and I guess it was inspired by all the existing `by_author` scopes :thinking:.

app/models/activity.rb
```diff
@@ -10,7 +10,6 @@ class Activity < ApplicationRecord
   scope :on_debates, -> { where(actionable_type: "Debate") }
   scope :on_users, -> { where(actionable_type: "User") }
   scope :on_comments, -> { where(actionable_type: "Comment") }
-  scope :on_budget_investments, -> { where(actionable_type: "Budget::Investment") }
```
javierm, line 13: Instead of removing this scope, what do you think about adding an "investments" tab/filter to the admin activity? Currently there are filters for everything except investments, which is probably unintentional :thinking:.

app/models/proposal.rb
```diff
@@ -89,7 +89,6 @@ class Proposal < ApplicationRecord
   scope :draft,          -> { excluding(published) }
 
   scope :not_supported_by_user, ->(user) { where.not(id: user.find_voted_items(votable_type: "Proposal")) }
-  scope :created_by,            ->(author) { where(author: author) }
```
javierm, line 93: The commit message says:

> It was introduced in commit 64258baf977 ("Refactor getting the public activity information") but has never been referenced since.

If I understand it correctly, what actually happened was that we stopped using this method in that commit :thinking:. The scope itself was introduced in commit 77dd60427.

app/models/debate.rb
```diff
@@ -38,7 +38,6 @@ class Debate < ApplicationRecord
   scope :sort_by_hot_score,        -> { reorder(hot_score: :desc) }
   scope :sort_by_confidence_score, -> { reorder(confidence_score: :desc) }
   scope :sort_by_created_at,       -> { reorder(created_at: :desc) }
-  scope :sort_by_most_commented,   -> { reorder(comments_count: :desc) }
```
javierm, line 41: Should we also remove the translations for `debates.index.orders.most_commented` as well? :thinking:

PR #6126, opened 2025-10-24
Newsletter recipients

johanndev1208:
## Objectives
feat(newsletter): add newsletter sign-up form for homepage and account integration

Why:
Many visitors want to receive updates without creating an account. Until now, only logged-in users could subscribe to newsletters, which limited outreach and engagement.

Adding a public sign-up form increases accessibility and allows non-registered users to stay informed about platform updates.

The change also improves GDPR compliance by introducing a clear, explicit opt-in mechanism with double confirmation. This ensures that no newsletter is sent without verified consent.

Finally, placing the sign-up option on the homepage enhances visibility, making it easier for both new and existing users to manage their subscriptions and stay connected with the platform.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -88,6 +88,7 @@
 //= require markdown_editor
 //= require html_editor
 //= require cocoon
+//= require newsletter_recipient
 //= require options
 //= require questions
 //= require legislation_admin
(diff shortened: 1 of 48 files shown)


lucialuzuriaga (commented):
For the moment we’ve decided to not implement this feature, as most users don’t enjoy receiving Newsletters — regardless of whether they have an account or not. For the most part, a user that wants a newsletter is a user prone to signing up in the first place. We will revisit this again in the future if needed.

PR #6144, opened 2025-11-07
Fix most Axe serious accessibility issues

javierm:
## References

* Our [test run 10130](https://github.com/consuldemocracy/consuldemocracy/actions/runs/19172920541) includes Axe tests for this code.

## Objectives

* Make it easier for people using screen readers to identify map markers.
* Make it easier to use the fullscreen markdown editor for people using a keyboard.
* Make "sign out" more accessible for people using screen readers.

## Notes

Among the "serious" accessibility issues reported by Axe, we aren't fixing the ones related to color contrast. Those issues will be addressed in a future pull request, since fixing them will take a while.

### DIFF
--- a/app/assets/javascripts/account_menu.js
+++ b/app/assets/javascripts/account_menu.js
@@ -1,0 +1,8 @@
+(function() {
+  "use strict";
+  App.AccountMenu = {
+    initialize: function() {
+      $(".account-menu > li > form button").attr("role", "menuitem");
+    }
+  };
+}).call(this);
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -117,6 +117,7 @@
 //= require budget_edit_associations
 //= require budget_hide_money
 //= require datepicker
+//= require account_menu
 //= require authenticity_token_refresh
 //= require_tree ./admin
 //= require_tree ./polls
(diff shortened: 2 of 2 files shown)


app/assets/javascripts/map.js
```diff
@@ -28,17 +28,19 @@
...
-      createMarker = function(latitude, longitude) {
+      markerIcon = function(alt_text) {
+        return L.divIcon({
+          className: "map-marker",
+          iconSize: [30, 30],
+          iconAnchor: [15, 40],
+          html: '<div class="map-icon" aria-label="' + alt_text + '"></div>'
+        });
+      };
+      createMarker = function(latitude, longitude, text) {
```
taitus, line 39: I noticed `createMarker` now takes a third `text` argument, but in line 65 we still call it like this:

```  
marker = createMarker(e.latlng.lat, e.latlng.lng);
```

This seems to end up rendering an `aria-label="undefined"`.

app/assets/javascripts/map.js
```diff
@@ -28,17 +28,19 @@
...
-        iconAnchor: [15, 40],
-        html: '<div class="map-icon"></div>'
-      });
-      createMarker = function(latitude, longitude) {
+      markerIcon = function(alt_text) {
+        return L.divIcon({
+          className: "map-marker",
+          iconSize: [30, 30],
+          iconAnchor: [15, 40],
+          html: '<div class="map-icon" aria-label="' + alt_text + '"></div>'
```
taitus: Here we concatenate the user provided title inside an `aria-label` attribute.
Should we take into account that users might include quotes or HTML that could break the attribute or the markup?

taitus (commented):
Hi @javierm

I've just added two small comments to the PR. Let me know what you think.

PR #6148, opened 2025-11-17
Add ARIA labels to admin settings map marker

javierm:
## References

* Continues pull request #6144

## Objectives

* Make sure people using screen readers know what a marker is about when there's a form to edit the marker.
* Use a legend instead of a label pointing nowhere in map fields.
* Simplify the code rendering the admin settings map.

### DIFF
--- a/app/assets/javascripts/map.js
+++ b/app/assets/javascripts/map.js
@@ -193,7 +193,6 @@
 
           if (App.Map.validCoordinates(coordinates)) {
             marker = createMarker(coordinates.lat, coordinates.long, coordinates.title);
-            marker.options.id = coordinates.investment_id;
             marker.bindPopup(App.Map.getPopupContent(coordinates));
           }
         });
--- a/app/models/map_location.rb
+++ b/app/models/map_location.rb
@@ -16,15 +16,6 @@ class MapLocation < ApplicationRecord
     mappable&.title
   end
 
-  def json_data
-    {
-      investment_id: investment_id,
-      proposal_id: proposal_id,
-      lat: latitude,
-      long: longitude
-    }
-  end
-
   def self.from_heading(heading)
     new(
       zoom: Budget::Heading::OSM_DISTRICT_LEVEL_ZOOM,
--- a/spec/system/budgets/budgets_spec.rb
+++ b/spec/system/budgets/budgets_spec.rb
@@ -305,9 +305,8 @@ describe "Budgets" do
         {
           lat: map_location[:latitude],
           long: map_location[:longitude],
-          investment_title: investment.title,
-          investment_id: investment.id,
-          budget_id: budget.id
+          title: investment.title,
+          link: "/budgets/#{budget.id}/investments/#{investment.id}"
         }
       end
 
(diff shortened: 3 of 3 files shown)


spec/shared/system/mappable.rb
```diff
@@ -219,22 +221,6 @@
...
-      fill_in "#{mappable_factory_name.camelize} title", with: "New title"
-      click_button "Save changes"
-
-      expect(page).not_to have_button "Save changes"
-      expect(page).to have_css ".map-location"
-      expect(page).to have_css ".map-location[data-marker-latitude='#{original_latitude}']"
-      expect(page).to have_css ".map-location[data-marker-longitude='#{original_longitude}']"
-    end
-
     scenario "Can not display map on #{mappable_factory_name} edit when remove map marker" do
```
taitus, line 222: Since we have a commit to group tests "Group similar map tests together", do you think it would make sense to also group the tests "Can not display map on edit when remove map marker" (lines 231-240) and "No need to skip map on update" (lines 254-263)?

spec/system/admin/settings_spec.rb
```diff
@@ -65,33 +59,14 @@
                                        'on "Features" tab.'
     end
 
-    scenario "Should show successful notice" do
+    scenario "Should update marker" do
```
taitus: Since we have a commit to group tests "Group similar map tests together", do you think it would make sense to also group the tests "Should be able when map feature activated" (lines 50-60) and "Should update marker" (lines 62-82)?

spec/shared/system/mappable.rb
```diff
@@ -33,7 +33,10 @@
       visit send(mappable_new_path, arguments)
 
       send("fill_in_#{mappable_factory_name}")
-      find("#new_map_location").click
+
+      within_fieldset "Map location" do
+        find("#new_map_location").click
+      end
```
taitus: Since in the commit "Use a legend instead of a label in map location fields" we're adding within_fieldset "Map location" for `find("#new_map_location").click`, would it make sense to do the same for consistency on lines 110 and 143, which use the same selector `find("#new_map_location")`?

And on the other hand, would it make sense to also add it to the `click_button "Remove map marker"` on lines 235 and 258?

app/components/budgets/investments/form_component.rb
```diff
@@ -21,12 +21,7 @@ def categories
     end
 
     def map_location_fields_component(form)
```
taitus: Could we remove this method and render directly in the component's view?

This could also be applied in app/components/budgets/investments/form_component.rb.

If we end up removing these two methods, we should remember to update the commit message of the commit "Move map location fields partial to a component" since it will no longer make sense.

app/components/shared/map_location_component.rb
```diff
@@ -62,12 +62,20 @@ def data
...
+        marker_title: map_location.title.presence || marker_coordinates_text,
         marker_clustering: feature?("map.feature.marker_clustering"),
         geozones: geozones_data
       }.merge(input_selectors)
     end
 
+    def marker_coordinates_text
+      return unless map_location.available?
+
+      t("proposals.form.map_marker_coodinates",
```
taitus: It looks like there's a small typo in the translation key name.

app/assets/javascripts/map.js
```diff
@@ -193,7 +193,6 @@
 
           if (App.Map.validCoordinates(coordinates)) {
             marker = createMarker(coordinates.lat, coordinates.long, coordinates.title);
-            marker.options.id = coordinates.investment_id;
```
taitus, line 196: In the commit "Update old usages of investments JSON data", you reference commit 3fa3c90db. Since that commit, the instance method json_data in app/models/map_location.rb appears to be unused.

Should we remove it in this commit, or leave it for later cleanup?

PR #6178, opened 2025-12-10
Make moderation specs faster

taitus:
## References
- Make nested documentable specs faster #5989

## Objectives

As we did in PR #5369, #5797 and #5989, we reduce the number of tests we are going to run. In this case we remove duplicate moderation spec files and create a unified spec file that runs the same tests but with a different factory in each execution for each test.

**Additional improvements:**
- Unify moderation JavaScript handlers into a single `moderator.js` file
- Fix redirection when blocking authors from legislation proposals (legislation proposals don't have a public index route, redirect to `legislation_process_proposals_path` instead)
- Fix redirection when blocking authors from comments (comments don't have a public index route, redirect to the commentable's show page instead)

### DIFF
--- a/spec/system/moderation_spec.rb
+++ b/spec/system/moderation_spec.rb
@@ -223,22 +223,22 @@ describe "Moderation" do
         end
 
         visit moderation_resource_index_path(filter: "all")
-        expect(page).to have_content content_for(regular_resource)
-        expect(page).to have_content content_for(pending_resource)
-        expect(page).not_to have_content content_for(hidden_resource)
-        expect(page).to have_content content_for(ignored_resource)
+        expect(page).to have_content regular_resource.human_name
+        expect(page).to have_content pending_resource.human_name
+        expect(page).not_to have_content hidden_resource.human_name
+        expect(page).to have_content ignored_resource.human_name
 
         visit moderation_resource_index_path(filter: pending_filter)
-        expect(page).not_to have_content content_for(regular_resource)
-        expect(page).to have_content content_for(pending_resource)
-        expect(page).not_to have_content content_for(hidden_resource)
-        expect(page).not_to have_content content_for(ignored_resource)
+        expect(page).not_to have_content regular_resource.human_name
+        expect(page).to have_content pending_resource.human_name
+        expect(page).not_to have_content hidden_resource.human_name
+        expect(page).not_to have_content ignored_resource.human_name
 
         visit moderation_resource_index_path(filter: ignored_filter)
-        expect(page).not_to have_content content_for(regular_resource)
-        expect(page).not_to have_content content_for(pending_resource)
-        expect(page).not_to have_content content_for(hidden_resource)
-        expect(page).to have_content content_for(ignored_resource)
+        expect(page).not_to have_content regular_resource.human_name
+        expect(page).not_to have_content pending_resource.human_name
+        expect(page).not_to have_content hidden_resource.human_name
+        expect(page).to have_content ignored_resource.human_name
       end
 
       context "Budget Investments, Comments, Debates and Proposals" do
@@ -251,21 +251,21 @@ describe "Moderation" do
 
           visit moderation_resource_index_path(order: order)
 
-          expect(content_for(flagged_new_resource)).to appear_before content_for(flagged_resource)
+          expect(flagged_new_resource.human_name).to appear_before flagged_resource.human_name
 
           visit moderation_resource_index_path(order: "flags")
 
-          expect(content_for(flagged_resource)).to appear_before content_for(flagged_new_resource)
+          expect(flagged_resource.human_name).to appear_before flagged_new_resource.human_name
 
           visit moderation_resource_index_path(filter: "all", order: order)
 
-          expect(content_for(newer_resource)).to appear_before content_for(flagged_new_resource)
-          expect(content_for(flagged_new_resource)).to appear_before content_for(flagged_resource)
+          expect(newer_resource.human_name).to appear_before flagged_new_resource.human_name
+          expect(flagged_new_resource.human_name).to appear_before flagged_resource.human_name
 
           visit moderation_resource_index_path(filter: "all", order: "flags")
 
-          expect(content_for(flagged_resource)).to appear_before content_for(flagged_new_resource)
-          expect(content_for(flagged_new_resource)).to appear_before content_for(newer_resource)
+          expect(flagged_resource.human_name).to appear_before flagged_new_resource.human_name
+          expect(flagged_new_resource.human_name).to appear_before newer_resource.human_name
         end
       end
 
@@ -299,7 +299,7 @@ describe "Moderation" do
         visit moderation_resource_index_path(filter: "all")
 
         expect(page).to have_content "Moderation"
-        expect(page).to have_content content_for(reported_resource)
+        expect(page).to have_content reported_resource.human_name
 
         if factory == :comment
           click_link reported_resource.commentable.title
@@ -308,16 +308,12 @@ describe "Moderation" do
         end
 
         expect(page).not_to have_content "Moderation"
-        expect(page).to have_content content_for(reported_resource)
+        expect(page).to have_content reported_resource.human_name
       end
     end
   end
 
   def moderation_resource_index_path(params = {})
     namespaced_polymorphic_path(:moderation, resource.class, **params)
-  end
-
-  def content_for(resource)
-    factory == :comment ? resource.body : resource.title
   end
 end


spec/system/moderation_spec.rb
```diff
@@ -1,88 +1,380 @@
...
+
+          visit moderation_proposal_notifications_path(filter: "all", order: "moderated")
+
+          expect(old_moderated_notification.title).to appear_before newer_notification.title
+        end
+      end
+
+      scenario "Visit flagged resources" do
+        if factory == :proposal_notification
+          reported_resource = create(factory)
```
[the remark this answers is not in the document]
javierm: False positive.

app/assets/javascripts/moderator.js
```diff
@@ -0,0 +1,24 @@
+(function() {
+  "use strict";
+  App.Moderator = {
+    add_class_faded: function(id) {
+      var isComment = id.startsWith("comment_");
+      var isProposalNotification = id.startsWith("proposal_notification_");
```
javierm: We usually don't use camelcase for variable names :thinking:. How about:

```suggestion
      var is_comment = id.startsWith("comment_");
      var is_proposal_notification = id.startsWith("proposal_notification_");
```

Back in 2019, when I checked ESLint rules, there wasn't an option to enforce snake case in variable names, but maybe now there's a rule we can enable :thinking:.

app/assets/javascripts/moderator.js
```diff
@@ -0,0 +1,24 @@
...
+
+      if (isComment) {
+        $("#" + id + " .comment-body:first").addClass("faded");
+      } else {
+        var $element = $("#" + id);
+        $element.addClass("faded");
```
javierm: To make it easier to understand that this part of the code hasn't changed when compared to the previous code, maybe we could leave it like:

```suggestion
        $("#" + id).addClass("faded");
```

app/controllers/moderation/proposal_notifications_controller.rb
```diff
@@ -9,7 +9,12 @@ class Moderation::ProposalNotificationsController < Moderation::BaseController
   load_and_authorize_resource
 
   def hide
-    ProposalNotification.find(params[:id]).update(moderated: true)
+    resource.update(moderated: true)
+    respond_to do |format|
+      format.js do
+        render "moderation/shared/hide", locals: { resource: resource }
+      end
+    end
```
javierm: Would removing the `hide` method and overwriting `hide_resource` work as well? :thinking:

```suggestion
  private
  
    def hide_resource(resource)
      resource.update(moderated: true)
    end
```

spec/system/moderation_spec.rb
```diff
@@ -1,88 +1,380 @@
...
+
+      login_as moderator.user
+      visit resource_path
+      click_link "Notifications (1)" if factory == :proposal_notification
+
+      within "##{dom_id(resource)}" do
+        expect(page).not_to have_button "Hide"
+        expect(page).not_to have_button "Block author"
+      end
+    end
```
javierm: Since we aren't testing using interaction, would it make sense to move this test to a component test? :thinking:

spec/system/moderation_spec.rb
```diff
@@ -1,88 +1,380 @@
...
+                     "and all their contents.") do
+        click_button "Block author"
+      end
+
+      expect(page).to have_content "The user has been blocked"
+      expect(page).not_to have_content resource_content
+    end
+
+    describe "/moderation/ screen" do
+      let(:active_link_text) do
```
javierm: Why "active"? :thinking: Would it make sense to use something like "newest" or "recent" instead?

spec/system/moderation_spec.rb
```diff
@@ -1,88 +1,380 @@
...
+            expect(page).to have_link "Most flagged"
+          end
+
+          expect(page).to have_current_path(/filter=all/)
+          expect(page).to have_current_path(/page=2/)
+          expect(page).to have_current_path(/order=#{order}/)
+        end
+      end
+
+      scenario "Current filter is properly highlighted" do
```
javierm: Since we aren't testing user interaction, would it make sense to move this test to a component test?

spec/system/moderation_spec.rb
```diff
@@ -1,88 +1,380 @@
...
+        expect(page).not_to have_link "Pending"
+        expect(page).to have_link "Marked as viewed"
+
+        visit moderation_resource_index_path(filter: ignored_filter)
+        expect(page).to have_link "All"
+        expect(page).to have_link "Pending"
+        expect(page).not_to have_link "Marked as viewed"
+      end
+
+      scenario "Filtering resources" do
```
javierm, line 214: Since we aren't testing user interaction, would it make sense to move this test to a component test?

spec/system/moderation_spec.rb
```diff
@@ -1,88 +1,380 @@
...
+        if factory == :proposal_notification
+          pending_resource = create(factory, :moderated)
+          ignored_resource = create(factory, :moderated, :ignored)
+        else
+          pending_resource = create(factory, :flagged)
+          ignored_resource = create(factory, :flagged, :with_ignored_flag)
+        end
+
+        visit moderation_resource_index_path(filter: "all")
+        expect(page).to have_content content_for(regular_resource)
```
javierm: Would `regular_resource.human_name` work in this case? :thinking: Same for every use of `content_for`.

spec/system/moderation_spec.rb
```diff
@@ -1,88 +1,380 @@
...
+        expect(page).not_to have_content content_for(regular_resource)
+        expect(page).not_to have_content content_for(pending_resource)
+        expect(page).not_to have_content content_for(hidden_resource)
+        expect(page).to have_content content_for(ignored_resource)
+      end
+
+      context "Budget Investments, Comments, Debates and Proposals" do
+        let(:factory) { (factories - [:proposal_notification]).sample }
+
+        scenario "Sorting resources" do
```
javierm, line 247: Since we aren't testing user interaction, would it make sense to move this test to a component test? :thinking:

app/controllers/moderation/proposal_notifications_controller.rb
```diff
@@ -9,7 +9,12 @@ class Moderation::ProposalNotificationsController < Moderation::BaseController
   load_and_authorize_resource
 
   def hide
-    ProposalNotification.find(params[:id]).update(moderated: true)
+    resource.update(moderated: true)
```
javierm: In the "Adapt "Hide" test to include proposal_notification as factory" commit, where we're changing validation rules, do you think it make sense to replace `update` with `update!` here?

spec/system/moderation_spec.rb
```diff
@@ -1,88 +1,380 @@
...
+      expect(page).not_to have_content resource_content
+
+      if factory == :proposal_notification
+        expect(page).to have_content "Notifications (0)"
+      else
+        expect(page).to have_content "Not found"
+      end
+    end
+
+    scenario "Can not hide own resource" do
```
javierm: Maybe it's a good time to fix this typo :thinking:.

```suggestion
    scenario "Cannot hide own resource" do
```

spec/system/moderation_spec.rb
taitus: I named it `active_link_text` to refer to the link's state (the one with the `is-active` class). Since the name seems confusing, I'll rename it to `order_link_text` to better reflect what it contains 🤔 (instead of "newest" or "recent" since it can vary depending on the resource type, and even though they mean the same thing, I think `order_link_text` is less confusing).

spec/components/moderation/shared/index_component_spec.rb
```diff
@@ -0,0 +1,123 @@
+require "rails_helper"
+
+describe Moderation::Shared::IndexComponent, controller: Moderation::DebatesController do
```
javierm: Do we need the `controller:` part? :thinking: I mean, we're already using `with_request_url` and `allow(vc_test_controller)` :thinking:.

```suggestion
describe Moderation::Shared::IndexComponent do
```

spec/components/moderation/shared/index_component_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+  let(:moderator) { create(:moderator) }
+  let(:records) { Debate.none.page(1) }
+
+  before do
+    sign_in(moderator.user)
```
javierm: Looks like this component doesn't do any authorization checks, so we can omit this line :thinking:.

spec/components/shared/moderation_actions_component_spec.rb
```diff
@@ -3,9 +3,9 @@
 describe Shared::ModerationActionsComponent do
   include Rails.application.routes.url_helpers
 
-  before { sign_in(create(:administrator).user) }
-
   describe "Hide button" do
+    before { sign_in(create(:administrator).user) }
+
```
javierm: Any particular reason for this change? This `before` block is still affecting every test in the file :thinking:.

spec/system/moderation_spec.rb
```diff
@@ -1,88 +1,326 @@
...
+        expect(page).not_to have_css "#admin_menu"
+        expect(page).not_to have_css "#valuation_menu"
+        expect(page).not_to have_content "You do not have permission to access this page"
+      end
+    end
 
-  scenario "Access as manager is not authorized" do
-    create(:manager, user: user)
-    login_as(user)
+    context "for administrators" do
```
javierm: For consistency, should we group the moderators/administrators tests together as well? :thinking:

spec/system/moderation_spec.rb
```diff
@@ -1,88 +1,326 @@
...
+    let(:moderator) { create(:moderator) }
+    let(:resource_path) do
+      if factory == :proposal_notification
+        polymorphic_path(resource.proposal, anchor: "tab-notifications")
+      else
+        polymorphic_path(resource)
+      end
+    end
+    let(:faded_selector) { factory == :comment ? "> .comment .faded" : ".faded" }
+    let(:resource_content) { factory == :comment ? resource.body : resource.title }
```
javierm: Would it make sense to replace `resource_content` with `resource.human_name` here? :thinking:

spec/system/moderation_spec.rb
```diff
@@ -1,88 +1,326 @@
...
-    visit root_path
-    click_link "Menu"
-    click_link "Moderation"
+        scenario "remembering page, filter and order" do
+          stub_const("#{ModerateActions}::PER_PAGE", 2)
+          create_list(factory, 4)
+
+          visit moderation_resource_index_path(filter: "all", page: "2", order: order)
+
+          accept_confirm("Are you sure? Mark as viewed") { click_button "Mark as viewed" }
```
javierm, line 190: The test passes without this line. Shall we add an extra expectation to make sure the request has finished? :thinking:

PR #6185, opened 2025-12-11
Singularise model names in activerecord.yml locales

digitalWestie:
## Objectives

Fixes a number of model names in the es/en activerecord locales by using singular keys. 

For example, the key `es.activerecord.models.legislation/questions` won't be recognised.

```ruby
I18n.locale = :es
Legislation::Question.model_name.human #=> "Question"
```

With singular keys e.g. `es.activerecord.models.legislation/question`:

```ruby
I18n.locale = :es
Legislation::Question.model_name.human #=> "Pregunta"
```

### DIFF
--- a/config/locales/en/activerecord.yml
+++ b/config/locales/en/activerecord.yml
@@ -106,22 +106,22 @@ en:
       legislation/proposal:
         one: "Proposal"
         other: "Proposals"
-      legislation/draft_versions:
+      legislation/draft_version:
         one: "Draft version"
         other: "Draft versions"
-      legislation/questions:
+      legislation/question:
         one: "Question"
         other: "Questions"
-      legislation/question_options:
+      legislation/question_option:
         one: "Question option"
         other: "Question options"
-      legislation/answers:
+      legislation/answer:
         one: "Answer"
         other: "Answers"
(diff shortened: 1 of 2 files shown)


config/locales/en/activerecord.yml
```diff
@@ -106,22 +106,22 @@ en:
...
-      legislation/question_options:
+      legislation/question_option:
         one: "Question option"
         other: "Question options"
-      legislation/answers:
+      legislation/answer:
         one: "Answer"
         other: "Answers"
-      documents:
+      document:
```
taitus: This creates a duplicate `document:` key. `document:` already exists at L100-102. Same in `es/activerecord.yml`. Suggest removing this block (the old `documents:` entry was effectively dead code)

taitus (commented):
Hi @digitalWestie 

Thanks a lot for the PR 👏 . Nice catch. I've left one small comment on the code, let me know what you think.

config/locales/en/activerecord.yml
digitalWestie: ah good spot, I've now removed the duplicate

taitus (approved):

PR #6190, opened 2025-12-20
Image suggestions with LLM

stirkac:
## Objectives

Since users might not have their own image to include on their purpose, we introduce AI-powered image suggestions next to the upload image button, to allow stock images from Pexels API to be added to the resource.

We reuse as much as possible of the existing infrastructure:
- expand upon `Images::NestedComponent` to render the suggestions button
- reuse DirectUploads model for storing everything related to the images

## Visual Changes

* Added setting to LLM Settings
llm-configuration-en

* Suggest image with AI button on Image input
upload-form-with-button-en

* Image suggestions
suggested-images-grid-en

* Attached image when image is selected
selected-image-preview-en

## Notes
Read through [docs/en/features/image_suggestions.md](https://github.com/consuldemocracy/consuldemocracy/blob/c477394190171da3155619f9212207aa8c048e6b/docs/en/features/image_suggestions.md) for a deeper explanation.


app/components/attachable/fields_component.html.erb
```diff
@@ -16,8 +16,11 @@
 
     <div class="small-9 column action-add attachment-errors <%= singular_name %>-attachment">
       <%= file_field %>
-    </div>
 
+      <% unless attachable.attachment.attached? && attachable.attachment.image? %>
```
taitus: I noticed that this `unless` has been added, but I'm not sure it's necessary.

I think this is the same case as in line 48 of `app/components/attachable/fields_component.rb`.

It might be possible to remove both pieces of code since the visibility is already handled by the CSS in uploads.scss (lines 90-92) 🤔 :

```
    &:not(:empty) + .document-attachment,
    &:not(:empty) + .image-attachment {
      display: none;
    }
```

In this case, could we simplify by removing these conditions and let the CSS handle the visibility?

app/components/budgets/investments/form_component.html.erb
```diff
@@ -37,7 +37,9 @@
     <legend><%= t("shared.optional") %></legend>
 
     <% if feature?(:allow_images) %>
-      <%= render Images::NestedComponent.new(f) %>
+      <%= render Images::NestedComponent.new(f) do |content| %>
```
taitus: Could we simplify this by removing the unused content parameter from the block?

```suggestion
      <%= render Images::NestedComponent.new(f) do %>
```

If the block doesn't use content, we can remove it.

app/components/images/suggest_images_component.html.erb
```diff
@@ -0,0 +1,4 @@
+<div class="suggested-images-container">
+  or
```
taitus: The text "or" is hardcoded. We should move it to the translation files (.yml) so it can be localized.

app/components/images/suggest_images_component.html.erb
```diff
@@ -0,0 +1,4 @@
...
+  <button type="button" class="button hollow js-suggest-image" data-resource-type="<%= resource_type %>" data-resource-id="<%= resource_id %>"><%= t("images.form.suggest_image_with_ai") %></button>
```
taitus: This line is too long. Consider breaking it into multiple lines for better readability:

```suggestion
<button type="button" 
        class="button hollow js-suggest-image" 
        data-resource-type="<%= resource_type %>" 
        data-resource-id="<%= resource_id %>">
  <%= t("images.form.suggest_image_with_ai") %>
</button>
```

app/components/images/suggest_images_component.rb
```diff
@@ -0,0 +1,24 @@
+module Images
+  class SuggestImagesComponent < ApplicationComponent
```
taitus: We don't usually use module Images with a nested class in components. For consistency with the rest of the code, consider using:

```suggestion
class Images::SuggestImagesComponent < ApplicationComponent
```
instead of the module structure.

app/components/images/suggested_images_component.html.erb
```diff
@@ -0,0 +1,21 @@
...
+      <button class="close-button" aria-label="<%= t("application.close") %>" type="button" data-close>
+        <span aria-hidden="true">&times;</span>
+      </button>
+      <%= error_messages.join(", ") %>
+    </div>
+  <% else %>
+    <div class="small-12 column">
+      <% suggested_images.each_with_index do |image, index| %>
+        <div class="small-3 column">
+          <button type="button" class="js-attach-suggested-image" data-image-id="<%= image.id %>" data-resource-type="<%= resource_type %>" data-resource-id="<%= resource_id %>" id="suggested-image-<%= image.id %>" aria-label="Attach suggested image <%= index + 1 %> of <%= suggested_images.count %>" aria-describedby="suggested-image-desc-<%= image.id %>">
```
taitus: This line is too long. Consider breaking it into multiple lines for better readability.

app/components/images/suggested_images_component.rb
```diff
@@ -0,0 +1,36 @@
+module Images
+  class SuggestedImagesComponent < ApplicationComponent
```
taitus: We don't usually use module Images with a nested class in components. For consistency with the rest of the code, consider using:
```suggestion
class Images::SuggestedImagesComponent < ApplicationComponent
```
instead of the module structure.

app/components/images/suggested_images_component.rb
```diff
@@ -0,0 +1,36 @@
...
+    attr_reader :resource_type, :resource_id, :resource_attributes
+
+    def initialize(resource_type:, resource_id: nil, resource_attributes: {})
+      @resource_type = resource_type
+      @resource_id = resource_id
+      @resource_attributes = resource_attributes
+    end
+
+    def suggested_images
+      results = response&.results
```
taitus: Is the safe navigation operator (&) necessary here? Since response always returns a Response object and is never nil (as ImageSuggestions::Llm::Client.call always returns a Response instance), it seems unnecessary. Is there a specific case where we need to keep it?

app/controllers/image_suggestions_controller.rb
```diff
@@ -0,0 +1,46 @@
+class ImageSuggestionsController < ApplicationController
+  include DirectUploadsHelper
+  include ActionView::Helpers::UrlHelper
+
+  before_action :authenticate_user!
+  skip_authorization_check # like direct uploads does too
```
taitus:
```suggestion
  skip_authorization_check
```
We don't usually add comments in the code. If something needs extra explanation, we typically put it in commit messages instead. 

This comment and the remaining comments throughout the code should be safe to remove.

app/controllers/image_suggestions_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  def create
+    @resource_type = params[:resource_type]
+    @resource_id = params[:resource_id]
+    @resource_attributes = params.require(@resource_type.parameterize.underscore).permit!.except(:subtitle)
```
taitus: Is the `.except(:subtitle)` exclusion needed because when we use `permit! `(which allows all parameters), the subtitle field from the form's `invisible_captcha` honeypot also gets included? If subtitle has a value, the invisible_captcha would detect it as a bot and reject the request, right?

If that's the reason, could we avoid using `permit!` and instead use `permit` with a specific list of allowed parameters?

We could even extract this parameter management into a method:
`def resource_attributes_params` 🤔

app/lib/image_suggestions/llm/client.rb
```diff
@@ -0,0 +1,83 @@
...
+    class Client
+      class Response
+        attr_accessor :results
+        attr_reader :errors
+
+        def initialize
+          @results = []
+          @errors = []
+        end
+      end
```
taitus: Response classes like this one are typically placed right before the private section.

app/lib/image_suggestions/llm/client.rb
```diff
@@ -0,0 +1,83 @@
...
+
+      NUMBER_OF_IMAGES = 4
+      attr_reader :model_instance, :chat, :prompt, :response
```
taitus: Constants and `attr_reader` declarations are typically placed at the beginning of the class.

app/lib/image_suggestions/llm/client.rb
```diff
@@ -0,0 +1,83 @@
...
+
+      def self.call(model_instance)
+        new(model_instance).call
+      end
+
+      def initialize(model_instance)
+        @model_instance = model_instance
+        @prompt = load_prompt
+        @response = Response.new
+      end
```
taitus: Noticed that `chat` is declared in `attr_reader` but isn't being used as an instance variable. Currently it's being created locally inside `generate_search_query`.

I'd suggest building the chat instance in initialize and reusing it, similar to how RemoteTranslation::Llm::Client does it. 

```suggestion
      def initialize(model_instance)
        @model_instance = model_instance
        @prompt = load_prompt
        @response = Response.new
        @chat = build_chat
      end      
```

And add new method:
```
        def build_chat
          ::Llm::Config.context.chat(provider: llm_provider, model: llm_model)
        end
```

And remove this line from `generate_search_query` method:
```
chat = ::Llm::Config.context.chat(provider: llm_provider, model: llm_model)
```

This would:
- Make use of the attr_reader :chat declaration
- Avoid creating the chat instance on every call
- Keep consistency with other LLM clients in the codebase

Would this approach work for you? If you prefer to remove chat from the attr_reader instead, that works too.

app/lib/image_suggestions/pexels.rb
```diff
@@ -0,0 +1,63 @@
+module ImageSuggestions
+  class Pexels
+    class PexelsError < StandardError; end
+    IMAGE_VARIANT_POSTFIX = "?auto=compress&cs=tinysrgb&h=900".freeze
+    attr_reader :photo
```
taitus: Noticed that `attr_reader :photo` is declared, but I don't see `@photo` being set or accessed anywhere in the class.

Should we remove this `attr_reader` if it's not needed?

app/lib/image_suggestions/pexels.rb
```diff
@@ -0,0 +1,63 @@
+module ImageSuggestions
+  class Pexels
+    class PexelsError < StandardError; end
+    IMAGE_VARIANT_POSTFIX = "?auto=compress&cs=tinysrgb&h=900".freeze
```
taitus, line 6: I noticed we're concatenating `IMAGE_VARIANT_POSTFIX` to `src["original"]` on line 28. According to the Pexels API docs, the src object provides several pre-sized variants like large, medium, small, landscape, etc.

Do you think we could use one of these pre-sized variants  (e.g., src["large"] or src["medium"]) instead of the original with query parameters?

app/lib/image_suggestions/pexels.rb
```diff
@@ -0,0 +1,63 @@
...
+        temp_file = Tempfile.new(uri.path, binmode: true)
+
+        Net::HTTP.start(uri.host, uri.port, use_ssl: true) do |http|
+          http.request(Net::HTTP::Get.new(uri)) do |response|
+            raise PexelsError, "Network request failed" unless response.is_a?(Net::HTTPSuccess)
+
+            response.read_body { |chunk| temp_file.write(chunk) }
+            temp_file.rewind
+          end
+        end
```
taitus: The current download_file method uses Net::HTTP manually. We could simplify this using Ruby's open-uri, which handles SSL and redirects.

Here's a simpler approach:

```suggestion
       require "open-uri"

        def download_file(image_url)
          URI.open(image_url, "rb")
        rescue OpenURI::HTTPError, SocketError => e
          raise PexelsError, "Failed to download image from Pexels: #{e.message}"
        end
```

app/views/image_suggestions/attach.js.erb
```diff
@@ -0,0 +1,2 @@
+// Handle the attach response
+// This will be implemented based on what the controller returns
```
taitus: Is this file needed?

config/locales/en/images.yml
```diff
@@ -6,11 +6,17 @@ en:
       delete_button: Remove image
       note: "You can upload one image of following content types: %{accepted_content_types}, up to %{max_file_size} MB."
       add_new_image: Add image
+      suggest_image_with_ai: Suggest an image with AI ✨
```
taitus: I noticed the emoji ✨ in the translation key `suggest_image_with_ai`. We typically don't include icons/emojis in translation strings. Should we remove it to keep consistency with the rest of the code? 
If we want to keep the visual element, we could add it via CSS instead 🤔 

Same in spanish version.

app/lib/image_suggestions/pexels.rb
stirkac: the predefined sizes crop the image to certain dimensions via width `&w=.`
```
    "large2x": "https://images.pexels.com/photos/2014422/pexels-photo-2014422.jpeg?auto=compress&cs=tinysrgb&dpr=2&h=650&w=940",
    "large": "https://images.pexels.com/photos/2014422/pexels-photo-2014422.jpeg?auto=compress&cs=tinysrgb&h=650&w=940",
    "medium": "https://images.pexels.com/photos/2014422/pexels-photo-2014422.jpeg?auto=compress&cs=tinysrgb&h=350",
    "small": "https://images.pexels.com/photos/2014422/pexels-photo-2014422.jpeg?auto=compress&cs=tinysrgb&h=130",
    "portrait": "https://images.pexels.com/photos/2014422/pexels-photo-2014422.jpeg?auto=compress&cs=tinysrgb&fit=crop&h=1200&w=800",
    "landscape": "https://images.pexels.com/photos/2014422/pexels-photo-2014422.jpeg?auto=compress&cs=tinysrgb&fit=crop&h=627&w=1200",
    "tiny": "https://images.pexels.com/photos/2014422/pexels-photo-2014422.jpeg?auto=compress&cs=tinysrgb&dpr=1&fit=crop&h=200&w=280"
```

we don't want the full original, since it can be (tens) of megabytes - and 'medium'-type of implementation is the only one constraining the cropping without specified witdth, but 350px for a on-page displayed photo is a little bit too little. since official links use the same approach, I think it's a good compromise.

app/lib/image_suggestions/pexels.rb
stirkac: I generally agree but I've used Net::HTTP because it's used everywhere else in the project as well, and open-uri is not

app/components/images/suggested_images_component.html.erb
```diff
@@ -0,0 +1,43 @@
...
+    <div class="small-12 column suggested-images-grid">
+      <% suggested_images.each_with_index do |image, index| %>
+        <div class="small-3 column">
+          <button type="button"
+                  class="js-attach-suggested-image"
+                  data-image-id="<%= image.id %>"
+                  data-resource-type="<%= resource_type %>"
+                  data-resource-id="<%= resource_id %>"
+                  id="suggested-image-<%= image.id %>"
+                  aria-label="Attach suggested image <%= index + 1 %> of <%= suggested_images.count %>"
```
taitus: There's a hardcoded English text in the `aria-label`. We should move it to the translation files (.yml) so it can be localized.

app/controllers/image_suggestions_controller.rb
taitus: Any thoughts on this one? 🤔

app/controllers/image_suggestions_controller.rb
taitus: Any thoughts on this one? 🤔

app/lib/image_suggestions/llm/client.rb
```diff
@@ -0,0 +1,87 @@
...
+
+      def call
+        validate_llm_settings!
+        return response if response.errors.any?
+
+        search_query = generate_search_query
+        return response if response.errors.any?
+
+        results = ImageSuggestions::Pexels.search(search_query, size: :small,
+                                                                per_page: NUMBER_OF_IMAGES)
```
taitus: I might be missing something here 🙏, but I was looking at Pexels' docs/guidelines (the 4MP minimum for uploads) and wanted to better understand how this filter is expected to behave in practice.

My understanding is that `size: :small` acts as a minimum threshold (≥4MP). If that's the case, would the behavior be effectively the same as not specifying `size` at all, or am I missing a subtle difference here?

app/lib/image_suggestions/pexels.rb
```diff
@@ -0,0 +1,49 @@
...
+    end
+
+    private
+
+      def get_photo(photo_id)
+        @client.photos.find(photo_id)
+      end
+
+      def download_file(image_url)
+        URI.open(image_url, "rb") # rubocop:disable Security/Open
```
taitus: According to the [RuboCop Security/Open](https://docs.rubocop.org/rubocop/cops_security.html#safety-securityopen) docs, we can use `URI.parse(image_url).open("rb")` instead of `URI.open(image_url, "rb")`. This avoids the disable and follows the recommended pattern 👌 .

app/assets/javascripts/imageable.js
```diff
@@ -125,6 +128,94 @@
...
+    initializeSuggestImage: function() {
+      // we serialize the entire parent form and submit to the image suggestions endpoint
+      $("body").on("click", ".js-suggest-image", function() {
+        var form, resourceType, dataString, button;
+        button = $(this);
+        form = button.closest("form");
+
+        // Add spinner and disable button
+        button.prop("disabled", true);
+        button.prepend('<i class="fa fa-spinner fa-spin" aria-hidden="true"></i> ');
```
taitus: I've noticed that the spinner isn't showing up. Would it make sense to use the `has-fa-icon` mixin pattern we've been using lately? This would align with our current approach and ensure the spinner displays correctly.

Here's what I'm thinking:
```suggestion
        button.addClass("is-loading");
```

We'd also need to:

1. Add `@import "font-awesome/animated";` to `mixins_and_variables.scss`
2. Add the styles in `uploads.scss`:
```
.js-suggest-image.is-loading {
  @include has-fa-icon(spinner, solid);

  &::before {
    animation: fa-spin 2s infinite linear;
  }
}
```
This should make the spinner visible.

app/assets/javascripts/imageable.js
```diff
@@ -125,6 +128,94 @@
...
+        // Sync CKEditor instances before serializing the form
+        if (typeof CKEDITOR !== "undefined") {
+          for (var name in CKEDITOR.instances) {
+            CKEDITOR.instances[name].updateElement();
+          }
+        }
+
+        resourceType = button.data("resource-type");
+        dataString = App.Imageable.imageSuggestionsParams(form, resourceType);
+        var uploadData = App.Imageable.buildData([], $(".direct-upload").first());
```
taitus: I've noticed we're relying on `.first()` here. Since there can be multiple `.direct-upload` elements on the page (e.g., images and documents), even if this currently picks the right one, it might be worth scoping it a bit more explicitly.

Would it make sense to target the `.direct-upload` within the button's context? For example, something similar to:

```suggestion
        var uploadData = App.Imageable.buildData([], button.closest(".image-fields.direct-upload")
);
```

This way we're selecting the intended element within the image context and avoiding any dependency on DOM order.

If this approach makes sense to you, we could apply the same idea to the other `.first()` usages introduced here as well.

app/assets/javascripts/imageable.js
```diff
@@ -125,6 +128,94 @@
...
+        App.Imageable.clearInputErrors(uploadData);
+        $.ajax({
+          url: "/image_suggestions",
+          type: "POST",
+          data: dataString,
+          dataType: "script"
+        });
```
taitus, line 164: Just to confirm my understanding: the first call uses dataType: "script" and the controller responds with `format.js`, which renders `create.js.erb` and executes it directly. In that setup, everything is handled on the server side, so there’s no need for explicit success or error callbacks on the client.

The second call, on the other hand, uses dataType: "json" with explicit success and error callbacks, which matches the direct-upload approach mentioned in the controller comment (`"mimicking the direct uploads behavior"`) and allows more control over how the response is handled.

Is that the reasoning behind using two different approaches here?

app/assets/javascripts/imageable.js
```diff
@@ -125,6 +128,94 @@
         $(this).closest(".direct-upload").remove();
       });
     },
+    imageSuggestionsParams: function(form, resourceType) {
+      var params;
+      params = form.serializeArray().filter(function(item) {
+        return item.name !== "_method";
+      });
```
taitus, line 135: Would you mind adding a short note in the commit message explaining why we need to exclude it?. That will make it clearer for future maintainers.

app/assets/javascripts/imageable.js
```diff
@@ -125,6 +128,94 @@
...
+        resourceType = $(this).data("resource-type");
+        resourceId = $(this).data("resource-id");
+        uploadData = App.Imageable.buildData([], $(".direct-upload").first());
+        App.Imageable.clearProgressBar(uploadData);
+        $(uploadData.progressBar).find(".loading-bar").css("width", "100%");
+
+        dataString = "resource_type=" + encodeURIComponent(resourceType);
+        if (resourceId) {
+          dataString += "&resource_id=" + encodeURIComponent(resourceId);
+        }
```
taitus: We could simplify this a bit by passing an object as data instead of building a query string manually. jQuery will take care of serializing and encoding it for us:

```suggestion
        dataString = { resource_type: resourceType };
        if (resourceId) {
          dataString.resource_id = resourceId;
        }
```
This should behave the same, but avoids the manual encodeURIComponent.

app/components/images/suggested_images_component.html.erb
```diff
@@ -0,0 +1,43 @@
...
+            <%= image_tag(image.src["small"], alt: image.user.name, class: "suggested-image") %>
+            <span id="suggested-image-desc-<%= image.id %>" class="show-for-sr">
+              <%= t("images.form.attach_suggested_image") %>
+            </span>
+          </button>
+        </div>
+      <% end %>
+    </div>
+  <% end %>
+  <% if suggested_images.blank? %>
```
taitus: Not a blocker, but this might read a bit nicer as an `if / else`.

app/components/images/suggested_images_component.html.erb
```diff
@@ -0,0 +1,43 @@
...
+    <div>
+      <small class='error'><%= t("images.form.no_images_found") %></small>
```
taitus: Could we use double quotes here? We usually avoid single quotes in this codebase. Same for line 40.

app/components/images/suggested_images_component.rb
```diff
@@ -0,0 +1,53 @@
...
+          next if key.to_s == "id"
+
+          result[key] = attributes_without_ids(value)
+        end
+      when Array
+        attrs.map { |item| attributes_without_ids(item) }
+      else
+        attrs
+      end
+    end
```
taitus, line 30: One thing I noticed is that this component is building a full model_instance (safe_constantize, new(...), plus the recursive attributes_without_ids) just to call the LLM client. Since the client only uses title / description, could we pass those explicitly instead? That would simplify the component.

To make this easier to discuss, I've put together a small patch with a possible approach applying these ideas (simplifying the component and passing only the required fields to the client).

As part of this, the patch also avoids duplicating data-resource-type / data-resource-id on each suggested image button, since the parent wrapper already has them (we could read them from there), along with a few minimal changes needed to adapt to this approach.

[0001-Possible-approach.patch](https://github.com/user-attachments/files/25232643/0001-Possible-approach.patch)

app/assets/javascripts/imageable.js
```diff
@@ -125,6 +128,94 @@
...
+      var data = App.Imageable.buildData([], $(".direct-upload").first());
+      data.jqXHR = xhr;
+      App.Imageable.setProgressBar(data, "errors");
+      App.Imageable.clearInputErrors(data);
+      App.Imageable.setInputErrors(data);
+    },
+    initializeAttachSuggestedImage: function() {
+      $("body").on("click", ".js-attach-suggested-image", function(event) {
+        var imageId, resourceType, resourceId, dataString, uploadData;
+        event.stopPropagation();
```
taitus: Curious whether we still need `event.stopPropagation()` here. If nothing relies on preventing the click, we could consider removing it to keep the handler simpler.

app/assets/javascripts/imageable.js
```diff
@@ -125,6 +128,94 @@
...
+        imageId = $(this).data("image-id");
+        resourceType = $(this).data("resource-type");
+        resourceId = $(this).data("resource-id");
+        uploadData = App.Imageable.buildData([], $(".direct-upload").first());
+        App.Imageable.clearProgressBar(uploadData);
+        $(uploadData.progressBar).find(".loading-bar").css("width", "100%");
```
taitus: If this bar doesn't reflect real upload progress and we're always setting it to 100%, it might be clearer not to show it at all. A fake progress bar can be a bit misleading for users. Would it make sense to remove this?

taitus (commented):
Hi @stirkac 👋

Thanks again for the PR! I've left a few comments on the code (mostly small suggestions, questions, and potential simplifications). 

I've also attached a small patch in one of the comments with a possible alternative approach, just as a proposal to make the discussion easier.

Let me know what you think, and happy to iterate or clarify anything 👍

app/assets/javascripts/imageable.js
stirkac: You are right
In first instance, I want to to render the SuggestedImagesComponent, doing everything over JSON would be a tremendous overhead.
The attach request however uses dataType: "json" to directly mimic how direct uploads work (I didn't want to reinvent the wheel).

app/assets/javascripts/imageable.js
stirkac: sure, I've added a comment. It causes routing issue because we (ab)use submitting the form data into another endpoint (image_suggestions)

app/components/images/suggested_images_component.rb
stirkac: Sure, I've applied the patch. My idea was to be very open ended, and then use AR's native methods to take care of stuff like translations, should this ever expand into other resources. But, it's a bridge we can cross later.

app/assets/javascripts/imageable.js
taitus: Thanks for adding the explanation. I actually meant a short note in the commit message (not in the code). That way the "why" is in the git history and, for consistency with the rest of the codebase, we keep the code free of commented notes. So that same explanation in the commit message is enough. 👍

app/controllers/image_suggestions_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+    end
+  end
+
+  private
+
+    def image_suggestion_params
+      @image_suggestion_params ||= params.require(@resource_type.parameterize.underscore)
+                                         .permit(translations_attributes: [:title, :description])
+                                         .fetch(:translations_attributes, {})
+                                         .values&.first || {}
```
taitus: Small suggestion: in values&.first, I believe the safe navigation operator isn't necessary, since values always returns an Array.

docs/en/features/image_suggestions.md
```diff
@@ -0,0 +1,201 @@
...
+
+![LLM Configuration with Image Suggestions enabled](../../img/image_suggestions/llm-configuration-en.png)
+
+## How It Works
+
+### User Experience
+
+When creating or editing content (proposals, budget investments, debates, etc.) that supports images:
+
+1. **User fills in title and description**: The feature requires both title and description fields to be filled in order to generate meaningful image suggestions.
```
taitus: The feature seems to work as long as either title or description is present, while the docs mention both are required. Might be good to align the wording with the actual behavior.

docs/en/features/image_suggestions.md
```diff
@@ -0,0 +1,201 @@
...
+
+2. **User clicks "Suggest an image with AI ✨"**: A button appears next to the image upload field (only when no image is currently attached).
```
taitus:
```suggestion
2. **User clicks "Suggest an image with AI"**: A button appears next to the image upload field (only when no image is currently attached).
```

docs/en/features/image_suggestions.md
```diff
@@ -0,0 +1,201 @@
...
+
+   ![Selected image preview](../../img/image_suggestions/selected-image-preview-en.png)
+
+### Technical Flow
+
+1. **Frontend**: User clicks the suggestion button, which sends an AJAX request with the form data (title, description, resource type, etc.)
+
+2. **Backend Processing**:
+   - `ImageSuggestionsController#create` receives the request
+   - Creates a model instance with the provided attributes
```
taitus: Since the recent refactor removed the model instantiation from this step, the current wording is slightly outdated. The controller now only extracts the relevant attributes and delegates to the LLM client. Maybe we can update it to something like:

```suggestion
   - Extracts the relevant attributes from the request
```

docs/en/features/image_suggestions.md
```diff
@@ -0,0 +1,201 @@
...
+   - LLM model is selected
+   - Pexels API key is present in secrets
+
+![LLM Configuration with Image Suggestions enabled](../../img/image_suggestions/llm-configuration-en.png)
+
+## How It Works
+
+### User Experience
+
+When creating or editing content (proposals, budget investments, debates, etc.) that supports images:
```
taitus: Given that this PR only applies to the Budget Investments module for now, perhaps we could narrow this sentence to refer specifically to budget investments, and update it later when support for other resources is added.

docs/en/features/image_suggestions.md
```diff
@@ -0,0 +1,201 @@
...
+     - Validates LLM configuration
+     - Validates that title and/or description are present
+     - Sends the prompt to the LLM with title and description
+     - Receives a search query string
+     - Searches Pexels API with the query
+     - Returns up to 4 image results
+
+3. **Image Attachment**:
+   - When user clicks a suggested image, `ImageSuggestionsController#attach` is called
+   - Downloads the full-resolution image from Pexels
```
taitus: Since we actually fetch an optimized variant, maybe we can just say:

```suggestion
   - Downloads the image from Pexels
```

spec/components/images/suggest_images_component_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+    end
+
+    context "when object is not persisted" do
+      it "returns nil" do
+        component = Images::SuggestImagesComponent.new(form_builder)
+        expect(component.send(:resource_id)).to be(nil)
+      end
+    end
+  end
+end
```
taitus, line 97: I'm adding here the same spec file with a few changes (mainly for consistency with the rest of the project) and some additional tests that render the component.

Some of the changes you'll notice:
- For consistency with the current scope of the feature (Budget Investments), I switched from: `let(:form_object) { build(:proposal) }` to `let(:budget_investment) { build(:budget_investment) }`


- To stay aligned with other component specs where we pass a form object, I'm instantiating a ConsulFormBuilder:

```
let(:form) { ConsulFormBuilder.new(:budget_investment, budget_investment, ApplicationController.new.view_context, {}) }
```
instead of using:

```
let(:form_builder) { instance_double(ActionView::Helpers::FormBuilder, object: form_object) }
```
- Instead of testing `component.render?` directly:

```
component = Images::SuggestImagesComponent.new(form_builder)
expect(component.render?).to be false
```
I followed the pattern we usually use in component specs:
```
let(:component) { Images::SuggestImagesComponent.new(form) }

render_inline component
expect(page).to be_rendered
```
Here's the suggested update:

```suggestion
require "rails_helper"

describe Images::SuggestImagesComponent do
  let(:budget_investment) { build(:budget_investment) }
  let(:form) { ConsulFormBuilder.new(:budget_investment, budget_investment, ApplicationController.new.view_context, {}) }
  let(:component) { Images::SuggestImagesComponent.new(form) }

  before do
    Setting["llm.provider"] = nil
    Setting["llm.model"] = nil
    Setting["llm.use_ai_image_suggestions"] = nil
  end

  describe "rendering" do
    context "when all LLM settings are present" do
      before do
        Setting["llm.provider"] = "OpenAI"
        Setting["llm.model"] = "gpt-4o"
        Setting["llm.use_ai_image_suggestions"] = true
      end

      it "renders the suggest images wrapper with data attributes and button" do
        render_inline component

        expect(page).to be_rendered
        expect(page).to have_css ".suggested-images-wrapper"
        expect(page).to have_css ".suggested-images-container"
        expect(page).to have_css "button.js-suggest-image")
        expect(page).to have_css ".suggested-images-wrapper[data-resource-type='Budget::Investment']"
        expect(page).to have_content "Suggest an image with AI"
      end

      context "when budget_investment is persisted" do
        let(:budget_investment) { create(:budget_investment) }

        it "renders data-resource-id with the object id" do
          render_inline component

          expect(page).to have_css(".suggested-images-wrapper[data-resource-type='Budget::Investment'][data-resource-id='#{budget_investment.id}']")
        end
      end
    end

    context "when LLM settings are not configured" do
      it "does not render the component" do
        render_inline component

        expect(page).not_to be_rendered
      end
    end

    context "when provider is missing" do
      before do
        Setting["llm.provider"] = nil
        Setting["llm.model"] = "gpt-4o"
        Setting["llm.use_ai_image_suggestions"] = true
      end

      it "does not render the component" do
        render_inline component

        expect(page).not_to be_rendered
      end
    end

    context "when model is missing" do
      before do
        Setting["llm.provider"] = "OpenAI"
        Setting["llm.model"] = nil
        Setting["llm.use_ai_image_suggestions"] = true
      end

      it "does not render the component" do
        render_inline component

        expect(page).not_to be_rendered
      end
    end

    context "when use_ai_image_suggestions is missing" do
      before do
        Setting["llm.provider"] = "OpenAI"
        Setting["llm.model"] = "gpt-4o"
        Setting["llm.use_ai_image_suggestions"] = nil
      end

      it "does not render the component" do
        render_inline component

        expect(page).not_to be_rendered
      end
    end
  end

  describe "#resource_type" do
    it "returns the class name of the form object" do
      expect(component.send(:resource_type)).to eq "Budget::Investment"
    end
  end

  describe "#resource_id" do
    context "when budget_investment is persisted" do
      let(:budget_investment) { create(:budget_investment) }

      it "returns the object id" do
        expect(component.send(:resource_id)).to eq budget_investment.id
      end
    end

    context "when budget_investment is not persisted" do
      it "returns nil" do
        expect(component.send(:resource_id)).to be nil
      end
    end
  end
end
```

spec/components/images/suggested_images_component_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+      end
+
+      it "uses pointer cursor class on attach buttons" do
+        render_inline component
+
+        expect(page).to have_css(".suggested-image-button")
+      end
+    end
+  end
+end
```
taitus, line 139: In the `.patch` I've shared, the following changes are included:
- early returns when title/description are blank (and ensuring the LLM is not called),
- handling blank results from the LLM,
- rendering when there are no results and no errors,
- and the memoization of the LLM client call.

It also adds explicit accessibility assertions (role, aria-labels, alt text) and removes unnecessary setup like sign_in.

spec/controllers/image_suggestions_controller_spec.rb
```diff
@@ -0,0 +1,187 @@
...
+          resource_type: resource_type
+        }
+
+        expect(response).to have_http_status(:unprocessable_entity)
+        json_response = response.parsed_body
+        expect(json_response["errors"]).to eq("File is too large")
+      end
+    end
+  end
+end
```
taitus, line 110: In the `.patch` I've shared, the following changes are included:

Switches the main scope to Budget::Investment, removes redundant examples, adopts fixture_file_upload for consistency, and reduces unnecessary doubles to keep the spec focused on observable behavior and easier to maintain.

spec/lib/image_suggestions/llm/client_spec.rb
```diff
@@ -0,0 +1,187 @@
...
+      response.results = ["result1", "result2"]
+      expect(response.results).to eq(["result1", "result2"])
+    end
+
+    it "allows adding errors" do
+      response.errors << "error1"
+      expect(response.errors).to include("error1")
+    end
+  end
+end
```
taitus, line 188: This is essentially the same spec file with small changes like private logic is now tested through #call, and a few I18n.t checks were replaced with explicit messages. No behavioral changes 👌 


```suggestion
require "rails_helper"

describe ImageSuggestions::Llm::Client do
  let(:title) { "Test Proposal" }
  let(:description) { "Test description" }
  let(:chat) { instance_double(RubyLLM::Chat) }
  let(:context) { instance_double(RubyLLM::Context, chat: chat) }
  let(:prompt_template) { "Generate a search query for: %{title} - %{description}" }
  let(:search_query) { "test proposal image" }
  let(:pexels_results) { instance_double(::Pexels::PhotoSet, photos: []) }

  before do
    Setting["llm.provider"] = "OpenAI"
    Setting["llm.model"] = "gpt-4o"
    Setting["llm.use_ai_image_suggestions"] = true
    allow(Llm::Config).to receive(:context).and_return(context)
    allow(YAML).to receive(:load_file).and_return({ "image_suggestion_prompt" => prompt_template })
    allow(chat).to receive(:ask).and_return(double(content: search_query))
    allow(ImageSuggestions::Pexels).to receive(:search).and_return(pexels_results)
  end

  describe ".call" do
    it "creates a new instance and calls it with title and description" do
      expect(ImageSuggestions::Llm::Client).to receive(:new).with(
        title: title,
        description: description
      ).and_call_original

      ImageSuggestions::Llm::Client.call(title: title, description: description)
    end
  end

  describe "#call" do
    let(:client) { ImageSuggestions::Llm::Client.new(title: title, description: description) }

    it "generates a search query using LLM" do
      expect(chat).to receive(:ask).with(
        "Generate a search query for: Test Proposal - Test description"
      ).and_return(double(content: search_query))

      client.call
    end

    it "searches Pexels with the generated query" do
      expect(ImageSuggestions::Pexels).to receive(:search).with(
        search_query,
        per_page: 4
      ).and_return(pexels_results)

      result = client.call
      expect(result.results).to eq(pexels_results)
    end

    it "returns a Response object with results" do
      result = client.call
      expect(result).to be_a ImageSuggestions::Llm::Client::Response
      expect(result.results).to eq pexels_results
      expect(result.errors).to be_empty
    end

    context "when title and description are blank" do
      let(:title) { "" }
      let(:description) { "" }
      let(:client) { ImageSuggestions::Llm::Client.new(title: title, description: description) }

      it "adds error and returns early" do
        result = client.call
        expect(result.errors).to include "To suggest images, please write a title and description first."
        expect(result.results).to be_empty
      end

      it "does not call LLM or Pexels" do
        expect(chat).not_to receive :ask
        expect(ImageSuggestions::Pexels).not_to receive :search
        client.call
      end
    end

    context "when only description is present" do
      let(:title) { "" }
      let(:client) { ImageSuggestions::Llm::Client.new(title: title, description: description) }

      it "uses empty string for title in prompt" do
        expect(chat).to receive(:ask).with(
          "Generate a search query for:  - Test description"
        ).and_return(double(content: search_query))

        client.call
      end
    end

    context "when LLM response has surrounding whitespace" do
      before do
        allow(chat).to receive(:ask).and_return(double(content: "  search query  "))
      end

      it "passes stripped query to Pexels" do
        expect(ImageSuggestions::Pexels).to receive(:search).with(
          "search query",
          per_page: 4
        ).and_return(pexels_results)
        client.call
      end
    end

    context "when Pexels::APIError is raised" do
      before do
        allow(ImageSuggestions::Pexels).to receive(:search)
          .and_raise(::Pexels::APIError.new("API error"))
      end

      it "catches error and adds to response errors" do
        result = client.call
        expect(result.errors).to be_present
        expect(result.errors).to include "API error"
      end
    end

    context "when RubyLLM::Error is raised" do
      before do
        llm_error = RubyLLM::Error.new
        allow(llm_error).to receive(:message).and_return("LLM error")
        allow(chat).to receive(:ask).and_raise(llm_error)
      end

      it "catches error and adds to response errors" do
        result = client.call
        expect(result.errors).to include "LLM error"
      end
    end

    context "when provider is missing" do
      before { Setting["llm.provider"] = nil }

      it "adds error to response and returns early" do
        result = client.call

        expect(result.errors).to include("LLM is not properly configured. Please contact an administrator.")
        expect(result.results).to be_empty
      end

      it "does not call LLM or Pexels" do
        expect(chat).not_to receive :ask
        expect(ImageSuggestions::Pexels).not_to receive :search
        client.call
      end
    end

    context "when model is missing" do
      before { Setting["llm.model"] = nil }

      it "adds error to response and returns early" do
        result = client.call
        expect(result.errors).to include("LLM is not properly configured. Please contact an administrator.")
        expect(result.results).to be_empty
      end
    end

    context "when use_ai_image_suggestions is disabled" do
      before { Setting["llm.use_ai_image_suggestions"] = false }

      it "adds error to response and returns early" do
        result = client.call
        expect(result.errors).to include("LLM is not properly configured. Please contact an administrator.")
        expect(result.results).to be_empty
      end
    end
  end

  describe "Response" do
    let(:response) { ImageSuggestions::Llm::Client::Response.new }

    it "initializes with empty results and errors" do
      expect(response.results).to eq []
      expect(response.errors).to eq []
    end

    it "allows setting results" do
      response.results = ["result1", "result2"]
      expect(response.results).to eq ["result1", "result2"]
    end

    it "allows adding errors" do
      response.errors << "error1"
      expect(response.errors).to include "error1"
    end
  end
end
```

taitus, line 188: This is essentially the same spec file with small refinements: private logic is now tested through #call, and a few I18n.t checks were replaced with explicit messages. No behavioral changes.

Here's the suggested update:

spec/lib/image_suggestions/pexels_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+      end
+
+      it "raises PexelsError" do
+        expect do
+          pexels_instance.download(photo_id)
+        end.to raise_error(ImageSuggestions::Pexels::PexelsError, /Failed to download image/)
+      end
+    end
+  end
+end
```
taitus, line 115: This is essentially the same spec file with small refinements.

The refactor removes the `.search` and `.download` describe blocks and reorganizes the let definitions to make each context more focused.

No behavioral changes.

Here's the suggested update:

spec/shared/system/remotely_translatable.rb
```diff
@@ -180,7 +180,11 @@
 
       scenario "request a translation of an already translated text" do
         response = generate_response(resource)
-        expect_any_instance_of(translation_client_class).to receive(:call).and_return(response)
+        # Refactored from expect_any_instance to allow_any_instance_of
+        # Since the test possibly uses different browser sessions:
+        # Browser one click translation button → creates a RemoteTranslation job → creates a Client instance
+        # Browser two clicks the translation button → another RemoteTranslation job → creates another Client
+        allow_any_instance_of(translation_client_class).to receive(:call).and_return(response)
```
taitus: If this change was introduced to fix a flaky test and is not directly related to this PR, I would prefer to remove it.

We recently fixed a flaky issue in a very similar spec in "Cache the caller for a remote translation"(#6239), which was merged into master not long ago. I haven't had the chance to fully investigate whether this is related to that issue or not.

Given that, I'm not sure whether it makes more sense to open a separate PR for this change right now, or to wait and see if the flaky test reproduces again. If it does, I can then properly address it and create a dedicated PR for it 👌 .

spec/lib/image_suggestions/llm/client_spec.rb
taitus, line 188: In the .patch I've shared, the following small adjustments are included:

Private logic is now tested through #call, and a few I18n.t checks were replaced with explicit messages. No behavioral changes.

spec/lib/image_suggestions/pexels_spec.rb
taitus, line 115: In the .patch I've shared, the following small adjustments are included:

Removes the `.search` and `.download` describe blocks and simplifies the setup by restructuring the let definitions per context.

No behavioral changes.

spec/shared/system/remotely_translatable.rb
taitus: If this change was introduced to fix a flaky test and is not directly related to this PR, I would prefer to remove it.

We recently fixed a flaky issue in a very similar spec in "Cache the caller for a remote translation" (#6239), which was merged into master not long ago. I haven't had the chance to fully investigate whether this is related to that issue or not.

Given that, I'm not sure whether it makes more sense to open a separate PR for this change right now, or to wait and see if the flaky test reproduces again. If it does, I can then properly address it and create a dedicated PR for it 👌 .

docs/es/SUMMARY.md
```diff
@@ -51,6 +51,7 @@
   * [Local Census](features/local_census.md)
   * [Multitenancy](features/multitenancy.md)
   * [Traducciones de contenido de usuario](features/user_content_translations.md)
+  * [Sugerencias de Imágenes con IA](features/image_suggestions.md)
```
javierm, line 54: For consistency with the previous line:

```suggestion
  * [Sugerencias de imágenes con IA](features/image_suggestions.md)
```

docs/es/features/image_suggestions.md
```diff
@@ -0,0 +1,201 @@
...
+### El botón para solicitar imágenes no aparece
+
+- Verifica que el proveedor LLM y el modelo estén configurados en **Admin > Configuración Global > Configuración LLM**
+- Verifica que la configuración de Sugerencias de imágenes esté habilitada
+- Asegúrate de que la clave API de Pexels esté presente en `secrets.yml`
+- Revisa la consola del navegador para errores de JavaScript
+
+### No se sugieren imágenes
+
+- Pexels podría no tener fotos de stock para ciertos términos.
```
javierm: I guess we should either omit the period at the end or add it to the other lines 🤔.

```suggestion
- Pexels podría no tener fotos de stock para ciertos términos
```

docs/es/features/image_suggestions.md
```diff
@@ -0,0 +1,201 @@
...
+### Costos de Pexels
+
+Pexels ofrece un plan gratuito que debería ser suficiente para la mayoría de las instalaciones:
+
+- 200 solicitudes por hora
+- 20,000 solicitudes por mes
+
+Para mayor uso, Pexels ofrece planes de pago. Consulta [precios de la API de Pexels](https://www.pexels.com/api/pricing/) para más detalles.
+
+## Solución de Problemas
```
javierm:
```suggestion
## Solución de problemas
```

docs/es/features/image_suggestions.md
```diff
@@ -0,0 +1,201 @@
...
+
+La funcionalidad de Sugerencias de Imágenes con IA permite a los usuarios obtener recomendaciones de imágenes de stock mediante una búsqueda generada por IA al crear o editar contenido en Consul Democracy. En lugar de buscar y subir imágenes manualmente, los usuarios pueden hacer clic en un botón para recibir sugerencias relevantes de imágenes de stock de Pexels basadas en el título y/o la descripción de su contenido.
+
+La funcionalidad utiliza un Modelo de Lenguaje Grande (LLM) para analizar los campos de título y descripción de un recurso, extraer conceptos significativos y generar una consulta de búsqueda optimizada. Esta consulta se utiliza luego para buscar en la API de Pexels imágenes de stock relevantes. Cuando un usuario selecciona una imagen sugerida, se descarga y se adjunta a su contenido como si fuera una imagen subida por el usuario.
+
+## Prerrequisitos
+
+Para usar esta funcionalidad, necesitas:
+
+1. **Cuenta de Proveedor LLM**: Una cuenta con un proveedor LLM (OpenAI, Anthropic, DeepSeek, etc.) o un endpoint Ollama auto-hospedado
```
javierm: Looks like the technical term is "autoalojamiento" 🤔. So:

```suggestion
1. **Cuenta de Proveedor LLM**: Una cuenta con un proveedor LLM (OpenAI, Anthropic, DeepSeek, etc.) o un endpoint Ollama auto-alojado
```

docs/es/features/image_suggestions.md
```diff
@@ -0,0 +1,201 @@
...
+
+### Obtener una Clave API de Pexels
+
+1. Visita [Pexels API](https://www.pexels.com/api/)
+2. Regístrate para una cuenta gratuita o inicia sesión
+3. Navega a tu [panel de API](https://www.pexels.com/api/new/)
+4. Crea una nueva aplicación para obtener tu clave API
+5. Copia la clave API para usarla en la configuración
+
+**Nota**: Pexels ofrece un generoso plan gratuito con 200 solicitudes por hora y 20,000 solicitudes por mes, que debería ser suficiente para la mayoría de los casos de uso.
```
javierm: Small typo with the Spanish thousands:

```suggestion
**Nota**: Pexels ofrece un generoso plan gratuito con 200 solicitudes por hora y 20.000 solicitudes por mes, que debería ser suficiente para la mayoría de los casos de uso.
```

docs/es/features/image_suggestions.md
```diff
@@ -0,0 +1,201 @@
...
+```yml
+apis: &apis
+  microsoft_api_key: ""
+  census_api_end_point: ""
+  # ... otras configuraciones de API ...
+  pexels_access_key: "tu-clave-api-pexels"
+  llm: *llm
+```
+
+**Para configuraciones de multientidad**: Puedes sobrescribir la clave API de Pexels para inquilinos específicos:
```
javierm:
```suggestion
**Para configuraciones de multientidad**: Puedes sobrescribir la clave API de Pexels para entidades específicas:
```

docs/es/features/image_suggestions.md
```diff
@@ -0,0 +1,201 @@
...
+```
+
+El prompt usa los marcadores `%{title}` y `%{description}` que serán reemplazados con el contenido real del recurso que se está creando o editando.
+
+### Paso 3: Habilitar la Funcionalidad
+
+Una vez que hayas configurado tanto el proveedor LLM como la clave API de Pexels:
+
+1. Navega a **Admin > Configuración Global > Configuración LLM**
+2. Habilita el interruptor de funcionalidad **Sugerencias de Imágenes con IA**
```
javierm: "Interruptor de funcionalidad" sounds a bit strange to me 🤔. How about:

```suggestion
2. Habilita la funcionalidad **Sugerencias de Imágenes con IA**
```

docs/es/features/image_suggestions.md
```diff
@@ -0,0 +1,201 @@
...
+   - El modelo LLM está seleccionado
+   - La clave API de Pexels está presente en los secretos
+
+![Configuración LLM con Sugerencias de Imágenes habilitadas](../../img/image_suggestions/llm-configuration-es.png)
+
+## Cómo Funciona
+
+### Experiencia del Usuario
+
+Al crear o editar inversiones presupuestarias:
```
javierm:
```suggestion
Al crear o editar proyectos de gasto:
```

docs/es/features/image_suggestions.md
```diff
@@ -0,0 +1,201 @@
...
+   - La adjunta al formulario como si fuera subida por el usuario
+   - Reemplaza la interfaz de carga con la vista previa de la imagen seleccionada
+
+   ![Vista previa de imagen seleccionada](../../img/image_suggestions/selected-image-preview-es.png)
+
+### Flujo Técnico
+
+1. **Frontend**: El usuario hace clic en el botón de sugerencia, que envía una solicitud AJAX con los datos del formulario (título, descripción, tipo de recurso, etc.)
+
+2. **Procesamiento Backend**:
```
javierm:
```suggestion
2. **Procesamiento en el servidor**:
```

docs/es/features/image_suggestions.md
```diff
@@ -0,0 +1,201 @@
...
+
+## Recursos Soportados
+
+La funcionalidad de sugerencias de imágenes funciona con cualquier recurso que:
+
+- Tenga atributos `title` y `description` (o responda a estos métodos)
+- Soporte adjuntos de imágenes a través del sistema estándar de adjuntos de imágenes
+- Use `Images::NestedComponent`
+
+¡Actualmente, esto solo está implementado con la funcionalidad de Inversiones Presupuestarias!
```
javierm:
```suggestion
¡Actualmente, esto solo está implementado con la funcionalidad de proyectos de gasto!
```

docs/es/features/image_suggestions.md
```diff
@@ -0,0 +1,201 @@
...
+     - Recibe una cadena de consulta de búsqueda
+     - Busca en la API de Pexels con la consulta
+     - Devuelve hasta 4 resultados de imágenes
+
+3. **Adjuntar Imagen**:
+   - Cuando el usuario hace clic en una imagen sugerida, se llama a `ImageSuggestionsController#attach`
+   - Descarga la imagen de Pexels
+   - Crea un `ActionDispatch::Http::UploadedFile` a partir de la imagen descargada
+   - Usa el sistema `DirectUpload` existente para adjuntar la imagen
+   - Devuelve JSON con los detalles del adjunto para su procesamiento en el frontend
```
javierm:
```suggestion
   - Devuelve JSON con los detalles del adjunto para su procesamiento en el cliente
```

docs/es/features/image_suggestions.md
```diff
@@ -0,0 +1,201 @@
...
+
+- Entrada: ~50-100 tokens (título + descripción + prompt)
+- Salida: ~5-15 tokens (consulta de búsqueda)
+
+### Costos de Pexels
+
+Pexels ofrece un plan gratuito que debería ser suficiente para la mayoría de las instalaciones:
+
+- 200 solicitudes por hora
+- 20,000 solicitudes por mes
```
javierm:
```suggestion
- 20.000 solicitudes por mes
```

taitus (approved):

PR #6237, opened 2026-02-12
Add support for Vertex AI using gcloud and key file authentication

digitalWestie:
## Objectives

The goal of the changes here is to add support for VertexAI as an LLM provider by upgrading the RubyLLM gem. VertexAI doesn't use an API key like most of the other providers, therefore this PR introduces changes to authenticate either by specifying the path to a service account key file OR gcloud cli authentication methods.  

## Visual Changes

None

### Authentication

Vertex auth is done via [Application Default Credentials (ADC)](https://docs.cloud.google.com/docs/authentication/application-default-credentials), Google Cloud's automatic way of finding and using credentials. 

**Limitations**

Unfortunately the way ruby-llm uses googleauth is process global. Therefore, we can't use different credentials for different tenants.

**Environment variable method**

`GOOGLE_APPLICATION_CREDENTIALS` is an environment variable to provide the location of a credential JSON file. This can be provided by setting it in the environment, or, setting the `apis.google_application_credentials` secret (see: `secrets.yml.example`. 

For example:

```yml
# secrets.yml
llm: &llm
  vertexai_project_id: "sensemaker-466109"
  vertexai_location: "us-central1"

apis: &apis
  google_application_credentials: "/home/digitalwestie/service-account-key.json"
```

**CLI** 

Make sure [the gcloud cli is installed](https://docs.cloud.google.com/sdk/docs/install-sdk) then run:

```bash
gcloud auth application-default login
```

This opens a browser window that you can use to login with a google account. Once that process is complete credentials will be saved by the CLI.  RubyLLM in Consul can use Vertex AI simply by providing a vertex config e.g.:

```yml
# secrets.yml
llm: &llm
  vertexai_project_id: "sensemaker-466109"
  vertexai_location: "us-central1"
```

### DIFF
--- a/app/lib/llm/config.rb
+++ b/app/lib/llm/config.rb
@@ -3,7 +3,7 @@ module Llm
     class << self
       def context
         @context = RubyLLM.context do |config|
-          ENV["GOOGLE_APPLICATION_CREDENTIALS"] ||= Rails.application.secrets[:google_application_credentials]
+          ENV["GOOGLE_APPLICATION_CREDENTIALS"] ||= Rails.application.secrets.google_application_credentials
 
           Tenant.current_secrets.llm&.each do |key, value|
             config.send("#{key}=", value)


Gemfile
```diff
@@ -29,6 +29,7 @@ gem "file_validators", "~> 3.0.0"
 gem "font-awesome-sass", "~> 5.15.1" # Remember to update vendor/assets/images/fontawesome when updating this gem
 gem "globalize", "~> 7.0.0"
 gem "globalize-accessors", "~> 0.3.0"
+gem "googleauth", "~> 1.16"
```
javierm: For consistency with the rest of the file, what do you think about:

```suggestion
gem "googleauth", "~> 1.16.1"
```

app/lib/llm/config.rb
```diff
@@ -3,6 +3,11 @@ class Config
     class << self
       def context
         @context = RubyLLM.context do |config|
+          credentials_path = Rails.application.secrets[:google_application_credentials]
```
digitalWestie: Switched to using Rails secrets rather than Tenant as an intentional cue we're not supporting multi-tenancy.

app/lib/llm/config.rb
```diff
@@ -3,6 +3,11 @@ class Config
...
+          if credentials_path.present?
+            ENV["GOOGLE_APPLICATION_CREDENTIALS"] = credentials_path
+          end
```
javierm: Here's a possible simplification:

```suggestion
          ENV["GOOGLE_APPLICATION_CREDENTIALS"] ||= Rails.application.secrets[:google_application_credentials]
```

This way people could still pass the `GOOGLE_APPLICATION_CREDENTIALS` environment variable some other way, and it'd take precedence over the one in the secrets.

spec/lib/llm/config_spec.rb
```diff
@@ -12,6 +12,26 @@
...
+      context = double("RubyLLM::Context", config: config)
+      expect(RubyLLM).to receive(:context).and_yield(config).and_return(context)
+
+      original = ENV["GOOGLE_APPLICATION_CREDENTIALS"]
+      begin
+        Llm::Config.context
+        expect(ENV["GOOGLE_APPLICATION_CREDENTIALS"]).to eq("/tmp/dummy.json")
+      ensure
+        ENV["GOOGLE_APPLICATION_CREDENTIALS"] = original
+      end
```
javierm: While this is fine 👌, we usually don't use `begin...ensure` blocks but use `after` blocks instead. So what do you think about:

```ruby
    context "google_application_credentials is present" do
      before do
        stub_secrets(
          llm: { openai_api_key: "1234" },
          google_application_credentials: "/tmp/dummy.json"
        )
      end

      let!(:original_google_application_credentials) { ENV["GOOGLE_APPLICATION_CREDENTIALS"] }
      after { ENV["GOOGLE_APPLICATION_CREDENTIALS"] = original_google_application_credentials }

      it "sets GOOGLE_APPLICATION_CREDENTIALS" do
        config = instance_double(RubyLLM::Configuration)
        allow(config).to receive(:openai_api_key=)
        context = double("RubyLLM::Context", config: config)
        expect(RubyLLM).to receive(:context).and_yield(config).and_return(context)

        Llm::Config.context

        expect(ENV["GOOGLE_APPLICATION_CREDENTIALS"]).to eq("/tmp/dummy.json")
      end
    end
```

spec/lib/llm/config_spec.rb
```diff
@@ -12,6 +12,26 @@
...
+    it "sets GOOGLE_APPLICATION_CREDENTIALS when google_application_credentials is present" do
+      stub_secrets(
+        llm: { openai_api_key: "1234" },
+        google_application_credentials: "/tmp/dummy.json"
+      )
+
+      config = instance_double(RubyLLM::Configuration)
+      allow(config).to receive(:openai_api_key=)
+      context = double("RubyLLM::Context", config: config)
+      expect(RubyLLM).to receive(:context).and_yield(config).and_return(context)
```
javierm: Since these lines are pretty much identical to the ones in the previous test, would it make sense to extract them to a `before` block? 🤔

app/lib/llm/config.rb
```diff
@@ -3,6 +3,8 @@ class Config
     class << self
       def context
         @context = RubyLLM.context do |config|
+          ENV["GOOGLE_APPLICATION_CREDENTIALS"] ||= Rails.application.secrets[:google_application_credentials]
```
javierm: Sorry I forgot about this one :pray:. For consistency with the rest of the code:

```suggestion
          ENV["GOOGLE_APPLICATION_CREDENTIALS"] ||= Rails.application.secrets.google_application_credentials
```

PR #6331, opened 2026-04-20
Make remotely translatable specs faster

taitus:
## References
- Make moderation specs faster #6178
- Make nested documentable specs faster #5989

## Objectives
As we did in PR #5369, #5797, #5989 and #6178, we reduce the number of tests we are going to run. In this case we remove the shared example file and create a unified spec file that runs the same tests but with a different factory, provider and path in each execution for each test. As they are the same tests and based on the same functionality, we use this related PR technique (random factory) to make sure that everything still works.


spec/system/remotely_translatable_spec.rb
```diff
@@ -0,0 +1,254 @@
...
+  def generate_response(resource)
+    field_text = Faker::Lorem.characters(number: 10)
+    resource.translated_attribute_names.map { field_text }
+  end
+
+  def show_path(**opts)
+    case factory
+    when :budget_investment then budget_investment_path(resource.budget, resource, **opts)
+    else polymorphic_path(resource, **opts)
+    end
```
javierm: Is the condition necessary? 🤔

```suggestion
    polymorphic_path(resource, **opts)
```

spec/system/remotely_translatable_spec.rb
```diff
@@ -0,0 +1,254 @@
...
+  end
+
+  def index_path(**opts)
+    case factory
+    when :budget_investment then budget_investments_path(resource.budget, **opts)
+    else polymorphic_path(collection_symbol, **opts)
+    end
```
javierm: To be consistent with what we usually do, what do you think about:

```suggestion
    if factory == :budget_investment
      budget_investments_path(resource.budget, **opts)
    else
      polymorphic_path(collection_symbol, **opts)
    end
```

spec/system/remotely_translatable_spec.rb
```diff
@@ -0,0 +1,254 @@
...
+    :legislation_process,
+    :proposal
+  ]
+
+  let(:factory) { factories.sample }
+  let!(:resource) { create(factory) }
+  let(:provider) { [RemoteTranslations::Llm, RemoteTranslations::Microsoft].sample }
+  let(:client) { provider::Client }
+  let(:available_locales) { provider::AvailableLocales }
+  let(:collection_symbol) { factory.to_s.pluralize.to_sym }
```
javierm: To simplify the setup a little bit, what do you think about removing this line, since it's only used once, and changing the line that uses `collection_symbol`?

spec/system/remotely_translatable_spec.rb
```diff
@@ -0,0 +1,252 @@
...
+  let(:path) do
+    paths = []
+    paths = [show_path, index_path] if factory != :legislation_process
+    paths << root_path if factory != :budget_investment
+    paths.sample
+  end
+
+  let(:untranslated_path) { "#{path}?locale=es" }
```
javierm: Unless we've got a strong preference about the name, maybe we could use the name we were using before 🤔. What do you think?

```suggestion
  let(:path_in_spanish) { "#{path}?locale=es" }
```

spec/system/remotely_translatable_spec.rb
```diff
@@ -0,0 +1,252 @@
...
+      visit untranslated_path
+
+      expect(page).to have_button "Traducir página"
+    end
+
+    scenario "should not be present when new current locale translation exists" do
+      add_translations(resource, :es)
+      visit path
+
+      expect(page).not_to have_button "Traducir página"
```
javierm: Is there a typo here? The original said:

```suggestion
      expect(page).not_to have_button "Translate page"
```

spec/system/remotely_translatable_spec.rb
```diff
@@ -0,0 +1,252 @@
...
+
+          select "Español", from: "Language:"
+
+          expect(page).to have_select "Idioma:"
+          expect(page).not_to have_button "Traducir página"
+        end
+      end
+
+      describe "should evaluate missing translations on featured resources" do
+        let(:factory) { (factories - [:budget_investment, :legislation_process]).sample }
```
javierm: "Featured" is something exclusive to debates and proposals, right? If so:

```suggestion
        let(:factory) { [:debate, :proposal].sample }
```

spec/system/remotely_translatable_spec.rb
```diff
@@ -0,0 +1,252 @@
...
+    comment = create(:comment, commentable: resource)
+    add_translations(comment, locale)
+  end
+
+  def generate_response(resource)
+    field_text = Faker::Lorem.characters(number: 10)
+    resource.translated_attribute_names.map { field_text }
+  end
+
+  def show_path
```
javierm: What do you think about moving this method (and the `index_path` method) to a `let` clause now that it doesn't receive parameters?

spec/shared/system/remotely_translatable.rb
```diff
@@ -1,233 +0,0 @@
-require "sessions_helper"
```
javierm, line 1: Looks like there are some tests that call the `in_browser` method without requiring the `sessions_helper` file. Those tests have stopped working after removing this file 🤔.

There are three possible solutions:

1. Include `sessions_helper` in every file that uses `in_browser`.
2. Remove every reference to `sessions_helper` and include it inside `rails_helper`.
3. Remove `sessions_helper` and replace `in_browser` with `using_session`.

I'd vote for the third option.

javierm (commented):
Noticed a test had failed. So one more comment!

spec/system/admin/stats_spec.rb
```diff
@@ -24,7 +24,7 @@
 
       allow_any_instance_of(ActionDispatch::Request).to receive(:user_agent).and_return("Different")
 
-      in_browser(:different) do
+      Capybara.using_session(:different) do
```
javierm: Do we need the `Capybara.` prefix? 🤔

```suggestion
      using_session(:different) do
```

javierm (approved):

Issue #5485, opened 2024-04-08
Replace Turbolinks with Turbo or drop it entirely


javierm:
## Background

We use [Turbolinks](https://github.com/turbolinks/turbolinks) to emulate a single-page application, which was the default Rails configuration for years.

However, Turbolinks is no longer being developed, and its last release took place in 2018. It has been replaced by [Turbo](https://github.com/hotwired/turbo).

On the other hand, Turbolinks adds complexity to our application in many ways, and we expect Turbo to behave similarly. It also provides a different user experience than most pages since it doesn't use the browser's default mechanism showing that a page is being loaded, which might confuse some people. What would happen if we removed it? Would we have a performance penalty?

## Definition of Done

1. A prototype (i.e. not everything needs to work) replacing Turbolinks with Turbo is provided
2. A prototype removing Turbolinks is provided (you can find one in a commit in pull request #5484)
3. A performance comparison between using Turbo and removing Turbolinks is provided

PR #6353, opened 2026-05-06
Replace Turbolinks with @hotwired/turbo 

AleVeraM:
Replaces the turbolinks gem with the @hotwired/turbo                                                                         
                                                                                                                        
  Changes:                                                                                                              
  - Add @hotwired/turbo via npm, remove turbolinks gem                                                                  
  - Update all JS event listeners from turbolinks:* to turbo:*                                                          
  - Update data-turbolinks attributes to data-turbo in views and components                                             
  - Update asset tracking from data-turbolinks-track to data-turbo-track                                                
  - Delete turbolinks_anchors.js (Turbo handles anchor navigation natively)                                              
  - Update ESLint globals and specs accordingly


lucialuzuriaga (commented):
Hi @AleVeraM, thank you for your pull request. It seems that all tests are currently failing. Could you take a look? Thanks!

AleVeraM:
Hi @lucialuzuriaga, thanks for the heads up! I've fixed the failing tests and all CI checks are now passing. Could you take another look when you have a moment? Thanks!

lucialuzuriaga (commented):
Hi @AleVeraM! We will review your PR soon and add it to our next release. However, it seems that during these weeks some merge conflicts took place. Could you kindly check? Thanks!

javierm:
Hi, @AleVeraM :smile:.

One initial question: why are we using the NPM package instead of the turbo-rails gem? :thinking: Mind you, I'm not saying we should use the gem - I haven't checked the advantages and disadvantages. Just curious about the reasons.

javierm:
By the way, let's use Turbo 7 for now instead of Turbo 8. We usually upgrade one major version at a time, and there should be better browser support in Turbo 7.

AleVeraM:
Hi @javierm!

turbo-rails brings extra functionality we don't need — we just want the Turbo JS runtime, served via Sprockets like our other JS deps. That's why npm over the gem.

Downgraded to ^7.3.0 as requested.

app/components/layout/legal_links_component.html.erb
```diff
@@ -5,7 +5,7 @@
   <li><%= repo_link %></li>
   <% if feature?(:cookies_consent) %>
     <li class="cookies-consent-management-footer-content">
-      <a href="#cookies_consent_management" data-open="cookies_consent_management">
+      <a href="#cookies_consent_management" data-open="cookies_consent_management" data-turbo="false">
```
javierm, line 8: Instead of adding this attribute, would it be possible to do something to prevent Turbo from managing anchor links, like we did in the `turbolinks_anchors.js` file? That way, it would work for every anchor link. If so, we could also remove the `data: { turbo: false }` attribute in the skip to main content link.

package.json
```diff
@@ -3,6 +3,7 @@
   "dependencies": {
     "@deltablot/dropzone": "^7.4.3",
     "@fortawesome/fontawesome-free": "^6.7.2",
+    "@hotwired/turbo": "^7.3.0",
```
javierm, line 6: Could we use the gem instead of the Node package? Here's the reason: Turbo 7.x reports a security vulnerability. This vulnerability doesn't seem to affect us, but it generates noise when installing Turbo as a Node package 😅.

app/assets/javascripts/application.js
```diff
@@ -165,5 +164,11 @@ var destroy_non_idempotent_modules = function() {
   App.SocialShare.destroy();
 };
 
-$(document).on("turbolinks:load", initialize_modules);
-$(document).on("turbolinks:before-cache", destroy_non_idempotent_modules);
+// Opt-in mode: Turbo only intercepts forms with data-turbo="true".
+// Existing forms rendered directly by controllers (200 OK instead of redirect)
+// are left to the browser/jquery-ujs. Remove this once all controllers follow
+// the Post/Redirect/Get pattern and forms are progressively opted in.
```
javierm, line 170: We prefer using the commit messages instead of adding comments to the code. Since the commit message already explains the reason for this code, we can remove this comment.

app/views/admin/budget_investments/show.html.erb
```diff
@@ -1,5 +1,5 @@
 <%= link_to admin_budget_budget_investments_path(Budget::Investment.filter_params(params).to_h),
-            class: "back", data: { turbolinks: false } do %>
+            class: "back", data: { turbo: false } do %>
```
javierm, line 2: Could yo add a similar change to `app/views/notifications/_notification.html.erb`? There's still a link with `turbolinks: false` there.

javierm:
@AleVeraM Thank you very much for this pull request! It's helped us realize that the problem is more complex than we first though. Back in 2024, we though Turbo was a replacement for Turbolinks. But it's actually a replacement for Turbolinks **and** rails-ujs. Having Turbo and rails-ujs at the same time but be possible, but it seems not to be well-supported.

So we've decided we'll only use Turbo when we've also changed the code that uses rails-ujs. We've updated issue #5485 accordingly.

I'm going to close this pull request, but if you'd like to reopen it and keep working on it to replace rails-ujs with Turbo and provide a performance comparison between using Turbo or simply removing Turbolinks, be my guest. In any case, this pull request has already been helpful and the contribution is much appreciated :pray:.

javierm (commented):
Leaving a few comments so we don't forget about them in the future.

PR #6367, opened 2026-05-21
Add creation date and comments to CSV exports

taitus:
## References
Related PR's: #6302  and #6189 

## Objectives

Extend the admin CSV exports for debates, proposals and budget investments so admins can analyze content outside the application:

* Add a `Created at` column to all three exports.
* Add associated comments to the CSV, using one row per comment and repeating the parent record data on each row.
* Include comment metadata: ID, author (email), body (without HTML), parent ID and creation date.

## Release Notes
⚠️  https://github.com/consuldemocracy/consuldemocracy/pull/6367#discussion_r3324853745
> The investments CSV export format has changed. A single investment can now generate multiple rows in the CSV file (one line per comment related to that investment). If you're using custom software to process this CSV export, you might have to modify it accordingly.

### DIFF
--- a/app/models/budget/investment/exporter.rb
+++ b/app/models/budget/investment/exporter.rb
@@ -14,7 +14,7 @@ class Budget::Investment::Exporter
 
   private
 
-    def headers
+    def model_headers
       [
         I18n.t("admin.budget_investments.index.list.id"),
         I18n.t("admin.budget_investments.index.list.title"),
@@ -32,7 +32,7 @@ class Budget::Investment::Exporter
       ]
     end
 
-    def csv_values(investment)
+    def record_csv_values(investment)
       [
         investment.id.to_s,
         investment.title,
--- a/app/models/budget/investment/exporter.rb
+++ b/app/models/budget/investment/exporter.rb
@@ -27,7 +27,8 @@ class Budget::Investment::Exporter
         I18n.t("admin.budget_investments.index.list.valuation_finished"),
         I18n.t("admin.budget_investments.index.list.selected"),
         I18n.t("admin.budget_investments.index.list.visible_to_valuators"),
-        I18n.t("admin.budget_investments.index.list.author_username")
+        I18n.t("admin.budget_investments.index.list.author_username"),
+        Budget::Investment.human_attribute_name(:created_at)
       ]
     end
 
@@ -44,7 +45,8 @@ class Budget::Investment::Exporter
         investment.valuation_finished? ? I18n.t("shared.yes") : I18n.t("shared.no"),
         investment.selected? ? I18n.t("shared.yes") : I18n.t("shared.no"),
         investment.visible_to_valuators? ? I18n.t("shared.yes") : I18n.t("shared.no"),
-        investment.author.username
+        investment.author.username,
+        I18n.l(investment.created_at, format: :datetime)
       ]
     end
 
--- a/app/models/concerns/csv_exporter.rb
+++ b/app/models/concerns/csv_exporter.rb
@@ -2,7 +2,63 @@ module CsvExporter
   def to_csv
     CSV.generate(headers: true) do |csv|
       csv << headers
-      records.each { |record| csv << csv_values(record) }
+      records.each do |record|
+        comments = record.comments.sort_by(&:created_at)
+
+        if comments.any?
+          comments.each { |comment| csv << csv_values(record, comment) }
+        else
+          csv << csv_values(record)
+        end
+      end
     end
   end
+
+  private
+
+    def csv_values(record, comment = nil)
+      record_csv_values(record) + comment_csv_values(comment)
+    end
+
+    def headers
+      model_headers + comment_csv_headers
+    end
+
+    def record_csv_values(record)
+      raise "This method must be implemented in class #{self.class.name}"
+    end
+
+    def model_headers
+      raise "This method must be implemented in class #{self.class.name}"
+    end
+
+    def comment_csv_headers
+      [
+        I18n.t("admin.comments.index.id"),
+        I18n.t("admin.comments.index.author"),
+        I18n.t("admin.comments.index.content"),
+        Comment.human_attribute_name(:parent_id),
+        Comment.human_attribute_name(:created_at)
+      ].map { |name| comment_header(name) }
+    end
+
+    def comment_csv_values(comment)
+      return empty_comment_csv_values unless comment
+
+      [
+        comment.id.to_s,
+        comment.author&.email.to_s,
+        comment.body,
+        comment.parent_id.to_s,
+        I18n.l(comment.created_at, format: :datetime)
+      ]
+    end
+
+    def empty_comment_csv_values
+      Array.new(comment_csv_headers.size, "")
+    end
+
+    def comment_header(name)
+      "#{Comment.model_name.human} #{name}"
+    end
 end
--- a/app/models/concerns/csv_exporter.rb
+++ b/app/models/concerns/csv_exporter.rb
@@ -3,7 +3,7 @@ module CsvExporter
     CSV.generate(headers: true) do |csv|
       csv << headers
       records.each do |record|
-        comments = record.comments.sort_by(&:created_at)
+        comments = comments_in_reading_order(record)
 
         if comments.any?
           comments.each { |comment| csv << csv_values(record, comment) }
@@ -60,5 +60,11 @@ module CsvExporter
 
     def comment_header(name)
       "#{Comment.model_name.human} #{name}"
+    end
+
+    def comments_in_reading_order(record)
+      Comment.sort_by_ancestry(record.comments) do |comment, sibling|
+        comment.created_at <=> sibling.created_at
+      end
     end
 end
--- a/app/models/debate/exporter.rb
+++ b/app/models/debate/exporter.rb
@@ -13,7 +13,8 @@ class Debate::Exporter
       [
         I18n.t("admin.debates.index.id"),
         Debate.human_attribute_name(:title),
-        I18n.t("admin.debates.index.author")
+        I18n.t("admin.debates.index.author"),
+        Debate.human_attribute_name(:created_at)
       ]
     end
 
@@ -21,7 +22,8 @@ class Debate::Exporter
       [
         debate.id.to_s,
         debate.title,
-        debate.author.email
+        debate.author.email,
+        I18n.l(debate.created_at, format: :datetime)
       ]
     end
 end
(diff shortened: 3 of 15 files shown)


spec/system/admin/proposals_spec.rb
```diff
@@ -123,10 +123,28 @@
...
+      root_comment = create(:comment, commentable: proposal, body: "I think it should")
+      reply = create(:comment, commentable: proposal, body: "I disagree", parent_id: root_comment.id)
+
+      visit admin_proposals_path
+
+      click_link "Download current selection"
+
+      csv_contents = <<~CSV
+        ID,Proposal,Author,Summary,Created at,Comment ID,Comment Author,Comment Content,Comment Parent,Comment Created at
+        #{proposal.id},#{proposal.title},#{proposal.author.email},#{proposal.summary},#{I18n.l(proposal.created_at, format: :datetime)},#{root_comment.id},#{root_comment.author.email},#{root_comment.body},"",#{I18n.l(root_comment.created_at, format: :datetime)}
```
javierm: I've just opened #6372 to detect this kind of incredibly long lines 😌.

config/locales/en/activerecord.yml
```diff
@@ -1,5 +1,6 @@
 en:
   attributes:
+    created_at: Created at
```
javierm, line 3: Since we're adding this generic attribute, what do you think about removing `created_at` from the `site_customization/page` attributes? 🤔 

(I think we aren't using the attribute name in site customization pages anyway.)

Same in Spanish.

spec/system/admin/budget_investments_spec.rb
```diff
@@ -1686,12 +1686,44 @@
...
 
       csv_contents = "ID,Title,Supports,Administrator,Valuator,Valuation Group,Scope of operation," \
-                     "Feasibility,Val. Fin.,Selected,Show to valuators,Author username\n" \
+                     "Feasibility,Val. Fin.,Selected,Show to valuators,Author username,Created at," \
+                     "Comment ID,Comment Author,Comment Content,Comment Parent,Comment Created at\n" \
                      "#{first_investment.id},Le Investment,88,Admin,-,Valuator Group," \
                      "Budget Heading,Feasible (€99),Yes,Yes,Yes," \
-                     "#{first_investment.author.username}\n#{second_investment.id}," \
+                     "#{first_investment.author.username}," \
+                     "#{I18n.l(first_investment.created_at, format: :datetime)}," \
```
javierm: Instead of using `I18n.l`, what do you think about setting the `created_at` attribute in the records (or using `travel_to` to do so) and checking the actual date here? This way, when reading the test, it'll be more obvious what the expected result is.

spec/system/admin/budget_investments_spec.rb
```diff
@@ -1686,12 +1686,44 @@
...
                      "Alt Investment,66,No admin assigned,Valuator,-,Other Heading," \
-                     "Unfeasible,No,No,No,#{second_investment.author.username}\n"
+                     "Unfeasible,No,No,No,#{second_investment.author.username}," \
+                     "#{I18n.l(second_investment.created_at, format: :datetime)}," \
+                     "\"\",\"\",\"\",\"\",\"\"\n"
+
+      expect(page.body).to eq(csv_contents)
+    end
+
+    scenario "Downloading CSV file with comments" do
```
javierm: What do you think about moving this test to a model test for the `Budget::Investment::Exporter` class? This way we could simplify the test and make it faster 🤔. We'd still be testing the download link functionality thanks to the test preceding this one.

Same with debates and proposals.

app/models/debate/exporter.rb
```diff
@@ -1,5 +1,6 @@
 class Debate::Exporter
   include CsvExporter
+  include JsonExporter
```
javierm: We don't export debates to JSON anywhere, right? 🤔

I take it this is only used in order to use `strip_tags` in the CSV exporter? If so, I'd vote to either move that method somewhere else or simply using `ActionView::Base.full_sanitizer.sanitize(html_string)` 🤔.

app/models/concerns/csv_exporter.rb
```diff
@@ -2,7 +2,51 @@ module CsvExporter
...
+      ]
+    end
+
+    def comment_csv_values(comment)
+      return empty_comment_csv_values unless comment
+
+      [
+        comment.id.to_s,
+        comment.author&.email.to_s,
+        strip_tags(comment.body),
```
javierm: Why are we stripping tags here? 🤔

app/models/budget/investment/exporter.rb
```diff
@@ -27,11 +27,12 @@ def headers
...
         I18n.t("admin.budget_investments.index.list.visible_to_valuators"),
-        I18n.t("admin.budget_investments.index.list.author_username")
-      ]
+        I18n.t("admin.budget_investments.index.list.author_username"),
+        Budget::Investment.human_attribute_name(:created_at)
+      ] + comment_csv_headers
     end
 
-    def csv_values(investment)
+    def record_csv_values(investment)
```
javierm, line 35: Shall we add a release note, so people who are using the investments CSV export are aware that the format has changed and now there are multiple rows per investment? 🤔

app/models/concerns/csv_exporter.rb
```diff
@@ -2,7 +2,51 @@ module CsvExporter
...
+      ]
+    end
+
+    def empty_comment_csv_values
+      Array.new(comment_csv_headers.size, "")
+    end
+
+    def comment_header(name)
+      "#{Comment.model_name.human} #{name}"
+    end
```
javierm, line 63: Maybe we could something similar to what the JsonExporter does 🤔.

```
    def record_csv_values(record)
      raise "This method must be implemented in class #{self.class.name}"
    end

    def headers
      raise "This method must be implemented in class #{self.class.name}"
    end

File: app/models/debate/exporter.rb [hunk 9]
```diff
@@ -13,15 +14,17 @@ def headers
       [
         I18n.t("admin.debates.index.id"),
         Debate.human_attribute_name(:title),
-        I18n.t("admin.debates.index.author")
-      ]
+        I18n.t("admin.debates.index.author"),
+        Debate.human_attribute_name(:created_at)
+      ] + comment_csv_headers
```
javierm: What do you think about removing this line and renaming this method to `record_headers`? Then add a `headers` method to the `CsvExporter` module. Same thing we've done with values.

javierm (commented):
@taitus I've left a few comments. Let me know what you think! :wink:

app/models/concerns/csv_exporter.rb
```diff
@@ -2,7 +2,63 @@ module CsvExporter
...
+
+    def comment_csv_headers
+      [
+        comment_header(I18n.t("admin.comments.index.id")),
+        comment_header(I18n.t("admin.comments.index.author")),
+        comment_header(I18n.t("admin.comments.index.content")),
+        comment_header(Comment.human_attribute_name(:parent_id)),
+        comment_header(Comment.human_attribute_name(:created_at))
+      ]
```
javierm: What do you think about:

```suggestion
      [
        I18n.t("admin.comments.index.id"),
        I18n.t("admin.comments.index.author"),
        I18n.t("admin.comments.index.content"),
        Comment.human_attribute_name(:parent_id),
        Comment.human_attribute_name(:created_at)
      ].map { |name| comment_header(name) }
```

app/models/concerns/csv_exporter.rb
```diff
@@ -2,7 +2,63 @@ module CsvExporter
   def to_csv
     CSV.generate(headers: true) do |csv|
       csv << headers
-      records.each { |record| csv << csv_values(record) }
+      records.each do |record|
+        comments = record.comments.sort_by(&:created_at)
```
javierm: Would it make sense to sort the comments in the same order people would read them if they're reading the comment tree? 🤔

Not sure how hard it is 🤔.

javierm (commented):

spec/models/csv_exporter_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+                                    body: "First thread",
+                                    created_at: Time.zone.local(2026, 6, 1, 15, 0, 0))
+      second_root = create(:comment, commentable: resource,
+                                     body: "Second thread",
+                                     created_at: Time.zone.local(2026, 6, 1, 15, 5, 0))
+      reply = create(:comment, commentable: resource,
+                               body: "Reply to first",
+                               parent_id: first_root.id,
+                               created_at: Time.zone.local(2026, 6, 1, 15, 10, 0))
+      reply2 = create(:comment, commentable: resource,
```
javierm: Since we use `second_root`:

```suggestion
      second_reply = create(:comment, commentable: resource,
```

spec/models/csv_exporter_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    let(:exporter_class) do
+      case factory
+      when :proposal
+        Proposal::Exporter
+      when :debate
+        Debate::Exporter
+      when :budget_investment
+        Budget::Investment::Exporter
+      end
+    end
```
javierm, line 20: To make it easy to follow, what do you think about defining `let(:factory)` before defining `let(:exporter_class)`? 🤔

spec/models/csv_exporter_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    let(:factory) { factories.sample }
+    let!(:resource) { create(factory) }
```
javierm: Do we need the exclamation mark here? 🤔

```suggestion
    let(:resource) { create(factory) }
```

spec/models/csv_exporter_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+      nested_reply = create(:comment, commentable: resource,
+                                      body: "Reply to first reply",
+                                      parent_id: reply.id,
+                                      created_at: Time.zone.local(2026, 6, 1, 15, 15, 0))
+
+      csv = exporter_class.new([resource]).to_csv
+      rows = CSV.parse(csv, headers: true)
+      comment_content_column = rows.map { |row| row["Comment Content"] }
+
+      expect(comment_content_column).to eq [
```
javierm:
```suggestion

      expect(rows.by_col["Comment Content"]).to eq [
```

javierm (commented):
`CSV.parse` is so much better :tada:. I've left one more commet.

javierm (approved):

PR #6401, opened 2026-06-10
Replace capybara-webmock with Puffing Billy

taitus:
## References
Related PR: #6399 

While working on the Rack 3 upgrade, we found that `capybara-webmock` needs local patches to keep working with Rack 3. The gem is used for system specs, so this PR replaces it with Puffing Billy instead of adding a temporary patch for `capybara-webmock`.

## Objectives

Replace `capybara-webmock` with Puffing Billy for system specs.

This keeps the same general purpose: Chrome/Selenium traffic goes through a test proxy, local application requests are allowed, and external browser requests are blocked so the test suite does not depend on external services.

Require `billy/capybara/rspec` so the proxy is stopped after the suite; without it the CI process hangs even when all tests pass.

## Notes

One behaviour difference is that Puffing Billy blocks non-whitelisted external requests, while `capybara-webmock` returned an empty 200 response.

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -19,7 +19,12 @@ require "custom_spec_helper"
 require "capybara/rails"
 require "capybara/rspec"
 require "selenium/webdriver"
+require "billy/capybara/rspec"
 require "view_component/test_helpers"
+
+Billy.configure do |config|
+  config.non_whitelisted_requests_disabled = true
+end
 
 module ViewComponent
   module TestHelpers
@@ -75,7 +80,9 @@ Capybara.register_driver :headless_chrome do |app|
     opts.add_argument "--headless"
     opts.add_argument "--no-sandbox"
     opts.add_argument "--window-size=1200,800"
-    opts.add_argument "--proxy-server=#{Capybara.app_host}:#{Capybara::Webmock.port_number}"
+    opts.add_argument "--ignore-certificate-errors"
+    opts.add_argument "--proxy-server=#{Billy.proxy.host}:#{Billy.proxy.port}"
+    opts.add_argument "--proxy-bypass-list=127.0.0.1;localhost;lvh.me;*.lvh.me"
   end
 
   Capybara::Selenium::Driver.new(app, browser: :chrome, options: options)
--- a/spec/system/notifications_spec.rb
+++ b/spec/system/notifications_spec.rb
@@ -154,6 +154,8 @@ describe "Notifications" do
     end
 
     scenario "With external link" do
+      proxy.stub("https://www.external.link.dev:443/").and_return(body: "<html></html>", code: 200)
+
       visit notifications_path
       expect(page).to have_content("Notification title")
       expect(page).to have_content("Notification body")
--- a/Gemfile
+++ b/Gemfile
@@ -82,9 +82,9 @@ end
 
 group :test do
   gem "capybara", "~> 3.40.0"
-  gem "capybara-webmock", "~> 0.7.0"
   gem "email_spec", "~> 2.3.1"
   gem "pdf-reader", "~> 2.15.1"
+  gem "puffing-billy", "~> 4.0"
   gem "rspec-rails", "~> 8.0.4"
   gem "selenium-webdriver", "~> 4.43.0"
   gem "simplecov", "~> 0.22.0", require: false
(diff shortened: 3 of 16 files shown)


spec/rails_helper.rb
```diff
@@ -19,8 +19,14 @@
 require "capybara/rails"
 require "capybara/rspec"
 require "selenium/webdriver"
+require "billy/capybara/rspec"
 require "view_component/test_helpers"
 
+Billy.configure do |c|
```
javierm: For clarity and consistency with similar code:

```suggestion
Billy.configure do |config|
```

spec/rails_helper.rb
```diff
@@ -19,8 +19,14 @@
...
+  c.whitelist << /lvh\.me/
```
javierm: Maybe this line is no longer necessary, now that we use the `--proxy-bypass-list` option 🤔. What do you think?

spec/rails_helper.rb
```diff
@@ -75,7 +81,9 @@ def visit(url, ...)
     opts.add_argument "--headless"
     opts.add_argument "--no-sandbox"
     opts.add_argument "--window-size=1200,800"
-    opts.add_argument "--proxy-server=#{Capybara.app_host}:#{Capybara::Webmock.port_number}"
+    opts.add_argument "--ignore-certificate-errors"
+    opts.add_argument "--proxy-server=#{Billy.proxy.host}:#{Billy.proxy.port}"
+    opts.add_argument "--proxy-bypass-list=127.0.0.1;localhost;lvh.me;*.lvh.me"
```
javierm, line 85: Do we need the first two? Maybe we can simplify it 🤔.

```suggestion
    opts.add_argument "--proxy-bypass-list=lvh.me;*.lvh.me"
```

spec/rails_helper.rb
```diff
@@ -89,5 +97,6 @@ def visit(url, ...)
 OmniAuth.config.test_mode = true
 
 def with_subdomain(subdomain, &)
-  Capybara.with(app_host: "http://#{subdomain}.lvh.me", &)
+  app_port = Capybara.current_session.server.port
+  Capybara.with(app_host: "http://#{subdomain}.lvh.me:#{app_port}", &)
```
javierm: Are these changes necessary? 🤔 Haven't noticed anything different without them.

spec/system/notifications_spec.rb
```diff
@@ -154,6 +154,11 @@
     end
 
     scenario "With external link" do
+      proxy.stub(/www\.external\.link\.dev/, method: "ALL").and_return(
+        headers: { "Content-Type" => "text/html" },
+        body: ""
+      )
+
```
javierm, line 158: Are these lines necessary? The test seems to pass without them 🤔. We also follow external links in the `external_links_spec.rb` tests.

spec/rails_helper.rb
```diff
@@ -19,8 +19,14 @@
 require "capybara/rails"
 require "capybara/rspec"
 require "selenium/webdriver"
+require "billy/capybara/rspec"
```
javierm, line 22: If we end up removing the line that calls `proxy.stub`, maybe we can remove this line too 🤔.

javierm (commented):
So great that there's an alternative :tada:. I've left a few comments. Let me know what you think! :wink:

spec/rails_helper.rb
taitus: I added it because it is included in Billy's documentation, in the [Setup for Capybara](https://github.com/oesmith/puffing-billy#setup-for-capybara) section.

I tried removing it to check if it was really needed. Without this require, the tests passed, but the CI job stayed running after the suite finished until it timed out. [[1]](https://github.com/consuldemocracy/consuldemocracy/actions/runs/27362534855/job/80853424379) [[2]](https://github.com/consuldemocracy/consuldemocracy/actions/runs/27399382573/job/80973653772)

Looking at the gem code, it seems that this require includes:
`config.after(:suite) { Billy.proxy.stop }`

So it seems that this require fixes the timeout problem by stopping the proxy properly at the end of the suite. So I would keep it to make sure the proxy is stopped properly 🤔 .

javierm (approved):

javierm (approved):

PR #6403, opened 2026-06-16
Disable Turbolinks for notification links

taitus:
## References
Related PR: #6401 

## Objectives
Notification links first open `/notifications/:id`. That route marks the notification as read and then redirects the user to the final link.

After adding Puffing Billy, external links go through the test proxy. This can make the redirect flow a bit slower in CI and expose a flaky system spec.

Turbolinks is not useful for this flow because there is no intermediate page to render. This PR disables Turbolinks for notification links, so the browser uses a normal navigation and follows the redirect directly.


javierm (approved):
Let's try this approach and see whether it's got any effect on the CI 👌.

PR #6414, opened 2026-06-26
Replace blueimp-file-upload with Dropzone for direct uploads

taitus:
## References

- Related to #6224 (Bump jquery from 3.7.1 to 4.0.0).
- Context: making file uploads compatible with jQuery 4 would require vendoring `jquery.fileupload.js` from `blueimp-file-upload` and maintaining a local patch (jQuery 4 removed helpers such as `$.trim`, `$.isNumeric`, `$.type` and `$.isArray`, which blueimp relies on). Instead of maintaining that patch, this PR removes blueimp and migrates uploads to Dropzone (`@deltablot/dropzone`).

## Objectives

- Remove the `blueimp-file-upload` dependency so we don't have to vendor or patch `jquery.fileupload.js` for jQuery 4.0.0.
- Migrate the direct uploads for documents and images to Dropzone while keeping the same markup, the same UI and the same backend contract.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -52,6 +52,7 @@
 //= require jquery-ui/ui/widgets/sortable
 //= require blueimp-file-upload/js/jquery.iframe-transport
 //= require blueimp-file-upload/js/jquery.fileupload
+//= require @deltablot/dropzone/dist/dropzone-min
 //= require foundation-sites
 //= require turbolinks
 //= require turbolinks_anchors
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -92,6 +92,7 @@
 //= require legislation_annotatable
 //= require legislation_draft_versions
 //= require followable
+//= require attachable
 //= require documentable
 //= require imageable
 //= require tree_navigator
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -50,8 +50,6 @@
 //= require jquery-ui/ui/i18n/datepicker-en-GB
 //= require jquery-ui/ui/widgets/autocomplete
 //= require jquery-ui/ui/widgets/sortable
-//= require blueimp-file-upload/js/jquery.iframe-transport
-//= require blueimp-file-upload/js/jquery.fileupload
 //= require @deltablot/dropzone/dist/dropzone-min
 //= require foundation-sites
 //= require turbolinks
--- a/app/assets/javascripts/attachable.js
+++ b/app/assets/javascripts/attachable.js
@@ -1,0 +1,120 @@
+(function() {
+  "use strict";
+
+  App.Attachable = {
+    setupInput: function(config) {
+      var $input, $container, uploadData, dropzone, $zone, dropzoneOptions;
+
+      $input = $(config.input);
+      $container = $input.closest(config.attachmentContainer);
+      $zone = $("<div>", { class: "hidden-dropzone-upload" });
+      $container.append($zone);
+
+      uploadData = {};
+
+      dropzoneOptions = {
+        url: $input.data("url"),
+        paramName: "attachment",
+        maxFiles: 1,
+        clickable: false,
+        headers: { "X-CSRF-Token": $("meta[name=csrf-token]").attr("content") },
+        previewTemplate: "<div></div>"
+      };
+
+      dropzone = new Dropzone($zone[0], dropzoneOptions);
+
+      $input.on("change", function() {
+        uploadData = App.Attachable.buildData(config.input);
+        App.Attachable.setFilename(uploadData, this.files[0].name);
+        dropzone.addFile(this.files[0]);
+      });
+
+      dropzone.on("addedfile", function() {
+        uploadData = App.Attachable.buildData(config.input);
+        App.Attachable.clearProgressBar(uploadData);
+        App.Attachable.setProgressBar(uploadData, "uploading");
+      });
+
+      dropzone.on("uploadprogress", function(_file, progress) {
+        $(uploadData.progressBar).find(".loading-bar").css("width", progress + "%");
+      });
+
+      dropzone.on("success", function(_file, response) {
+        var destroyAttachmentLink;
+
+        $(uploadData.cachedAttachmentField).val(response.cached_attachment);
+        App.Attachable.setTitleFromFile(uploadData, response.filename);
+        App.Attachable.setProgressBar(uploadData, "complete");
+        App.Attachable.setFilename(uploadData, response.filename);
+        App.Attachable.clearInputErrors(uploadData);
+        destroyAttachmentLink = $(response.destroy_link);
+        $(uploadData.destroyAttachmentLinkContainer).html(destroyAttachmentLink);
+
+        if (config.onSuccess) {
+          config.onSuccess(uploadData, response);
+        }
+      });
+
+      dropzone.on("error", function(file, message) {
+        $(uploadData.cachedAttachmentField).val("");
+        App.Attachable.clearFilename(uploadData);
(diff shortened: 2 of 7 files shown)


app/assets/javascripts/dropzone_autodiscover_off.js
```diff
@@ -0,0 +1 @@
+Dropzone.autoDiscover = false;
```
javierm: Is this necessary? 🤔 According to dropzone's changelog:

> `Dropzone.autoDiscover` has been removed! If you want to auto discover your elements, invoke Dropzone.discover()` after your HTML has loaded and it will do the same.

The only reference to autoDiscover in dropzone's source code is a piece of commented code 🤔.

app/assets/javascripts/attachable_dropzone.js
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  App.AttachableDropzone = {
+    setupInput: function(config) {
+      var $input, $container, uploadData, dropzone, $zone, attachable, changeNamespace, dropzoneDataKey, dropzoneOptions;
+
+      $input = $(config.input);
+      attachable = config.attachable;
+      changeNamespace = "." + config.namespace;
+      dropzoneDataKey = config.namespace + "Dropzone";
```
javierm: Do we need this? 🤔

app/assets/javascripts/attachable_dropzone.js
```diff
@@ -0,0 +1,105 @@
...
+        $(uploadData.addAttachmentLabel).addClass("error");
+
+        if (config.onError) {
+          config.onError(uploadData);
+        }
+
+        dropzone.removeFile(file);
+      });
+
+      $input.data(dropzoneDataKey, dropzone);
```
javierm: I take it this line is used to prevent adding a dropzone twice, right? Do we need to prevent it? 🤔

app/assets/stylesheets/mixins/uploads.scss
```diff
@@ -5,6 +5,13 @@
     max-width: rem-calc(150);
   }
 
+  .hidden-dropzone-upload {
+    height: 0;
+    overflow: hidden;
+    position: absolute;
+    width: 0;
+  }
+
```
javierm: Do we need this code? 🤔 I haven't noticed any differences when removing it.

app/assets/javascripts/attachable_dropzone.js
```diff
@@ -0,0 +1,105 @@
...
+      if ($input.data(dropzoneDataKey)) {
+        return;
+      }
+
+      $container = $input.closest(config.attachmentContainer);
+      $zone = $container.find("." + config.dropzoneClass);
+      if ($zone.length === 0) {
+        $zone = $("<div>", { class: config.dropzoneClass + " hidden-dropzone-upload" });
+        $container.append($zone);
+      }
```
javierm: Do we need this condition? 🤔


```suggestion
      $container = $input.closest(config.attachmentContainer);
      $zone = $("<div>", { class: config.dropzoneClass });
      $container.append($zone);
```

app/assets/javascripts/attachable_dropzone.js
```diff
@@ -0,0 +1,105 @@
+(function() {
+  "use strict";
+
+  function uploadDataWithErrors(uploadData, xhr, message) {
```
javierm: Why is this function outside the definition of `App.AttachableDropzone`? 🤔

app/assets/javascripts/attachable_dropzone.js
```diff
@@ -0,0 +1,105 @@
...
+        paramName: "attachment",
+        maxFiles: 1,
+        clickable: false,
+        headers: { "X-CSRF-Token": $("meta[name=csrf-token]").attr("content") },
+        previewTemplate: "<div></div>"
+      };
+
+      dropzone = new Dropzone($zone[0], dropzoneOptions);
+
+      $input.on("change" + changeNamespace, function() {
```
javierm: Do we need the `changeNamespace` here? 🤔

```suggestion
      $input.on("change", function() {
```

app/assets/javascripts/attachable_dropzone.js
```diff
@@ -0,0 +1,105 @@
...
+
+  function uploadDataWithErrors(uploadData, xhr, message) {
+    var errors;
+
+    errors = xhr && xhr.responseJSON && xhr.responseJSON.errors || message;
+    uploadData.jqXHR = xhr || {};
+    uploadData.jqXHR.responseJSON = uploadData.jqXHR.responseJSON || {};
+    uploadData.jqXHR.responseJSON.errors = uploadData.jqXHR.responseJSON.errors || errors;
+
+    return uploadData;
```
javierm: We're returning `uploadData`, but in the function calling this one, we aren't using this returned value 🤔. Is it on purpose?

app/assets/javascripts/attachable_dropzone.js
```diff
@@ -0,0 +1,105 @@
...
+      dropzoneDataKey = config.namespace + "Dropzone";
+
+      if ($input.data(dropzoneDataKey)) {
+        return;
+      }
+
+      $container = $input.closest(config.attachmentContainer);
+      $zone = $container.find("." + config.dropzoneClass);
+      if ($zone.length === 0) {
+        $zone = $("<div>", { class: config.dropzoneClass + " hidden-dropzone-upload" });
```
javierm: Do we need `config.dropzoneClass`? 🤔 Haven't found where we're using it.

app/assets/javascripts/attachable_dropzone.js
```diff
@@ -0,0 +1,105 @@
...
+        headers: { "X-CSRF-Token": $("meta[name=csrf-token]").attr("content") },
+        previewTemplate: "<div></div>"
+      };
+
+      dropzone = new Dropzone($zone[0], dropzoneOptions);
+
+      $input.on("change" + changeNamespace, function() {
+        if (this.files.length === 0) {
+          return;
+        }
```
javierm: Do we need this condition? When can the input change without having any files?

app/assets/javascripts/attachable_dropzone.js
```diff
@@ -0,0 +1,105 @@
...
+
+        uploadData = attachable.buildData([], config.input);
+        attachable.setFilename(uploadData, this.files[0].name);
+        dropzone.removeAllFiles(true);
```
javierm: Do we need this line? 🤔

javierm (commented):
@taitus Here's part 1 of the review. Let me know what you think! :wink:

app/assets/javascripts/attachable_dropzone.js
taitus: No, it's redundant. I'll remove it.

app/assets/javascripts/application.js
```diff
@@ -50,8 +50,7 @@
 //= require jquery-ui/ui/i18n/datepicker-en-GB
 //= require jquery-ui/ui/widgets/autocomplete
 //= require jquery-ui/ui/widgets/sortable
-//= require blueimp-file-upload/js/jquery.iframe-transport
-//= require blueimp-file-upload/js/jquery.fileupload
+//= require @deltablot/dropzone/dist/dropzone-min
```
javierm: Since we're already minify the JavaScript with `assets:precompile`, and for consistency, what do you think about:

```suggestion
//= require @deltablot/dropzone/dist/dropzone
```

javierm (commented):
Looking good! Just one small detail comes to mind.

javierm (approved):

app/assets/javascripts/application.js
javierm: Looks like this doesn't work, though... Oops!

PR #6438, opened 2026-07-03
Restore image cancel behavior after validation errors

taitus:
## References

Related PR: #6414 , specifically commit 00ffe208e3 ("Migrate image uploads from blueimp to Dropzone").

## Objectives

Fix a regression in the nested image field. After a validation error, clicking "Cancel" on the image block removed the fields but did not restore the "Add image" button, leaving the image section empty.

### DIFF
--- a/spec/system/nested_imageable_spec.rb
+++ b/spec/system/nested_imageable_spec.rb
@@ -68,13 +68,17 @@ describe "Nested imageable" do
       expect(cached_attachment_field.value).to be_empty
     end
 
-    scenario "Should show image errors after invalid form submit" do
+    scenario "Shows image errors after invalid submit and restores add link on cancel" do
       click_link "Add image"
       click_button submit_button_text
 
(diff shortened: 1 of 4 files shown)


spec/system/nested_imageable_spec.rb
```diff
@@ -68,13 +68,17 @@
...
       click_link "Add image"
       click_button submit_button_text
 
       within "#nested-image .image-fields" do
         expect(page).to have_content("can't be blank", count: 2)
+        click_link "Cancel"
       end
+
+      expect(page).not_to have_css "#nested-image .image-fields"
+      expect(page).to have_link id: "new_image_link"
```
javierm: In order to test it from the user's point of view:

```suggestion
      expect(page).not_to have_content "Choose image"
      expect(page).to have_link "Add image"
```

javierm (approved):

PR #6301, opened 2026-04-01
Upgrade C3.js datavis package to v0.7.20

digitalWestie:
## Objectives

Following on from #6300, to upgrade dependencies towards modernising visualisation packages.

## Visual Changes

None. I have tested manually to ensure the graphs and control behaviour still works. This includes panning, zooming, toggling labels, and hovering.

image

image

image

PR #6439, opened 2026-07-06
Replace C3.js with billboard.js

taitus:
## References

Supersedes #6301, which attempted to upgrade the unmaintained C3.js dependency while keeping the old D3 3.x stack.

C3 is no longer maintained and is tied to outdated D3 versions. [billboard.js](https://github.com/naver/billboard.js) is its actively maintained successor and keeps a compatible API (`bb.generate` with the same options we were using).

## Objectives

- Replace C3.js and D3 3.5.17 with billboard.js 3.x and preserve current chart behaviour; no redesign of the visualization layer.
- Remove dead code (`c3ext.js`) that was required but never used.
- Remove the `c3` and `d3` npm dependencies. This clears the `d3-color` advisory that `npm audit` reported with the old D3 3.x stack (verified with `npm audit` / `npm audit --omit=dev` before and after the change).

## billboard.js version

We are using billboard.js 3.x (`^3.18.0`), not 4.x.

Version 4 was released recently. It would likely work with our pkgd + Sprockets setup, but v3 is the safer choice for this migration: it is the established C3 replacement path and has been in production use for longer. We can evaluate upgrading to 4.x in a follow-up once this change is stable.


app/assets/javascripts/stats.js
```diff
@@ -1,6 +1,6 @@
 (function() {
   "use strict";
-  // Helper for generate C3.js graphs
+  // Helper for generate billboard.js graphs
```
javierm: Since we don't usually add comments to the code, maybe now it's a good time to remove this one 🤔. What do you think?

By the way, there's also a comment in `app/models/ahoy/data_source.rb` that references C3.

```suggestion
```

app/assets/stylesheets/dashboard.scss
```diff
@@ -345,7 +345,7 @@
 
 .proposal-graph {
 
-  .c3-ygrid-lines {
+  .bb-ygrid-lines {
```
javierm, line 348: Is this class used? There were elements with the `c3-ygrid-lines` class in the old version, but I can't find any elements with the `bb-ygrid-lines` class in the new one 🤔.

taitus: It is only rendered on the proposal dashboard progress chart (`dashboard_graphs.js`), not on the admin stats charts.

javierm: I couldn't find it there either :thinking:.

javierm: Oh, so apparently now you need to create a resource in order for this class to be there :relieved:.

Looks good to me, then.

javierm (approved):

Issue #2415, opened 2018-02-01
Race Condition due to lack of UNIQUE constrain on "votes" table


dreadlocked:
By default, the tuple (user_id, proposal_id) is not UNIQUE on votes table. This makes ActAsVotable vulnerable to race condition, where the same user can vote twice (or more) the same proposal. 

	votes = find_votes_by(options[:voter], options[:vote_scope])

  	if votes.count == (0) || options[:duplicate]
    	# this voter has never voted
    	vote = ActsAsVotable::Vote.new(
      	votable: self,
      	voter: options[:voter],
      	vote_scope: options[:vote_scope]
    	)
  	else
    	# this voter is potentially changing his vote
    	vote = votes.last

A thread can request database if a voter have already voted a proposal while the first thread have not executed insert statement yet leading to a race condition where the database responds votes.count = 0, the second thread will then enter into the condition and insert a second register because the database definition permits this behavior by default.

javierm:
A possible plan for fixing this issue:

1. Avoid duplicate votes in the future, maybe by locking `current_user` every time we use `vote_by` (note: I'm not sure this is the best option)
2. Add a task removing existing duplicate votes
3. Release a version of CONSUL so duplicate votes are removed before the unique index is added
4. Add a unique index to the votes table or upgrade to a version of acts_as_votable implementing this restriction (see ryanto/acts_as_votable#211)
5. Remove the `current_user` locks in the code (we might also look for ways to recover from a `RecordNotUnique` exception :thinking_face:)
6. Release another version of CONSUL

lucialuzuriaga:
Edit: reopening as there is still pending work to be done for this issue.

lucialuzuriaga:
Closing as fixed. Adressed in #6441.

PR #6441, opened 2026-07-08
Avoid creating duplicate votes

taitus:
## References

* Fixes part of #2415
* Similar to #5532

## Objectives

* Prevent duplicate vote records
* Remove existing duplicate votes as part of the release tasks, so we can later add a unique index
* Handle vote collisions when taking votes from erased users
* Fix anonymous votes on debates not being properly updated

## Notes

This pull request follows the first steps of the plan described in [comment](https://github.com/consuldemocracy/consuldemocracy/issues/2415#issuecomment-1167424959) from #2415:

1. avoid duplicate votes in the future by locking the user when calling `vote_by`;
2. add a task to remove existing duplicate votes;
3. release a version of CONSUL so duplicate votes are removed before adding the unique index.

We can't introduce a unique index yet, otherwise, the migration introducing it could crash if there are already duplicate records in existing installations.

The unique index will be added in a later pull request, after this release has been published and the `votes:remove_duplicate_votes` task has run.

For now, this pull request uses `user.with_lock` that call `vote_by`, following the approach used previously for duplicate poll voters.

This is a temporary measure until the database can enforce uniqueness directly.

## Release notes

ℹ️  Due to a race condition in `acts_as_votable`, previous versions could create duplicate records in the `votes` table when the same user voted the same resource simultaneously.

This release prevents new duplicate votes in the main application flows and, as part of the release tasks, runs `votes:remove_duplicate_votes` to remove existing duplicate vote records.

When duplicate votes are removed, affected cached vote counters are recalculated.

A later release will add a unique index to the `votes` table to enforce this restriction at the database level. As usual, backup your database before upgrading.

### DIFF
--- a/app/models/budget/investment.rb
+++ b/app/models/budget/investment.rb
@@ -314,7 +314,7 @@ class Budget
     end
 
     def register_selection(user)
-      vote_by(voter: user, vote: "yes") if selectable_by?(user)
+      user.with_lock { vote_by(voter: user, vote: "yes") } if selectable_by?(user)
     end
 
     def calculate_confidence_score
--- a/app/models/proposal.rb
+++ b/app/models/proposal.rb
@@ -183,7 +183,7 @@ class Proposal < ApplicationRecord
 
   def register_vote(user, vote_value)
     if votable_by?(user) && !archived?
-      vote_by(voter: user, vote: vote_value)
+      user.with_lock { vote_by(voter: user, vote: vote_value) }
     end
   end
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -312,7 +312,19 @@ class User < ApplicationRecord
     end
 
     Budget::Ballot.where(user_id: other_user.id).update_all(user_id: id)
-    Vote.where("voter_id = ? AND voter_type = ?", other_user.id, "User").update_all(voter_id: id)
+
+    with_lock do
+      Vote.where(voter_id: other_user.id, voter_type: "User").find_each do |vote|
+        votable = vote.votable
+        if Vote.where(voter_id: id, voter_type: "User", votable: votable).any?
+          vote.delete
+          votable&.update_cached_votes
+        else
+          vote.update_column(:voter_id, id)
+        end
+      end
+    end
+
     data_log = "id: #{other_user.id} - #{Time.current.strftime("%Y-%m-%d %H:%M:%S")}"
     update!(former_users_data_log: "#{former_users_data_log} | #{data_log}")
   end
--- a/lib/tasks/consul.rake
+++ b/lib/tasks/consul.rake
@@ -3,8 +3,10 @@ namespace :consul do
   task execute_release_tasks: ["settings:rename_setting_keys",
                                "settings:add_new_settings",
                                "cache:clear",
-                               "execute_release_2.5.1_tasks"]
+                               "execute_release_2.6.0_tasks"]
 
-  desc "Runs tasks needed to upgrade from 2.4.1 to 2.5.1"
-  task "execute_release_2.5.1_tasks": []
+  desc "Runs tasks needed to upgrade from 2.5.1 to 2.6.0"
+  task "execute_release_2.6.0_tasks": [
+    "votes:remove_duplicate_votes"
+  ]
 end
--- a/lib/tasks/votes.rake
+++ b/lib/tasks/votes.rake
@@ -2,6 +2,7 @@ namespace :votes do
   desc "Removes duplicate votes"
   task remove_duplicate_votes: :environment do
     logger = ApplicationLogger.new
+    duplicate_records_logger = DuplicateRecordsLogger.new
     logger.info "Removing duplicate votes"
 
     Tenant.run_on_each do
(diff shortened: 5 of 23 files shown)


app/models/budget/investment.rb
```diff
@@ -314,7 +314,7 @@ def voted_in?(heading, user)
     end
 
     def register_selection(user)
-      vote_by(voter: user, vote: "yes") if selectable_by?(user)
+      user.with_lock { vote_by(voter: user, vote: "yes") } if selectable_by?(user)
```
javierm, line 317: Shall we add a test for this change? 🤔

app/models/proposal.rb
```diff
@@ -183,7 +183,7 @@ def retired?
 
   def register_vote(user, vote_value)
     if votable_by?(user) && !archived?
-      vote_by(voter: user, vote: vote_value)
+      user.with_lock { vote_by(voter: user, vote: vote_value) }
```
javierm, line 186: Shall we add a test for this change? 🤔

app/models/user.rb
```diff
@@ -312,7 +312,17 @@ def take_votes_from(other_user)
     end
 
     Budget::Ballot.where(user_id: other_user.id).update_all(user_id: id)
-    Vote.where("voter_id = ? AND voter_type = ?", other_user.id, "User").update_all(voter_id: id)
+    with_lock do
```
javierm, line 316: Maybe we could take this chance to add blank lines before and after this block 🤔. What do you think?

spec/models/debate_spec.rb
```diff
@@ -215,6 +215,18 @@
         expect { debate.register_vote(user, "yes") }.not_to change { debate.reload.total_anonymous_votes }
       end
     end
+
+    it "does not create two votes when calling the method twice at the same time", :race_condition do
+      debate.update(cached_anonymous_votes_total: 42, cached_votes_total: 100)
```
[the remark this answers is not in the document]
javierm:
```suggestion
      debate.update!(cached_anonymous_votes_total: 42, cached_votes_total: 100)
```

spec/models/user_spec.rb
```diff
@@ -824,6 +824,41 @@
       expect(Poll::Voter.where(user: user)).to match_array [voter, other_poll_voter]
       expect(Poll::Voter.where(user: other_user)).to eq []
     end
+
+    it "does not reassign votes if the user has already voted" do
```
javierm, line 828: This description is identical to the one in the previous example 🤔. What do you think about changing one of these descriptions? Should the previous example by "voters" instead of "votes"? Maybe we should add the `RSpec/RepeatedDescription` rule 🤔.

spec/models/user_spec.rb
```diff
@@ -824,6 +824,41 @@
...
+      other_vote = create(:vote, voter: other_user, votable: proposal)
+      other_debate_vote = create(:vote, voter: other_user, votable: debate)
+
+      expect(Vote.where(voter: user)).to eq [vote]
+      expect(Vote.where(voter: other_user)).to match_array [other_vote, other_debate_vote]
+
+      user.take_votes_from(other_user)
+
+      expect(Vote.where(voter: user)).to match_array [vote, other_debate_vote]
+      expect(Vote.where(voter: other_user)).to eq []
```
javierm, line 844: Shall we add one expectation to check that `other_vote` has been deleted?

lib/tasks/consul.rake
```diff
@@ -3,7 +3,12 @@ namespace :consul do
...
-                               "execute_release_2.5.1_tasks"]
+                               "execute_release_2.6.0_tasks"]
+
+  desc "Runs tasks needed to upgrade from 2.5.1 to 2.6.0"
+  task "execute_release_2.6.0_tasks": [
+    "votes:remove_duplicate_votes"
+  ]
 
   desc "Runs tasks needed to upgrade from 2.4.1 to 2.5.1"
   task "execute_release_2.5.1_tasks": []
```
javierm: I take it we can now delete this code, right?

lib/tasks/votes.rake
```diff
@@ -1,4 +1,40 @@
 namespace :votes do
+  desc "Removes duplicate votes"
+  task remove_duplicate_votes: :environment do
+    logger = ApplicationLogger.new
+    logger.info "Removing duplicate votes"
```
javierm, line 6: Shall we add a `DuplicateRecordsLogger` as well, like we did with polls?

lib/tasks/votes.rake
```diff
@@ -1,4 +1,40 @@
...
+          vote.delete
+
+          tenant_info = " on tenant #{Tenant.current_schema}" unless Tenant.default?
+          logger.info "Deleted duplicate record with ID #{vote.id} " \
+                      "from the #{Vote.table_name} table " \
+                      "with voter_id #{voter_id}, voter_type #{voter_type}, " \
+                      "votable_id #{votable_id} and votable_type #{votable_type}" \
+                      + tenant_info.to_s
+
+          votable&.update_cached_votes
```
javierm: Shall we move this line before the block with the logger? This way we'd group related logic together 🤔.

javierm (commented):
@taitus Nice! I've left a few comments. Let me know what you think!

spec/models/user_spec.rb
```diff
@@ -824,6 +824,42 @@
...
+      expect(Vote.where(id: other_vote.id)).to eq []
```
javierm: Since we're using `id:`, which can only return one record at most:

```suggestion
      expect(Vote.find_by(id: other_vote.id)).to be nil
```

javierm (commented):
One more thing :joy:.

javierm (approved):

javierm (approved):

PR #6445, opened 2026-07-16
fix(i18n): missing french translations

M3lkior:
This PR add some missing french translations since recents features

### DIFF
--- a/config/locales/fr/general.yml
+++ b/config/locales/fr/general.yml
@@ -195,6 +195,7 @@ fr:
       conditions: Conditions légales
       consul: Application CONSUL DEMOCRACY
       consul_url: https://github.com/consuldemocracy/consuldemocracy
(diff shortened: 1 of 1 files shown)


lucialuzuriaga (approved):
Thanks, @M3lkior! Will be added to the next release.

lucialuzuriaga (commented):
@M3lkior  Apologies, but upon double-checking and as stated in our `CONTRIBUTING.md`, all translations must be added via Crowdin in order to persist in the platform. 

_Add or modify i18n translations only in the base languages, English (en) and Spanish (es); we manage all other languages through the [Crowdin integration](https://translate.consuldemocracy.org/)._

Even if we merged this pull request, Crowdin would overwrite the translations, so it's best that you add them there. Thank you 🙏

PR #6448, opened 2026-07-23
Fix image suggestions button alignment after failed upload

taitus:
## References
Related PR's: [#6190](https://github.com/consuldemocracy/consuldemocracy/pull/6190) and [#6325](https://github.com/consuldemocracy/consuldemocracy/pull/6325)

## Objectives

- Fix the attachment action row layout on validation errors so Choose and Suggest stay on one line, Cancel stays on the right, and the error appears below.

### DIFF
--- a/app/assets/stylesheets/mixins/uploads.scss
+++ b/app/assets/stylesheets/mixins/uploads.scss
@@ -26,6 +26,9 @@
 
     .document-attachment,
     .image-attachment {
+      column-gap: 0.7ch;
+      display: flex;
+      flex-wrap: wrap;
       padding-left: 0;
 
       p {
@@ -35,6 +38,10 @@
       &:focus-within label {
         @include focus-outline;
       }
+
+      .suggested-images-wrapper {
+        flex-grow: 1;
+      }
     }
 
     [type=file] {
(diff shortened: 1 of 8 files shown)


app/assets/stylesheets/mixins/uploads.scss
```diff
@@ -26,6 +26,10 @@
 
     .document-attachment,
     .image-attachment {
+      align-items: center;
+      display: flex;
+      flex-wrap: wrap;
+      gap: calc($line-height / 4);
```
javierm: We usually use `$line-height` for vertical margins 🤔. Since here we want approximately one character as gap:

```suggestion
      gap: calc($line-height / 4) 0.7ch;
```

app/assets/stylesheets/mixins/uploads.scss
```diff
@@ -35,6 +39,15 @@
       &:focus-within label {
         @include focus-outline;
       }
+
+      > div:has([type=file]) {
```
javierm: Maybe we can simplify it a little bit 🤔.

```suggestion
      > :has([type=file]) {
```

app/assets/stylesheets/mixins/uploads.scss
```diff
@@ -35,6 +39,15 @@
...
+        display: none;
+      }
+
+      > span[id$="_error"] {
```
javierm: Same here.

```suggestion
      > [id$="_error"] {
```

javierm (approved):

PR #6484, opened 2026-08-06
Replace Uglifier with Terser for JS compression

taitus:
## References
[#6439](https://github.com/consuldemocracy/consuldemocracy/pull/6439) Introduced billboard.js, which broke assets:precompile with Uglifier on staging/production.

## Objectives
Uglifier cannot minify modern JavaScript such as billboard.js, which breaks assets:precompile in staging and production. Terser handles ES6+ syntax and matches what Rails recommends for current asset pipelines.


.gitlab-ci.yml
```diff
@@ -33,6 +33,20 @@ tests:
     paths:
       - tmp/screenshots/
 
+production_assets:
+  image: "ruby:3.4.9"
+  stage: test
```
javierm: Do we think i'd make sense to add the new task to the `stages` array and change the stage? 🤔

.gitlab-ci.yml
```diff
@@ -33,6 +33,20 @@ tests:
...
+  cache:
+    key: consul
+    paths:
+      - vendor/
+  script:
+    - apt-get update && apt-get install -y nodejs npm
+    - for i in config/*.example; do cp "$i" "${i/.example}"; done
+    - bundle --without development
```
javierm:
```suggestion
    - bundle --without development test
```

.github/workflows/production_assets.yml
```diff
@@ -0,0 +1,32 @@
...
+      - i18n_master
+
+permissions:
+  contents: read
+
+jobs:
+  production_assets:
+    runs-on: ubuntu-24.04
+    timeout-minutes: 15
+    steps:
```
javierm, line 19: What do you think about an extra environment variable? I'm not sure about the syntax 🤔. Would be something like:

```yaml
    env:
      BUNDLE_WITHOUT: "test development"
```

javierm (commented):
Very nice! I've left a couple of comments about CI.

javierm (approved):

